PR #5, opened 2026-05-21
docs(guardrails): Focus skills on content-safety flow; docs updates

JashG:
- **Skill updates:** Lead with content-safety input/output rails as the canonical example instead of self-check.
- **Docs updates:** Update self-check prompts, add note about using reasoning models with self-check


gabwow (approved):

JashG:

PR #6, opened 2026-05-21
feat(evaluator): port metric output protocol

SandyChapman:
## Summary

Ports the Metric protocol contract from NVIDIA-NeMo/Evaluator#950 into `nemo_evaluator_sdk`.

This intentionally replaces the SDK runtime metric author contract with:

- `compute_scores(input: MetricInput) -> MetricResult(outputs=...)`
- `output_spec() -> list[MetricOutputSpec]`

The old SDK metric author contract using `compute_scores(item, sample)`, `score_names()`, and `MetricResult(scores=...)` is removed.

## What Changed

- Adds SDK-owned metric protocol types in `nemo_evaluator_sdk.metrics.protocol`.
- Plumbs `MetricInput` / `MetricResult(outputs=...)` through local metric execution, benchmark execution, row results, aggregation, exports, and service integration.
- Migrates built-in SDK metrics to the new protocol, including deterministic metrics, BLEU/ROUGE/F1, LLM judge, RAGAS, remote metrics, and tool-calling metrics.
- Moves shared aggregation helpers into the SDK and updates the Evaluator service to import them directly.
- Keeps `MetricOutputSpec` focused on output contract shape only. LLM judge rubric metrics emit a numeric output plus a companion label output, e.g. `quality` and `quality.label`; rubric aggregate metadata is derived separately from metric score config at aggregation time.
- Refreshes generated OpenAPI, web SDK, Python SDK, and the vendored beta Evaluator SDK copy.
- Fixes the repo-root unit-suite collection path for SDK metric helper imports and tolerates both argparse choice renderings in the platform startup test.

## Commit Structure

This PR was originally developed as a layered stack, but has been squashed for review into a single commit:

- `feat(evaluator): port metric output protocol`

## Validation

- `LOG_FORMAT=plain make test-unit-ci`
  - `9382 passed, 57 skipped, 2 xfailed, 9 subtests passed`
- `uv run --frozen pytest packages/nemo_evaluator_sdk/tests -q`
  - `791 passed, 1 warning`
- `uv run --frozen pytest plugins/nemo-evaluator/tests -q`
  - `135 passed, 5 warnings`
- `uv run --frozen pytest services/evaluator/tests/nmp/evaluator/api/v2/metrics/schemas/test_evaluation.py services/evaluator/tests/nmp/evaluator/app/metrics/test_metric_protocol.py -q`
  - `50 passed`
- `uv run --frozen ruff check`
- `uv run --frozen ruff format --check`
- Targeted `ty check` on PR-owned SDK/service/plugin files
- `bash tools/lint/lint-openapi.sh`
- `bash tools/lint/lint-python-sdk.sh`
- `mise exec node@22.19.0 -- pnpm gen:check`

## Notes

- This is intentionally breaking for SDK metric authors.
- Aggregation is derived from numeric/boolean output specs.
- Non-score outputs are preserved in row-level results and exports, but excluded from aggregate statistics.
- Full-repo `ty check` still has unrelated baseline diagnostics outside this PR; targeted typecheck on PR-owned files passes.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/protocol.py
```diff
@@ -0,0 +1,241 @@
...
+
+    row: DatasetRow
+    candidate: CandidateOutput
+
+
+class ContinuousScore(RootModel[float]):
+    """Continuous numeric metric value."""
+
+
+class DiscreteScore(RootModel[int]):
```
ngoncharenko, line 62: Great to see our score types becoming more versatile

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/protocol.py
```diff
@@ -0,0 +1,241 @@
...
+        return sample
+
+
+class MetricInput(BaseModel):
+    """Complete per-row scoring input passed to a metric."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    row: DatasetRow
+    candidate: CandidateOutput
```
ngoncharenko, line 55: wouldn't naming it `prediction: PredictionOutput` make it semantically more clear?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/rouge.py
```diff
@@ -56,12 +56,12 @@ def _metric(self, item: dict, sample: dict) -> dict:
         )
         return self._scorer.score(ground_truth, prediction)
 
-    async def compute_scores(self, item: dict, sample: dict) -> MetricResult:
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
         """Compute structured score output for one item/sample pair."""
-        scores = self._metric(item, sample)
+        scores = self._metric(input.row.data, input.candidate.as_sample())
```
ngoncharenko: I wonder if we could keep `sample` arg still typed and only flatten it where it actually needs to be flattened 

```python
def _metric(self, item: dict, sample: CandidateOutput) -> dict
              ground_truth, prediction = render_reference_and_candidate(...., sample=sample)


## in the packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/template_rendering.py
def render_reference_and_candidate(..., sample: CandidateOutput):
      context = build_template_context(item, sample)

def build_template_context(item: dict[str, Any], sample: CandidateOutput) -> dict[str, Any]:
    """Build the template context shared by item and sample rendering."""
    return {**item, **sample, "item": item, "sample": sample.as_sample()}
```

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/aggregation.py
```diff
@@ -317,3 +413,29 @@ def aggregate_metrics(items: list[MetricResult]) -> AggregatedMetricResult:
...
+                count=0,
+            )
+            for rubric in score.rubric
+        ]
+    return definitions
+
+
+def rubric_definitions_from_metric(metric: object) -> dict[str, list[RubricScoreStat]]:
+    """Return rubric bucket definitions for metrics that carry score config."""
+    scores = getattr(metric, "scores", None)
```
ngoncharenko: what kind of Metric has metric.scores? I wonder if we could do this instead and avoid cast:

```python
@runtime_checkable
class MetricWithScores(Protocol):
    @property
    def scores(self) -> Sequence[Score]:
        ...

def rubric_definitions_from_metric(metric: object) -> dict[str, list[RubricScoreStat]]:
    if not isinstance(metric, MetricWithScores):
        return {}
    return rubric_definitions_from_scores(metric.scores)
```
and then invoke it after checking instance

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
```diff
@@ -9,30 +9,82 @@
...
 
 logger = getLogger(__name__)
 
 
-def nan_metric_result(score_names: str | Iterable[str]) -> MetricResult:
-    """Build the NaN score payload used for ignored scoring failures.
+_CANDIDATE_SAMPLE_FIELDS = frozenset({"output_text", "response", "trajectory"})
+
+
+def build_metric_input(row: dict[str, Any], sample: dict[str, Any], index: int | None = None) -> MetricInput:
```
ngoncharenko, line 41: would it worth making a factory method? MetricInput.build_from_row_and_candidate?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
```diff
@@ -9,30 +9,82 @@
...
+            response=sample.get("response"),
+            trajectory=sample.get("trajectory"),
+            metadata=metadata,
+        ),
+    )
+
 
-    Accepts either a single score name (single-metric pipelines) or an
-    iterable of score names (multi-score/benchmark pipelines).
+def nan_metric_result(outputs: Iterable[MetricOutputSpec]) -> MetricResult:
```
ngoncharenko, line 60: we can also make this factory method potentially

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
```diff
@@ -9,30 +9,82 @@
...
+    Only aggregateable outputs receive NaN placeholders; non-score outputs are
+    not synthesized for failed rows.
     """
-    names: Iterable[str] = (score_names,) if isinstance(score_names, str) else score_names
-    return MetricResult(scores=[MetricScore(name=name, value=float("nan")) for name in names])
+    return MetricResult(
+        outputs=[
+            MetricOutput(name=output.name, value=float("nan"))
+            for output in outputs
+            if output.value_json_schema().get("type") in {"number", "integer", "boolean"}
```
ngoncharenko: we could probably re-use `_is_aggregateable_spec(output)` from packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/aggregation.py

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/llm_judge.py
```diff
@@ -141,7 +159,12 @@ def _handle_invalid_output(self, error: Exception, fallback: MetricResult, messa
         raise error
 
     def _nan_result(self) -> MetricResult:
-        return MetricResult(scores=[MetricScore(name=name, value=float("nan")) for name in self._parsers])
+        outputs: list[MetricOutput] = []
+        for score in self.scores:
+            outputs.append(MetricOutput(name=score.name, value=float("nan")))
+            if isinstance(score, RubricScore):
+                outputs.append(MetricOutput(name=f"{score.name}.label", value=""))
```
ngoncharenko, line 170: does it mean RubricScore will have two MetricOutput entries?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
SandyChapman: In general, I prefer free functions over static methods. While putting it in the class helps with namespacing, we already have the module to namespace.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
SandyChapman: Same here. If we find we need to re-use them we can move under MetricResult as a factory method.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/llm_judge.py
SandyChapman: Yes, intentionally. A rubric score emits the numeric score output, e.g. `quality`, plus a companion label output, e.g. `quality.label`. The numeric output participates in aggregation, while the label output lets aggregation reconstruct the rubric distribution/mode without putting rubric-specific metadata into the generic metric output protocol.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/protocol.py
SandyChapman: Good question. I’m going to keep `candidate` because the scored value is not always a model prediction: it can be an offline dataset field, an agent trajectory, a retrieval response, or another candidate artifact. I did update the docstring to clarify that this is the candidate/prediction output being scored.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/protocol.py
SandyChapman: Agreed! The type of score impacts how we do aggregations so I think this is a nice change.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
SandyChapman: Agreed. I promoted the helper to `is_aggregateable_output_spec(...)` and reused it from `nan_metric_result(...)`, rather than importing a private helper across modules.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/rouge.py
SandyChapman: Agreed. I updated the template rendering helpers to accept `CandidateOutput` directly and centralize the `.as_sample()` conversion there. Most metric implementations now pass `input.candidate` through to the template helpers instead of flattening it themselves.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/aggregation.py
SandyChapman: Agreed. The `scores` attribute is specifically for metrics whose config carries score definitions, especially rubric score definitions for LLM judge-style metrics. I added a `MetricWithScores` protocol and use `isinstance(metric, MetricWithScores)` before reading `metric.scores`, which removes the `getattr`/`cast` shape.

PR #9, opened 2026-05-21
feat(intake): add traces API to intake

BrianNewsom:
Also fixes a number of subtle bugs, including duplicate ATIF token counting.

Validation:

- `uv run --frozen pytest services/intake/tests/test_traces_api.py services/intake/tests/test_traces_clickhouse_repository.py services/intake/tests/test_spans_schemas.py services/intake/tests/test_atif_v17.py services/intake/tests/integration/spans/test_chat_completions_ingest.py services/intake/tests/integration/spans/test_atif_ingest.py services/intake/tests/integration/spans/test_traces_read.py -q`
- `uv run --frozen pytest sdk/python/nemo-platform/tests/api_resources/intake/test_traces.py -q`
- `uv run --frozen ruff check services/intake/src/nmp/intake services/intake/tests`
- `pnpm --filter @nemo/sdk test`
- `pnpm gen:check`


BrianNewsom:

services/intake/src/nmp/intake/spans/domain.py
```diff
@@ -82,6 +82,82 @@ class SpanListFilter(BaseModel):
...
+    workspace: str
+    session_id: str
+    source_format: str
+    name: str | None = None
+    input: str | None = None
+    output: str | None = None
+    environment: str | None = None
+    tags: list[str] | None = None
+    metadata: dict[str, Any] | None = None
+    project: str | None = None
```
shanaiabuggy: These are declared on the domain model but:
- Not selected by the rollup SQL in trace_repository.py
- Not populated in Trace.from_domain (the response schema doesn't have them anyway)

shanaiabuggy: Also do we want to promote environment / tags / metadata to the span attribute catalog?

services/intake/src/nmp/intake/spans/trace_repository.py
```diff
@@ -0,0 +1,530 @@
...
+            {root_alias}.source_format AS source_format,
+            nullIf({root_alias}.external_span_id, '') AS root_span_id,
+            nullIf({root_alias}.name, '') AS name,
+            nullIf({root_alias}.input, '') AS input,
+            nullIf({root_alias}.output, '') AS output,
+            {root_alias}.start_time AS started_at,
+            nullIf({root_alias}.end_time, {_ZERO_DATETIME_SQL}) AS ended_at,
+            {root_alias}.attributes_string AS root_attributes_string,
+            {root_alias}.event_ts AS ingested_at
+        FROM {table} AS {root_alias} FINAL
```
shanaiabuggy: My agent says that `FINAL` is expensive at scale. Alternative is to replace FINAL with argMax-based deduplication. The idea is to do the "pick the latest version per row" work inside the query as a normal aggregation, rather than asking ClickHouse to merge the underlying parts at query time.

shanaiabuggy (approved):

services/intake/src/nmp/intake/spans/domain.py
BrianNewsom: Good catch. i think we should just remove these from traces and should not add to spans. We explicitly removed environment, etc.

services/intake/src/nmp/intake/spans/trace_repository.py
BrianNewsom: Good point.. I did some perf testing and it wasn't far off. But I think this is likely long term more performant.

PR #12, opened 2026-05-21
fix(agents): Fix calculator agent by bumping to latest version of NAT

matthewgrossman:
## Description

Restores the `calculator-agent` example end-to-end after a regression introduced when `plugins/nemo-agents` bumped its NAT floor to `1.7.0rc1`. Without this PR, `nemo agents evaluate run --agent calculator-agent` scores `accuracy = 0.00` against the deployed agent, with every `generated_answer` extracted as `null`.

Two stacked upstream NAT bugs are at play; this PR fixes the example by bumping NAT to a release where the wire-format issue is resolved, and by toggling a NAT config flag that bypasses the parser issue.

### Bug 1 — `/generate/full` wire format regressed in NAT 1.7

NAT 1.6 emitted one `data: {"value": "<answer>"}\n\n` line per request from `/generate/full`. The NAT eval client at `nat.plugins.eval.runtime.remote_workflow.py` extracts answers via `chunk_data.get("value")`, and `test_remote_evaluate.py`'s server fixture documents that contract.

[NVIDIA/NeMo-Agent-Toolkit#1851](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1851) (token-streaming support for `react_agent`) added a `_stream_fn` that yields `ChatResponseChunk` (OpenAI envelope shape). The chunk's full envelope was emitted on the wire, no top-level `value` field, and the eval client extracted `None`.

Fixed upstream in [NVIDIA/NeMo-Agent-Toolkit#1980](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1980), shipped in `nvidia-nat-core==1.7.0rc3`. This PR bumps every NAT pin to that release.

### Bug 2 — NAT 1.7 ReAct text parser rejects trailing prose

NAT 1.7's `react_agent` text-mode parser greedy-captures everything to end-of-string when the LLM emits tool-call JSON followed by prose (e.g. Nemotron-3 emits `{"numbers":[50,35]}\n\nThen we get ...`). The full prose-suffixed string is then handed to the tool's Pydantic input schema, which rejects it as a string-instead-of-list, the agent retries 3× and gives up, and the eval scores 0 even with the wire format fixed.

NAT supports a config flag `use_native_tool_calling: true` that bypasses the regex parser entirely — `bind_tools()` is called on the LLM and structured `tool_calls` are read straight off the OpenAI-compatible chat completions API. This is the supported NAT path for tool-calling-capable models; it is not a workaround.

This PR sets the flag in `calculator-agent.yml`. The flag is independent of the wire-format fix; both are needed for a non-zero score against Nemotron-3 specifically.

## Changes

| File | Change |
|---|---|
| `pyproject.toml` | `nvidia-nat-atif` floor → `>=1.7.0rc3,<1.8` |
| `plugins/nemo-agents/pyproject.toml` | All six `nvidia-nat-*` pins → `>=1.7.0rc3,<1.8` |
| `plugins/nemo-agents/examples/calculator-agent/pyproject.toml` | `nvidia-nat-core>=1.7.0rc3,<1.8` |
| `packages/nemo_platform/pyproject.toml` | Generated `bundle-package` tables bumped in lockstep |
| `plugins/nemo-agents/examples/calculator-agent/src/calculator_agent/calculator-agent.yml` | `use_native_tool_calling: true` |
| `uv.lock` | Regenerated; all six NAT packages locked to `1.7.0rc3` |

### Why `<1.8`

`>=1.7.0rc3` alone matches future `1.8.0aXXX` nightlies on PyPI (today: `1.8.0a20260521`). The `<1.8` upper bound excludes those without preventing future `1.7.0rc4` / `1.7.0` / `1.7.x` patch releases from being picked up automatically.

## Verification

End-to-end on a clean checkout of this branch:

```bash
make bootstrap
source .venv/bin/activate
nemo setup
export NEMO_DEFAULT_MODEL=nvidia-nemotron-3-super-120b-a12b
uv pip install -e plugins/nemo-agents/examples/calculator-agent
nemo agents create --name calculator-agent \
  --agent-config plugins/nemo-agents/examples/calculator-agent/src/calculator_agent/calculator-agent.yml
nemo agents deploy --agent calculator-agent --wait --timeout 240
nemo agents evaluate run \
  --eval-config plugins/nemo-agents/examples/calculator-agent/src/calculator_agent/calculator-eval.yml \
  --agent calculator-agent
```

| State | Score | `generated_answer` shape |
|---|---|---|
| `main` (NAT 1.7.0rc1) | `0.00` | `null` for every item |
| This branch (NAT 1.7.0rc3 + `use_native_tool_calling`) | `0.83` | full final answers, all three items |

Wire shape probed directly against the deployed agent on this branch:

```
$ curl -s http://127.0.0.1:<port>/generate/full \
    -H 'content-type: application/json' \
    -d '{"input_message":"What is 3 times 7?"}' | grep '^data:'
data: {"value": " 21"}
```

`uv lock` resolves all six `nvidia-nat-*` packages to exactly `1.7.0rc3`; no 1.8 alphas leak in.


matthewgrossman:

mckornfield (approved):
just need a license update (and also... I wonder if that workflow has to stay the same now that we're OS lol)

tylersbray (approved):
lgtm

PR #14, opened 2026-05-21
feat(agents): add `nemo agents package` container packaging command

arpitsardhana:
Introduces a render → validate → build → publish pipeline for packaging NAT agents as OCI container images, driven by flags on a single command.

- New `container/` module: Jinja2 Dockerfile renderer (template.py), python-on-whales builder (builder.py) and publisher (publisher.py), structural config validator (validator.py), OCI metadata extractor (metadata.py), and a sample sandbox policy (openshell_policy.yaml).
- CLI: registers `nemo agents package` with `--no-build`, `--publish`, `--registry`, `--format docker|whl`, `--platform`, plus reproducibility knobs (`--nat-version`, `--base-image-*`, `--python-version`) and security defaults (`--allow-root` opt-out, generated `.dockerignore`).
- pyproject: declares `[container]` optional extra (jinja2, python-on-whales) and adds jinja2 to the test extra.
- Tests: 84 unit tests covering render, build, publish, validate, metadata, and CLI flag-combination handling.
- Example: minimal react_agent workflow (`examples/hello_world.yaml`).

## Summary by CodeRabbit

* **New Features**
  * Add `nemo agents package` CLI to render → validate → build → (optional) publish container images with project/config modes, image tagging, and push.
* **Documentation**
  * Detailed packaging docs: pipeline stages, CLI examples, flags, tagging/label conventions, validation checks, Dockerfile strategies, and example image/Dockerignore/config.
* **Chores**
  * New optional `container` extras for packaging tooling.
* **Bug Fixes**
  * Validator reports warnings alongside errors and improves YAML/file error handling.
* **Tests**
  * Extensive unit and end-to-end tests for rendering, metadata, validation, build, publish, and CLI flows.

### DIFF
--- a/plugins/nemo-agents/README.md
+++ b/plugins/nemo-agents/README.md
@@ -10,6 +10,7 @@ Agents are NAT workflow YAML files. The plugin provides:
 - **Gateway** — reverse-proxy agent traffic through `/apis/agents/…/-/…`
 - **CLI** — `nemo agents` subcommand for platform-managed workflows
 - **Evaluation** — delegate to `nat eval` against live agent endpoints
+- **Packaging** — containerize agents with a single `nemo agents package` command that progressively renders, builds, and publishes
 
 ---
 
@@ -224,6 +225,268 @@ The job pre-flights every LLM `model_name` against
 `sdk.inference.virtual_models.retrieve` before invoking `nat eval`, so a
 missing or mistyped model fails fast with a message naming the model and
 suggesting recovery options instead of an opaque subprocess error.
+
+---
+
+## Packaging command — containerize agents as Docker images
+
+The plugin ships a single `package` command that encapsulates the render →
+validate → build → publish pipeline.  Flags control how far the pipeline
+runs, so one command covers the entire inner-loop → outer-loop transition.
+The command works locally — no running platform is required.
+
+| Requirement | Notes |
+|---|---|
+| Docker | A running Docker daemon (Docker Desktop / Podman / etc.) |
+| `jinja2` | `uv pip install 'nemo-agents-plugin[container]'` |
+| `python-on-whales` | included in the `[container]` extra |
+
+### Progressive pipeline
+
+| Invocation | Stages run | Output |
+|---|---|---|
+| `package --no-build` | render | `Dockerfile` + `.dockerignore` |
+| `package` *(default)* | render → validate → build | Local Docker image |
+| `package --publish --registry <url>` | render → validate → build → publish | Local image + registry push |
+
+Validation always runs before a build unless `--skip-validation` is passed.
+`--no-build` skips the build (and therefore validation) and only emits files.
+
+### `package` — render, build, and publish in one command
+
+Render-only (inspect or edit the Dockerfile before building):
+
+```bash
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
+    --no-build
+```
+
+Output:
+```
+Dockerfile written to examples/Dockerfile
+.dockerignore written to examples/.dockerignore
+```
+
+Build (default — render + validate + build):
+
+```bash
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
+    --tag my-agent:1.0
+```
+
+Output:
+```
+Building image 'my-agent:1.0' from context examples ...
+Successfully built my-agent:1.0
+Image ready: my-agent:1.0
+```
+
+Full pipeline — build and publish in one call:
+
+```bash
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
+    --tag my-agent:1.0 \
+    --publish --registry nvcr.io/my-org
+```
+
+Output:
+```
+Building image 'my-agent:1.0' from context examples ...
+Successfully built my-agent:1.0
+Image ready: my-agent:1.0
+Tagging my-agent:1.0 -> nvcr.io/my-org/my-agent:1.0
+Pushing nvcr.io/my-org/my-agent:1.0 ...
+Successfully pushed nvcr.io/my-org/my-agent:1.0
+Published: nvcr.io/my-org/my-agent:1.0
+```
+
+**With an existing Dockerfile** (skip the render stage entirely):
+
+```bash
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --dockerfile examples/Dockerfile \
+    --nat-version 1.5.0 \
+    --tag my-agent:1.0
+```
+
+**Project mode** — if the agent ships inside a Python project, pass `--pyproject`:
+
+```bash
+nemo agents package \
+    --agent configs/agent.yaml \
+    --pyproject pyproject.toml \
+    --nat-version 1.5.0
+```
+
+### Flag reference
+
+**Pipeline control:**
+
+| Flag | Default | Description |
+|---|---|---|
+| `--no-build` | `False` | Stop after render; emit Dockerfile + `.dockerignore` only |
+| `--publish` | `False` | After building, tag and push to `--registry` |
+| `--registry`, `-r` | *(none)* | Remote registry URL (required when `--publish` is set) |
+| `--push-tag` | `<registry>/<tag>` | Override the fully-qualified remote tag |
+
+**Source inputs:**
+
+| Flag | Default | Description |
+|---|---|---|
+| `--agent`, `-c` | *(required)* | Path to the NAT workflow YAML |
+| `--pyproject` | *(none)* | Path to `pyproject.toml` (enables project mode) |
+| `--format` | `docker` | Packaging format. Only `docker` (Jinja2 Dockerfile) is implemented; `whl` is reserved for future wheel-based builds and is rejected at flag-validation time. |
+| `--dockerfile` | *(render on-the-fly)* | Use an existing Dockerfile instead of rendering |
+| `--template` | built-in | Path to an external Jinja2 Dockerfile template |
+
+**Build options:**
+
+| Flag | Default | Description |
+|---|---|---|
+| `--tag`, `-t` | `<agent-name>-<agent-id>:<agent-version>` | Image tag |
+| `--platform` | local daemon's native platform | Target platform (e.g. `linux/amd64`). At most one value — multi-arch builds via buildx are not yet implemented and are rejected. |
+| `--nat-version` | `$NAT_VERSION` env | NAT package version to install |
+| `--output`, `-o` | `<config-dir>/Dockerfile` | Where to write Dockerfile (with `--no-build`) |
+| `--skip-validation` | `False` | Bypass `validate_agent_config` before build |
+
+**Hardening overrides:**
+
+| Flag | Default | Description |
+|---|---|---|
+| `--allow-root` | `False` | Disable non-root `USER` hardening |
+| `--no-ignore` | *(generates by default)* | Skip `.dockerignore` generation |
+
+**OCI labels:**
+
+| Flag | Default | Description |
+|---|---|---|
+| `--agent-version` | from pyproject or `YY.MM.DD` | Override agent version label |
+| `--agent-author` | from `git config user.name` | Override agent author label |
+
+### Full example — inspect, build, publish
+
+```bash
+# 1. Render the Dockerfile so you can review it
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
+    --agent-version 1.0.0 \
+    --no-build
+
+# 2. (Optional) Edit examples/Dockerfile, then build against the edited file
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --dockerfile examples/Dockerfile \
+    --nat-version 1.5.0 \
+    --tag my-react-agent:1.0.0
+
+# 3. Build & publish in one step (when skipping the review step)
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
+    --tag my-react-agent:1.0.0 \
+    --publish --registry nvcr.io/my-org
+```
+
+### Image tagging convention
+
+When `--tag` is not provided, the image tag is computed automatically as:
+
+```
+<agent-name>-<agent-id>:<agent-version>
+```
+
+Each component is resolved through a fallback chain:
+
+| Component | Resolution order |
+|---|---|
+| **agent-name** | `pyproject.toml` `[project].name` → config file stem (e.g. `react-agent` from `react-agent.yml`) |
+| **agent-version** | `--agent-version` flag → `pyproject.toml` `[project].version` → today's date as `YY.MM.DD` |
+| **agent-id** | Truncated (12-char) SHA-256 hash of the config YAML content (+ `pyproject.toml` content when present) |
+
+Examples:
+
+| Scenario | Tag |
+|---|---|
+| Standalone `react-agent.yml`, no flags | `react-agent-f7e8d9c0b1a2:26.04.10` |
+| With pyproject (`name=calculator`, `version=2.3.0`) | `calculator-a1b2c3d4e5f6:2.3.0` |
+| With `--agent-version 1.0.0` override | `react-agent-f7e8d9c0b1a2:1.0.0` |
+
+The agent ID is **content-addressable** — identical config (and pyproject) content
+always produces the same ID. Changing any line in either file produces a
+different ID, giving every build a traceable fingerprint.
+
+### OCI image labels
+
+Generated Dockerfiles include image labels that follow the
+[OCI Image Spec annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md).
+Standard OCI keys are used where a mapping exists; agent-specific metadata uses
+the `com.nemo.agent.*` namespace.
+
+**Standard OCI labels:**
+
+| Label | Value |
+|---|---|
+| `org.opencontainers.image.title` | Agent name (same as tag name component) |
+| `org.opencontainers.image.version` | Agent version (same as tag version component) |
+| `org.opencontainers.image.authors` | `--agent-author` → `git config user.name` → `"unknown"` |
+| `org.opencontainers.image.created` | Build timestamp (ISO 8601 / RFC 3339) |
+| `org.opencontainers.image.description` | `pyproject.toml` `[project].description` → `"{workflow._type} agent"` → `""` |
+| `org.opencontainers.image.revision` | `git rev-parse HEAD` → `""` |
+| `org.opencontainers.image.source` | `git remote get-url origin` → `""` |
+| `org.opencontainers.image.licenses` | `pyproject.toml` `[project].license` (SPDX expression) — omitted when absent |
+
+**Custom agent labels:**
+
+| Label | Value |
+|---|---|
+| `com.nemo.agent.id` | Content-addressable SHA-256 hash (12 chars) |
+| `com.nemo.agent.framework` | `"nemo_agent_toolkit"` when config has a `workflow` key, else `"unknown"` |
+| `com.nemo.agent.nat-version` | NAT version used at build time |
+| `com.nemo.agent.contract-version` | Packaging format version (`"1.0"`) |
+
+### Agent config validation
+
+The `package` command validates the agent config before building (skip with
+`--skip-validation`). Validation checks:
+
+- File is valid YAML that parses to a mapping (dict)
+- Top-level `workflow` key exists and is a mapping
+- `workflow._type` is a known NAT agent type (`react_agent`, `tool_calling_agent`, `reasoning_agent`, `rewoo_agent`)
+- Every name in `workflow.tool_names` is defined in `functions` or `function_groups`
+- `workflow.llm_name` is defined in `llms`
+
+Multiple errors are collected and reported together rather than failing on the first.
+
+### Security defaults
+
+Generated Dockerfiles apply several hardening measures by default:
+
+| Default | Override |
+|---|---|
+| Non-root `USER agent` (uid 1000) | `--allow-root` |
+| `apt-get --no-install-recommends` | *(none — always applied)* |
+| `rm -rf /var/lib/apt/lists/*` after install | *(none — always applied)* |
+| `.dockerignore` excludes `.env`, `.git/`, `*.pem`, `credentials.json`, `__pycache__/`, `.venv/`, `node_modules/` | `--no-ignore` |
+
+### Rendering modes
+
+The Dockerfile template has two modes, selected automatically:
+
+| Mode | Trigger | Install strategy |
+|---|---|---|
+| **Config-only** | No `--pyproject` | `uv pip install "nvidia-nat[most]==${NAT_VERSION}"` |
+| **Project** | `--pyproject` provided | `uv sync` + `uv pip install .` (installs the project and its deps) |
+
+In project mode the entire project directory is the build context and the config
+path is resolved relative to the `pyproject.toml` parent directory.
 
 ---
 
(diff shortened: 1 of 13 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -197,6 +198,337 @@ def run(
...
+            None,
+            "--tag",
+            "-t",
+            help="Image tag.  Defaults to '<agent-name>-<agent-id>:<agent-version>'.",
+        ),
+        platform: Optional[list[str]] = typer.Option(
+            None,
+            "--platform",
+            help="Target platform (repeatable).  Default: linux/amd64 + linux/arm64.",
+        ),
```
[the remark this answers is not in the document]
arpitsardhana: Issue fixed

plugins/nemo-agents/src/nemo_agents_plugin/container/builder.py
```diff
@@ -0,0 +1,193 @@
...
+        return docker_build(
+            context_dir=context_dir,
+            dockerfile=tmp_dockerfile,
+            tag=tag,
+            build_args=build_args,
+        )
+    finally:
+        tmp_dockerfile.unlink(missing_ok=True)
+        if ignore_file is not None:
+            ignore_file.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
arpitsardhana: The data-loss bug the finding describes was real in the original PR but was fixed in commit de7a4e1d (which the user has already committed). The current implementation prevents it via the sentinel-header / Path | None return mechanism rather than the alternate (path, created_flag) tuple the finding proposes, but the user-facing contract is the same and is exercised by test_dockerignore_preserves_user_files_and_overwrites_its_own. Re-implementing the same guard a second way would only add lines without strengthening any invariant.

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -0,0 +1,336 @@
...
+
+    has_pyproject = pyproject is not None and pyproject.exists()
+
+    if has_pyproject:
+        assert pyproject is not None
+        try:
+            relative_config = agent_config.resolve().relative_to(pyproject.resolve().parent)
+        except ValueError:
+            relative_config = Path(agent_config.name)
+        config_file_path = f"/workspace/{relative_config.as_posix()}"
```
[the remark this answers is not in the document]
arpitsardhana: this is fixed

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
```diff
@@ -0,0 +1,231 @@
...
+        return desc
+    try:
+        import yaml
+
+        data = yaml.safe_load(config_text)
+        if isinstance(data, dict):
+            wf = data.get("workflow", {})
+            if isinstance(wf, dict) and wf.get("_type"):
+                return f"{wf['_type']} agent"
+    except Exception:
```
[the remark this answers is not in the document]
arpitsardhana: this is fixed

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -0,0 +1,439 @@
...
+    ``dockerfile_escape`` filter is invoked explicitly at every label
+    interpolation site. ``StrictUndefined`` so a typo in an external
+    template fails at render time instead of producing a Dockerfile with
+    silently empty ``ARG`` / ``LABEL`` lines.
+    """
+    env = jinja2.Environment(
+        autoescape=False,
+        keep_trailing_newline=True,
+        undefined=jinja2.StrictUndefined,
+    )
```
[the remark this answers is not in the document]
arpitsardhana: This is not applicable here since there is no html rendered

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def _detect_framework(config_text: str) -> str:
+    try:
+        import yaml
+
+        data = yaml.safe_load(config_text)
+        if isinstance(data, dict) and "workflow" in data:
+            return "nemo_agent_toolkit"
+    except Exception:
```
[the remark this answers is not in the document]
arpitsardhana: this is fixed

plugins/nemo-agents/agent_container_commands_3964e748.plan.md
```diff
@@ -0,0 +1,509 @@
...
+
+| File | Coverage |
+|---|---|
+| `test_container.py` | 56 existing tests: render modes, secure defaults, `.dockerignore`, OCI labels, validation, metadata, build, publish, e2e pipeline, external template. **Plus new tests for `package` flag combinations, multi-arch, whl format.** |
+
+---
+
+## 1. The `package` Subcommand
+
+```
```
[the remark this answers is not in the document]
arpitsardhana: this is not valid anymore

plugins/nemo-agents/examples/Dockerfile
```diff
@@ -0,0 +1,58 @@
+ARG BASE_IMAGE_URL=nvcr.io/nvidia/base/ubuntu
+ARG BASE_IMAGE_TAG=22.04_20240212
```
mckornfield: guess we have another year, but any reason not to do 24.x for ubuntu?

plugins/nemo-agents/README.md
```diff
@@ -227,6 +228,268 @@ suggesting recovery options instead of an opaque subprocess error.
...
+`--no-build` skips the build (and therefore validation) and only emits files.
+
+### `package` — render, build, and publish in one command
+
+Render-only (inspect or edit the Dockerfile before building):
+
+```bash
+nemo agents package \
+    --agent examples/react-agent.yml \
+    --nat-version 1.5.0 \
```
SandyChapman, line 262: It's interesting that the nat-version isn't encoded in the workflow yaml.

plugins/nemo-agents/examples/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+ARG PYTHON_VERSION=3.13
+ARG NAT_VERSION=1.6.0
+FROM ${BASE_IMAGE_URL}:${BASE_IMAGE_TAG}
+ARG PYTHON_VERSION
+ARG NAT_VERSION
+
+COPY --from=ghcr.io/astral-sh/uv:0.8.15 /uv /uvx /bin/
```
mckornfield, line 10: this is a pretty old version of uv no?

plugins/nemo-agents/examples/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+    uv pip install --prerelease=allow "nvidia-nat[most]==${NAT_VERSION}" && \
+    chmod -R a+rX /opt/uv /workspace/.venv
+
+LABEL org.opencontainers.image.title="hello_world" \
+      org.opencontainers.image.version="26.04.17" \
+      org.opencontainers.image.authors="Arpit Singh (SW-CLOUD)" \
+      org.opencontainers.image.created="2026-04-17T23:50:42.868232+00:00" \
+      org.opencontainers.image.description="react_agent agent" \
+      org.opencontainers.image.revision="8d0654a8b8e1edccce579223272e31dcee1608cd" \
+      org.opencontainers.image.source="ssh://git@gitlab-master.nvidia.com:12051/aire/microservices/nmp.git" \
```
mckornfield: ruh roh

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
```diff
@@ -0,0 +1,315 @@
...
+    if pyproject is None or not pyproject.exists():
+        return {}
+    try:
+        import tomllib
+    except ModuleNotFoundError:
+        import tomli as tomllib  # type: ignore[no-redef]
+    try:
+        return tomllib.loads(pyproject.read_text(encoding="utf-8"))
+    except Exception:
+        return {}
```
mckornfield, line 101: will these not be error cases?

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
```diff
@@ -0,0 +1,315 @@
...
+        return datetime.now(UTC).isoformat()
+    return datetime.now(UTC).isoformat()
+
+
+def _detect_framework(config_text: str) -> str:
+    try:
+        import yaml
+
+        data = yaml.safe_load(config_text)
+        if isinstance(data, dict) and "workflow" in data:
```
mckornfield: seems like a pretty rudimentary detection but 🤷

plugins/nemo-agents/src/nemo_agents_plugin/container/publisher.py
```diff
@@ -0,0 +1,69 @@
...
+            ``<registry>/<local_tag>``.
+
+    Returns:
+        The remote image tag that was pushed.
+
+    Raises:
+        typer.Exit: On tag or push failure.
+    """
+    try:
+        from python_on_whales import docker  # type: ignore[unresolved-import]
```
mckornfield, line 41: curious why this import is used to check?

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -0,0 +1,449 @@
...
+    "uv_version": "NAT_UV_VERSION",
+}
+
+# -- Jinja2 template --------------------------------------------------------
+
+DOCKERFILE_TEMPLATE = (
+    f"""\
+{DOCKERFILE_SENTINEL}
+"""
+    + """\
```
mckornfield, line 96: 🤔 would this be better as a jinja file loaded in separately?

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -0,0 +1,449 @@
...
+
+_ENV_MAP: dict[str, str] = {
+    "base_image_url": "NAT_BASE_IMAGE_URL",
+    "base_image_tag": "NAT_BASE_IMAGE_TAG",
+    "python_version": "NAT_PYTHON_VERSION",
+    "nat_version": "NAT_VERSION",
+    "uv_version": "NAT_UV_VERSION",
+}
+
+# -- Jinja2 template --------------------------------------------------------
```
SandyChapman, line 90: Personal preference, but I like to see these templates in their own *.j2 file or similar.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -197,6 +198,397 @@ def run(
...
+            None,
+            "--nat-version",
+            envvar="NAT_VERSION",
+            help=(
+                "NAT release to install (e.g. '1.6.1').  Strongly recommended: "
+                "pin explicitly so image tags/labels/deps are reproducible.  "
+                "When omitted, a baked-in default is used and a warning is printed."
+            ),
+        ),
+        uv_version: Optional[str] = typer.Option(None, "--uv-version", envvar="NAT_UV_VERSION"),
```
mckornfield, line 309: seems like a pretty full up arglist, but don't know if a config file holding it would be any better

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -197,6 +198,397 @@ def run(
...
+        skip_validation: bool = typer.Option(
+            False, "--skip-validation", help="Bypass validate_agent_config before build."
+        ),
+        agent_version: Optional[str] = typer.Option(None, "--agent-version", help="Override agent version OCI label."),
+        agent_author: Optional[str] = typer.Option(None, "--agent-author", help="Override agent author OCI label."),
+        template: Optional[str] = typer.Option(
+            None, "--template", help="Path to an external Jinja2 Dockerfile template."
+        ),
+    ) -> None:
+        """Package a NAT agent — render → validate → build → publish.
```
mckornfield: should we avoid the unicode arrows?

plugins/nemo-agents/README.md
```diff
@@ -227,6 +228,268 @@ suggesting recovery options instead of an opaque subprocess error.
 
 ---
 
+## Packaging command — containerize agents as Docker images
+
+The plugin ships a single `package` command that encapsulates the render →
+validate → build → publish pipeline.  Flags control how far the pipeline
```
mckornfield, line 234: is this a standard, vs. having "build" and "push" as separate steps? I'd rather do build and push together than one whole thing, but maybe that's because I'm used to docker/helm

plugins/nemo-agents/src/nemo_agents_plugin/container/publisher.py
```diff
@@ -0,0 +1,69 @@
...
+    except ImportError:
+        typer.echo(
+            "Error: 'python-on-whales' is required for publishing images.  "
+            "Install it with:  pip install 'nemo-agents-plugin[container]'",
+            err=True,
+        )
+        raise typer.Exit(code=1)
```
SandyChapman, line 48: I wonder if just using subprocess to invoke `docker` is simpler? This way we don't require another external dep for nemo-agents?

SandyChapman (approved):

plugins/nemo-agents/src/nemo_agents_plugin/container/validator.py
```diff
@@ -0,0 +1,115 @@
...
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import yaml
+
+_KNOWN_WORKFLOW_TYPES = frozenset(
```
benmccown, line 17: Flagging a concern that Claude caught. Not sure if it's worth addressing or if it's just a nitpick.

---

**Hardcoded workflow type list.** 

`validator.py:17-24` only knows about `react_agent`, `tool_calling_agent`, `reasoning_agent`, `rewoo_agent`. NAT can register more types via its plugin system (and new ones land regularly). Anyone using a valid-but-unlisted type will see a validation failure and have to `--skip-validation` to proceed. The "structured error instead of opaque NAT traceback" intent is good, but the implementation goes the wrong way — it's a denylist of "things we recognize" disguised as an allowlist of "things NAT supports." Options:
- Query NAT's registered workflow types at runtime
- Soft-warn ("workflow type X not in known list — proceeding") instead of hard-failing
- Demote this check to opt-in (`--strict-validation`) so the default behavior doesn't break on valid configs

benmccown (approved):

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: This is an example Dockerfile.Base image , uv and NAT version all work together.
I will correct the default value though

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: Same as above, this is meant for demonstraton

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: Fixed it

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
arpitsardhana: Added Decode specific exception instead of general exception catch

plugins/nemo-agents/src/nemo_agents_plugin/container/publisher.py
arpitsardhana: python on whales get installed when the particular plugin is installed ie "pip install agents[containers]"

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
arpitsardhana: This avoid bit of file management tax. It is template and handled as string efficiently. Also wanted to avoid someone mucking up template by editing file

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
arpitsardhana: It was intentional.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
arpitsardhana: Each flag has user driven justification. Essential one design decision from RFC was to have one command to handle render/build/publish so led to proliferation of flags to customize and set up preference of main versions during dockerfile render

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
arpitsardhana: fixed

plugins/nemo-agents/README.md
arpitsardhana: The command allows build publish as one whole thing.

plugins/nemo-agents/README.md
arpitsardhana: This is standard for open source libraries. That version is available when it is queried but it is not part of config files, since they are portable across versions

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
arpitsardhana: It allows easier string operations , avoid file/path mgmt ops and wanted to avoid unpredictability in behaviour when a packaged file is mucked around.

plugins/nemo-agents/src/nemo_agents_plugin/container/publisher.py
arpitsardhana: The dep is already declared in the optional [container] extra (not a hard requirement), and builder.py uses it too

plugins/nemo-agents/src/nemo_agents_plugin/container/validator.py
arpitsardhana: This is fair point. I introduced and demoted it to warning

plugins/nemo-agents/examples/Dockerfile
```diff
@@ -0,0 +1,59 @@
+# Managed by `nemo agents package` — safe to delete if you take ownership.
+ARG BASE_IMAGE_URL=nvcr.io/nvidia/base/ubuntu
```
gabwow, line 2: What are the pro's and con's for using a full ubuntu base image vs alpine?

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
```diff
@@ -0,0 +1,320 @@
...
+        return desc
+    try:
+        import yaml
+
+        data = yaml.safe_load(config_text)
+        if isinstance(data, dict):
+            wf = data.get("workflow", {})
+            if isinstance(wf, dict) and wf.get("_type"):
+                return f"{wf['_type']} agent"
+    except Exception:
```
gabwow: why are we catching bare exceptions here?

plugins/nemo-agents/README.md
```diff
@@ -227,6 +228,268 @@ suggesting recovery options instead of an opaque subprocess error.
...
+| `.dockerignore` excludes `.env`, `.git/`, `*.pem`, `credentials.json`, `__pycache__/`, `.venv/`, `node_modules/` | `--no-ignore` |
+
+### Rendering modes
+
+The Dockerfile template has two modes, selected automatically:
+
+| Mode | Trigger | Install strategy |
+|---|---|---|
+| **Config-only** | No `--pyproject` | `uv pip install "nvidia-nat[most]==${NAT_VERSION}"` |
+| **Project** | `--pyproject` provided | `uv sync` + `uv pip install .` (installs the project and its deps) |
```
gabwow: Drift with what's actually done by the template, which only runs uv pip install "# ``uv sync`` is deliberately not used because it honors..."

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -0,0 +1,449 @@
...
+    """Return the ``nemo-agents-plugin`` package version."""
+    from importlib.metadata import PackageNotFoundError, version
+
+    try:
+        return version("nemo-agents-plugin")
+    except PackageNotFoundError:
+        return "0.0.0"
+
+
+def render_dockerignore(output_dir: Path) -> Path | None:
```
gabwow, line 437: A point codex raised that looks valid:

build_agent_image() treats .dockerignore as a temp artifact during build-mode packaging:

It calls render_dockerignore(context_dir).
If that function returns a path, build_agent_image() stores it in ignore_file.
In finally, it does ignore_file.unlink(missing_ok=True).
The subtle part is what render_dockerignore() returns:

If .dockerignore is user-owned (no sentinel), it returns None (so nothing gets deleted).
If .dockerignore is plugin-managed (has sentinel), it returns the file path — even when that file already existed before this run.
So a pre-existing plugin-managed file can be deleted at cleanup time, even if the repo intentionally wanted to keep it.

Example scenario
Repo already contains .dockerignore with plugin sentinel header.
You run nemo agents package (build mode, not --no-build).
Build code receives that path as ignore_file.
finally unlinks it.
Result: .dockerignore is gone after the command.
Why this feels risky
It mixes two cases:

“file created temporarily for this run” (safe to delete), and
“file already existed before command” (usually should preserve).

gabwow (approved):
Approving but I want you to think how tightly coupled we want this implementation to NAT. Is it too early to think about what a harness protocol would look like?

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: This comes since above is set as default BASE_IMAGE. It is variable which user can specify.
Alpine is smaller it will require lot of dependency installation. Above base image is cuda enabled.  So, it allows to do lot of operations without having to separately install all the dependencies which are tricky to install on fly

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: Updated the default ubuntu version

plugins/nemo-agents/examples/Dockerfile
arpitsardhana: The other aspect is: this is nvidia blessed https://catalog.ngc.nvidia.com/orgs/nvidia/teams/base/containers/ubuntu/tags?version=jammy-20260217 

We dont have official alpine images. 
this is good place for default, we can iterate and replace with thin images preferably python slim images

plugins/nemo-agents/README.md
arpitsardhana: Thanks for catching the issue. This is fixed. Removed the uv sync which can interfere with install

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
arpitsardhana: Thanks for catching it. 
Added a flag which persist the presence of file and ensures it is not unlinked once operations are done

plugins/nemo-agents/src/nemo_agents_plugin/container/metadata.py
arpitsardhana: Fixed the exception to catch specific exception instead of broad exception handling

arpitsardhana:
> Approving but I want you to think how tightly coupled we want this implementation to NAT. Is it too early to think >about what a harness protocol would look like?

@gabwow Things have certainly changed between when feature request was made versus now. We dont have a view yet for what a generic harness protocol looks like. We can address it iteratively across our agent plugin where deploy also need to be decoupled from framework

arpitsardhana:
@coderabbitai resume

PR #16, opened 2026-05-21
chore: refactor data-designer plugin tests

mikeknep:
### What's here

I find integration-level tests, in which we exercise the system through end user-facing interfaces as much as possible, provide greater confidence that the plugin is working as intended, especially nowadays when agents feel perfectly at liberty to change source code **and** (usually unit) test code when implementing some feature. This PR mostly reorganizes and adds tests to increase the integration-level coverage. There's still more work to be done here (always!), but this is a step in the right direction and should reduce the feeling of needing to manually run some tests as a sanity check at least a little bit.

### Coverage stats

There are two important packages to pay attention to:
- `plugins/nemo-data-designer/src/nemo_data_designer_plugin`: where the plugin entry-point interfaces are implemented
- `packages/data_designer_nemo/src/data_designer_nemo`: the "pure-library" glue code that knows how to use the Data Designer library in the NeMo Platform context

The plugin integration test coverage of these packages is:

| Branch | Coverage |
| --- | --- |
| main | 68.58% |
| this PR | 78.71% (+10 pp) |


plugins/nemo-data-designer/src/nemo_data_designer_plugin/sdk/errors.py
```diff
@@ -18,3 +30,27 @@ class DataDesignerPreviewError(DataDesignerClientError):
...
+    try:
+        response.read()
+    except Exception:
+        pass
+
+    detail = response.text
+    try:
+        body = response.json()
+    except Exception:
+        body = None
```
mckornfield: nit- could this just be None before? and do something  like

```
from contextlib import suppress

body = None
with suppress(Exception):
    body = response.json()
```

plugins/nemo-data-designer/tests/integration/test_job_sdk.py
```diff
@@ -0,0 +1,361 @@
...
+        dd_client = DataDesignerResource(ctx.sdk)
+        job_resource = dd_client.get_job_resource(_JOB_NAME, workspace="default")
+
+        assert isinstance(job_resource, DataDesignerJobResource)
+
+        job = job_resource.get_job()
+        assert job["name"] == _JOB_NAME
+
+
+@pytest.mark.integration
```
mckornfield: can we just add this mark to the whole file?

plugins/nemo-data-designer/tests/unit/test_errors.py
```diff
@@ -0,0 +1,87 @@
...
+    assert detail == "Job result not found"
+
+
+def test_extract_http_error_info_falls_back_to_raw_body_when_not_json() -> None:
+    exc = _make_status_error(500, body="<html>oh no</html>")
+
+    status_code, detail = extract_http_error_info(exc)
+
+    assert status_code == 500
+    assert detail == "<html>oh no</html>"
```
mckornfield, line 41: lol

mckornfield (approved):
some nice tests there

plugins/nemo-data-designer/src/nemo_data_designer_plugin/sdk/errors.py
mikeknep: I do love `contextlib.suppress` as well. Done in 1204064

plugins/nemo-data-designer/tests/integration/test_job_sdk.py
mikeknep: TIL! Thanks! This has been annoying to set on every test. Done in d3e60ac

PR #22, opened 2026-05-22
build(studio): generate SDK on install instead of committing output

steramae-nvidia:
As we move to a public release, I figured this would be a good change to make to prevent thrashing the repo with lines changed. We now defer the SDK generation to a local post install script. All SDK generated code is now git ignored 

## Summary
- The orval-generated SDK tree under `web/packages/sdk/generated/` (1,261 files / 13 MB / ~100k LOC) is now gitignored and regenerated on `pnpm install`. PRs that touch OpenAPI specs no longer drown reviewers in generated-file diffs.
- A content-hash sentinel (`generated/.input-hash`) makes the install hook idempotent: ~2s on cache hit vs ~45s cold. The hash covers the input YAMLs, orval version, and generator source — any of those changing invalidates the cache.
- New CI job `web-sdk-gen-deterministic` runs `pnpm gen:check`, which generates twice and asserts byte-identical output. Catches template/post-process changes that would produce non-reproducible builds.

### Mental model for teammates
- `pnpm install` regenerates the SDK only if inputs changed (transparent).
- `pnpm gen` or `pnpm --filter @nemo/sdk gen:all` → force regen (manual).
- `pnpm --filter @nemo/sdk gen:if-stale` → what the install hook calls (skip if hash matches).

### Out of scope
- Studio package has pre-existing typecheck failures from `react-hook-form ^7.58.1` resolving to 7.76.0 on main. They're independent of this PR — `git checkout main` reproduces them — and tracked separately.

## Test plan
- [x] `rm -rf packages/sdk/generated && pnpm install --frozen-lockfile` regenerates the tree (~45s).
- [x] Second `pnpm install` short-circuits via hash (~2s, prints "SDK is up to date").
- [x] `pnpm gen:check` passes (deterministic output verified).
- [x] `pnpm --filter @nemo/sdk test` passes (26/26).
- [x] `pnpm --filter @nemo/scripts typecheck` passes.
- [ ] CI: `web-sdk-gen-deterministic` and all other studio-ci jobs pass.
- [ ] Confirm fresh clone on a teammate's machine: `git clone` → `pnpm install` → SDK present and usable.


web/packages/sdk/package.json
```diff
@@ -5,7 +5,8 @@
   "description": "A collection of generated types and hooks for NeMo Platform.",
   "scripts": {
     "gen": "tsx ./orval/generate.ts",
-    "gen:all": "tsx ./generateAll.ts",
+    "gen:all": "tsx ./generateAll.ts --force",
```
aray12: nit: I think it makes more sense to have `gen:all` and `gen:all:force`

walston (approved):

aray12 (approved):
Might be cool to do per-service caching so we could partial regenerations. But this looks good overall

web/packages/sdk/package.json
steramae-nvidia: Sure! I can make that change

PR #30, opened 2026-05-22
fix: avoid encoded PyPI summary for nemo-platform-plugin

svvarom:
## Summary

PyPI currently renders the `nemo-platform-plugin` short summary as an encoded metadata header:

`=?utf-8?q?Public_plugin_contract_for_the_NeMo_Platform_=E2=80=94_the_only_package_plugin_authors_need=2E?=`

That comes from the non-ASCII em dash in the package description. The wheel build rewrites `METADATA`, and Python's email header serializer encodes that header instead of keeping it readable.

## Change

Replace the em dash with an ASCII hyphen in the package description so future wheels publish a readable `Summary` field.

## Validation

Not run; metadata-only text change.


crookedstorm (approved):
🎉

PR #32, opened 2026-05-22
feat(studio): Studio Tools in Chat

steramae-nvidia:
Adds rich interactive content in Studio chat through web search chat. This optional tool lets any tool-enabled model use the web-search endpoint on the Studio service for querying a headless duckduckgo. Results are forwarded to the user.  

ns local aire nvidia com_5173_workspaces_default_inference-providers(iPhone 14 Pro Max)


steramae-nvidia:
Closing to reduce scope

PR #33, opened 2026-05-22
fix(agents): pin boto3/langchain-aws to fix pip resolution-too-deep (ASTD-164)

marcusds:
## Summary

- `pip install nemo-platform[services]` fails during dependency resolution with `Dependency resolution exceeded maximum depth`. The resolver backtracks across the full `boto3` release history while resolving `langchain-aws` (pulled in via `nvidia-nat-langchain` in `nemo-agents-plugin`).
- The workspace already constrains these via uv `override-dependencies` in the root `pyproject.toml`, but those overrides don't reach the published wheel. This PR carries the same bounds in `plugins/nemo-agents/pyproject.toml` so they propagate to the `nemo-platform[services]` extra:
  - `langchain-aws==1.1.0`
  - `boto3>=1.40.46,<1.40.62`
  - `botocore>=1.40.46,<1.40.62`
- The `<1.40.62` upper bound matches `aiobotocore 2.25.1`'s cap (required by `aioboto3 15.5.0` in files-service). A wider pin lets pip backtrack on `aiobotocore` + `wrapt` + `yarl` simultaneously and time out.

Linear: https://linear.app/nvidia/issue/ASTD-164

## Test plan

- [x] Build wheel with `uv build --wheel packages/nemo_platform`.
- [x] Create fresh venv on Python 3.12, `pip install --dry-run dist/nemo_platform-*.whl[services]` — resolution completes cleanly (334 packages, exit 0). Before the fix this aborted with `resolution-too-deep`.
- [x] `uv lock --check` passes.
- [ ] Verify Linux wheel resolves cleanly (CI).


mckornfield (approved):
each day the deptree grows

PR #34, opened 2026-05-25
fix(jobs): read subprocess secrets from value field

SandyChapman:
## Summary
- Update subprocess secret injection to read the current secrets access response field, value
- Add coverage that rejects the old data-shaped response so this does not silently regress

## Context
The subprocess executor was introduced in NVIDIA-NeMo/Platform commit aaa45abf6 (#259) when secret access responses used data. The secrets API later renamed data to value in commit c68ea061e (#466), but the subprocess runtime consumer was not updated.

## Validation
- uv run --frozen ruff check services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess_runtime.py services/core/jobs/tests/controllers/test_subprocess_runtime.py
- uv run --frozen --extra cpu ty check services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess_runtime.py services/core/jobs/tests/controllers/test_subprocess_runtime.py
- uv run --frozen pytest services/core/jobs/tests/controllers/test_subprocess_runtime.py -q


stefan-kickoff (approved):
I'm surprised anything was working without this?

PR #37, opened 2026-05-25
chore(guardrails): speed up IGW middleware integration test fixtures

albcui:
Big idea: convert the heavy ASGI stack from function-scoped to module-scoped, so that they can be re-used across tests. The heavy stuff includes:
- SQLite database initialization + migrations
- 3 services: (Models, Entities, IGW)
- FastAPI app
- default workspace + project seeding
- (for loopback) a fresh uvicorn thread per test

Results using `time -p ...`
| Configuration | main | this branch |
|---|---|---|
| sequential | ~26s | ~19s |
| xdist `--dist=loadfile -n 4` | ~17s | ~14s |

The main speedup is experienced when you run the tests sequentially. This is often the case when you are in a dev loop and just running locally with `pytest ...`. The change is less noticeable when you use `xdist`, which already parallelizes the per-test rebuild across workers, so module-scoping on top yields less benefits.

## Summary by CodeRabbit

* **New Features**
  * Added harness helper to create/track secrets and made per-harness workspace accessible.

* **Tests**
  * Integration harness now builds a module-scoped app context for faster, isolated runs.
  * Tests updated to use the dynamic harness.workspace instead of a static default.

* **Documentation**
  * Expanded test-harness README with module-scope guidance, xdist notes, and workspace usage.

* **Chores**
  * Reorganized fixtures and test helpers for clearer lifecycle and teardown behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/37?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/services/core/inference-gateway/src/nmp/core/inference_gateway/testing/fixtures.py
+++ b/services/core/inference-gateway/src/nmp/core/inference_gateway/testing/fixtures.py
@@ -3,39 +3,30 @@
 
 """Pytest fixtures for the IGW middleware test harnesses.
 
-Importing :func:`igw_plugin_harness` (or :func:`igw_loopback_harness`)
-into a test module — or re-exporting from a project ``conftest.py`` —
-registers the fixture for the surrounding scope. Both piggyback on
-``pytest_httpserver``'s function-scoped ``httpserver`` fixture so each
-test gets an isolated socket and clean handler state.
+Import :func:`igw_plugin_harness` or :func:`igw_loopback_harness` into a
+test module (or re-export from a project ``conftest.py``) to register
+the fixture. Both use ``pytest_httpserver``'s function-scoped
+``httpserver`` so each test gets a fresh mock-NIM socket.
 
-Scope split (AIRCORE-585):
+The ASGI stack is split across two scopes so each test only pays for
+what changes between tests:
 
-* :func:`_igw_app_context` — **module-scoped**. Wraps the expensive
-  ``create_test_client`` call (SQLite DB, FastAPI app + IGW + Models
-  services, dependency wiring, ``/health/ready`` polling, workspace +
-  project seeding) once per test file. The ``InferenceGatewayService``
-  config disables the periodic ``refresh_model_cache_task`` so it
-  cannot wake mid-test and re-populate the model cache with stale
-  rows from the previous test.
-* :func:`_igw_loopback_context` — **module-scoped**. Adds a uvicorn
-  thread + ``per_request_http_client`` dependency override + platform
-  base URL patch on top of the module app. Only entered when a test
-  asks for the loopback variant so plain modules don't pay the
-  uvicorn startup cost.
-* :func:`igw_plugin_harness` / :func:`igw_loopback_harness` —
-  **function-scoped**. Build a fresh :class:`IGWPluginHarness` per
-  test on top of the module context: pre/post global-cache resets,
-  per-test ``app.state.pending_post_response_tasks`` re-init,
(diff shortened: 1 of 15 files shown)


services/core/inference-gateway/src/nmp/core/inference_gateway/testing/fixtures.py
```diff
@@ -21,139 +40,224 @@
...
+
+    Two module-scope hazards are neutralised here:
+
+    1. The 3-second background ``refresh_model_cache_task``. ``on_startup``
+       reads ``refresh_model_cache_interval_sec`` from the module-level
+       config snapshot (captured at first import), so a ``service_configs``
+       override is too late. Patch the snapshot field to 0 *before*
+       entering ``create_test_client`` and ``on_startup`` never schedules
+       the loop.
+    2. The shared SDK HTTP client's ``aclose``. Plugins like
```
JashG, line 77: [non-blocking] In the test_client [here](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/packages/nmp_testing/src/nmp/testing/client.py#L367), should we set the HTTP client to a simple wrapper that overrides `aclose`? i.e.:

```
class _TestAsyncHttpClient(httpx.AsyncClient):
    async def aclose(self) -> None:
        pass

    async def _real_close(self) -> None:
        await super().aclose()
...
async_http_client = _TestAsyncHttpClient(transport=transport, base_url="http://testserver", timeout=pdp_timeout)
```

If it's only relevant to Guardrails, doing the override it in this fixture is fine too.

services/core/inference-gateway/src/nmp/core/inference_gateway/testing/harness.py
```diff
@@ -147,37 +162,102 @@ def _build(
...
             _model_cache=model_cache,
             _vm_cache=vm_cache,
             _cache_accessor=cache_accessor,
             _virtual_models=[],
+            _providers=[],
+            _secrets=[],
             **extra_fields,
         )
 
     def _cleanup(self) -> None:
```
JashG, line 176: [nit] Since the harness is now module-scoped, could `_cleanup` instead simply list and delete secrets/providers/etc. in `self.workspace`? This would cover cases where a resource is created without using the harness helpers.

JashG (approved):

services/core/inference-gateway/src/nmp/core/inference_gateway/testing/fixtures.py
albcui: Hmm, I'd rather isolate my changes to the IGW middleware test fixtures. Right now, `_build_app_context` acts as the caller that correctly restores the original `aclose` method on teardown.  We'd need a similar mechanism that wraps `_TestAsyncHttpClient`, and restores the `_real_close` as `aclose`,  which would be a bigger change to `nmp_testing`.

services/core/inference-gateway/src/nmp/core/inference_gateway/testing/harness.py
albcui: I'm just worried about stuff in `self.workspace` that we shouldn't delete... like seeded data, or data that was **meant** to be shared across tests, like a shared provider or something. This is the main benefit of the helper methods like `create_secret`, `add_provider`, etc... ofc anything created out-of-band via the sdk would not be tracked...

The downside is noted, maybe we can try to come up with a better tracking method? However, I do want to keep the ability to track-and-delete entities created by each test explicitly.

PR #46, opened 2026-05-26
feat(evaluator): bundle metrics for plugin execution

SandyChapman:
## Summary
- add bundle-native metric wire models and cloudpickle bundling primitives
- make evaluator plugin backend submit bundled metrics and hydrate by payload kind at execution time
- add plugin-native job compilation, task entrypoint, and smoke coverage

Note: rebased onto main after the resolver branch landed in #38. Remote result-registration fix will be split into a separate main-based PR.

## Summary by CodeRabbit

* **New Features**
  * Users can bundle and submit custom Python metrics for remote evaluation; example updated to demonstrate packaging and submission.
  * Evaluation API now accepts bundled metrics so multiple metrics can be submitted and hydrated at runtime.
  * Packaged metrics include validation (output contract checks) and runtime compatibility checks to prevent invalid or incompatible bundles.

* **Chores**
  * Added cloudpickle dependency to support metric serialization.


plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
```diff
@@ -0,0 +1,242 @@
...
+        ...
+
+    @property
+    @abstractmethod
+    def digest(self) -> str:
+        """Format-specific digest for the payload contents."""
+        ...
+
+
+class MetricPayloadBundler(Protocol):
```
arpitsardhana: Nit: There is little name confusion due to similarity in naming of "MetricPayloadBundler" and "MetricBundlePayload".

arpitsardhana (approved):
Can you add how input spec will look like from user when submitting a custom metric

plugins/nemo-evaluator/src/nemo_evaluator/jobs/compiler.py
```diff
@@ -0,0 +1,127 @@
...
+def _evaluate_step(spec: EvaluateSpec, profile: str | None) -> PlatformJobStep:
+    return PlatformJobStep(
+        name=EVALUATE_STEP_NAME,
+        executor=CPUExecutionProviderSpec(
+            profile=profile or "default",
+            provider="cpu",
+            container=ContainerSpec(
+                image=get_qualified_image("nmp-cpu-tasks"),
+                entrypoint=["python", "-m"],
+                command=["nemo_evaluator.tasks.evaluate"],
```
arpitsardhana, line 89: We may need to make sure that:
This evaluate is ---> also calling file download
FileSetStep is also calling --> file download.

It may be possible that double file download may happen for evaluator native metrics

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/cloudpickle.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def bundle(self, metric: Metric) -> MetricBundlePayload:
+        """Serialize a runtime metric object to a cloudpickle payload."""
+        if not isinstance(metric, Metric):
+            raise MetricBundlingError("object does not satisfy the Metric protocol")
+
+        blob = cloudpickle.dumps(metric, protocol=pickle.HIGHEST_PROTOCOL)
+        return CloudpickleMetricPayload.from_blob(blob)
+
+    def unbundle(self, payload: MetricBundlePayload) -> Metric:
```
arpitsardhana: nit: We probably should enforce size limits on pickle bundle ie should be less than lets say 10 MB or whatever json limit allows to be mounted as file

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
```diff
@@ -0,0 +1,242 @@
...
+
+def unbundle_metric(bundle: MetricBundle) -> Metric:
+    """Hydrate a runtime metric from a standard metric bundle envelope."""
+    payload_bundler = metric_payload_bundler_for_payload(bundle.payload)
+    hydrated_metric = payload_bundler.unbundle(bundle.payload)
+    _validate_metric_matches_bundle(hydrated_metric, bundle)
+    return hydrated_metric
+
+
+def _validate_metric_matches_bundle(metric: object, bundle: MetricBundle) -> None:
```
arpitsardhana, line 208: Is there a way to check python version as well before unbundling at server side and ensure late metric failure does not happen

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
SandyChapman: Agreed. I renamed the strategy side to `MetricBundlePackager` / `CloudpickleMetricBundlePackager` and kept the data model as `MetricBundlePayload`. I also changed the strategy methods from `bundle` / `unbundle` to `package` / `load`, which should leave room for future implementations that package metrics as container-image references rather than serialized bytes.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/compiler.py
SandyChapman: Good catch. I removed the compile-time `dataset-download` step. The evaluate task now owns FilesetRef resolution, which avoids double download and keeps the door open for a future streaming/source-backed dataset path.

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/cloudpickle.py
SandyChapman: Added a decoded cloudpickle payload limit of 10 MiB. That matches the suggested order of magnitude and fails early before the base64-expanded job config becomes too large. I added coverage for oversized payload rejection.

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
SandyChapman: Added an early Python compatibility check before `cloudpickle.loads()`. It compares Python major/minor and raises a clear `MetricBundlingError` if the bundle was produced on a different runtime. I did not enforce `cloudpickle_version` yet since compatibility there may be less strict than Python runtime compatibility.

SandyChapman:
Replying to the review note: "Can you add how input spec will look like from user when submitting a custom metric"

Added a small custom metric example that shows the user-facing flow: define a metric object, generate the cloudpickle bundle from code with `CloudpickleMetricBundlePackager`, and build the `EvaluateSpec` from that generated payload. The example calls out that the base64 payload is not intended to be hand-authored.

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
```diff
@@ -0,0 +1,242 @@
...
+    payload_type: type[MetricBundlePayload],
+    packager_factory: Callable[[], MetricBundlePackager],
+) -> None:
+    """Register the payload model and packager factory for a bundle kind."""
+    if not kind:
+        raise ValueError("metric bundle payload kind must not be empty")
+    _BUNDLE_REGISTRY[kind] = _MetricBundleRegistration(
+        payload_type=payload_type,
+        packager_factory=packager_factory,
+    )
```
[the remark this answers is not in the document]
SandyChapman: Addressed in 4e5b80f0f: bundle-kind registration is now idempotent for identical registrations and raises on conflicting payload/packager registrations instead of silently overwriting the existing kind.

plugins/nemo-evaluator/src/nemo_evaluator/shared/metric_bundles/bundles.py
```diff
@@ -0,0 +1,242 @@
...
+    """Validate that bundle metadata still describes the hydrated metric."""
+    if not isinstance(metric, Metric):
+        raise MetricBundlingError("unbundled object does not satisfy the Metric protocol")
+
+    output_names = [output.name for output in metric.output_spec()]
+    bundled_output_names = [output.name for output in bundle.outputs]
+    if output_names != bundled_output_names:
+        raise MetricBundlingError("unbundled metric output spec does not match bundle metadata")
+    if validate_metric_type(metric) != bundle.metric_type:
+        raise MetricBundlingError("unbundled metric type does not match bundle metadata")
```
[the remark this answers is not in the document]
SandyChapman: Addressed in 4e5b80f0f: unbundle validation now compares the full `BundledMetricOutputSpec` projection from the hydrated metric against the bundle metadata, not just output names.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/compiler.py
```diff
@@ -0,0 +1,91 @@
...
```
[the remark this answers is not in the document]
SandyChapman: I am leaving this as `python -m` intentionally. The `uv run` guideline applies to local developer commands in this repo; this is a production job container entrypoint, where the image should run the installed module directly without adding a runtime dependency on uv or uv environment resolution.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/compiler.py
```diff
@@ -0,0 +1,91 @@
...
+            raise TypeError("agent target requires RunConfigOnline")
+    elif not isinstance(spec.params, RunConfig):
+        raise TypeError("offline evaluation requires RunConfig")
+
+
+def _add_secret_ref(secret_refs: dict[str, str], env_name: str, secret_name: str) -> None:
+    existing = secret_refs.get(env_name)
+    if existing is not None and existing != secret_name:
+        raise ValueError(f"conflicting secret references for environment variable {env_name!r}")
+    secret_refs[env_name] = secret_name
```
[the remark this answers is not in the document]
SandyChapman: Addressed in 4e5b80f0f: `_add_secret_ref()` now rejects reserved job environment names before adding secret refs, which prevents bundles or target credentials from shadowing `PERSISTENT_JOB_STORAGE_PATH_ENVVAR`.

SandyChapman:
Addressed the two CodeRabbit outside-diff notes in 4e5b80f0f as well: sync/async `Evaluator.submit()` now fail fast when `metric_bundle_packager` is missing, and sync/async standalone remote backends now validate `metric_bundle_packager` before delegating to the executor.

PR #55, opened 2026-05-26
fix(studio): Operators type for query filter

steramae-nvidia:
This PR addresses https://linear.app/nvidia/issue/ASTD-87/improve-filter-typing-for-dataview by adding a type that captures the mongo like operators that are supported by the backend but not included in the openapi definition. There is a new utility that creates a type union between string and the operator type.

 

## Summary by CodeRabbit

* **Refactor**
  * Standardized filtering across the app: introduced operator-based filters and a central helper, and made data-view hooks and many components accept typed filter shapes.
  * Improved filter typing for lists, searches and route query generation (jobs, deployments, evaluations, metrics, benchmarks, models, filesets, providers, exports, etc.), increasing consistency and reliability of search/filter behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/55?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/packages/common/src/api/filterOperators.ts
+++ b/web/packages/common/src/api/filterOperators.ts
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Mongo-style comparison operators accepted by NeMo Platform's unified filter
+ * syntax (e.g. `{ name: { $like: '%foo%' } }`, `{ created_at: { $gte, $lte } }`).
+ *
+ * The OpenAPI-generated SDK types model filter fields as bare scalars
+ * (`name?: string`) and do not expose the operator-object form. Use
+ * {@link WithFilterOperators} to widen a generated filter shape so call sites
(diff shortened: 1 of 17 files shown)


marcusds (approved):

web/packages/common/src/api/filterOperators.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Mongo-style comparison operators accepted by NeMo Platform's unified filter
+ * syntax (e.g. `{ name: { $like: '%foo%' } }`, `{ created_at: { $gte, $lte } }`).
+ *
+ * The OpenAPI-generated SDK types model filter fields as bare scalars
+ * (`name?: string`) and do not expose the operator-object form. Use
+ * {@link WithFilterOperators} to widen a generated filter shape so call sites
+ * can build operator filters without `as unknown as` casts.
+ */
+export interface FilterOperators<V> {
```
marcusds, line 13: It would be nice if Orval could do this, but that is a whole big thing probably from top down.

steramae-nvidia: That would require a backend update which I'm not sure is prioritized at the moment. This is a bandaid but won't be necessary once the openapi spec has the correct union type.

PR #58, opened 2026-05-26
fix: jobs tests and repo config yaml

mikeknep:
### What was happening

If you set `NMP_CONFIG_FILE_PATH="packages/nmp_platform/config/local.yaml"` **and** have docker running when you spin up `nemo services run`, the Jobs service registers the docker executor but not the subprocess executor, and all jobs you attempt to create try to go through the docker backed only to fail with an Image Not Found error (or, if you have a stale version of `nmp-cpu-tasks` laying around, some other error since that image hasn't been rebuilt).

### What's here

This PR registers the subprocess executor on that YAML config so that its executors match the bundled YAML config that gets shipped in the wheel and that serves as the default when the env var is unset: `packages/nmp_platform_runner/src/nmp/platform_runner/config/local.yaml`

### Longer term

It'd be nice to consolidate these two YAML files into a single source of truth, but this is a larger effort.

### Unit test cleanup

Fixes some tests failing on `main` by:
- Deduping a bunch of constants appearing in two places
- Ensuring the test executor sets an entrypoint and command (required by the subprocess executor)

## Summary by CodeRabbit

* **Refactor**
  * Job creation now converts CPU container execution steps into subprocess-compatible steps prior to storage.
  * Default executor selection temporarily uses a subprocess-only default.
  * Test helpers and fixtures reorganized; test container specs now include explicit entrypoint and command.

* **Tests**
  * Updated tests and assertions to match the translation behavior and subprocess-only defaults, including GPU/Docker scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/58?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/services/evaluator/tests/integration/tasks/conftest.py
+++ b/services/evaluator/tests/integration/tasks/conftest.py
@@ -63,7 +63,10 @@ class JobContext:
                         "executor": {
                             "provider": "cpu",
                             "profile": "default",
-                            "container": {"image": "test:latest"},
+                            "container": {
+                                "image": "test:latest",
+                                "command": ["command"],
+                            },
                         },
                     }
                 ]
--- a/services/evaluator/tests/integration/tasks/conftest.py
+++ b/services/evaluator/tests/integration/tasks/conftest.py
@@ -65,6 +65,7 @@ class JobContext:
                             "profile": "default",
                             "container": {
                                 "image": "test:latest",
+                                "entrypoint": ["entrypoint"],
                                 "command": ["command"],
                             },
                         },
--- a/packages/nmp_platform/config/local.yaml
+++ b/packages/nmp_platform/config/local.yaml
@@ -39,6 +39,30 @@ entities: {}
 
 # Jobs service configuration
 jobs:
+  # Explicitly register the subprocess executor at profile "default". This opts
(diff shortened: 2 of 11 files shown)


services/evaluator/tests/integration/tasks/conftest.py
```diff
@@ -63,7 +63,10 @@ def create(self, job_name: str):
                         "executor": {
                             "provider": "cpu",
                             "profile": "default",
-                            "container": {"image": "test:latest"},
+                            "container": {
+                                "image": "test:latest",
+                                "command": ["command"],
+                            },
```
mikeknep, line 70: For some reason when I try to add `entrypoint` to this one (matching all the others) I get this error from `ty` 🤔 

> Expected `PlatformJobSpecParam`, found `dict[Unknown | str, Unknown | list[Unknown | dict[Unknown | str, Unknown | str | dict[Unknown | str, Unknown | str | dict[Unknown | str, Unknown | str | list[Unknown | str]]]]]]`

mikeknep: 🤦 it's a seq too, fixed

benmccown (approved):

PR #59, opened 2026-05-26
test(studio): Add deployments e2e spec and recording mode

steramae-nvidia:
[video.webm](https://github.com/user-attachments/assets/db82513d-f504-4d9e-9d74-76bb13847245)

## Summary
- Adds a Playwright e2e spec covering the full model deployment lifecycle (create NGC deployment → view details panel → delete), tagged `@record` for inclusion in curated recordings
- Adds a `test:e2e:record` script and `RECORD=1` env flag in `playwright.config.ts` that flips `video` and `trace` to `'on'` for tagged tests, leaving normal CI behavior (`retain-on-failure` / `on-first-retry`) untouched
- Introduces a workspace-scoped page object (`workspace-deployments.ts`) and a minimal `DeploymentsAPI` for cleanup, following the patterns from the existing project-scoped specs

## Test plan
- [ ] `pnpm --filter nemo-studio-ui test:e2e:record` produces a video at `test-results/<test-name>/video.webm` for the deployments spec
- [ ] `pnpm --filter nemo-studio-ui test:e2e` (without `RECORD=1`) keeps existing video/trace behavior
- [ ] `pnpm --filter nemo-studio-ui typecheck` is clean

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests for model deployment flows (create, verify, open details, delete) with cleanup and long-operation waits.
  * Added supporting test helpers and page objects for workspace deployments and API cleanup.

* **Chores**
  * Playwright now conditionally enables trace/video when recording.
  * Added a script to run recorded E2E tests.
  * Updated E2E test base URL default for non-local environments.


walston (commented):
What happens if a test fails before cleanup? do the cleanup steps happen?

When I set these types of things up in Cypress before I would do some manual API driven cleanup after the test, but I don't see that happening here

steramae-nvidia:
> What happens if a test fails before cleanup? do the cleanup steps happen?
> 
> When I set these types of things up in Cypress before I would do some manual API driven cleanup after the test, but I don't see that happening here

@walston  Yup, there was a cleanup stage but it required the test to finish. I've added a tracking fixture to ensure that in the case of test failure that we cleanup any entities that we needed

web/packages/studio/e2e-tests/utils/environment.ts
```diff
@@ -17,7 +17,8 @@ export const {
  * - In CI with ephemeral deployments: Uses VSERVICE_URL_STUDIO_UI from the deployment job
  * - Local development: Falls back to local development server
  */
-export const STUDIO_URL = process.env.VSERVICE_URL_STUDIO_UI || 'http://localhost:8080/studio/';
+export const STUDIO_URL =
+  process.env.VSERVICE_URL_STUDIO_UI || 'https://ns.local.aire.nvidia.com:5173/';
```
aray12: I thought we were trying to make sure there were no nvidia specific artifacts in the repo

aray12 (approved):

web/packages/studio/e2e-tests/utils/environment.ts
steramae-nvidia: dang yeah I missed that. I'll change it back

PR #65, opened 2026-05-27
fix(models): Prefix digit-leading upstream model ids

matthewgrossman:
## What

When the models controller reconciler discovers models from an external provider, it normalizes each upstream id to an internal entity name that satisfies the entity store's RFC 1035-style `NAME_PATTERN` (leading `[a-z]`, length 2-63, only `[a-z0-9-]`). Upstream catalogs occasionally publish digit-leading ids — e.g. NVIDIA Build's `01-ai/yi-large` — which the normalizer rejects because `01-ai-yi-large` doesn't start with a letter. The reconciler then logs `Skipped 1 model(s) with invalid names ...` on every cycle (~every 5s) and the model never becomes routable.

This PR teaches `normalize_model_entity_name` to prepend an internal `m-` prefix when the cleaned form starts with a digit, so digit-leading upstream ids become valid entity names instead of being skipped.

## Why this approach

* **Internal-only, not user-facing.** The normalized string is the entity store's `model_entity_id`. The `served_model_name` mapping still preserves the original id (`provider_reconciler.py:707`), so the IGW continues to forward `01-ai/yi-large` (the original) to the backend. End users keep calling the model exactly as the upstream advertises it.
* **Minimal & deterministic.** A constant `m-` prefix is mechanical, idempotent (`m-01-ai-yi-large` → `m-01-ai-yi-large`), and preserves the source id verbatim after the prefix, so the entity name still reads like the upstream id.
* **Length-safe.** The prefix is applied *before* the existing 63-char truncate-with-hash step, so long digit-leading names still produce length-bounded results via the existing machinery.

Considered and rejected: provider-derived prefixes (collide with the existing `workspace/name` convention, lengthen names), spelling out leading digits (lossy/weird), dropping leading digits (collisions across ids that differ only in the digit segment), hash prefixes (ugly, inconsistent with the existing policy of hashing only on overflow).

## Verification

* New positive test cases for `01-ai/yi-large`, `9model`, `123`, `2pac/rapper-model`.
* New dedicated tests for idempotence and the digit-leading + 63-char-overflow interaction.
* Updated `test_entity_name_from_discovered_model_invalid_raises` and added `test_entity_name_from_discovered_model_digit_leading_gets_prefix` so the reconciler's caller-side contract is also pinned.
* Full `services/core/models/tests/unit` suite: 992 passed, 1 skipped.
* `ruff check` / `ruff format --check` clean. `ty` clean for the change (5 pre-existing diagnostics in unrelated functions in the same file).

## Tracking

* Linear: [AIRCORE-678](https://linear.app/nvidia/issue/AIRCORE-678/models-controller-prefix-digit-leading-upstream-model-ids-so-they)
* NVBug: [6215677](https://nvbugspro.nvidia.com/bug/6215677/1)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of model identifiers that start with digits to ensure proper system compatibility.
  * Updated validation logic to handle digit-leading model IDs with automatic prefix handling.

* **Tests**
  * Added test coverage for digit-leading model identifier normalization and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/65?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/services/core/models/src/nmp/core/models/app/utils.py
+++ b/services/core/models/src/nmp/core/models/app/utils.py
@@ -186,6 +186,12 @@ def normalize_model_entity_name(model_name: str) -> str:
     normalizes when possible; if the result would not match, it raises ValueError
     so callers can skip or fail explicitly.
 
+    Names that would otherwise normalize cleanly but begin with a digit (e.g.
+    upstream catalog ids like ``01-ai/yi-large``) are prefixed with ``m-`` so they
+    satisfy the leading-letter requirement. The prefix is purely internal to the
+    entity store / routing layer; the original id is preserved as the
+    ``served_model_name`` and remains the user-facing handle.
+
     Args:
         model_name: The original model name (e.g., "meta/llama-3.2-1b-instruct")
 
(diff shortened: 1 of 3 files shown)


mckornfield:
lmao jinx https://github.com/NVIDIA-NeMo/nemo-platform/pull/67

mckornfield (approved):
`m-` is probably better than `md-` and mine has lint failures

PR #77, opened 2026-05-27
fix(services): sweep orphaned child processes on `nemo services stop`

tylersbray:
## Summary

- `nemo services stop` only killed the platform PID, leaving agent deployment subprocesses (`nat start fastapi`) alive across stop/start cycles — potentially causing OOM
- Added `_snapshot_children()` / `_sweep_orphans()` helpers that capture the child process tree before SIGTERM and sweep survivors after the parent exits
- Graceful shutdown still gets first shot; the psutil-based sweep is a belt-and-suspenders safety net
- Updated CLI output to report swept child count separately

## Test plan

- [x] Unit tests for `_snapshot_children()` (live children, dead PID, childless process)
- [x] Unit tests for `_sweep_orphans()` (surviving children, already-exited children, SIGKILL escalation)
- [x] Integration tests for `stop_instance()` (swept children populated vs empty on clean shutdown)
- [x] All 46 tests pass
- [x] Ruff lint + format clean
- [x] Pre-commit hooks pass
- [x] Vendored SDK copy in sync

Closes AIRCORE-675

## Summary by CodeRabbit

* **New Features**
  * Stop operations now report how many orphaned child processes were swept and include those PIDs in results.

* **Bug Fixes**
  * Improved cleanup of orphaned child processes during stop flows; sweeping runs reliably after stop attempts.

* **Tests**
  * Added tests covering snapshotting, sweeping of child processes, and stop behavior including swept-child reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/77?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
@@ -28,7 +28,7 @@ import subprocess
 import sys
 import tempfile
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Literal
@@ -322,6 +322,53 @@ def log_path_for(scope: str, *, base_dir: Path | None = None) -> Path:
 @dataclass
 class StopResult:
     stopped_pids: list[int]
+    swept_children: list[int] = field(default_factory=list)
+
+
+def _snapshot_children(pid: int) -> list[psutil.Process]:
+    """Return all descendant processes of *pid*.
+
+    Must be called while the parent is still alive; once it exits,
+    children are reparented to init and won't appear in the tree.
+    """
+    try:
+        return psutil.Process(pid).children(recursive=True)
+    except (psutil.NoSuchProcess, psutil.AccessDenied):
+        return []
+
+
+def _sweep_orphans(children: list[psutil.Process], timeout: float = 5.0) -> list[int]:
+    """Terminate any still-alive processes from a prior snapshot.
+
+    Sends SIGTERM, waits up to *timeout*, then SIGKILL survivors.
+    Returns PIDs that were signaled.  Handles ``NoSuchProcess`` gracefully
+    since children may have already exited during graceful shutdown.
+    """
+    alive_children = [c for c in children if c.is_running()]
+    if not alive_children:
+        return []
+
+    killed: list[int] = []
+    for child in alive_children:
+        try:
+            child.terminate()
+            killed.append(child.pid)
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            pass
+
+    _, still_alive = psutil.wait_procs(alive_children, timeout=timeout)
+    for child in still_alive:
+        try:
+            child.kill()
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            pass
+
+    if killed:
+        logger.info(
+            "Swept %d orphaned child %s: %s", len(killed), "process" if len(killed) == 1 else "processes", killed
+        )
+
+    return killed
 
 
 def stop_instance(
@@ -342,6 +389,8 @@ def stop_instance(
     stopped via Ctrl-C in their own terminal.  Pass *force=True* to override.
 
     Sends SIGTERM, waits up to *timeout* seconds, then escalates to SIGKILL.
+    After the parent exits, any surviving child processes (e.g. agent
+    deployments) are swept up via SIGTERM/SIGKILL.
     """
     desc = read_descriptor(scope, base_dir=base_dir)
     alive = is_instance_alive(scope, base_dir=base_dir)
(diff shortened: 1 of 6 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/cli.py
```diff
@@ -409,7 +409,11 @@ def stop_services_cmd(
         typer.echo("Platform services are not running.")
         return
     pids_str = ", ".join(str(p) for p in result.stopped_pids)
-    typer.echo(f"Stopped platform services (pid {pids_str})")
+    msg = f"Stopped platform services (pid {pids_str})"
+    if result.swept_children:
+        n = len(result.swept_children)
+        msg += f" and {n} child {'process' if n == 1 else 'processes'}"
```
mckornfield: nit- could take ternary out of the fmt, kinda hard to read

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
```diff
@@ -322,6 +322,53 @@ def log_path_for(scope: str, *, base_dir: Path | None = None) -> Path:
...
+    since children may have already exited during graceful shutdown.
+    """
+    alive_children = [c for c in children if c.is_running()]
+    if not alive_children:
+        return []
+
+    killed: list[int] = []
+    for child in alive_children:
+        try:
+            child.terminate()
```
mckornfield, line 354: lol the names are accurate but tragic

mckornfield (approved):

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
matthewgrossman: won't you think of the orphans?!

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
```diff
@@ -322,6 +322,53 @@ def log_path_for(scope: str, *, base_dir: Path | None = None) -> Path:
 @dataclass
 class StopResult:
     stopped_pids: list[int]
+    swept_children: list[int] = field(default_factory=list)
+
+
+def _snapshot_children(pid: int) -> list[psutil.Process]:
```
matthewgrossman, line 328: did we look into using a process group for terminating everything? I feel like I ran into this somewhat recently, and as long as we ensure all the procs are in the same group we could just let the OS handle this (maybe?)

not educated enough to say the full solution, just curious if its something you looked at

matthewgrossman (approved):

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
```diff
@@ -322,6 +322,53 @@ def log_path_for(scope: str, *, base_dir: Path | None = None) -> Path:
...
+
+    killed: list[int] = []
+    for child in alive_children:
+        try:
+            child.terminate()
+            killed.append(child.pid)
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            pass  # Already exited or not owned by us — skip.
+
+    _, still_alive = psutil.wait_procs(alive_children, timeout=timeout)
```
maxdubrinsky, line 359: Image

maxdubrinsky (approved):
morbid but looks good to me

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
tylersbray: Yep, the jobs backend is using process groups but there is a wrinkle here.  Here is the long answer:

Good question, and yes we looked at it. The jobs backend is the right comparison point -- it uses os.killpg because its setup is ideal for it: each job task gets start_new_session=True, creating a dedicated process group that is isolated from everything else. Killing that group is clean and contained.

The platform stop path has a different topology that makes process groups a poor fit:

The platform itself starts with start_new_session=True (line 520 of _process.py). Agent processes spawned by InMemoryRunnerBackend._spawn() do not set start_new_session (line 316 of in_memory.py), so they inherit the platform's process group. An os.killpg on that group would hit every process simultaneously -- platform and all agents get the signal at the same time with no ordering control.

That kills graceful shutdown. Our whole design is: SIGTERM the parent first, let it run its on_shutdown handlers (which call _terminate() on each agent deployment), and only sweep orphans that survive. os.killpg is atomic -- there's no "signal the parent, wait, then mop up." You'd lose the graceful path entirely.

It's fragile against future backends. Any plugin or backend that spawns processes with start_new_session=True (like the jobs backend does) would escape the platform's process group. The psutil.children(recursive=True) walk doesn't care about group membership -- it follows the parent-child tree regardless of session/group boundaries.

In short: os.killpg works great for the jobs case (isolated, self-contained groups where you want atomic teardown), but for stopping the platform it would sacrifice graceful shutdown ordering and break on any subprocess that creates its own session.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/cli.py
tylersbray: Updated for readability, thanks.

PR #78, opened 2026-05-27
feat(build): Add CI test that builds+tests the wheels

matthewgrossman:
## What

Adds a CI job that builds and tests every shipping wheel — `nemo-platform` and `nemo-platform-plugin` — across every supported Python (3.11, 3.12, 3.13). Each row builds the wheel with the exact same code path the release workflow uses, installs it into an isolated env, and exercises it end-to-end. Today CI only runs against the editable workspace install, so wheel-install breakages (missing `policy.wasm`, vendor-metadata drift, entry-point typos, namespace-package gotchas, broken `Requires-Dist`, generated-extra mismatches) only surface on manual post-release smoke checks.

The build itself is factored into a new composite action, `.github/actions/build-nemo-platform-wheel`, which is also called by `release-bundle.yaml`'s `build-sdks` matrix. CI test wheels and published wheels now come out of one code path.

## Shape

```
policy-wasm  ──▶  wheel-test  ──▶  wheel-test-aggregate
                  matrix(package × python)        (required-check pin)
                  6 rows
```

- **`wheel-test`** matrix is `package: [nemo-platform, nemo-platform-plugin] × python-version: [3.11, 3.12, 3.13]` = 6 rows. `fail-fast: false` so a single-row failure doesn't hide the others' logs.
- **`wheel-test-aggregate`** is a single-job umbrella that succeeds iff every matrix row succeeded. Branch protection should pin *this* check name (`Wheel build + test`); the matrix can grow without anyone touching protection settings.
- **Composite action** owns the build path (uv setup, conditional Studio asset compile, version stamping via `stamp_sdk_version.py`, `uv build --wheel`, single-wheel guard, `rm -rf` safety check on caller-provided `out-dir`). Stamping always runs regardless of caller, with `cadence: nightly` + sentinel epoch `19700101000000` for CI tests so the test wheel version is well-formed PEP 440 dev but obviously synthetic.

## Per-package test surface

- **`nemo-platform`**: `uv tool install ./*.whl[services]`, then `script/test-nemo-cli.sh` boots the platform via `nemo services run` (no `--services` filter — every bundled service module loads, ~17 services + 4 controllers) and polls `nemo workspaces list -f json` until both `default` and `system` (the platform-seeded workspaces) appear, with SIGTERM-then-SIGKILL teardown of the whole process group via bash job control (`set -m`).
- **`nemo-platform-plugin`**: library-only (no `[project.scripts]`), so `uv venv` + `uv pip install` + `python -c "import nemo_platform_plugin; import nemo_platform_plugin.cli; import nemo_platform_plugin.commands"`. That still catches the bug classes a wheel-test is supposed to catch (broken pyproject.toml, broken `Requires-Dist`, `hatch_build.py` regressions, namespace-package gotchas).

`script/test-nemo-cli.sh` is install-method agnostic — anywhere `nemo` is on PATH (`uv tool install` from a wheel, `uv sync` in dev, `pipx`, distro package), the script can run and prove the CLI works.

## What this does NOT do (yet)

- **Not yet a required check.** The aggregator name `Wheel build + test` is set up to be the single pin point; flipping branch-protection-required is a one-line admin change once this has been green for a few working days.
- **Doesn't gate `release-stable.yaml`.** That's the second admin step: extend the stable-release dispatch to fail if `Wheel build + test` didn't pass on the source SHA.
- **Platform-Deploy migration is separate.** `Platform-Deploy/.github/workflows/release-sdk.yaml` still has its own inline build path that drifts from this action; migrating it to consume `build-nemo-platform-wheel@<sha>` is a Platform-Deploy PR.
- **No vendor-drift handling.** The test deliberately runs against committed metadata; vendor-drift detection belongs in `lint`.

## Coderabbit feedback addressed

- `rm -rf "${abs_out_dir}"` in the action now rejects empty/`/` paths and requires the resolved path to live inside source-root or `$GITHUB_WORKSPACE`.
- Readiness probe now requires both `default` and `system` workspaces (the entities service seeds them back-to-back, so missing one is a real signal that init partially failed).
- `github.run_started_at` reference removed in favor of the sentinel epoch (the context property doesn't exist).
- SHA-pinning of `actions/checkout@v6` etc. and `persist-credentials: false` deferred — every other job in `ci.yaml` uses major-version tags, so policy-pinning the wheel-test job alone would be inconsistent. Belongs in a separate repo-wide PR.

## Verification

- All YAML files parse with `yaml.safe_load`.
- `script/test-nemo-cli.sh` is `shellcheck` and `bash -n` clean.
- Pre-commit clean on all changed files.
- Most recent CI run on this branch: 7 jobs green (6 matrix rows + aggregator).

## Tracking

- Linear: [AIRCORE-682](https://linear.app/nvidia/issue/AIRCORE-682/ci-build-wheel-install-in-fresh-env-smoke-test-nemo-platform-end-to)
- Plan: `docs/plans/aircore-682-wheel-test.md`

### DIFF
--- a/.github/actions/build-nemo-platform-wheel/action.yaml
+++ b/.github/actions/build-nemo-platform-wheel/action.yaml
@@ -23,7 +23,7 @@ inputs:
     required: false
     default: nemo-platform
   out-dir:
-    description: Absolute path where the wheel is written. Must be absolute.
+    description: Absolute path where the wheel is written.
     required: true
   cadence:
     description: >
@@ -72,6 +72,43 @@ runs:
         version-file: ${{ inputs.source-root }}/pyproject.toml
         cache-dependency-glob: ${{ inputs.source-root }}/uv.lock
 
+    # Stamp early, before pnpm/node setup, so bad cadence/label/timestamp
+    # inputs fail fast instead of after spending ~20s on tool setup. The
+    # version is written into packages/nmp_common/.../version.py and the
+    # generated _version.py before `uv build` runs (hatchling reads
+    # version.py via its regex source but does not write it). The
+    # --print-version flag emits just the resolved version on stdout;
+    # the human banner goes to stderr.
+    - name: Stamp SDK version
+      id: stamp
+      shell: bash
+      env:
+        SOURCE_ROOT: ${{ inputs.source-root }}
+        PACKAGE: ${{ inputs.package }}
+        CADENCE: ${{ inputs.cadence }}
+        RELEASE_LABEL: ${{ inputs.release-label }}
+        NIGHTLY_TIMESTAMP: ${{ inputs.nightly-timestamp }}
+        # Run the stamp script from the action's own checkout (the workflow
+        # ref), not from inputs.source-root. For ci.yaml these are the same;
+        # for release-bundle.yaml the source checkout can be at an older
+        # source_sha (RC/stable rerun of an old tag) and we want today's
+        # stamping logic to apply.
+        STAMP_SCRIPT_PATH: ${{ github.action_path }}/../../scripts/stamp_sdk_version.py
+      run: |
+        set -euo pipefail
+        args=(
+          --source-root "${SOURCE_ROOT}"
+          --sdk-id "${PACKAGE}"
+          --cadence "${CADENCE}"
+          --nightly-timestamp "${NIGHTLY_TIMESTAMP}"
+          --print-version
+        )
+        if [[ -n "${RELEASE_LABEL}" ]]; then
+          args+=(--release-label "${RELEASE_LABEL}")
+        fi
+        wheel_version="$(uv run --no-project python "${STAMP_SCRIPT_PATH}" "${args[@]}")"
+        echo "wheel-version=${wheel_version}" >>"${GITHUB_OUTPUT}"
+
     # Studio assets are only force-included by the nemo-platform wrapper.
     # The hatch hook in packages/nemo-platform/hatch_build.py compiles them
     # via pnpm during `uv build`; we set up pnpm/node here so the hook can
@@ -90,52 +127,13 @@ runs:
         cache: pnpm
         cache-dependency-path: ${{ inputs.source-root }}/web/pnpm-lock.yaml
 
-    # Stamp the version into packages/nmp_common/.../version.py and the
-    # generated _version.py before `uv build` runs (hatchling reads
-    # version.py via its regex source but does not write it). The
-    # --print-version flag emits just the resolved version on stdout;
-    # the human banner goes to stderr.
-    - name: Stamp SDK version
-      id: stamp
-      shell: bash
-      env:
-        SOURCE_ROOT: ${{ inputs.source-root }}
-        SDK_ID: ${{ inputs.package }}
-        CADENCE: ${{ inputs.cadence }}
-        RELEASE_LABEL: ${{ inputs.release-label }}
-        NIGHTLY_TIMESTAMP: ${{ inputs.nightly-timestamp }}
-        # Run the stamp script from the action's own checkout (the workflow
-        # ref), not from inputs.source-root. For ci.yaml these are the same;
-        # for release-bundle.yaml the source checkout can be at an older
-        # source_sha (RC/stable rerun of an old tag) and we want today's
-        # stamping logic to apply.
-        STAMP_SCRIPT: ${{ github.action_path }}/../../scripts/stamp_sdk_version.py
-      run: |
-        set -euo pipefail
-        args=(
-          --source-root "${SOURCE_ROOT}"
-          --sdk-id "${SDK_ID}"
-          --cadence "${CADENCE}"
-          --nightly-timestamp "${NIGHTLY_TIMESTAMP}"
-          --print-version
-        )
-        if [[ -n "${RELEASE_LABEL}" ]]; then
-          args+=(--release-label "${RELEASE_LABEL}")
-        fi
-        wheel_version="$(uv run --no-project python "${STAMP_SCRIPT}" "${args[@]}")"
-        if [[ -z "${wheel_version}" ]]; then
-          echo "::error::stamp_sdk_version.py produced empty output" >&2
-          exit 1
-        fi
-        echo "wheel-version=${wheel_version}" >>"${GITHUB_OUTPUT}"
-
     - name: Build wheel
       id: build
       shell: bash
       env:
         SOURCE_ROOT: ${{ inputs.source-root }}
         OUT_DIR: ${{ inputs.out-dir }}
-        SDK_ID: ${{ inputs.package }}
+        PACKAGE: ${{ inputs.package }}
       run: |
         set -euo pipefail
         if [[ "${OUT_DIR}" != /* ]]; then
@@ -145,13 +143,13 @@ runs:
 
         uv build --wheel \
           --directory "${SOURCE_ROOT}" \
-          --package "${SDK_ID}" \
+          --package "${PACKAGE}" \
           --out-dir "${OUT_DIR}"
 
         shopt -s nullglob
         wheels=("${OUT_DIR}"/*.whl)
         if [[ "${#wheels[@]}" -ne 1 ]]; then
-          echo "::error::expected exactly one wheel for ${SDK_ID}, found ${#wheels[@]}" >&2
+          echo "::error::expected exactly one wheel for ${PACKAGE}, found ${#wheels[@]}" >&2
           printf '  %s\n' "${wheels[@]}" >&2
           exit 1
         fi
--- a/.github/actions/build-nemo-platform-wheel/action.yaml
+++ b/.github/actions/build-nemo-platform-wheel/action.yaml
@@ -1,0 +1,158 @@
+name: Build nemo-platform wheel
+description: >
+  Set up the build toolchain (uv, plus pnpm/node when building nemo-platform
+  so the hatch hook can compile Studio assets), stamp the SDK version, and
+  run `uv build --wheel --package <pkg>`. The build itself — including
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -202,6 +202,73 @@ jobs:
...
+    needs: [policy-wasm]
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Download policy WASM
+        uses: actions/download-artifact@v8
```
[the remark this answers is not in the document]
matthewgrossman: Skipping in this PR: every existing job in `ci.yaml` uses major-version tags (`@v6`, `@v8`, `@v4`) for `actions/*` actions, and none set `persist-credentials: false`. Pinning only the new `wheel-test` job to commit SHAs would create policy inconsistency across the file without solving the underlying gap.

If we want repo-wide SHA pinning + zizmor-clean checkout configuration, that should be a separate, single-purpose PR (likely with a dependabot config so the SHAs auto-update), not snuck into this one.

.github/workflows/ci.yaml
```diff
@@ -202,6 +202,118 @@ jobs:
...
+    name: ${{ matrix.package }} wheel build + test (py${{ matrix.python-version }})
+    needs: [policy-wasm]
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        package: [nemo-platform, nemo-platform-plugin]
```
svvarom, line 221: Can we somehow load this from [release/assets.yaml](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/release/assets.yaml) instead? I’m worried that if we add many new SDKs and forget to add them here, we won’t test them and will only find failures during release workflows.

svvarom: Reading further, it matters less, ig, if we need to include a separate test for each SDK - unless we just want to test building the wheel

.github/actions/build-nemo-platform-wheel/action.yaml
```diff
@@ -0,0 +1,202 @@
...
+        RELEASE_LABEL: ${{ inputs.release-label }}
+        NIGHTLY_TIMESTAMP: ${{ inputs.nightly-timestamp }}
+        # Always run the stamp script from the same checkout the composite
+        # action lives in (the workflow ref), not from the release-target
+        # source checkout. For ci.yaml these are the same; for
+        # release-bundle.yaml the source checkout can be at an older
+        # source_sha (RC/stable rerun of an old tag), and we want today's
+        # stamping logic to apply to that source tree.
+        STAMP_SCRIPT: ${{ github.action_path }}/../../scripts/stamp_sdk_version.py
+      run: |
```
mckornfield: nit- can this be from the repo root or smth?

.github/actions/build-nemo-platform-wheel/action.yaml
```diff
@@ -0,0 +1,202 @@
...
+        if [[ -n "${RELEASE_LABEL}" ]]; then
+          args+=(--release-label "${RELEASE_LABEL}")
+        fi
+
+        output="$(uv run --no-project python "${STAMP_SCRIPT}" "${args[@]}")"
+
+        echo "${output}"
+
+        # The script prints "Stamped sdk:<id> version <version>." — extract <version>.
+        sdk_version="$(awk '/^Stamped sdk:/ { sub(/\.$/, "", $NF); print $NF }' <<<"${output}")"
```
mckornfield: lol... would be nice if it could be read like a property or smth

script/test-nemo-cli.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
mckornfield, line 1: kinda existential question (this PR might be fine too)... can we not do this all via a unit/integration test? i.e., does it merit another CI job, or can we just run the stamping, wheel maker and install in an existing job, and still get what we need? And then we just need to put the wheel creation into a script completely instead of having it be spread across github action steps

mckornfield (commented):
two minors and a major, but probably fine (can stamp after)

.github/workflows/ci.yaml
svvarom: We might want to add `wheel-test-aggregate` here

script/test-nemo-cli.sh
matthewgrossman: Two questions hidden in here, I think:

(a) **Could this be a unit/integration test?** No — the test deliberately runs against a wheel installed into a fresh, source-tree-free env. Pytest and friends presume cwd is the repo and the workspace is editable-installed, which is the exact setup we're trying to *not* trust. The bug class this catches is 'wheel installs cleanly in isolation'; you can't catch that from inside a job that already `uv sync`ed the workspace.

(b) **Could the wheel-creation be one self-contained script instead of spread across action steps?** Fair. The action splits across steps because each maps to a distinct GHA concept (`setup-uv`, `pnpm/action-setup`, `setup-node`). The Studio compile and node setup *can't* sensibly move into a script (they need first-class actions), but the Python-side logic — stamping, `uv build`, exact-one-wheel guard, the `rm -rf` safety check — could fold into a `script/build-wheel.sh` that the action invokes. That's a real cleanup.

Filing this as a follow-up rather than landing it here. Tracking item: 'consolidate composite-action bash steps into script/build-wheel.sh'.

.github/workflows/ci.yaml
matthewgrossman: Real concern, and you talked yourself most of the way there in the self-reply: each package has its own *test* surface (today `nemo-platform` boots a platform; `nemo-platform-plugin` does an import), so a fully-data-driven matrix would only be honest for the **build** half.

Two reasonable shapes for closing the gap:

(a) Read `release/assets.yaml` in a small upstream job that emits a matrix output (mirroring `release-bundle.yaml`'s `plan-release` job — it's already doing this). Downstream test-step gates stay explicit per package; a new SDK gets its build verified for free, and we'd have to add a test gate manually, but the build coverage is now data-driven.

(b) Cheap version: comment + lint check that `release/assets.yaml`'s `sdk` entries are a subset of this matrix's `package` list.

I'd vote (a) but it's a meaningful refactor (~50 lines), so I'd file it as a follow-up rather than land it here. If you'd rather block on it I'll do it now — let me know.

.github/actions/build-nemo-platform-wheel/action.yaml
matthewgrossman: The path traversal is intentional, and the comment block immediately above it (lines 86-91) covers the why, but the comment is dense — let me unpack it inline.

There are two checkouts at runtime:
- For `ci.yaml`: action lives at `<repo>/.github/actions/build-nemo-platform-wheel/`, source-root is `<repo>`. They're the same checkout.
- For `release-bundle.yaml`: it does two checkouts — `workflow/` (this repo at the workflow ref) and `source/` (the release-target source at `source_sha`, which can be an old tag). The action runs out of `workflow/.github/actions/...`.

We want to run **today's** stamp script (the one in `workflow/`), not the release-target's (which could be weeks old). `${{ github.action_path }}` is the only context value that's guaranteed to point at the action's own checkout regardless of which caller invoked it. Adding 2 `../`s gets us back to that checkout's `.github/scripts/`.

Alternatives I considered:
- `${{ github.workspace }}/.github/scripts/` — wrong for release-bundle, where `github.workspace` is the runner workspace root and the action's checkout is at `workflow/`.
- `${{ inputs.source-root }}/.github/scripts/` — wrong for release-bundle for the opposite reason: would point at the release-target's (old) script.

So `action_path/../../scripts/` is doing real work, just from an awkward starting point. Open to better names if there's a context value I missed.

.github/actions/build-nemo-platform-wheel/action.yaml
matthewgrossman: Done — added `--print-version` to `stamp_sdk_version.py` and dropped the awk parsing. Action now does `sdk_version=$(uv run … --print-version)`. Commits 3a9775b + 65853d3.

.github/workflows/ci.yaml
matthewgrossman: Done — added `wheel-test-aggregate` to `notify-ci-consumer`'s `needs:` list. Now we won't dispatch a 'CI passed' signal downstream on a SHA whose wheel test failed. Commit 57ebb0d.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
matthewgrossman: @coderabbitai yes please — open a tracking issue for repo-wide SHA pinning of `actions/*` + `persist-credentials: false` + Dependabot config. Out of scope for this PR.

.github/actions/build-nemo-platform-wheel/action.yaml
matthewgrossman: ya this is ugly but I think flox will make it better

.github/workflows/ci.yaml
matthewgrossman: can do this as a fast-follow. I also think once builds are easier via flox this will get smaller surface area too

script/test-nemo-cli.sh
matthewgrossman: ya I think I mostly agree with agent-matt above, this this can't be a unit/integration test because the goal is to run this in a completely isolated environment as the "CLI" level? but maybe I'm not fully understanding

mckornfield: I think this is clearly cleaner. But you would run the unit test in a tmpdir etc

mckornfield (approved):

svvarom (approved):

PR #85, opened 2026-05-28
chore(intake): Remove unused intake surface area

BrianNewsom:
## Summary by CodeRabbit

* **Refactor**
  * Intake refocused on telemetry: spans, traces, annotations, and evaluator-results; legacy apps/entries/tasks/exports surfaces and related CLI commands removed.

* **New Features**
  * OTLP/trace ingest and ClickHouse-backed span/trace read APIs added; stricter OpenAI-style chat-completions ingest validation.

* **Documentation**
  * README rewritten with v2 workspace endpoints, local dev startup steps, OTLP example, and OpenAPI refresh guidance.

* **Tests**
  * Tests updated to reflect span-era API surface and revised readiness/bootstrap behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/85?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


services/intake/alembic/versions/2025_10_14_1101-a7e3b1e7355b_initial_migration.py
```diff
@@ -1,124 +0,0 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
asutermo, line 1: Is this okay to remove?

services/intake/src/nmp/intake/api/v2/health/endpoints.py
```diff
@@ -44,19 +41,11 @@ async def health_live() -> dict:
...
-    Checks entity store connectivity.
+    Health check endpoint to verify the status of the application.
     """
-    try:
-        # Try a simple query to verify entity store connectivity
-        await entities_client.list(App, page=1, page_size=1)
-        return {"status": "ready", "entity_store": "connected"}
-    except Exception:
-        return {"status": "not_ready", "entity_store": "disconnected"}
+    return {"status": "ready"}
```
asutermo: does having a clickhouse check matter here too?

asutermo (approved):

services/intake/src/nmp/intake/api/v2/health/endpoints.py
BrianNewsom: The answer long term is yes, but right now it starts up w/o clickhouse so we can lazily provision as to prevent breaking running instances (not totally failing on startup)..

So leaving this as is for now.

services/intake/alembic/versions/2025_10_14_1101-a7e3b1e7355b_initial_migration.py
asutermo: I guess I ask because if this is tied to a greater alembic setup or is just intake specific

BrianNewsom: it's a good question. This migration doesn't "revise" anything so I basically think this is local to intake only (even though it's technically shared storage)

BrianNewsom: Confirmed this is local to intake.

PR #90, opened 2026-05-28
feat(build): Cleanup vendor script, add new `nemo-platform[all]` group

matthewgrossman:
Closes [AIRCORE-654](https://linear.app/nvidia/issue/AIRCORE-654/add-all-extra-alias-for-packaged-nemo-platform-installs).

## Summary

Adds a new `nemo-platform[all]` extra as the recommended user-facing install path, and uses the opportunity to clean up `vendor_package.py`.

### `nemo-platform[all]` extra

`pip install nemo-platform[all]` now resolves to the same dependency set as `[services]`. The existing `[services]` extra is preserved for backwards compatibility. User-facing docs, tutorials, and the CLI's "missing pyleak" hint message have all been updated to recommend `[all]`.

`all` is declared as a hand-written extra directly in `packages/nemo_platform/pyproject.toml`, sitting alongside the auto-generated extras in the same table.

### Vendor flow simplification

The vendor flow already supported a "comment-marked = generated, no-comment = hand-written" convention for `[project.optional-dependencies]`, but the implementation had grown to several passes (annotate, reset-marked, copy-from-bundle, sort) with a lot of incidental complexity around carrying comment trivia through tomlkit table rewrites.

This PR consolidates that into a single function (`_refresh_bundle_owned_optional_dependencies`) that:

- Walks the existing table once.
- Classifies each key by whether it has the `# Generated from [tool.bundle-package]; do not edit by hand.` marker comment immediately above it.
- Refreshes vendor-owned keys (those whose names are still produced by `[tool.bundle-package]`).
- Drops stale vendor-owned keys (marker present but bundle no longer claims the name).
- Preserves hand-written keys (no marker) in their original position.
- Emits vendor-owned keys alphabetically below the hand-written ones.

The marker comment is the load-bearing signal that distinguishes vendor-owned from hand-written. This is the same convention the vendor flow was already using, just implemented more directly.

### Dead code removal

Removed ~10 helpers in `vendor_package.py` that had no callers (`_merge_project_scripts`, `_merge_project_entrypoints`, four `_remove_*` script/entry-point helpers, `_merge_client_source_imports`, `_build_module_rewrites`, `_remove_empty_generated_project_tables`, `_merge_project_optional_dependencies`). Some were dead before this PR; some became dead as a result of the refresh-function consolidation.

Net: `vendor_package.py` is ~110 lines smaller.

## Verification

- `make vendor` is idempotent on the wrapper, `nemo-data-designer`, and `nemo-platform-plugin`.
- Stale-cleanup verified by injecting a fake `# Generated...`-marked extra in both wrapper and plugin pyprojects; both are removed on the next vendor run.
- Hand-written extras (`all` on the wrapper, `test` on `nemo-data-designer`) survive across vendor runs.
- 38/38 services CLI tests pass (`packages/nemo_platform_ext/tests/cli/commands/test_services.py`).
- `uv run pre-commit run -a` passes (ruff, format, ty, uv lock).
- `uv lock --check` confirms `nemo-platform[all]` resolves to the same dependency set as `[services]`.

## Acceptance criteria (from AIRCORE-654)

- [x] `uv tool install 'nemo-platform[all]'` installs the same service-capable package dependencies currently provided by `nemo-platform[services]`.
- [x] Existing `nemo-platform[services]` installs continue to work.
- [x] User-facing docs and setup output prefer `nemo-platform[all]` for the full packaged install path.
- [x] Lock and packaging metadata updated consistently.

## Summary by CodeRabbit

* **Documentation**
  * Updated all installation instructions to use `nemo-platform[all]` instead of `nemo-platform[services]` across CLI docs, tutorials, and SDK guidance.

* **Tests**
  * Updated test assertions to reflect the new installation command references.

* **Chores**
  * Updated package configuration and CLI error hints to reference the new installation extras.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/90?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/nemo_platform/pyproject.toml
```diff
@@ -31,19 +31,18 @@ classifiers = [
...
   "nmp-common",
 ]
 
-# AUTO-GENERATED by `make vendor` — do not edit by hand.
-# optional-dependencies, scripts, and entry-points are written
-# by tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/vendor/vendor_package.py.
-# Changes below this comment will be overwritten on the next `make vendor` run.
-
 [project.optional-dependencies]
+all = ["nemo-platform[services]"]
```
matthewgrossman, line 44: this was really the only "manual" change in this whole PR

mckornfield: sed doing the heavy liftin'

mckornfield: why don't we just rename services to all where it is? or better to indirect?

uv.lock
```diff
@@ -5057,6 +5057,94 @@ aiohttp = [
...
 ]
```
mckornfield, line 5059: lol funny how many lines it adds to just copy paste an extra

mckornfield (approved):

packages/nemo_platform/pyproject.toml
matthewgrossman: ya it's a fair question; I think there exists a future where we do want some indirection, so I did this instead ¯\_(ツ)_/¯

uv.lock
matthewgrossman: right, uv be crazy

packages/nemo_platform/pyproject.toml
matthewgrossman: meaning, it might not always be `services` that we want, but perhaps something else, or services AND something else. So I like `all` being a "virtual" extra I guess?

PR #91, opened 2026-05-28
fix(security): address high+ CodeQL alerts across plugins, services, SDK

marcusds:
## Summary

Addresses **28 of 66 open high+ CodeQL findings**. Remaining 38 `py/clear-text-logging-sensitive-data` alerts are confirmed false positives (they log secret references like `workspace/name` or already-masked values, never secret material) and are left open for triage in the GitHub Security tab.

## Alerts addressed

| Severity | Rule | # | Problem | How we addressed it |
| --- | --- | --- | --- | --- |
| Critical | `py/partial-ssrf` | 1 | `nemo-agents` proxy concatenated `endpoint + "/" + trailing_uri`, where `trailing_uri` is a user-controlled FastAPI path param. A value like `//evil.com/x` or `http://evil.com/x` could shift the request to a different host (SSRF). | Switched the join to `urljoin(endpoint + "/", trailing_uri)`, then parsed the result and rejected with **HTTP 400** if `scheme` or `netloc` no longer matches the trusted deployment endpoint. Added a parametrized unit test (`test_cross_origin_trailing_uri_rejected`) that drives `_proxy` directly with hostile inputs (`starlette` normalizes the URL before route handlers see it, so the test calls the proxy function rather than the route). |
| High | `py/tarslip` | 4 | Anonymizer / evaluator / data-designer SDKs unpacked job artifact tarballs with `tar.extractall(path=...)`. The anonymizer and evaluator validated members in a separate loop before extracting, but CodeQL can't trace per-member validation to a bulk `extractall`; the data-designer SDK had no validation at all. | Refactored all sites to extract member-by-member: validate each member's resolved path stays under `output_path`, reject `issym()`/`islnk()` and non-file/non-dir types, then call `tar.extract(member, path=output_path)`. Added a `_safe_extract_tar` helper to the data-designer SDK (sync + async paths) since it had no validation. |
| High | `py/jinja2/autoescape-false` | 3 | `services/core/models` schema validator, inference-gateway auth-header renderer, and the SDK CLI generator all instantiated `jinja2.Environment()` without an `autoescape` argument, which defaults to `False`. CodeQL flags this even when output is not HTML. | Imported `select_autoescape` and passed `autoescape=select_autoescape()` at each `Environment(...)` call. For all three sites the runtime value is still `False` (no HTML filenames involved), so rendering of HTTP header values / generated Python source is unchanged, but the call signals deliberate consideration to scanners. |
| High | `py/weak-sensitive-data-hashing` | 2 | `endpoint_identity()` in the evaluator SDK resilience module passed an API token (`auth_identity`) through `hashlib.sha256(...).hexdigest()[:16]` to build a stable dict key for scheduler state. CodeQL flagged this as password hashing (variable-name heuristic) regardless of `usedforsecurity=False`, because its "passwords need slow hashing" sub-rule does not respect that flag. | This is bookkeeping, not credential storage — there is no comparison-against-stored-hash. Swapped the SHA-256 truncation for `zlib.crc32(...)` formatted as 8 hex chars. CRC32 is not cryptographic, so the password-hashing rule cannot reach it; identity behaviour (stable, unique per token) is preserved. Mirrored the change to the SDK-vendored copy. |
| High | `py/incomplete-url-substring-sanitization` | 18 | Substring `"nvcr.io" in image_lower` / `"github.com" in url` checks could be spoofed by paths or repo names containing the literal host (e.g. `https://github.com.evil.example.com`, an image tag with `nvcr.io` in its repo path). 9 production sites and 9 test-assertion sites. | Production code: replaced `"nvcr.io" in registry` with `registry == "nvcr.io"` once the registry host was already extracted; added an `_image_registry_host_from_ref()` helper to the quickstart prompts so we compare an extracted host equality; replaced the agents' forge detection (`"github.com" in url`) with a new `_git_remote_host()` helper that handles both HTTPS (`urlparse`) and SSH (`user@host:path`) git URL forms. Test assertions: switched to `urlparse(call_args.kwargs["url"]).hostname == "expected.host"`. Mirrored every quickstart change in the SDK-vendored copy. |
| High | `py/clear-text-logging-sensitive-data` | 38 | CodeQL flagged 38 logging sites where variable names matching `*api_key*`, `*secret*`, `*auth*` appear in the format string. | **Not fixed in this PR.** All 38 sites were audited: every one logs only an identifier (`workspace/name`, provider name, env-var name) or a pre-masked value — none log secret material. Recommend bulk-dismissing in the Security tab with reason "false positive: only secret reference logged, not value". |

## Test plan

- [x] `uv run pytest plugins/nemo-agents/tests/unit/test_gateway.py` — 20 pass (incl. new parametrized SSRF guard)
- [x] `uv run pytest plugins/nemo-anonymizer/tests/unit/test_sdk_job_resources.py -k 'extract or tar'` — 3 pass
- [x] `uv run pytest plugins/nemo-evaluator/tests/test_sdk_job_resources.py -k 'extract or tar or safe'` — 6 pass
- [x] `uv run pytest services/core/inference-gateway/tests/unit/test_proxy.py` — 82 pass
- [x] `uv run pytest services/core/inference-gateway/tests/unit/test_model_router.py services/core/inference-gateway/tests/unit/test_openai_router.py` — 72 pass
- [x] `uv run pytest packages/nemo_platform_ext/tests/quickstart/test_prompts.py test_container.py test_quickstart_config.py` — 118 pass
- [x] `uv run pytest plugins/nemo-agents/tests/unit/test_improvement_preflight.py test_improvement_jobs.py` — 17 pass
- [x] `pnpm --filter nemo-studio-ui typecheck` — clean
- [ ] Re-run CodeQL after merge; confirm the 28 alerts close
- [ ] Bulk-dismiss the 38 `py/clear-text-logging-sensitive-data` alerts as "false positive" in the Security tab

## Summary by CodeRabbit

* **New Features**
  * Added helpers for extracting registry and git remote hosts and a shared safe tar extraction utility; auth fingerprint generation now uses a different digest.

* **Bug Fixes**
  * Registry detection and credential selection use exact host matching; routing and URL checks now compare parsed hostnames.

* **Security Enhancements**
  * Hardened proxy target validation, safer archive extraction, worker-origin checks, and preserved raw header rendering.

* **Tests**
  * Added/updated tests for strict proxy, host-matching, and registry detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/91?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
@@ -5,7 +5,7 @@
 
 from __future__ import annotations
 
-import hashlib
+import zlib
 from datetime import UTC, datetime
 from email.utils import parsedate_to_datetime
 
@@ -19,10 +19,10 @@ _TRANSIENT_STATUS_CODES = frozenset({408, 500, 502, 504})
 
 def endpoint_identity(base_url: str, model_id: str | None = None, auth_identity: str | None = None) -> str:
     """Build a stable endpoint key for scheduler state and accounting."""
-    auth_hash = ""
+    auth_fingerprint = ""
     if auth_identity:
-        auth_hash = hashlib.sha256(auth_identity.encode("utf-8"), usedforsecurity=False).hexdigest()[:16]
-    return f"{base_url}|{model_id or '_'}|{auth_hash}"
+        auth_fingerprint = format(zlib.crc32(auth_identity.encode("utf-8")), "08x")
+    return f"{base_url}|{model_id or '_'}|{auth_fingerprint}"
 
 
 def _status_code_from_exception(exc: Exception) -> int | None:
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
@@ -21,7 +21,7 @@ def endpoint_identity(base_url: str, model_id: str | None = None, auth_identity:
     """Build a stable endpoint key for scheduler state and accounting."""
     auth_hash = ""
     if auth_identity:
-        auth_hash = hashlib.sha256(auth_identity.encode("utf-8")).hexdigest()[:16]
+        auth_hash = hashlib.sha256(auth_identity.encode("utf-8"), usedforsecurity=False).hexdigest()[:16]
     return f"{base_url}|{model_id or '_'}|{auth_hash}"
 
 
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
@@ -5,7 +5,7 @@
 
 from __future__ import annotations
 
-import zlib
+import hashlib
 from datetime import UTC, datetime
 from email.utils import parsedate_to_datetime
 
@@ -21,7 +21,7 @@ def endpoint_identity(base_url: str, model_id: str | None = None, auth_identity:
     """Build a stable endpoint key for scheduler state and accounting."""
     auth_fingerprint = ""
     if auth_identity:
-        auth_fingerprint = format(zlib.crc32(auth_identity.encode("utf-8")), "08x")
+        auth_fingerprint = hashlib.blake2b(auth_identity.encode("utf-8"), digest_size=8).hexdigest()
     return f"{base_url}|{model_id or '_'}|{auth_fingerprint}"
 
 
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/git_url.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/git_url.py
@@ -1,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Helpers for parsing git remote URLs (HTTPS and SSH forms)."""
+
+from __future__ import annotations
+
+from urllib.parse import urlparse
+
+
+def git_remote_host(url: str) -> str:
+    """Return the hostname from a git remote URL, or "" if unparseable.
+
+    Handles both schemed URLs (``https://github.com/org/repo``,
+    ``ssh://git@github.com/org/repo``) and SSH alt form (``git@github.com:org/repo``).
+    """
+    if "://" in url:
+        return (urlparse(url).hostname or "").lower()
+    if "@" in url and ":" in url:
+        return url.split("@", 1)[1].split(":", 1)[0].lower()
+    return ""
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/git_url.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/git_url.py
@@ -11,11 +11,21 @@ from urllib.parse import urlparse
 def git_remote_host(url: str) -> str:
     """Return the hostname from a git remote URL, or "" if unparseable.
 
-    Handles both schemed URLs (``https://github.com/org/repo``,
-    ``ssh://git@github.com/org/repo``) and SSH alt form (``git@github.com:org/repo``).
+    Handles schemed URLs (``https://github.com/org/repo``,
+    ``ssh://git@github.com/org/repo``) and SCP-style alt form, both with
+    (``git@github.com:org/repo``) and without (``github.com:org/repo``) a
+    user prefix. Windows-style absolute paths (``C:\\repo``) are rejected.
     """
     if "://" in url:
         return (urlparse(url).hostname or "").lower()
-    if "@" in url and ":" in url:
-        return url.split("@", 1)[1].split(":", 1)[0].lower()
+    if ":" in url:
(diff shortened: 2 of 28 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/resilience/classifier.py
```diff
@@ -21,7 +21,7 @@
     """Build a stable endpoint key for scheduler state and accounting."""
     auth_hash = ""
     if auth_identity:
-        auth_hash = hashlib.sha256(auth_identity.encode("utf-8")).hexdigest()[:16]
+        auth_hash = hashlib.sha256(auth_identity.encode("utf-8"), usedforsecurity=False).hexdigest()[:16]
```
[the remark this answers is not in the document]
marcusds: Claude says: Pushed follow-up — swapped `hashlib.sha256(..., usedforsecurity=False)` for `zlib.crc32`. CodeQL's `py/weak-sensitive-data-hashing` has a separate sub-rule ("password hashing must be slow") that the `usedforsecurity=False` flag does not clear; the variable name `auth_identity` triggers the password taint regardless. The hash here is a scheduler bookkeeping identifier (used as a dict key for retry/rate-limit state) rather than credential storage, so CRC32 is the right tool: not cryptographic, no password-hash rule can fire on it.

sdk/python/nemo-platform/src/nemo_platform/beta/evaluator/resilience/classifier.py
```diff
@@ -21,7 +21,7 @@
     """Build a stable endpoint key for scheduler state and accounting."""
     auth_hash = ""
     if auth_identity:
-        auth_hash = hashlib.sha256(auth_identity.encode("utf-8")).hexdigest()[:16]
+        auth_hash = hashlib.sha256(auth_identity.encode("utf-8"), usedforsecurity=False).hexdigest()[:16]
```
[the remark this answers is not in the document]
marcusds: Claude says: Pushed follow-up — swapped `hashlib.sha256(..., usedforsecurity=False)` for `zlib.crc32`. CodeQL's `py/weak-sensitive-data-hashing` has a separate sub-rule ("password hashing must be slow") that the `usedforsecurity=False` flag does not clear; the variable name `auth_identity` triggers the password taint regardless. The hash here is a scheduler bookkeeping identifier (used as a dict key for retry/rate-limit state) rather than credential storage, so CRC32 is the right tool: not cryptographic, no password-hash rule can fire on it.

plugins/nemo-data-designer/src/nemo_data_designer_plugin/sdk/job_resources.py
```diff
@@ -58,6 +58,20 @@ def _raise_for_status(resp: httpx.Response) -> None:
         raise DataDesignerJobError(detail, status_code=status_code) from exc
 
 
+def _safe_extract_tar(tar: tarfile.TarFile, output_path: Path) -> None:
```
mckornfield, line 62: ugh I just hate tar vulns

mckornfield: should this be lifted somewhere higher? I saw a similar but not the same one in another method, but realize there's different exceptions

services/core/models/src/nmp/core/models/schemas.py
```diff
@@ -337,7 +337,7 @@ def _validate_auth_header_format(v: str | None) -> str | None:
     """
     if v is None:
         return v
-    env = Environment()
+    env = Environment(autoescape=select_autoescape())
```
mckornfield: dumb q but do we have tests for these select_autoescape things we're turning on?

mckornfield (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/git_url.py
```diff
@@ -0,0 +1,21 @@
...
+def git_remote_host(url: str) -> str:
+    """Return the hostname from a git remote URL, or "" if unparseable.
+
+    Handles both schemed URLs (``https://github.com/org/repo``,
+    ``ssh://git@github.com/org/repo``) and SSH alt form (``git@github.com:org/repo``).
+    """
+    if "://" in url:
+        return (urlparse(url).hostname or "").lower()
+    if "@" in url and ":" in url:
+        return url.split("@", 1)[1].split(":", 1)[0].lower()
```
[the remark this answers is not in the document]
marcusds: Claude says: Good catch. Fixed in `ca6f9b9e10` — git_remote_host() now accepts bare `host:path` SCP form too, with a dot-in-host requirement so Windows drive letters (`C:\repo`) are rejected. Added 14 parametrized unit tests in `packages/nemo_platform_plugin/tests/test_git_url.py` covering the three suggested cases (`github.com:org/repo.git`, `gitlab.com:group/repo.git`, `C:\repo`) plus HTTPS / SSH / mixed-case variants.

plugins/nemo-data-designer/src/nemo_data_designer_plugin/sdk/job_resources.py
marcusds: safe_extract_tar now lives in packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/archive.py and takes an error_cls factory, so each plugin keeps its own exception type

services/core/models/src/nmp/core/models/schemas.py
marcusds: This got reverted.

mckornfield (approved):

PR #94, opened 2026-05-28
feat(agents): typed AgentSpec + explore-first onboarding chain (FP-161)

aleckhoury:
## Summary

Make the agent spec a typed, server-side-visible artifact instead of an unstructured local markdown file, and rebuild the onboarding skill chain around it.

- **New `AgentSpec` Pydantic model** (`nemo_agents_plugin.spec`) with hard validators on `Job` (vague-phrase rejection, min length) and `Framework` (`langgraph-nat` vs `needs-wrapper` enum). 12 body sections + 2 front-matter fields, sourced from the POR's existing + proposed sections table.
- **New `spec_render` module** with bidirectional `AgentSpec` ↔ markdown round-trip. Strict parser: rejects unknown sections, duplicate sections, and malformed labeled bullets before the value reaches the Pydantic model.
- **`spec_file_ref: FilesetRef | None` on the `Agent` entity** and `CreateAgentRequest`, with a `--spec-file-ref` flag on `nemo agents create`. Lets `nemo agents list` answer "which agents have an onboarded spec?" with a single field check.
- **`nemo-explore` SKILL.md** rewritten to be explore-first and schema-driven: scans the codebase, fills every inferable field, presents the full spec in a single review block, and only asks for gaps. Replaces the hardcoded 12-question Q&A loop.
- **`nemo-spec` SKILL.md** rewritten to validate against `AgentSpec`, refuse to write when `Job` or `Framework` is unresolved, upload the canonical copy to a per-agent Filesets fileset (Fileset is canonical, local file is a write-through cache), and hand `SPEC_FILE_REF` to `nemo-build-agent`.
- **`nemo-build-agent` SKILL.md** updated to thread `SPEC_FILE_REF` through all three `nemo agents create` invocations.
- **Spec template** rewritten to match the new schema + renderer. Drops the `Known Issues` section (handled by the Insights plugin's first-class entity instead).
- **Plugin OpenAPI** regenerated; **SDK skill copies** re-vendored.

## Test plan

- [x] 45 new unit tests across `test_spec.py` and `test_spec_render.py` covering schema, validators, JSON-schema export, render output shape, round-trip identity, and structural parse errors
- [x] `test_entities.py` / `test_cli.py` extended for `spec_file_ref` on `Agent`, `CreateAgentRequest`, and the new CLI flag
- [x] 102 tests pass across the touched files (`pytest plugins/nemo-agents/tests/unit/test_{spec,spec_render,entities,cli,agents_api}.py`)
- [x] `ruff check` + `ruff format --check` clean
- [x] `ty check` clean on changed files (two pre-existing diagnostics in `setup.py` unrelated, verified via stash)
- [x] `pre-commit run -a` passes all hooks except `studio-lint-staged` (pre-existing local node-version mismatch, only fires under `-a`)
- [x] Exercised end-to-end against an external AIQ checkout (NVIDIA-AI-Blueprints/aiq) via skill symlinks — `nemo-explore` codebase scan → single review pass → `nemo-spec` validate + Filesets upload → `nemo-build-agent` create with `--spec-file-ref` populated

## Notes for reviewers

- Field-level guidance ("what good looks like") deliberately lives in `nemo-explore/SKILL.md`, not in the schema docstrings, so the skill stays the single source of truth for interview prompts and the schema stays terse.
- `Framework` is marked planned-deprecation in JSON schema (`x-planned-deprecation`) — expected to relax as NAT supports more frameworks; flagged so future-us can delete the field cleanly.
- The `Known Issues` section was deliberately dropped from the schema per the design discussion: known issues / recurring failure patterns are first-class `Insight` entities owned by the insights plugin, not duplicated into markdown.
- Stacked: [PR2 (#pending)](#) follows up with platform-startup default fixes discovered while testing this end-to-end.

## Summary by CodeRabbit

* **Documentation**
  * Updated skill workflows (nemo-explore, nemo-spec, nemo-build-agent, nemo-skill-selection) with refined agent creation and specification processes.
  * Redesigned agent spec template with new structure, front-matter fields, and required sections for improved spec clarity and validation.

* **Chores**
  * Migrated agent spec file location from `agents/<name>.spec.md` to `agents/<name>-spec/AGENT-SPEC.md` across all documentation and examples.
  * Added spec schema validation and canonical spec location convention helpers.


plugins/nemo-agents/src/nemo_agents_plugin/entities.py
```diff
@@ -38,6 +39,16 @@ class Agent(NemoEntity, entity_type="agent"):
             "Currently only 'nat-workflow-v1' is supported."
         ),
     )
+    spec_file_ref: FilesetRef | None = Field(
```
nicot: Is it possible to have a unique spec defined by the agent registration name so we don't need an env variable for this or to save the path? for example fileset://workspace-id/specs/my-agent.spec.md

sdk/python/nemo-platform/src/nemo_platform/skills/nemo-build-agent/SKILL.md
```diff
@@ -41,16 +41,29 @@ NeMo Platform optimizes LangGraph agents wrapped in NVIDIA NeMo Agent Toolkit (N
...
+
+   ```bash
+   SPEC_FILE_REF="${SPEC_FILE_REF:-default/${AGENT_NAME}-spec}"
+   nemo files filesets get "${SPEC_FILE_REF#*/}" >/dev/null 2>&1 \
+     && echo "spec_fileset_ok" \
+     || echo "spec_fileset_missing — run nemo-spec to upload before continuing"
+   ```
+
+   If the fileset is missing, route back to `nemo-spec`. The ref is passed to every `nemo agents create` call below so the platform Agent entity is linked to the spec from creation.
+6. Check for an existing deployment: `.venv/bin/nemo agents deployments list 2>/dev/null | grep -q "$AGENT_NAME"`. If the agent is already deployed, ask the user whether to skip (idempotent path) or redeploy.
```
nicot, line 48: It seems like a lot of this content is written down in 3 or more files, is there a way to consolidate it? And/or reference the canonical skill from the other skills?

nicot (approved):
Looking great!

PR #98, opened 2026-05-28
fix(codeql): clear error-severity code-quality alerts (batch 2)

marcusds:
## Summary

Batch 2 of the CodeQL cleanup. Addresses **15 of 16 open error-severity code-quality alerts** (excludes 12 `py/unsafe-cyclic-import` refactor work). The remaining 1 (`py/illegal-raise` in vendored switchyard) was dismissed via API since direct edits get overwritten on the next snapshot refresh.

This branch follows [#91](https://github.com/NVIDIA-NeMo/nemo-platform/pull/91) (high+ security alerts) and targets the next tier — error-level diagnostics that surface real correctness issues rather than security policy.

## Alerts addressed

| Severity | Rule | # | Problem | How we addressed it |
| --- | --- | --- | --- | --- |
| Error | `py/uninitialized-local-variable` | 8 | Local variables (`df`, `benchmark_ref`, `entities`/`total`, `nss_job_config`, `client`, `commands`) read on paths CodeQL couldn't prove unreachable — e.g. `match`/`case` with a `raise` wildcard, `if`/`elif` chains without `else`, or `pytest.fail()`/`pytest.skip()` that the analyzer doesn't recognize as terminating. | `safe_synthesizer_sdk/job_builder.py`: swapped `match … case pd.DataFrame() as df` for explicit `isinstance` branches so `df` is assigned unambiguously. `safe_synthesizer/__main__.py`: same swap for the dict/wildcard match. `evaluator/metric_results.py`: added exhaustive `else: raise TypeError` so `benchmark_ref` is always set. `entities/endpoints.py`: hoisted `raise_if_workspace_inaccessible` to fail fast before the entities/total assignment. `nmp_testing/docker.py`: replaced `pytest.fail(…)` with explicit `raise pytest.fail.Exception(…) from e`. `jobs-execute-gpu-cli/tests`: added `return` after `pytest.skip()` so the unreachable tail is visible. |
| Error | `py/call/wrong-arguments` | 2 | `PatternSpec.mask = lambda s: _mask_middle(s)` flagged as a 0-arg "method" when called with `value`; `print_verbose("Unused schemas: ", ", ".join(...), style=...)` called with 2 positional args against a 1-arg signature. | `pii_scan.py`: moved `_mask_middle` above `PatternSpec` and used the named function as the dataclass default (no lambda). `openapi_tools.py`: concatenated the two message args into one f-string. |
| Error | `py/undefined-export` | 2 | `MOCK_NIM_IMAGE_TAG` and `MOCK_SIDECAR_IMAGE_TAG` were listed in `nmp_testing/__init__.py`'s `__all__` but never re-exported from `docker`. | Added both names to the `from .docker import (…)` block. |
| Error | `py/missing-call-to-init` | 1 | `ModelRouterMiddleware` overrode `__init__` without invoking `NemoInferenceMiddleware`'s. | Added `super().__init__()` as the first line. |
| Error | `py/side-effect-in-assert` | 1 | `assert pf.remove("status") is None` would lose its `remove` side-effect under `pytest -O`. | Split into a side-effecting call followed by a pure assert on the result. |
| Error | `py/call/wrong-named-class-argument` | 1 | `test_rejects_legacy_backend_argument` deliberately calls `Evaluator(backend=backend)` to verify the legacy kwarg is rejected — CodeQL flags it because `backend=` is not a real parameter. | Routed the bad kwarg through a `**kwargs` dict so the call is dynamic; runtime behaviour (`pytest.raises(TypeError, match="backend")`) unchanged. |
| Error | `py/illegal-raise` | 1 | `raise last_exc` where `last_exc` could be `None`, in vendored `switchyard/lib/backends/latency_service_llm_backend.py`. | **Dismissed via API as "won't fix"** — vendored upstream code. Edits would be overwritten on the next snapshot refresh. Track upstream fix separately. |

## Drive-by typecheck fixes

The `ty` pre-commit hook surfaces existing diagnostics in any file touched by a commit. Since this PR touches `metric_results.py` and `job_builder.py`, the following pre-existing errors are also fixed:

- `metric_results.py`: union-narrowed `job: MetricJob | BenchmarkJob` with `typing.cast` at the call sites; wrapped `benchmark_ref` in `app.BenchmarkRef(...)` so the typed Pydantic model matches the field signature instead of relying on implicit `str → RootModel` coercion.
- `job_builder.py`: moved the `# type: ignore` from the trailing `**kwargs` line (where it was unused) to the actual `spec=spec` argument (where the dict-vs-`SafeSynthesizerJobConfigParam` mismatch lives).

## Test plan

- [x] `uv run pytest services/core/entities/tests/ -k list` — 20 pass
- [x] `uv run pytest services/safe-synthesizer/tests/` — same 13 pre-existing failures as `main` (test_jobs / test_model_init / test_task_upload_results); no new failures
- [x] `uv run pytest packages/nmp_testing/tests/` — same 6 pre-existing failures as `main`; 61 pass (up from 8 on `main` because more files compile)
- [x] `uv run pytest packages/nemo_evaluator_sdk/tests/execution/test_evaluator.py::TestEvaluator::test_rejects_legacy_backend_argument` — pass (still verifies `TypeError, match="backend"`)
- [x] `uv run pytest packages/nmp_common/tests/api/test_parsed_filter.py` — pass
- [x] `ruff check` on all touched files — clean
- [x] `uv run --frozen --extra cpu ty check` on touched files — clean
- [ ] Re-run CodeQL after merge and confirm the 15 alerts close

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust Docker client failure reporting and stricter workspace access checks in APIs.
  * Clearer errors for unsupported benchmark types.

* **Refactor**
  * Simplified job configuration datasource handling and improved type casting.
  * Cleaner schema output formatting and initialization of middleware.

* **Tests**
  * Stabilized tests to avoid spurious failures and clarified assertions/skip behavior.

* **Chores**
  * Adjusted test exports and updated default PII masking behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA-NeMo/nemo-platform/pull/98?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/nmp_testing/src/nmp/testing/__init__.py
```diff
@@ -46,7 +46,9 @@
 from .client import TEST_ADMIN_EMAIL, TEST_USER_EMAIL, ClientContext, create_test_client
 from .docker import (
     DEFAULT_RETRY_CONFIG,
+    MOCK_NIM_IMAGE_TAG,
```
mckornfield: lol was something importing them from here?

packages/safe_synthesizer_sdk/src/safe_synthesizer_sdk/job_builder.py
```diff
@@ -227,13 +228,12 @@ def with_hf_token_secret(self, secret_name: str) -> Self:
...
         if self._data_source_path is not None:
             return  # already uploaded; reuse the cached result
-        match self._data_source:
-            case pd.DataFrame() as df:
-                pass
-            case str(url):
-                df = pd.read_csv(url, **kwargs)
-            case _:
-                raise ValueError("Data source must be a pandas DataFrame or a URL")
+        if isinstance(self._data_source, pd.DataFrame):
```
mckornfield, line 231: does it not like match / case?

packages/safe_synthesizer_sdk/src/safe_synthesizer_sdk/job_builder.py
mckornfield, line 1: probably fine, I'm about to delete this I believe

services/safe-synthesizer/src/nmp/safe_synthesizer/tasks/safe_synthesizer/__main__.py
```diff
@@ -302,12 +302,10 @@ def run_task():
     enable_synthesis: bool = job_config.get("enable_synthesis", True)
     logger.info(f"enable_synthesis={enable_synthesis}")
 
-    nss_job_config: SafeSynthesizerJobConfig
-    match job_config:
-        case dict():
-            nss_job_config = SafeSynthesizerJobConfig.model_validate(job_config)
-        case _:
-            raise ValueError(f"Config must be a dictionary or a string: {job_config}")
+    if isinstance(job_config, dict):
```
mckornfield, line 305: maybe it's the count of match case things?

mckornfield (approved):
match case sad. also I have an nss pr that will probably delete some of these files but I'll just rebase

packages/safe_synthesizer_sdk/src/safe_synthesizer_sdk/job_builder.py
marcusds: It likes be diffucult.

PR #111, opened 2026-05-29
chore(studio): Copy BE wiring for coding agents in Studio

dmariali:
## Summary by CodeRabbit

* **New Features**
  * Local coding-agent bridge: create/manage sessions, stream agent output via SSE, and retrieve session history
  * MCP JSON-RPC endpoint and interactive tool permission flow for runtime approvals

* **Bug Fixes**
  * Suppress sensitive startup errors during streaming and ensure session stream cleanup on failure

* **Documentation**
  * Added feature-flag config to enable coding-agent functionality

* **Tests**
  * Unit tests covering sessions, history, MCP flows, streaming, and feature-flag behavior


services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+    updated_input: dict[str, Any] | None = None
+
+
+class HistorySessionResponse(BaseModel):
+    """Summary of a Claude session stored on disk."""
+
+    session_id: str
+    mtime: float
+    first_prompt: str
+    message_count: int
```
mckornfield, line 62: Is something like token count/ tool calls also useful?

services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+    app.add_api_route(
+        PUBLIC_MCP_PATH,
+        mcp_endpoint,
+        methods=["POST"],
+        name=PUBLIC_MCP_ROUTE_NAME,
+        include_in_schema=False,
+    )
+
+
+def _validate_session_id(session_id: str) -> str:
```
mckornfield, line 117: We don't have to do some sort of lookup/ security? Or is that done at the api

services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+                    entry = json.loads(line)
+                except json.JSONDecodeError:
+                    continue
+                if entry.get("type") != "user" or entry.get("isSidechain"):
+                    continue
+                content = (entry.get("message") or {}).get("content")
+                if not isinstance(content, str):
+                    continue
+                count += 1
+                if first_prompt is None:
```
mckornfield: Do we need to json load the other lines if we have first prompt already?

services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+        "--mcp-config",
+        mcp_config,
+        "--permission-prompt-tool",
+        f"mcp__{CLAUDE_MCP_SERVER_NAME}__approval_prompt",
+        session_flag,
+        session_id,
+    ]
+
+
+def _claude_env() -> dict[str, str]:
```
mckornfield, line 379: Is there a case where we'd want to set it via env var?

services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+
+    if method == "tools/call":
+        name = params.get("name")
+        args = params.get("arguments") or {}
+        if name != "approval_prompt":
+            return JSONResponse(
+                {
+                    "jsonrpc": "2.0",
+                    "id": request_id,
+                    "error": {"code": -32601, "message": f"unknown tool: {name}"},
```
mckornfield, line 620: is this a standard error code?

services/studio/src/nmp/studio/coding_agents.py
```diff
@@ -0,0 +1,546 @@
...
+            }
+        )
+
+    return JSONResponse(
+        {
+            "jsonrpc": "2.0",
+            "id": request_id,
+            "error": {"code": -32601, "message": f"method not found: {method}"},
+        }
+    )
```
mckornfield, line 641: one thing to consider about this, and maybe for a follow up: it's very claude code specific, we probably want an interface that is implemented by something claude code specific, codex and then maybe a third (opencode?) should we support those

mckornfield (approved):
mostly nitty things, think the next time this coding_agents file is touched we probably need an abstraction above claude code. we can use tiktoken too or something like it if we want token estimates (or it might be stored since I know claude displays it)

services/studio/src/nmp/studio/coding_agents.py
dmariali: I think we want this MCP callback as an auth bypassed route so Claude CLI can call it.

services/studio/src/nmp/studio/coding_agents.py
dmariali: I think for now we can do it this way till Manjesh does the full setup to do it all properly

services/studio/src/nmp/studio/coding_agents.py
dmariali: yeah I believe it's the model not found error code.

services/studio/src/nmp/studio/coding_agents.py
dmariali: ah yes - Manjesh is gonna be working on an agnostic interface - for the first stage of this we decided to just at least get it working with Claude.

PR #112, opened 2026-05-29
feat(evaluator): support SDK field mappings

SandyChapman:
Adds SDK field_mapping support for evaluator runs and propagates it through local/plugin backends, including vendored SDK updates.

## Summary by CodeRabbit

* **New Features**
  * Exposes FieldMapping at the SDK surface and adds optional field-mapping support to evaluator APIs (run, run_sync, submit), propagating mappings through local and plugin execution paths.
  * Offline evaluation now builds richer per-row samples so metric workers receive populated sample data instead of empty payloads.

* **Tests**
  * Added and updated tests to validate field-mapping propagation and offline sample population during evaluations.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
```diff
@@ -38,6 +38,19 @@
 _CANDIDATE_SAMPLE_FIELDS = frozenset({"output_text", "response", "trajectory"})
 
 
+def build_offline_sample(row: dict[str, Any]) -> dict[str, Any]:
```
ngoncharenko: not sure if this is the right file for this function. Maybe it should be in utils?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/scoring.py
SandyChapman: I pulled it out to a new package called `samples.py`

PR #121, opened 2026-06-01
chore(evaluator): detangle plugin from legacy service

SandyChapman:
## Summary

- move plugin-only evaluator pieces, including FilesetRef and plugin examples, out of the standalone SDK
- remove active SDK/plugin references to the legacy evaluator service package
- keep the accepted nmp-common plugin dependency for get_qualified_image/get_task_sdk
- fix bundle metadata generation so stale generated extras stop retaining removed service deps

## Validation

- `uv run ruff check packages/nemo_evaluator_sdk plugins/nemo-evaluator tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/vendor/vendor_package.py tools/nemo-platform-sdk-tools/tests/sdk/vendor/test_vendor_package.py`
- `uv run --frozen --extra cpu ty check ...` focused evaluator files
- `uv run pytest packages/nemo_evaluator_sdk/tests/test_params.py plugins/nemo-evaluator/tests/test_sdk.py plugins/nemo-evaluator/tests/test_evaluate_job.py tools/nemo-platform-sdk-tools/tests/sdk/vendor/test_vendor_package.py -q`
- `uv run pytest packages/nemo_evaluator_sdk/tests/execution/test_evaluator.py plugins/nemo-evaluator/tests/test_standalone_sdk_backend.py -q`
- `tools/lint/lint-all.sh`

## Release Notes

* **New Features**
  * Added FilesetRef type with fragment composition support for dataset references
  * Improved task execution with typed exit codes and graceful SIGTERM shutdown handling

* **Refactor**
  * Consolidated SDK dataset types and utilities to SDK-owned modules
  * Removed remote execution support, focusing on local evaluator workflows
  * Enhanced task initialization error handling with explicit logging

* **Documentation**
  * Clarified backend adapter ownership in execution design notes

* **Chores**
  * Removed legacy migration comments across modules
  * Updated dependency declarations and improved dependency group management


packages/nemo_evaluator_sdk/examples/examples.py
SandyChapman, line 1: I pulled out the plugin specific examples from this file and instead put them with the plugin.

ngoncharenko: I iterated on skills and ended up with a recommended short version of a skill examples. I will sync with your changes

https://github.com/NVIDIA-NeMo/nemo-platform/blob/ngoncharenko/metric-types/skills/nemo-evaluator-plugin/assets/examples/plugin_sdk_examples.py

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
```diff
@@ -3,9 +3,6 @@
 
 """Result types for evaluator SDK runtime."""
 
-# Migrated from: services/evaluator/src/nmp/evaluator/app/values/results.py
```
ngoncharenko, line 6: thanks for cleaning this up

plugins/nemo-evaluator/src/nemo_evaluator/jobs/utils.py
```diff
@@ -6,19 +6,133 @@
...
+    return parse_fileset_ref(dataset.root, workspace_fallback=None)
+
+
+def _base_download_path(destination: str, workspace: str, fileset: str) -> Path:
+    return Path(destination) / workspace / fileset
+
+
+async def dataset_exists(sdk: AsyncNeMoPlatform, dataset: FilesetRef) -> bool:
+    """Return whether a persisted fileset reference resolves through the Files API."""
+    fs = FilesetFileSystem(sdk=sdk)
```
ngoncharenko, line 33: we could organize the functions under DatasetProcessor (or other name) class that will take sdk in init and initialize `fs = FilesetFileSystem(sdk=sdk)`. Then, we will methods dataset_exists, download_dataset, download_dataset_sync and don't need to pass sdk everywhere

plugins/nemo-evaluator/src/nemo_evaluator/sdk/__init__.py
```diff
@@ -26,3 +30,24 @@
     "FilesetRef",
     "PluginDatasetInput",
 ]
+
+
+def __getattr__(name: str) -> object:
+    """Load SDK resources lazily so type-only imports do not initialize jobs."""
```
ngoncharenko: would nice to expand on this - what chain exactly we want to prevent from being imported eagerly and link where the lightweight import is actually used. This is for humans and also for agents, so they don't remove it in the future

packages/nemo_evaluator_sdk/examples/examples.py
SandyChapman: Thanks, happy to make adjustments with my PR if yours lands first.

plugins/nemo-evaluator/src/nemo_evaluator/tasks/evaluate/__main__.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def main() -> int:
+    """Build the task SDK and dispatch to the evaluator plugin job."""
+    signal.signal(signal.SIGTERM, _shutdown_handler)
+    try:
+        sdk = get_task_sdk("evaluator")
+    except Exception:
+        logger.exception("Failed to build task SDK for evaluator")
+        return 2
```
ngoncharenko: What is meaning of `2` exit code? Should we make it a descriptive enum?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/utils.py
SandyChapman: I tried this locally, but if we made a `DatasetProcessor` class, its instances would not get reused in the current call graph: each caller constructs it for one operation, so the class mostly wraps the sdk argument without reducing lifecycle complexity. Since sync and async download paths are also distinct, I’d keep the explicit helper functions for now and revisit a processor if we later have a multi-step dataset workflow that can reuse the same instance/state.

ngoncharenko (approved):

plugins/nemo-evaluator/src/nemo_evaluator/sdk/__init__.py
SandyChapman: I've updated the docstring here with more details.

plugins/nemo-evaluator/src/nemo_evaluator/tasks/evaluate/__main__.py
SandyChapman: Good call. I replaced the bare `2` with `EvaluateTaskExitCode.SDK_INITIALIZATION_FAILED` and added test coverage for that failure path.

While addressing this, I also noticed `nemo_evaluator.tasks.evaluate` resolves to `tasks/evaluate.py`, not the sibling `tasks/evaluate/__main__.py`, so I moved the handling into the actual resolved entrypoint and removed the shadowed duplicate file.

PR #139, opened 2026-06-02
docs(skills): use platform health routes for probes

aleckhoury:
## Motivation
- Hit this issue when trying to document AIQ agent with NeMo Platform. My agent pulled from these skills, continually trying to get platform status from /v1/models at the root, instead of using the ready endpoint

## Change Summary
- Replace stale root `/v1/models` skill probes with the platform `/health/ready` readiness endpoint.
- Add `/status` to the `nemo-status` dashboard flow so services and controllers are surfaced from the platform health router.
- Keep source skill files and SDK-vendored skill copies in sync.


aleckhoury:
Closing because it was based on the feature branch by mistake. Replaced by https://github.com/NVIDIA-NeMo/nemo-platform/pull/140, which branches from latest main.

PR #140, opened 2026-06-02
docs(skills): use platform health routes for probes

aleckhoury:
## Motivation
- Hit this issue when trying to document AIQ agent with NeMo Platform. My agent pulled from these skills, continually trying to get platform status from /v1/models at the root, instead of using the ready endpoint

## Change Summary
- Replace stale root `/v1/models` skill probes with the platform `/health/ready` readiness endpoint.
- Add `/status` to the `nemo-status` dashboard flow so services and controllers are surfaced from the platform health router.
- Keep source skill files and SDK-vendored skill copies in sync.

## Summary by CodeRabbit

* **Documentation**
  * Standardized platform readiness checks to probe /health/ready and require HTTP 200 for “ready”.
  * Tightened interpretation so only HTTP 200 (when a listener exists) proceeds; other results route to status handling.
  * Clarified pre-flight guidance to allow using curl specifically for the health probe.
  * Enhanced status output to show Services (ready/not-ready counts) and Controllers health.
  * Updated verification and wedge-detection wording with clearer recovery instructions.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
@@ -31,11 +31,11 @@ NeMo Platform optimizes LangGraph agents wrapped in NVIDIA NeMo Agent Toolkit (N
 
 ## Pre-flight
 
-1. Confirm the platform is up using `lsof` (ground truth) + `curl` (functional). If either fails, route to `nemo-setup` and stop. Do not trust `nemo services status` — it reports stale "running" from held locks after the process has died.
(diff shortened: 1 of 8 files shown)


yamini (approved):
LGTM, approving. Nice catch on the root cause too. The old probe was reading the 404 from `/v1/models` as healthy (the regex allowed 4xx), so the agent thought the platform was up when the route just didn't exist. Moving to `/health/ready` with a strict 200 fixes that.

Couple of small things, none blocking:

- `/status` returns JSON, but the dashboard template shows formatted lines like `Controllers: healthy`. The agent has to parse `controllers.healthy` and `services.ready/not_ready` itself. Might be worth a one-liner in the skill saying the output is JSON to parse, so it doesn't expect a pre-formatted dashboard.
- During startup, services that aren't ready yet return 503, which now reads as `PLATFORM_WEDGED`. The recovery text covers "partially-started" so it's fine, just noting a mid-boot platform will show as wedged.
- Tiny cosmetic thing: the template rows are misaligned (`Services:` vs `Controllers:` have different trailing spaces).

PR #141, opened 2026-06-02
fix(cli): fail fast when services port is occupied (AIRCORE-607)

tylersbray:
## Summary

- Add bind preflight (`check_port_available_for_start`) before spawning local platform services.
- Surface port conflicts on the terminal immediately in `nemo setup`, `nemo services start`, `nemo services run`, and `nemo services restart` instead of waiting for health timeouts while bind errors sit in `services.log`.
- Distinguish foreign process vs existing NeMo instance for actionable messages (`lsof` / `nemo services stop`).

Fixes [AIRCORE-607](https://linear.app/nvidia/issue/AIRCORE-607/if-port-8080-is-occupied-nemo-setup-just-fails-after-120s-on-macos).

## Test plan

- [x] Unit tests: `test_setup.py` (`TestMaybeStartServices`), `test_services_lifecycle.py` (`TestPortAvailability`), `test_services.py` (start/run/restart port occupied)
- [x] `make vendor-nemo-platform-ext`
- [x] Manual: occupy `:8080`, run `nemo setup` with service start — expect immediate stderr error, no long health poll

## Summary by CodeRabbit

* **New Features**
  * Preflight port availability checks for start/run/restart and setup; uses configured default bind host.
  * Distinct, multi-line guidance when a required port is in use, distinguishing external processes vs. existing service instances.

* **Bug Fixes**
  * Services now fail fast on port conflicts to avoid partial or failed startups; startup is aborted with clear instructions.

* **Tests**
  * Added CLI and lifecycle tests that bind real sockets to validate port-preflight behavior and conflict messaging.


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
```diff
@@ -172,6 +176,105 @@ def __init__(self, scope: str, pid: int) -> None:
...
+        return False
+    if not infos:
+        return False
+    for family, socktype, proto, _, sockaddr in infos:
+        try:
+            with socket.socket(family, socktype, proto) as sock:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+                sock.bind(sockaddr)  # noqa: S104  # nosec B104
+            return True
+        except OSError:
```
mckornfield: could use `contextlib.suppress`

mckornfield (approved):

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
tylersbray: Done, thanks.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
```diff
@@ -172,6 +177,103 @@ def __init__(self, scope: str, pid: int) -> None:
...
+    """Return True when a live instance for *scope* is bound to *host*:*port*."""
+    if not is_instance_alive(scope, base_dir=base_dir):
+        return False
+    desc = read_descriptor(scope, base_dir=base_dir)
+    if desc is None:
+        return False
+    return desc.port == port and _normalize_bind_host(desc.host) == _normalize_bind_host(host)
+
+
+def is_port_bindable(host: str, port: int) -> bool:
```
anastasia-nesterenko, line 217: how would this work for ::1:8080 and 127.0.0.1:8080?

anastasia-nesterenko (approved):
left a small question on the localhost versions

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/services/_process.py
tylersbray: "Today we only normalize localhost → 127.0.0.1 for descriptor matching. ::1 and 127.0.0.1 are different strings, so we only classify nemo_instance when the descriptor’s host matches the host we’re checking. Default nemo setup / nemo services use 127.0.0.1, so the common path is consistent. If you start with --host ::1 while an instance is on 127.0.0.1:8080, the OS may still allow the bind (separate loopback addresses), and we wouldn’t treat it as the same NeMo instance. Happy to add loopback equivalence (127.0.0.1 / ::1 / localhost) in _normalize_bind_host for ownership checks if we want that for non-default hosts."

PR #156, opened 2026-06-03
fix: RAGAS output name alignment

mckornfield:
## Summary by CodeRabbit

* **Bug Fixes**
  * RAGAS-based metrics now reliably remap RAGAS score keys to each metric's declared output names and normalize missing/invalid values to match the declared output shape.

* **Tests**
  * Added and updated tests to cover score-name alignment (including answer and context metrics), NaN/missing-value handling, and consistent use of the topic_adherence output key across success and error scenarios.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/ragas/base.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/ragas/base.py
@@ -234,17 +234,31 @@ class BaseRAGASMetric(MetricBase):
         return getattr(self, "ignore_request_failure", False)
 
     def _nan_scores_for_metrics(self, metrics: list) -> dict[str, float]:
-        """Build a NaN score mapping for the active RAGAS metric objects."""
-        metric_names: list[str] = []
-        for metric in metrics:
-            metric_name = getattr(metric, "name", None)
-            if isinstance(metric_name, str) and metric_name:
-                metric_names.append(metric_name)
-
+        """Build a NaN score mapping using declared output_spec names."""
+        metric_names = [output.name for output in self.output_spec()]
         if not metric_names:
-            metric_names = [output.name for output in self.output_spec()]
+            for metric in metrics:
+                metric_name = getattr(metric, "name", None)
+                if isinstance(metric_name, str) and metric_name:
+                    metric_names.append(metric_name)
 
         return {metric_name: float("nan") for metric_name in metric_names}
+
+    def _align_scores_to_output_spec(self, scores: dict[str, float]) -> dict[str, float]:
+        """Map RAGAS internal metric keys (e.g. ``nv_accuracy``) to declared output names."""
+        declared = [output.name for output in self.output_spec()]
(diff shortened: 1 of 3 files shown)


arpitsardhana (approved):
mostly good with minor comments

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/ragas/base.py
```diff
@@ -234,18 +234,32 @@ def _ignore_request_failure(self) -> bool:
...
+        """Map RAGAS internal metric keys (e.g. ``nv_accuracy``) to declared output names."""
+        declared = [output.name for output in self.output_spec()]
+        if not declared or not scores:
+            return scores
+
+        if all(name in scores for name in declared):
+            return {name: scores[name] for name in declared}
+
+        undeclared = {name: value for name, value in scores.items() if name not in declared}
+        if len(declared) == 1 and len(undeclared) == 1:
```
arpitsardhana: Nit: I am worried that fix may be fragile and can break if RAGAS metrics list is appended later with more metrics. something like a static map ```json
# e.g. per subclass or a static map
RAGAS_KEY_TO_OUTPUT = {
    "nv_answer_accuracy": "answer_accuracy",
    "answer_relevancy": "response_relevancy",
    ...
}``` would be more predictable

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/ragas/base.py
```diff
@@ -327,14 +341,14 @@ def _run_evaluate(self, dataset: EvaluationDataset, metrics: list) -> dict[str,
                 "RAGAS evaluate failed with parse/output error; returning NaN score",
                 extra={"error": str(error), "metric_type": self.type},
             )
-            return self._nan_scores_for_metrics(metrics)
+            return self._align_scores_to_output_spec(self._nan_scores_for_metrics(metrics))
```
arpitsardhana: To my understanding: Calling both alingn and nan is duplicate right? Once nan finds the right keys, then calling align on its output is no-op

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/metrics/ragas/base.py
```diff
@@ -327,14 +341,14 @@ def _run_evaluate(self, dataset: EvaluationDataset, metrics: list) -> dict[str,
...
         except (httpx.HTTPError, TimeoutError) as error:
             if self._ignore_request_failure():
                 self._log.warning(
                     "RAGAS judge inference failed and is ignored by metric policy; returning NaN score",
                     extra={"error": str(error), "metric_type": self.type},
                 )
-                return self._nan_scores_for_metrics(metrics)
+                return self._align_scores_to_output_spec(self._nan_scores_for_metrics(metrics))
```
arpitsardhana: Same as above for below 4 instances

ngoncharenko (approved):

PR #164, opened 2026-06-03
fix(models): extend provider discovery timeout and reduce log noise

tylersbray:
## Summary

- Adds `provider_discovery_timeout_seconds` (default 180s) and `provider_discovery_max_retries` (default 0) to `ControllerConfig` for GET `/v1/models` autodiscovery via IGW
- Wires a discovery-scoped SDK client with zero SDK retries and passes the extended per-request timeout, eliminating multi-minute retry storms on slow NVIDIA Build catalog responses
- Downgrades duplicate transient discovery failure warnings in `_discover_models` to debug (READY providers still get status-appropriate info from `_on_transient_failure`)

Fixes [NVBugs 6215676](https://nvbugspro.nvidia.com/bug/6215676) / [AIRCORE-717](https://linear.app/nvidia/issue/AIRCORE-717/nvbugs-6215676-reduce-noisyslow-nvidia-build-provider-discovery).

## Test plan

- [x] `uv run --frozen pytest services/core/models/tests/unit/controllers/test_provider_reconciler.py services/core/models/tests/unit/test_config.py services/core/models/tests/unit/controllers/test_models_controller_unit.py::test_controller_initialization -v`
- [ ] Manual: with platform running and `default/nvidia-build` registered, confirm reconciliation no longer emits multi-minute `Retrying request` spam; discovery succeeds or fails once per cycle with quieter logs

## Summary by CodeRabbit

* **New Features**
  * Added configurable timeout and retry settings for provider model discovery via the inference gateway, allowing tuning for slow or large external provider responses (defaults provided).

* **Documentation**
  * Updated configuration reference to document the new provider discovery settings, their intended effects, and default values.

### DIFF
--- a/services/core/models/src/nmp/core/models/config.py
+++ b/services/core/models/src/nmp/core/models/config.py
@@ -336,6 +336,23 @@ class ControllerConfig(BaseModel):
         description="Maximum delay in seconds between drift recovery attempts (caps exponential backoff)",
     )
 
+    provider_discovery_timeout_seconds: int = Field(
+        default=180,
+        ge=1,
+        description=(
+            "Per-request timeout in seconds for GET /v1/models provider autodiscovery via the inference gateway. "
+            "External providers (e.g. NVIDIA Build) can return large model lists and need longer than the SDK default."
+        ),
+    )
+    provider_discovery_max_retries: int = Field(
+        default=0,
+        ge=0,
+        description=(
+            "SDK retry count for provider autodiscovery requests. The controller loop already retries on "
+            "each cycle; disabling SDK retries avoids multi-minute retry storms on slow upstreams."
+        ),
+    )
+
     @field_validator("backends", mode="before")
     @classmethod
     def validate_backends(cls, v: Any) -> dict[str, BackendConfig]:
--- a/services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
+++ b/services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
@@ -25,6 +25,7 @@ from nmp.core.models.app import (
     normalize_model_entity_name,
     parse_model_name_revision,
 )
+from nmp.core.models.config import ControllerConfig
 from nmp.core.models.controllers.context import ModelContext
 from nmp.core.models.schemas import BackendFormat, ModelProviderStatus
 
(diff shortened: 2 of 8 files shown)


services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
```diff
@@ -625,11 +632,17 @@ async def _discover_models(self, provider: ModelProvider) -> DiscoveryResult:
...
+            logger.debug(
+                "Failed to get models from provider via gateway",
+                extra={"provider": provider_id, "error": str(e), "status_code": e.status_code},
+            )
             return DiscoveryTransientError(f"Gateway error (HTTP {e.status_code})")
         except Exception as e:
             # Network errors, timeouts, etc. — do NOT clear served_models
-            logger.warning(f"Failed to get models from {provider_id} via gateway: {e}")
+            logger.debug(
+                "Failed to get models from provider via gateway",
```
mckornfield, line 643: assuming this downgrade was intentional?

services/core/models/src/nmp/core/models/config.py
```diff
@@ -336,6 +336,23 @@ class ControllerConfig(BaseModel):
         description="Maximum delay in seconds between drift recovery attempts (caps exponential backoff)",
     )
 
+    provider_discovery_timeout_seconds: int = Field(
+        default=180,
```
mckornfield, line 340: kinda crazy long, no?

mckornfield (approved):

services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
tylersbray: "Intentional. _discover_models was warning on every transient failure, then _on_transient_failure logs again at info/warn depending on provider status — so READY providers got duplicate warnings every ~10s, on top of SDK "Retrying request" lines. Discovery logs the detail at debug; the handler keeps the operator-visible signal. Happy to also downgrade the 404 gateway-cache path for consistency if you've seen that in logs."

services/core/models/src/nmp/core/models/config.py
tylersbray: "It's the prior worst-case budget: SDK default 60s timeout × 3 attempts (DEFAULT_MAX_RETRIES=2). We collapsed that into a single attempt with max_retries=0 so discovery doesn't block the reconciler in a multi-minute retry loop. 180s is configurable via provider_discovery_timeout_seconds if we want a lower default after we see Build p95s from QA."

PR #181, opened 2026-06-04
fix(studio): dataset schema and file browser polish

nv-odrulea:
Screenshot 2026-06-04 at 11 12 39 AM

  ## Summary

  Polish pass on dataset schema detection and file preview for external datasets (HuggingFace, NGC, S3
  mirrors), which contain many non-data files and large JSONL shards that broke the previous assumptions.

  **Schema column**
  - `isSchemaAssignableFile(path)` util (`@nemo/common`) — true for `.json` / `.jsonl` only. Non-data
  files (README, images, scripts) show a blank Schema cell instead of a misleading `default` marker.
  - `getSchemaCellLabel(path, metadata)` extracted as a pure helper with its own spec — covers string ref
  / inline / root-ref / inline-root-`default` / blank cases. Fixes the pre-existing bug where inline
  root schema showed blank instead of `default`.
  - `sharedReferrerCount` in `DatasetSchemaEditor` now filters to schema-assignable files only, so the
  "used by N files" confirm modal no longer inflates N with non-data files.

  **Schema inference (Generate button)**
  - Replaced full-file download with `useDownloadFileHead` — fetches the first 64 KB per file via HTTP
  `Range: bytes=0-65535`. Reduces per-file download from ~1 GB to 66 KB for large external shards.

  **File preview**
  - `BINARY_FILE_EXTENSIONS` blocklist replaces the old allowlist (`PREVIEWABLE_FILE_TYPES`). Text files
  of any extension (`.gitattributes`, `.py`, `.yaml`, shell scripts, etc.) render in the plain-text
  editor instead of returning "Unsupported file type". Genuinely binary formats (images, archives, model
  weights) show "Text preview not available for binary files" immediately without a download.
  - Preview is now capped at **512 KB** via HTTP Range, preventing OOM crashes on 1 GB JSONL shards.
  `useDatasetFileContent` sends a `Range` header instead of slicing the blob post-download.
  - Non-data files no longer show "Schema not applicable" — the schema panel is simply hidden.
  - Markdown preview gets `bg-surface-raised border rounded-lg` card styling, matching the JSON code
  editor.

  **Right panel layout**
  - Removed top padding from the right panel so the schema editor aligns with the top of the adjacent
  file browser / preview panel.

  **Breadcrumb fix**
  - `Breadcrumbs/index.tsx` strips `?` and `#` from `href` on all breadcrumb links. Previously,
  navigating from `?tab=files` to the Filesets list carried `?tab=files` into the list URL.

  ## Files

  **New**
  - `web/packages/common/src/utils/jsonSchema/schemaAssignable.ts` + `.spec.ts`
  - `web/packages/studio/src/routes/DatasetDetailRoute/FilesTab/schemaColumn.ts` + `.spec.ts`
  - `web/packages/studio/src/components/filesets/hooks/useDownloadFileHead.ts`

  **Modified**
  - `web/packages/common/src/components/FileContentPreview/index.tsx`
  - `web/packages/common/src/components/Breadcrumbs/index.tsx`
  - `web/packages/studio/src/api/datasets/constants.ts`
  - `web/packages/studio/src/api/datasets/useDatasetFileContent.ts`
  - `web/packages/studio/src/components/FilesetFilePreviewPanel/FilesetFilePreviewContent/index.tsx`
  - `web/packages/studio/src/routes/DatasetDetailRoute/DatasetSchemaEditor/index.tsx`
  - `web/packages/studio/src/routes/DatasetDetailRoute/FilesTab/index.tsx`

## Summary by CodeRabbit

* **New Features**
  * Binary-file detection that shows "Text preview not available for binary files."
  * Dataset schema editor now appears only for JSON/JSONL files

* **Improvements**
  * Byte-range fetching and HEAD-based checks for safer, faster text previews
  * Markdown preview receives bordered/raised styling
  * Schema column shows explicit labels and respects per-file/root mappings

* **Tests**
  * Added unit tests for schema-assignable detection, schema-label logic, and preview handling

### DIFF
--- a/web/packages/studio/src/api/datasets/useDatasetFileContent.ts
+++ b/web/packages/studio/src/api/datasets/useDatasetFileContent.ts
@@ -2,8 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { customFetch } from '@nemo/sdk/generated/fetchers/platform';
-import { filesDownloadFile, filesHeadFile } from '@nemo/sdk/generated/platform/api';
-import { EntityIdentifier } from '@studio/api/common/types';
+import {
+  filesDownloadFile,
+  filesHeadFile,
+  getFilesDownloadFileQueryKey,
+} from '@nemo/sdk/generated/platform/api';
+import type { EntityIdentifier } from '@studio/api/common/types';
 import { getDatasetFileContentQueryKey } from '@studio/api/datasets/invalidateDatasetCaches';
 import { isBinaryExtension } from '@studio/util/binaryFile';
 import { queryOptions, useQuery, UseQueryOptions, useSuspenseQuery } from '@tanstack/react-query';
@@ -71,8 +75,13 @@ export const datasetFileContentQueryOptions = ({
       } else {
         const start = range ? range[0] : 0;
         const end = range ? range[1] : FILE_PREVIEW_MAX_BYTES - 1;
+        const [fileUrl] = getFilesDownloadFileQueryKey(
+          encodeURIComponent(workspace!),
+          encodeURIComponent(name),
+          encodeURIComponent(path)
+        );
         const blob = await customFetch<Blob>({
-          url: `/apis/files/v2/workspaces/${encodeURIComponent(workspace!)}/filesets/${encodeURIComponent(name)}/-/${encodeURIComponent(path)}`,
+          url: fileUrl,
           method: 'GET',
           responseType: 'blob',
           headers: { Range: `bytes=${start}-${end}` },
--- a/web/packages/studio/src/api/datasets/useDatasetFileContent.ts
+++ b/web/packages/studio/src/api/datasets/useDatasetFileContent.ts
@@ -1,14 +1,13 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { customFetch } from '@nemo/sdk/generated/fetchers/platform';
 import { filesDownloadFile, filesHeadFile } from '@nemo/sdk/generated/platform/api';
 import { EntityIdentifier } from '@studio/api/common/types';
 import { getDatasetFileContentQueryKey } from '@studio/api/datasets/invalidateDatasetCaches';
-import { PLATFORM_BASE_URL } from '@studio/constants/environment';
 import { isBinaryExtension } from '@studio/util/binaryFile';
 import { queryOptions, useQuery, UseQueryOptions, useSuspenseQuery } from '@tanstack/react-query';
 import { parquetRead } from 'hyparquet';
-import { useAuth } from 'react-oidc-context';
 
 // Cap text-file preview at 512 KB. Enough to show meaningful JSONL content
 // while preventing OOM crashes on multi-GB external dataset shards.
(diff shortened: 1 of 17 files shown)


web/packages/studio/src/components/filesets/hooks/useIsBinaryFile.ts
```diff
@@ -0,0 +1,85 @@
...
+): { isBinary: boolean; isLoading: boolean } {
+  const auth = useAuth();
+
+  const blocklisted = filePath !== undefined && isBinaryExtension(filePath);
+
+  const { data: headBinary, isPending } = useQuery({
+    queryKey: ['file-content-type', workspace, filesetName, filePath],
+    queryFn: async (): Promise<boolean> => {
+      if (!filePath) return false;
+      const url = [
```
steramae-nvidia: this can be replaced with `getFilesDownloadFileQueryKey`?

web/packages/studio/src/components/filesets/hooks/useDownloadFileHead.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return useCallback(
+    async ({
+      workspace,
+      datasetName,
+      path,
+      bytes = 65536,
+    }: DownloadFileHeadArgs): Promise<ArrayBuffer | null> => {
+      const url = [
+        PLATFORM_BASE_URL,
```
steramae-nvidia: This can be replaced with `getFilesDownloadFileQueryKey`?

web/packages/studio/src/api/datasets/useDatasetFileContent.ts
```diff
@@ -3,14 +3,21 @@
...
 import { getDatasetFileContentQueryKey } from '@studio/api/datasets/invalidateDatasetCaches';
+import { PLATFORM_BASE_URL } from '@studio/constants/environment';
+import { isBinaryExtension } from '@studio/util/binaryFile';
 import { queryOptions, useQuery, UseQueryOptions, useSuspenseQuery } from '@tanstack/react-query';
 import { parquetRead } from 'hyparquet';
+import { useAuth } from 'react-oidc-context';
+
+// Cap text-file preview at 512 KB. Enough to show meaningful JSONL content
+// while preventing OOM crashes on multi-GB external dataset shards.
+const FILE_PREVIEW_MAX_BYTES = 512 * 1024;
```
steramae-nvidia, line 18: Would be interesting to see this for parquet files too, have you tried seeing if this could be integrated?

nv-odrulea: have not gotten to parquet files yet.  afaik, those are different because you do have to read to the end of the file in every case? but maybe there is some known solution to handle very large files, haven't looked there yet.

web/packages/studio/src/components/filesets/hooks/useIsBinaryFile.ts
nv-odrulea: `getFilesDownloadFileQueryKey` would conflict. Same key = shared cache entry, but filesDownloadFile stores a Blob and our HEAD query stores a boolean. TanStack Query would serve the wrong type depending on which query ran first.

Custom key `['file-content-type', workspace, filesetName, filePath]` is correct here because it's a distinct query with a distinct result type

web/packages/studio/src/components/filesets/hooks/useDownloadFileHead.ts
nv-odrulea: No, same reason, `getFilesDownloadFileQueryKey` is a cache key helper, not a URL builder. It returns something like `['/apis/files/v2/workspaces/ws/filesets/name/-/path']` for TanStack Query's cache, not a fully-formed fetch URL.

The URL construction in useDownloadFileHead is a raw fetch call (not a TanStack Query query), so no query key applies here.

We are actually bypassing tanstack here because the SDK doesn't support `Range: ` header (which allows us to limit response size).

UPDATE:
On the other hand, after digging deeper into this, it does appear that we could use `customFetch` provided by the generated SDK to achieve this same outcome, and that would allow us to route the request through tanstack, for one of the requests.

steramae-nvidia (approved):

PR #182, opened 2026-06-04
feat(auditor): add filtering to target and config list endpoints (AIRCORE-387)

maxdubrinsky:
## What

The auditor plugin's target and config list endpoints previously supported only `page` / `page_size` / `sort`. This adds query filtering through the standard plugin idiom.

- New `TargetFilter` (`type`, `model`, `description`, `project`, `created_at`, `updated_at`) and `ConfigFilter` (`description`, `project`, `created_at`, `updated_at`) in `api/v2/schemas.py` — both `NemoFilter` subclasses (`extra="forbid"`), with `DatetimeFilter` range support on the timestamps.
- `list_targets` / `list_configs` now take a `filter[...]` dependency (`make_filter_obj_dep`) and forward `filter_obj` to `entity_client.list`.
- A small auditor-local `_filters.make_filter_dep` wrapper maps the raw pydantic `ValidationError` raised on an unknown filter key to HTTP 422 — the platform registers no handler for it, so the bare dependency would surface a 500. Shared infra is untouched.

## Compatibility

Additive. With no filter the call is identical to before; `page` / `page_size` / `sort` and the response shape are unchanged (a `filter` echo key is added). Misspelled filter keys now return 422 (intended).

## Tests

15 new tests (8 target / 7 config): per-field filtering, `project` narrowing, `created_at` `$gte`/`$lte` ranges, 422 on unknown key, and an empty-filter regression. Plugin suite green (40 passed).

## Remaining before merge

Run `make refresh-openapi` to regenerate `plugins/nemo-auditor/openapi/openapi.yaml` (spec-from-Pydantic; no Stainless — auditor ships hand-written SDK resources).

AIRCORE-387

## Summary by CodeRabbit

* **New Features**
  * Add schema-driven filtering to config and target list endpoints, including date-range support; applied filters are forwarded to storage and echoed in list responses. OpenAPI now documents filter parameters and schemas.

* **Bug Fixes**
  * Unknown or misspelled filter keys now return HTTP 422 with clearer validation details.

* **Tests**
  * New tests covering config and target list filtering: parameter forwarding, range parsing, error handling, and response shape.

### DIFF
--- a/plugins/nemo-auditor/src/nemo_auditor/api/v2/targets.py
+++ b/plugins/nemo-auditor/src/nemo_auditor/api/v2/targets.py
@@ -66,11 +66,11 @@ async def list_targets(
     page: int = Query(default=1, ge=1),
     page_size: int = Query(default=20, ge=1, le=100),
     sort: str = Query(default="-created_at"),
-    filter: TargetFilter = Depends(_target_filter_dep),
+    parsed_filter: TargetFilter = Depends(_target_filter_dep),
     entity_client: NemoEntitiesClient = Depends(get_entity_client),
 ) -> dict:
     """List audit targets in the workspace with pagination and filter support."""
-    filter_dict = filter if isinstance(filter, dict) else filter.model_dump(exclude_none=True)
+    filter_dict = parsed_filter if isinstance(parsed_filter, dict) else parsed_filter.model_dump(exclude_none=True)
     try:
         result = await entity_client.list(
             AuditTarget,
@@ -87,7 +87,7 @@ async def list_targets(
         "data": [t.model_dump(mode="json") for t in result.data],
         "pagination": result.pagination.model_dump() if result.pagination else None,
         "sort": sort,
-        "filter": filter or None,
+        "filter": parsed_filter or None,
     }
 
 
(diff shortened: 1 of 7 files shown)


plugins/nemo-auditor/src/nemo_auditor/api/v2/targets.py
```diff
@@ -52,22 +56,29 @@ async def create_target(
...
+    "/targets",
+    tags=["Auditor Targets"],
+    openapi_extra=generate_openapi_extra_params(filter_schema=TargetFilter),
+)
 async def list_targets(
     workspace: str,
     page: int = Query(default=1, ge=1),
     page_size: int = Query(default=20, ge=1, le=100),
     sort: str = Query(default="-created_at"),
+    filter: TargetFilter = Depends(_target_filter_dep),
```
parkanzky: This is going to shadow the `filter` python built-in. If that's the only way to get the user-facing API we want that may be ok, but it would be best to not shadow built-ins.

maxdubrinsky: Good callout, we bounce between `filter` and `parsed_filter` and it looks like the former got picked. Updated here, will update other places in the next bit.

parkanzky (approved):

PR #187, opened 2026-06-04
feat(studio): Data sample panel

steramae-nvidia:
Screenshot 2026-06-08 at 10 21 02 AM

This PR adds a data sample to the bottom right panel in order to support quick previews of the data in the dataset card case.

This also slightly changes the panel layout in the FilesetCard component. It adds a Resizeable panel so users can choose which view to focus on (README or the dataset sample)

## Summary by CodeRabbit

* **New Features**
  * Data sample preview panel for dataset filesets with file selection, sampling method, and row-count controls.
  * Resizable two-pane layout for improved content organization.

* **Bug Fixes**
  * Parquet BigInt values now serialize safely as strings for downloads/previews.

* **Improvements**
  * Centered loading spinner during file fetches and inline error banners on load failure.
  * Added a loading scenario to component previews.

* **Tests**
  * New tests for sampling panel and file-content handling.


htolentino-nvidia:
I like the idea a lot, I'm just getting a bit of a weird behaviour with the scrolling on that section:
image

steramae-nvidia:
> I like the idea a lot, I'm just getting a bit of a weird behaviour with the scrolling on that section: image

@htolentino-nvidia taking a look!

steramae-nvidia:
> > I like the idea a lot, I'm just getting a bit of a weird behaviour with the scrolling on that section: image
> 
> @htolentino-nvidia taking a look!

Screenshot 2026-06-04 at 3 19 11 PM

Fixed by removing one scroll bar on the outer container (the inner data view still needs scroll bars)

htolentino-nvidia:
Sorry to be annoying about this, but there are a few things looking a bit off with the design that I think would be easy wins:

- The whole view scrolls a tiny bit event thought content actually is contained to the panels. (on screenshot)
- Loading and error views are a bit weird right now (on screenshot)
- I think there a bit too many panel borders nested and it looks a bit cluttered, could possibly use a a simple grid with just the content portion borders (on screenshot)
- When the screen is too narrow, could we stack the panels?
  



image

image

htolentino-nvidia (approved):

PR #191, opened 2026-06-04
feat(filter): add InMemoryFilterRepository and use it for jobs status filtering

maxdubrinsky:
## What

Adds `InMemoryFilterRepository` — a `FilterRepository` whose backend returns `bool` instead of a SQL expression — so a parsed filter tree can be evaluated against an in-memory Python object through the same `operation.apply(repo)` front door the SQL layer uses. Makes the jobs dispatcher use it.

**Primitive (AIRCORE-388):** `InMemoryFilterRepository` in `nemo_platform_plugin/in_memory_filter.py`, a peer of `SQLAlchemyFilterRepository`. Bind an entity, call `operation.apply(repo)`, get a `bool`. Covers `$eq` / `$like` / `$in` / `$nin` / `$lt` / `$lte` / `$gt` / `$gte` / `$and` / `$or` / `$not`; `$exists` / relationship operations inherit the base `NotImplementedError`. `filter_ops.py` is unchanged from `main` — there's no new method on the operation types, and operator dispatch stays in the existing `apply()`.

Semantics are **native Python, NOT a byte-for-byte mirror of the SQL backends** (documented on the class):

- Values compared by Python type — no JSON-to-text coercion, so `data.score` (stored int `5`) does not match the string `"5"`.
- Absent fields (missing key/attr) and explicit `None` are equivalent and satisfy only `$eq None` (standard SQL three-valued NULL logic, which both SQLite and PostgreSQL share).
- `$like` is a literal case-insensitive substring test; `%` / `_` are literal characters, not SQL wildcards.

Where the SQL backends disagree *with each other* (JSON boolean text is `"1"/"0"` on SQLite but `"true"/"false"` on PostgreSQL; SQLite casts non-numeric text to `0.0` while PostgreSQL errors), the in-memory path follows neither and compares natively. The underlying SQL-layer inconsistencies — unescaped `$like` wildcards and cross-backend JSON coercion — are tracked in **AIRCORE-749**.

**Consumer (AIRCORE-324):** `list_jobs` evaluates the full filter tree in-memory via `op.apply(InMemoryFilterRepository(virtual_job))` against a virtual job entity carrying the attempt status, replacing the hand-rolled status split. This **fixes AIRCORE-324**: `$or` / `$not` / `$nin` status expressions (and status mixed with other fields) now return correct results instead of being rejected or inverted. The store pre-query uses a provably-superset transform (`_status_free_superset`) so it never drops a row the in-memory pass would accept; that pass then narrows the page exactly.

## Compatibility

Additive. `InMemoryFilterRepository` is a new class in its own module; `filter_ops.py` and the `api.filter` facades are unchanged from `main`. For the status filters the old code already supported (`$eq` / `$in` under `$and`), results are unchanged. One minor delta: an *invalid* status value (e.g. `status=zzzz`) previously returned `([], 0)` and now returns `([], <superset count>)` — the data list is empty in both cases.

## Tests

- **54 unit tests** for `InMemoryFilterRepository` — every operator, JSON-path vs plain attribute, null/missing, and the cases where native semantics intentionally diverge from one or both SQL backends (pinned as documented behavior).
- A **SQL-parity test (27 cases)** that runs each filter through both `SQLAlchemyFilterRepository` and `InMemoryFilterRepository` via the same `apply` front door and asserts they select the identical row set. Scoped to the cases where native and SQL semantics agree (the divergent cases live in the unit tests, by design).
- `list_jobs` dispatcher tests assert correct results for the previously-rejected `$or` / `$not` / `$nin` status trees.

## Not in scope

Status-filtered pagination and `total_count` are inaccurate — the dispatcher pages a store superset then filters in-memory, so a page can under-fill and `total_count` reflects the superset. This is **pre-existing** (the old hand-rolled implementation had the identical structure) and is **unchanged** by this PR. Tracked separately in AIRCORE-738.

AIRCORE-388, AIRCORE-324, AIRCORE-749


packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
```diff
@@ -167,3 +235,193 @@ def to_dict(self) -> Dict[str, Any]:
...
+    For a plain column a NULL value never matches (``NULL ILIKE x`` is NULL).
+    For a JSON field, null/missing renders as the literal text ``"null"`` and
+    is matched as such.
+    """
+    needle = str(value).lower()
+    if is_json:
+        return needle in _json_text(field_value).lower()
+    if field_value is _MISSING or field_value is None:
+        return False
+    return needle in str(field_value).lower()
```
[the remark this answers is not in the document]
maxdubrinsky: This is a bug in the entities code rather than here, I will make the fix to prevent this type of matching.

packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
```diff
@@ -144,6 +162,56 @@ def to_dict(self) -> Dict[str, Any]:
...
+            data = _get_attr(entity, "data", _MISSING)
+            path = self.field.split(".")[1:]
+            return _walk_json_path(data, path), True
+
+        raise ValueError(f"Field '{self.field}' does not exist on {type(entity).__name__}")
+
+    def matches(self, entity: Any) -> bool:
+        field_value, is_json = self._resolve_field(entity)
+
+        if self.operator == FilterOperator.EQ:
```
mckornfield: could also see a case statement, but I think the code quality guy creates some false positives for them (as far as not meeting all the cases lol)

packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
```diff
@@ -167,3 +235,193 @@ def to_dict(self) -> Dict[str, Any]:
...
+    """Emulate SQLite ``CAST(<text> AS FLOAT)``.
+
+    SQLite performs a lenient leading-numeric parse (``"5abc"`` -> 5.0,
+    ``"abc"`` / ``"null"`` -> 0.0) rather than Python's strict ``float()``.
+    """
+    try:
+        return float(text)
+    except (TypeError, ValueError):
+        pass
+    match = re.match(r"\s*[+-]?(\d+\.?\d*|\.\d+)([eE][+-]?\d+)?", text)
```
mckornfield: lol this is fun

mckornfield: "null" -> 0 seems a bit odd but null also stinks

packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
```diff
@@ -167,3 +235,193 @@ def to_dict(self) -> Dict[str, Any]:
...
+    mismatch is treated as "no match" (mirrors SQL three-valued NULL results).
+    """
+    try:
+        if op == "lt":
+            return left < right
+        elif op == "lte":
+            return left <= right
+        elif op == "gt":
+            return left > right
+        else:  # "gte"
```
mckornfield: why not be explicit and throw if it's not one of the 4?

services/core/entities/tests/test_filter_matches_sql_parity.py
```diff
@@ -0,0 +1,125 @@
...
+    ("in_name", C(FilterOperator.IN, "name", ["llama", "mistral"])),
+    ("in_data_tier", C(FilterOperator.IN, "data.tier", ["pro", "free"])),
+    ("in_data_score", C(FilterOperator.IN, "data.score", [5, 10])),
+    ("nin_name", C(FilterOperator.NIN, "name", ["llama"])),
+    ("nin_data_tier", C(FilterOperator.NIN, "data.tier", ["pro"])),
+    ("nin_data_score", C(FilterOperator.NIN, "data.score", [5, 9])),
+    ("gt_data_score", C(FilterOperator.GT, "data.score", 9)),
+    ("gte_data_score", C(FilterOperator.GTE, "data.score", 10)),
+    ("lt_data_score", C(FilterOperator.LT, "data.score", 10)),
+    ("lte_data_score", C(FilterOperator.LTE, "data.score", 9)),
```
mckornfield, line 102: beautiful

services/core/jobs/src/nmp/core/jobs/app/dispatcher.py
```diff
@@ -83,102 +83,107 @@ def create_platform_job_response(job: PlatformJob, attempt: PlatformJobAttempt)
...
     return False
 
 
-def _subtree_all_status(operation: FilterOperation) -> bool:
+def _status_free_superset(operation: FilterOperation | None) -> FilterOperation | None:
+    """Build a status-free store filter that accepts a SUPERSET of ``operation``.
+
+    Status lives on the attempt, not the job, so it cannot be pushed to the
+    PlatformJob store query. We push down a relaxed, status-independent filter
+    that is guaranteed to keep every row the in-memory ``matches()`` would
```
mckornfield: but the in memory one throws doesn't it? or no?

mckornfield (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
maxdubrinsky: atwobu

services/core/jobs/src/nmp/core/jobs/app/dispatcher.py
maxdubrinsky: No the in-memory takes over `status` filtering. We do need to keep the whole filter context though since you could be combining with other operations.

packages/nemo_platform_plugin/src/nemo_platform_plugin/filter_ops.py
maxdubrinsky: That shouldn't be possible given how we use this `_compare` function, but it's a bit smelly that we're passing around raw `op: str`. Will do a bit of cleanup here

mckornfield (approved):
much better!

PR #197, opened 2026-06-04
feat(filter): add StringFilter type and widen per-field filter types

maxdubrinsky:
## What

Introduces `StringFilter` — a structured filter type for string fields that supports `$eq`, `$like`, `$in`, and `$nin` operators. Widens key filter fields (`name`, `description`, `type`, `project`, `source`, etc.) across services from bare `str` to `str | StringFilter`, enabling operator-based string matching on those fields.

**Core** (`packages/nmp_common`, `packages/nemo_platform_plugin`):

- `StringFilter(Filter)` added to `nmp.common.entities.values` with `eq`, `like`, `in_`, `nin` fields and `$`-aliased keys. Exported from `nmp.common.entities`.
- `StringFilter` also added to `nemo_platform_plugin.schema` (alongside `BooleanFilter`, `DatetimeFilter`) and re-exported in `__all__`.
- `sdk/stainless.yaml` gains `string_filter: StringFilter` under `$shared.models`.

**Schema widening** across services:

| Service | Fields widened |
|---------|---------------|
| `core/files` filesets | `name`, `purpose` |
| `core/jobs` jobs | `source`, `type` |
| `core/models` | `name`, `publisher`, `task`, `architecture`, `format` |
| `core/auth` IAM | `name`, `role` |
| `evaluator` benchmarks | `name`, `project`, `description`, `type` |
| `evaluator` metrics | `name`, `project`, `description`, `type` |
| `guardrails` configs | `name`, `project` |

**Tests**: new unit tests in `test_parsed_filter.py` and `test_client.py` covering operator round-trips and schema widening.

## SDK regen required before merge

This PR changes the OpenAPI-visible filter schemas. Run the following before merging:

```bash
make refresh-openapi   # regenerates openapi/openapi.yaml
make stainless         # pushes to Stainless, pulls updated SDK (requires STAINLESS_API_KEY)
make update-web-sdk
make update-cli
```

These cannot run inside the sandbox, so they're not included in this commit.

## Compatibility

`str` remains a valid filter value on all widened fields — `name: str` still works via the existing `str` arm of the union. Callers who want operator-based matching can now pass `{"$like": "llama"}` instead.

## Summary by CodeRabbit

* **New Features**
  * Expanded string filtering across many list endpoints: filters now accept either plain strings or a new structured StringFilter supporting equality, pattern match, inclusion, and exclusion.
  * OpenAPI/REST docs updated to expose the new structured string filter in query schemas.

* **Tests**
  * Added end-to-end and unit tests validating StringFilter parsing, aliasing, and behavior.


packages/nmp_common/src/nmp/common/entities/values.py
```diff
@@ -169,3 +169,36 @@ class DatetimeFilter(Filter):
...
+        serialization_alias="$eq",
+        description="Filter for results equal to this value.",
+    )
+    like: Optional[str] = Field(
+        None,
+        alias="$like",
+        serialization_alias="$like",
+        description="Filter for results matching this pattern.",
+    )
+    in_: Optional[list[str]] = Field(
```
mckornfield, line 254: is the `in_` just to match the convention (I remember seeing it or `_in` before, but noticed the others don't have it, and it still uses the $in below)

services/evaluator/src/nmp/evaluator/api/v2/metrics/schemas/metrics_resp.py
```diff
@@ -229,8 +229,8 @@ class SystemMetricResponse(app.SystemMetric, _OptionalEntity):
 class MetricsListFilter(Filter):
     """Filter for list metrics query."""
 
-    name: str | None = Field(default=None, description="Filter metrics by name.")
-    description: str | None = Field(default=None, description="Filter metrics by description.")
+    name: StringFilter | str | None = Field(default=None, description="Filter metrics by name.")
```
mckornfield, line 232: probably a super nit, but `NullableStringFilterOrStr` or something as a reusable union type... but meh it is probably nastier than what you have lol

mckornfield (approved):

services/evaluator/src/nmp/evaluator/api/v2/metrics/schemas/metrics_resp.py
maxdubrinsky: Yeah I was trying to balance import count + code ergonomics, this feels ok? If we eventually go crazy and have more things then we can always add it.

packages/nmp_common/src/nmp/common/entities/values.py
maxdubrinsky: Using `in_` here because `in` is reserved and the leading `_` means private. For normal dicts though we can use `$in`, hence the madness.

PR #217, opened 2026-06-08
feat(e2e): Add files external provider tests

matthewgrossman:
## Summary

Implements [AIRCORE-747](https://linear.app/nvidia/issue/AIRCORE-747): E2E tests for NGC-backed filesets with CI token, plus Hugging Face storage backend tests.

### E2E test suite (`e2e/files/`)

- **Reorganized** `e2e/test_files.py` → `e2e/files/test_files.py` (file upload/download tests).
- **New `e2e/files/test_storage_backends.py`** with tests for NGC and Hugging Face external storage backends:
  - `TestNGCFileset` — creates a secret + NGC-backed fileset pointing at a public resource (`nvidia/nemo-microservices/nemo-microservices-quickstart`), validates list/download/cache-status. Parametrized error tests verify that invalid key, wrong org, nonexistent resource, and missing secret all return `400 BadRequestError` with descriptive messages. Skipped when `NGC_API_KEY` is not set.
  - `TestHuggingFaceFileset` — creates an HF-backed fileset (`hf-internal-testing/tiny-random-bert`), validates list/download/cache-status. Uses `HF_TOKEN` for authenticated access when available. Error test for nonexistent repo.

### CI wiring (`.github/workflows/ci.yaml`)

- Passes `NGC_API_KEY` (from `NGC_REGISTRY_READ_TOKEN` secret) and `HF_TOKEN` to the `python-e2e-test` job.

### Bug fix: NGC backend error handling (`services/core/files/.../ngc.py`)

- **`_get_client()`**: Now catches `NgcException` (in addition to `ValueError`) during `client.configure()`. Previously, org mismatches raised `NgcException` which bubbled up as an unhandled 500.
- **`validate_storage()`**: Moved `_get_registry_api()` and `_get_target()` inside the try block. Previously these calls were outside error handling, so failures like `AccessDeniedException` from version resolution were unhandled 500s.

### E2E infrastructure improvements (`e2e/conftest.py`)

- **Services log on failure**: Added a `pytest_runtest_makereport` hook that appends the tail of the services log to the test report when a test fails. Previously, server-side errors only showed as opaque "500" from the client.
- **Per-session log paths**: `services_log_path` is now a session-scoped fixture with a UUID suffix, so parallel pytest processes never clobber each other's logs.

## Test plan

- [x] All 11 storage backend tests pass locally (`pytest e2e/files/test_storage_backends.py --run-e2e`)
- [x] All 5 existing file tests pass after move (`pytest e2e/files/test_files.py --run-e2e`)
- [x] All 29 NGC backend unit tests pass (`pytest services/core/files/tests/test_ngc_backend.py`)
- [ ] CI `python-e2e-test` job passes (NGC tests will skip until `NGC_REGISTRY_READ_TOKEN` secret is created)
- [ ] Create `NGC_API_KEY` repository secret (mapped from `NGC_REGISTRY_READ_TOKEN`)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests for NGC and Hugging Face storage backends covering listing, download, cacheability, and error cases; improved diagnostics to attach recent service logs on failures.

* **Bug Fixes**
  * Improved and standardized error handling for NGC storage operations.

* **Chores**
  * CI updated to provide credentials for external storage tests and to collect/upload per-test service log directories.


e2e/conftest.py
```diff
@@ -99,8 +118,37 @@ def background_process(args: list[str], stdout: IO[Any] | None = None) -> Iterat
...
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo):  # noqa: ARG001
+    """Append the services log tail to the report when a test fails.
+
+    This hook is the pytest-sanctioned way to add extra sections to test
+    reports (``report.sections``).  Fixtures cannot do this because they
+    don't have access to the report object.
+    """
+    outcome = yield
```
mckornfield, line 141: lol this looks so strange

mckornfield (approved):

e2e/files/test_storage_backends.py
```diff
@@ -0,0 +1,309 @@
...
+    )
+    yield fileset_name
+    try:
+        sdk.files.filesets.delete(fileset_name, workspace=workspace)
+    except Exception:
+        pass  # Best-effort cleanup; the workspace is deleted anyway
+
+
+@pytest.fixture
+def hf_secret(sdk: NeMoPlatform, workspace: str) -> Iterator[str | None]:
```
mikeknep: Why is this one optional but the NGC secret is required?

mikeknep: Oh I might have an answer actually: is it because the HF resource is public and a token isn't strictly necessary, but providing one will help avoid rate-limiting issues? Or something else?

mikeknep (approved):
LGTM, just the one non-blocking question I'm curious about

e2e/files/test_storage_backends.py
matthewgrossman: yep that's exactly it. though on 2nd thought, maybe we _should_ make it error out if the token somehow disappears? I think silently degrading to public-repo-only is a bad idea?

mikeknep: The only argument I can see for keeping it optional is that it'd be slightly easier to run the e2e tests locally. However, anyone doing that is going to *need* an NGC key in their env regardless—no way out of that one—so maybe requiring HF token as well is fine. I guess I lean slightly towards making it required but I don't feel strongly about it. If we keep the asymmetry, it's probably worth an explanatory comment.

matthewgrossman: fixing

e2e/conftest.py
matthewgrossman: I know it does 😭 , its just the structure of `pytest_runtest_makereport`. Basically this allows us to `yield` for the actual test run, then capture the output from it ¯\_(ツ)_/¯

crookedstorm (approved):
Seems good for now. This implies PRs will require copy-pr-bot as soon as we accept contributions. I can test implementing that as soon as we need it. (Folks have told me to wait for something).

PR #221, opened 2026-06-09
fix(auditor): Update auditor usages

anastasia-nesterenko:
## Summary

Fixes auditor agentic-use task instructions and verifiers to use the current `nemo auditor` CLI surface instead of the removed `nemo audit` / audit-job commands.

Linear: AIRCORE-739

## Changes

- Update auditor default/custom-probes tasks from `audit jobs create/get-status` to `auditor audit run`
- Update verifier code to use the plugin SDK namespace, `client.auditor`, for auditor configs and targets
- Verify audit invocation through agent trajectory instead of expecting persisted audit jobs
- Refresh related Auditor command guidance in the agents optimization skill and Studio suggestion text
- Replace stale SDK README auditor example with a valid guardrail config example
- Remove stale `services/auditor/tests` pytest discovery path

## Testing

- Not run: agentic-use tasks require the Harbor/platform test environment
- Checked command/API wiring during review against the local auditor plugin SDK and CLI registration

Closes: [AIRCORE-739](https://linear.app/nvidia/issue/AIRCORE-739/nvbug-310-baseline-tasks-fail-auditor-cli-command-mismatch-guardrails)

## Release Notes

* **Documentation**
  * Updated auditor CLI documentation to reflect new target and config creation workflow instead of audit jobs.
  * Added CLI command examples and revised success criteria for audit operations.

* **Tests**
  * Updated test suites to validate auditor target and config creation patterns.
  * Removed legacy audit job validation tests.
  * Added tests for agent audit invocation with custom configurations.


tylersbray (approved):
Looks good.  I did a pass with an AI code review, there were two things that popped out:

1) Run at least one updated task in Harbor (e.g. auditor-default-job-cli-easy) to confirm:
* client.auditor.* calls succeed against the eval container API
* Agents following new instructions pass the verifier

2) Could be a little tighter in auditor-default-job-cli-easy/tests/test_outputs.py.  There is this snippet:

```
has_audit_run = any(
    all(p in cmd for p in ["auditor", "audit", "run", CONFIG_NAME, TARGET_NAME])
    for cmd in commands
)
```

It quibbled about these issues:
* CONFIG_NAME = "default" is an extremely common substring — can false-pass on unrelated commands containing "default".
* No _has_token() — custom-probes task uses word-boundary regex; default-job does not.
* Does not verify qualified spec — instructions say default/audit-target; test only checks bare name substrings.
* No target-create trajectory check — only config interaction is trajectory-tested.

Recommendation: Port _has_token() from custom-probes and require audit-target + either default/audit-target or default/default in the run command.

I'm not sure how set up harbor is at this point, I think we were waiting on some NAT changes at one point?  Tightening up the tests might be good to do a little more than just look for strings.

PR #222, opened 2026-06-09
feat(evaluator): decouple sdk runner from a benchmark tasksets

ngoncharenko:
## Summary

Links NVIDIA AALGO-257: https://linear.app/nvidia/issue/AALGO-257/allow-running-user-defined-benchmark-taskset

- Add an agent-eval benchmark adapter contract to decouple runner from arbitrary custom task (given implemented interface)
- Add ProfBench as an adapter-backed example with offline, Docker-backed Codex, and local Codex invocation paths (this will be replaced by nat_runner)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
```diff
@@ -0,0 +1,122 @@
...
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+from nemo_evaluator_sdk.agent_eval.types import AgentEvalAttempt, AgentEvalRunResult, AgentEvalTask
+
+
+class AgentEvalBenchmarkEvaluationKind(StrEnum):
+    """How a benchmark bundle should be executed by :class:`AgentEvaluator`."""
+
+    STORED_ATTEMPTS = "stored_attempts"
+    LIVE_TARGET = "live_target"
```
SandyChapman: Why do we need this? Typically we've used the presence/absence of a target to determine the approach.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
```diff
@@ -0,0 +1,122 @@
...
+    """Experimental protocol for adapting external benchmarks into agent-eval."""
+
+    @property
+    def name(self) -> str:
+        """Stable benchmark name used in diagnostics, metadata, and user-facing output."""
+        ...
+
+    def load(self, config: AgentEvalBenchmarkLoadConfig) -> AgentEvalBenchmarkBundle:
+        """Load tasks and, for stored-attempt runs, recorded attempts for one evaluation shape."""
+        ...
```
SandyChapman, line 62: This is persisting benchmarks? In the SDK, we haven't dealt with persistence for model evals. We can explore this as an option, but I'm wondering if we leave this out of the SDK for now? We can leave `ProfBenchAgentEvalBenchmark` for loading the benchmark, but the main entrypoint still just uses the AgentEvaluator.run/run_sync calls? We can likely discuss in parking lot.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/docs/architecture.md
```diff
@@ -0,0 +1,291 @@
...
+---
+config:
+  theme: dark
+---
+flowchart TD
+    OfflineCmd["uv run --frozen python -m packages.nemo_evaluator_sdk.examples.profbench.examples offline"]
+    
+    LocalCmd["uv run --frozen python -m packages.nemo_evaluator_sdk.examples.profbench.examples local --agent-model gpt-5.5"]
+
+    subgraph ProfBenchCLI["ProfBench CLI and adapter"]
```
SandyChapman: Did we intend to add this md file? If so, it contains ProfBench specific content that shouldn't live in the SDK.

ngoncharenko: Yes, this is intentional for me or agent to understand the execution flow. I can remove it

ngoncharenko: moved it to the examples

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
ngoncharenko: Made main the entrypoint in example to be `AgentEvaluator.run` by moving the wrapper into the example.py. What do you mean by persistence here? Do you mean saving in the nemo-platform or saving results files?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
ngoncharenko: Yes, good point. Removed to use target/attempts presence as a deciding factor

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
SandyChapman: Yes, I mean, this seems like it's meant to save a benchmarkbundle to disk or something? Just curious since this seems like it's meant to mirror the `MetricBundle`

SandyChapman (approved):

arpitsardhana (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/benchmarks.py
ngoncharenko: No, the write_reports() function takes `AgentEvalRunResult` and creates dashboards `sdk-report.html`, `report.html`. The `load` is init of Benchmark - in case of ProfBench, it loads dataset from HF

PR #223, opened 2026-06-09
ci: publish RC/stable release bundles as GitHub Release assets

svvarom:
## Summary
RC and stable releases currently hand off the built bundle only as a transient GitHub Actions artifact. This attaches those artifacts to the **GitHub Release** for the tag instead — as **individual assets** (each SDK wheel, `release-manifest.json`, and `checksums.txt`) — so release outputs are discoverable from the repo and downloadable individually, matching common OSS practice. **Nightly is unchanged** and continues to use Actions artifacts.

## What changes
All in `.github/workflows/release-bundle.yaml`:
- **`assemble-release-bundle`**: for RC/stable, create a GitHub Release on the already-reserved tag and attach each artifact as its own asset (RC marked pre-release). The nightly path is unchanged aside from a cadence `if` guard.
- **`dispatch-release-bundle`**: the `release-bundle-produced` payload gains a `source` field (`artifact` | `release`) so consumers select the matching fetch path; the RC/stable payload identifies the bundle by release tag instead of artifact id.
- No change to how wheels are built, the manifest/checksum format, or the nightly flow.

## Summary by CodeRabbit

* **Chores**
  * Release pipeline now records whether a bundle comes from a nightly artifact or an RC/stable release and exposes the release tag and checksums for RC/stable bundles.
  * Nightly bundles continue to be uploaded as artifacts; RC/stable bundles are published as release assets with accompanying checksums.


.github/workflows/release-bundle.yaml
```diff
@@ -443,15 +449,56 @@ jobs:
...
+          # consumer restores wheels/ from the manifest paths. Re-run model: if a release already
+          # exists for this tag the command fails ("a release with the same tag name already
+          # exists"); delete it (gh release delete "${RELEASE_TAG}") to redo.
+          gh release create "${RELEASE_TAG}" \
+            release-bundle/release-manifest.json \
+            release-bundle/checksums.txt \
+            release-bundle/wheels/*.whl \
+            --title "${RELEASE_TAG}" \
+            --notes "Automated ${CADENCE} release bundle for ${RELEASE_TAG}. checksums.txt sha256=${checksums_digest}" \
+            "${prerelease[@]}"
```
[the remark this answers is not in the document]
svvarom: Note from Claude Code: this is intentional. We chose mutable releases + delete-to-redo over a `--clobber` upsert: a rerun on an existing tag is meant to fail, and recovery is deleting the release. The common path (GitHub's "re-run failed jobs") re-runs `dispatch`, not `assemble`, so it never re-hits `gh release create`. Leaving as-is.

mquong-ca (approved):
lgtm

.github/workflows/release-bundle.yaml
```diff
@@ -443,15 +449,60 @@ jobs:
...
+          # Upload each artifact as its own asset. GitHub flattens asset names to basenames; the
+          # consumer restores wheels/ from the manifest paths. Re-run model: if a release already
+          # exists for this tag the command fails ("a release with the same tag name already
+          # exists"); delete it (gh release delete "${RELEASE_TAG}") to redo.
+          gh release create "${RELEASE_TAG}" \
+            release-bundle/release-manifest.json \
+            release-bundle/checksums.txt \
+            release-bundle/wheels/*.whl \
+            --title "${RELEASE_TAG}" \
+            --notes "Automated ${channel} build of NeMo Platform ${RELEASE_TAG}." \
```
svvarom, line 498: This will be changed in a follow up PR

svvarom: Will use gh release note generation.

svvarom: This is just a placeholder, nightlies consume artifacts so this should be a no-op until a real release

crookedstorm (approved):
This looks good! Made some pie-in-the-sky comments offline.

PR #243, opened 2026-06-09
feat(studio): Filtering for ExperimentGroup DataView

nakolean:
https://github.com/user-attachments/assets/c679e16b-cd94-44ba-907b-ed06faeb2e1c

## Summary by CodeRabbit

* **New Features**
  * Advanced date/time filtering for experiment columns (Name, Created, Updated At).
  * Experiment-group view now applies the active DataView filter criteria for more accurate results.
  * Default column visibility streamlined to hide non-essential columns.

* **Bug Fixes / Improvements**
  * Text filter input now reliably reflects external clear/reset actions and uses a debounced input for smoother typing behavior.
  * Debounce default standardized across inputs for consistent responsiveness.

### DIFF
--- a/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/index.tsx
+++ b/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/index.tsx
@@ -97,10 +97,6 @@ export const ExperimentGroupDataView: FC<ExperimentGroupDataViewProps> = ({
     [experimentsData]
   );
 
-  if (groupError) {
-    return <ErrorMessage message="Failed to load experiment group." />;
-  }
-
   const makeColumns = useCallback<
     ComponentProps<typeof DataViewRoot<ExperimentRow>>['makeColumns']
   >(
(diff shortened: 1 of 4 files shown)


aray12 (approved):

rrhyne:
EDIT: IGNORE, these are violating existing studio patterns

Looking good! Let's make these changes:

* Reduce the space between the filter label and the text input
* The applied filters and the clear filter button should be on the same line as the filters button
* Let's add the count on the left both when filtered and when not filtered as per the designs

web/packages/studio/src/components/dataViews/ExperimentGroupDataView/index.tsx
```diff
@@ -47,8 +52,10 @@ export const ExperimentGroupDataView: FC<ExperimentGroupDataViewProps> = ({
   } = useGetExperimentGroup(workspace, experimentGroupName);
   const experimentGroupId = group?.id ?? '';
 
-  const dataViewState = useStudioDataViewState({
+  const dataViewState = useStudioDataViewState<ExperimentFilter>({
     defaultSort: { id: 'created_at', desc: true },
+    // created_by isn't returned by the API and updated_at isn't shown; both are filter-only.
+    columnVisibility: { created_by: false, updated_at: false },
```
shanaiabuggy, line 58: ah, I will update the API to return `created_by` so we can create a column for it, don't worry about that in this PR tho. 

honestly, lets just drop the `updated_at` filter. I'm not convinced I want it as a table column and I don't want filters to exist without an associated column.

shanaiabuggy (approved):
Gorg! Left 1 comment inline.

Also nit: Can we add placeholders to the text filters? Like "Filter by agent name"

nakolean:
Added search bar, moved `name` filter to search. Used $like so it will substring search instead of exact match filtering.

https://github.com/user-attachments/assets/24411f1a-5fa1-4075-8166-9de470a51fc7

PR #247, opened 2026-06-09
chore: Data Designer e2e tests

mikeknep:
### The main thing

Restores the Data Designer e2e tests. Includes a test for workloads that require Nemotron Personas data, which previously was untested at the e2e level.

### Also featuring

Fixes a Files service bug. Currently you cannot delete a Fileset that has an invalid (presumably: already deleted) secret, because a `StorageImpl` is a prerequisite to the actual data and entity deletion, and the `StorageImpl` requires secrets, so when we call out to Secrets service to fetch the secret we fail with a `SecretNotFoundError`.

~~Second: currently when deleting a Fileset, we do not delete cached data, we only delete the "primary" data (when owned; we no-op for NGC and HF).~~ <--This is expected behavior

## Release Notes

* **Bug Fixes**
  * Fileset deletion now conditionally manages underlying storage data based on backend ownership—local and S3 data are deleted; external storage is preserved.

* **Tests**
  * Added comprehensive Data Designer end-to-end tests covering config preview, job execution, and LLM model integration.
  * Enhanced test infrastructure for NGC-backed testing.


e2e/conftest.py
```diff
@@ -206,3 +232,37 @@ def workspace(sdk: NeMoPlatform) -> Iterator[str]:
...
+        cmd = ["uv", "run", "--project", str(repo_root), "--frozen", "nemo", "-f", "json", *args]
+        return subprocess.run(
+            cmd,
+            cwd=repo_root,
+            env=env,
+            timeout=timeout,
+            capture_output=capture_output,
+            stdin=stdin,
+            text=True,
+        )
```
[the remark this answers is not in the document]
mikeknep: This seems OK to me—I think it's plausible someone might want to test that a certain CLI invocation would produce an error, so leaving this up to the caller seems fine?

e2e/conftest.py
```diff
@@ -206,3 +229,37 @@ def workspace(sdk: NeMoPlatform) -> Iterator[str]:
     sdk.workspaces.create(name=name)
     yield name
     sdk.workspaces.delete(name)
+
+
+@pytest.fixture(scope="session")
+def nemo_run(_services: str) -> NemoRun:
```
matthewgrossman, line 235: this fixture is great.

services/core/files/src/nmp/core/files/api/v2/filesets/endpoints.py
```diff
@@ -414,10 +417,46 @@ async def delete_fileset(
...
+                f"Cannot delete fileset '{workspace}/{name}': its storage secret is "
+                f"unavailable, so the underlying data cannot be removed. Restore the "
+                f"secret and retry. ({exc})",
+            ) from exc
+        logger.warning(
+            f"Storage secret unavailable while deleting external fileset '{workspace}/{name}'; "
+            f"the source is not ours to delete, proceeding with entity deletion: {exc}"
+        )
+
+    # 1. Clean up any cached copies the platform materialized in default storage for
```
matthewgrossman: so one reason we put the cache in a totally separate directory de-coupled from the fileset was that you could have multiple filesets that point to the same data. personA makes a fileset pointing at llama3-8b, personB makes a fileset pointing at llama3-8b. 

If personA downloads it first, they'll fill the cache. Then when personB uses the fileset, they'll get a cache hit quickly. This was thought to be useful for things like ToT where various tests end up creating and using the same models.

Of course this *sounds* dangerous because then two filesets are pointing at the same data, but this only works for the "external" datasets where their cache path is delineated via a locked repo version. As in, the cache path for HF purposefully includes the commit/sha of the HF repo to make a pseudo content-addressed cache.

---

The reason I bring this is up is that I'm not sure that deleting a fileset should necessarily clear the cache for that repo? Especially if we have tests in ToT that are creating/deleting the same llama3-8b repo, we could save ourselves a lot of time by keeping the static data around in the cache.

So then the question comes "when DO we clear the cache". well, right now: never 🙃 But it feels like we should have some mechanism to audit/clean/prune the cache perhaps?

packages/nmp_common/src/nmp/common/files/storage_config.py
```diff
@@ -47,6 +47,40 @@ def get_secret_references(self) -> dict[str, SecretRef]:
         """Get the secret references for the storage config."""
         return {}
 
+    @property
+    def owns_storage_data(self) -> bool:
```
matthewgrossman, line 51: I'm glad we're adding something like this, I think we've really just been hacking along the s3/local vs ngc/hf in random if/else statements, when we should've been using a protocol here

I've definitely thought about this in the past, and where I thought about this going _some day_ was that it would actually be an attribute on fileset itself, rather than on the backend provider (though with some defaults perhaps)

The reason being certain s3 buckets: you might want an s3 bucket that is a normal-read-write bucket that you have access, but you also might've been given read-only access to some public s3 bucket. Thats why I think it's an attribute of the fileset itself, with perhaps some providers _only_ supporting read-only (like NGC/HF)

matthewgrossman: and in the case of "read only" I'd imagine that has the same semantics as "owns storage data" that you wrote here

services/core/files/src/nmp/core/files/api/v2/filesets/endpoints.py
mikeknep: Ahh, ok this is great context, I didn't know about that use case and thought this was a bug while exploring with Claude. I'll back out of the cache-deletion stuff here for now. Looking ahead, we'll need a larger discussion of course but my gut reaction is we might want to have two things:
- `delete_cache: bool = False` on the delete-fileset endpoint
- `prune_cache` as a standalone endpoint/operation

packages/nmp_common/src/nmp/common/files/storage_config.py
mikeknep: Great point. I'm on the fence about keeping it in this PR or dropping it; do you have an opinion? In this PR it's only used in one spot, and that spot is a place where we do have a `fileset` instance available, so if it became a Fileset property it'd be easy to change this one, but I don't know if there's a risk of us building up increased usage of this property as-is in places where we have a StorageImpl but not an actual Fileset, and it becomes harder to untangle...?

I'll see what things look like with/out it after I revert the cache deletion code (see other comment thread).

mikeknep: Update: I think it makes sense to keep the `owns_storage_data` on the storage configs as currently implemented here. Each Fileset instance has its own `storage: StorageConfig` instance, and ownership of storage data is **specific to storage type and instance**:
- HF and NGC = always False
- Local = always True
- S3 = varies by bucket

That suggests the property should live at the storage config level, with `S3StorageConfig` exposing a public `readonly: bool` field (exact name/impl TBD) that users set upon creation and that `owns_storage_data` delegates to.

services/core/files/src/nmp/core/files/api/v2/filesets/endpoints.py
mikeknep: Reverted cache-deletion logic in cf1c62a5b

packages/nmp_common/src/nmp/common/files/storage_config.py
matthewgrossman: sgtm!

matthewgrossman (approved):

PR #252, opened 2026-06-09
ci: add label-driven GitHub release notes for RC/stable releases

svvarom:
## What this does

RC and stable GitHub Releases currently ship a one-line body. This makes them produce **categorized release notes** (New Features, Bug Fixes, Documentation, …) that GitHub generates automatically from PR labels.

Three pieces:

1. **`.github/workflows/label-pr.yaml`** (new) — applies a conventional-commit type label (`feat`, `fix`, `docs`, …) to each PR, derived from its title. Complements `semantic-pull-requests.yaml`, which validates the title but doesn't label.
2. **`.github/release.yml`** (new) — groups those labels into release-note sections.
3. **`.github/workflows/release-bundle.yaml`** — the RC/stable release step now emits GitHub's generated, categorized notes, scoped to changes since the previous stable release.

## Notes for reviewers

- Only the release-notes **body** changes. Release assets, the downstream handoff, and the nightly path are unchanged.
- The labeler runs on `pull_request_target` so PRs from forks are labeled too; it only reads the PR title from the event payload and never checks out PR code.
- Type labels are created on first use (auto-assigned color, no description).
- Both RC and stable notes span from the previous stable tag, so each RC shows the full delta since the last stable.

## Summary by CodeRabbit

* **Chores**
  * Configured automated GitHub release notes generation with categorized changelog sections based on commit types.
  * Added automatic pull request labeling workflow to streamline release organization.


mquong-ca (approved):
Just the two files;lgtm

svvarom:
Thanks for approving, @mquong-ca. 

@crookedstorm, can you pls take a look as well? Just want to be mindful cause this will affect every future PR.

.github/workflows/release-bundle.yaml
```diff
@@ -200,11 +202,26 @@ jobs:
             release_date_json="\"${RELEASE_DATE}\""
           fi
 
+          # Generated release notes span from the previous stable release so RC and stable both
+          # show the full delta since the last stable. Empty when no prior stable tag exists.
+          # Hardened for `set -euo pipefail`: `|| true` covers the no-stable-tags case, and awk
+          # reads to EOF (no early `exit`/`head`) so `sort` is never killed by SIGPIPE.
```
crookedstorm, line 208: Won't this method break on patches?

.github/workflows/label-pr.yaml
```diff
@@ -0,0 +1,20 @@
+name: Label PR by type
```
crookedstorm, line 1: Only problem here: What about when the commit checker fails? Does this just fail as well? That's ok, if so.

svvarom: Yes - it fails in lockstep with the semantic check (had Fable confirm it)

.github/workflows/release-bundle.yaml
svvarom: tested this in a scratch repo. A patch like `1.0.1` generates notes against `1.0.0`, even when a newer release already exists (that's the case GitHub's default gets wrong, and why we pick the tag explicitly). Hotfixes cut from a release branch also come out clean. Verified end-to-end with `gh release create --generate-notes`.

.github/workflows/label-pr.yaml
svvarom: We could actually drop [semantic-pull-requests.yaml](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/.github/workflows/semantic-pull-requests.yaml) since this validates the same thing, but we'd lose its title-length check

.github/workflows/release-bundle.yaml
crookedstorm: Nice!

crookedstorm (approved):

PR #259, opened 2026-06-10
fix(tests): Mock HF in integration tests

matthewgrossman:
## Problem

`TestTrustRemoteCodePermission` integration tests create HuggingFace-backed filesets (`storage.type=huggingface`), which triggers real HF Hub API calls (`repo_info`, `get_hf_file_metadata`) during `validate_storage()` and `resolve_config()`. These calls are rate-limited by HuggingFace, causing all 6 tests to fail with:

```
nemo_platform.InternalServerError: Error code: 502 - {'detail': 'Storage backend unavailable:
Rate limited for https://huggingface.co/api/models/Qwen/Qwen3-0.6B/revision/main'}
```

Setting `HF_TOKEN` in CI (PR #202) reduced the frequency but didn't eliminate the issue — rate limits still apply to authenticated users, and concurrent CI runs sharing the same token compound the problem.

## Strategy

**Mock at the Files service layer, not in individual tests.** These tests verify authorization behavior (who can set `trust_remote_code`), not HuggingFace connectivity. The real HF API calls are an unnecessary side effect of fileset creation.

A `no_hf_network` fixture (defined in `conftest.py`) patches `HuggingfaceStorageImpl.validate_storage` and `resolve_config` to no-ops. This is the same pattern already used by the Files service auth tests in `tests_filesets_with_auth_secrets.py`. It's better than mocking `HfApi` directly because it only disables the two methods that egress to HF, leaving everything else (like `list_files`) intact.

`TestTrustRemoteCodePermission` uses this fixture via an autouse wrapper so every test in the class gets the mock automatically.

### What about other tests?

- **Files service HF endpoint tests** (`test_huggingface_endpoints.py`) — already safe, use ASGI transport (no real network)
- **Files external storage tests** (`test_huggingface_storage.py`) — gated behind `RUN_EXTERNAL_STORAGE_TESTS=1`, skipped by default
- **Files auth+secrets tests** (`tests_filesets_with_auth_secrets.py`) — already use `no_hf_network`
- **Models parallelism tests** — guarded by `pytest.importorskip("torch")`, skipped in CI (torch not installed)

The only unprotected tests hitting real HF in CI were the 6 `TestTrustRemoteCodePermission` tests fixed here.

## Test plan

- [x] All 6 `TestTrustRemoteCodePermission` tests pass locally (verified with `HF_HUB_OFFLINE=1`)
- [ ] CI integration tests pass
- [ ] No other integration test regressions

Fixes [AIRCORE-744](https://linear.app/nvidia/issue/AIRCORE-744)

## Release Notes

* **Tests**
  * Enhanced test isolation to prevent unintended external API calls during test execution
  * Improved integration test configuration for authorization behavior verification


mckornfield (approved):
serenity now!

services/core/models/tests/integration/conftest.py
```diff
@@ -52,6 +59,14 @@ async def _resolve_passthrough(self):
     monkeypatch.setattr(HuggingfaceStorageImpl, "validate_storage", _validate_noop)
     monkeypatch.setattr(HuggingfaceStorageImpl, "resolve_config", _resolve_passthrough)
 
+    def _poisoned_repo_info(self, *args, **kwargs):
```
mckornfield: lol love me some poison https://www.youtube.com/watch?v=yJw2qd3MtqE

matthewgrossman: this was silly and I removed it

PR #278, opened 2026-06-11
fix: Data Designer UX improvements

mikeknep:
### What's here

Some CLI improvements to prevent leaking ugly stack traces when we already are raising a clear custom error message.

Restores the old "save as batches complete" job results logic. This feature was temporarily removed when we switched to use the library's high-level `DataDesigner` interface object server-side—it previously did not expose the hook (we could only access it because we were operating "one level down").

Adds a `name` property to the high-level `[Async]DataDesignerJobResource` objects for easy access to the unique identifier for re-fetching that object later.

Lifts a Jobs integration test helper to a shared location so that Data Designer integration tests can also set it. I noticed that the DD integration tests were failing locally when run _without_ Docker running in the background. This fixes that.

## Release Notes

* **Bug Fixes**
  * Improved error messages for network connectivity and timeout issues
  * CLI now displays clear error messages without stack traces for configuration errors

* **New Features**
  * Added incremental artifact saving during batch job processing
  * Made job name accessible through public API

* **Tests**
  * Expanded integration test coverage for error handling and partial dataset scenarios

### DIFF
--- a/packages/data_designer_nemo/src/data_designer_nemo/model_provider.py
+++ b/packages/data_designer_nemo/src/data_designer_nemo/model_provider.py
@@ -10,7 +10,14 @@ from data_designer.engine.model_provider import ModelProvider as NDDModelProvide
 from data_designer.engine.model_provider import ModelProviderRegistry, resolve_model_provider_registry
 from data_designer_nemo.errors import NDDInternalError, NDDInvalidConfigError
 from data_designer_nemo.sdk_translation import sync_to_async_sdk
-from nemo_platform import AsyncNeMoPlatform, NeMoPlatform, NotFoundError, PermissionDeniedError
+from nemo_platform import (
+    APIConnectionError,
+    APITimeoutError,
+    AsyncNeMoPlatform,
+    NeMoPlatform,
+    NotFoundError,
+    PermissionDeniedError,
+)
 from nemo_platform.types.inference import ModelProvider as NMPModelProvider
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


packages/data_designer_nemo/src/data_designer_nemo/model_provider.py
```diff
@@ -205,6 +212,17 @@ async def _get_nmp_provider(
                 f"Cannot access provider {user_supplied_provider_name!r}. Check that it exists and you have access to it."
             )
             self.inaccessible_providers.add(user_supplied_provider_name)
+        except (APIConnectionError, APITimeoutError) as e:
```
mikeknep, line 215: Without this change, these errors' stacktraces would leak out to the CLI, **even though** the error was getting wrapped in an `NDDInternalError`, due to the `logger.exception` call.

I think we want to keep `logger.exception` for truly unknown/unexpected `Exception`, but these ones we don't need the full trace (unless debug mode is on)

plugins/nemo-data-designer/src/nemo_data_designer_plugin/jobs/run.py
```diff
@@ -86,8 +86,11 @@ def _run_step_config(
         dd_ctx=dd_ctx,
     )
     data_designer.set_run_config(dd.RunConfig(buffer_size=BUFFER_SIZE))
-    # TODO: set `on_batch_complete=lambda _: result_manager.save_artifacts()` once it is available here
-    dataset_creation_results = data_designer.create(config_builder, num_records=step_config.job_config.num_records)
+    dataset_creation_results = data_designer.create(
+        config_builder,
+        num_records=step_config.job_config.num_records,
+        on_batch_complete=lambda _: result_manager.save_artifacts(),
```
mikeknep, line 92: This became available when we bumped to `0.6.1`, I just forgot to fold it in to that PR

matthewgrossman (approved):
subprocess changes look fine to me!

PR #280, opened 2026-06-11
feat(deployments): scaffold plugin API and registry (AIRCORE-755)

tylersbray:
## Summary

Scaffolds `nemo-deployments-plugin` as the substrate-agnostic contract layer for [AIRCORE-755](https://linear.app/nvidia/issue/AIRCORE-755/deployments-plugin-scaffold-the-plugin-entities-api-backend-abc):

- **Entities:** `DeploymentConfig`, `Deployment`, `Volume` plus RFC supporting types
- **API v1:** CRUD for configs/deployments/volumes at `/apis/deployments/v1/workspaces/{workspace}/...`
- **Reconciler hooks:** bulk `status_in` list filter; controller-only `PUT .../status` (service-principal gated via `X-NMP-Principal-Id`)
- **Backend contract:** `DeploymentBackend` ABC + `BackendStatusUpdate` / `VolumeStatusUpdate`
- **Executor registry:** named-instance pattern; plugin starts cleanly with zero backends registered
- **Validation:** prerequisite cycle detection (paginated config graph)

No substrate backends or reconcile controller in this PR — those are AIRCORE-756/757/758.

## Test plan

- [x] `uv run pytest plugins/nemo-deployments/tests/unit -v` (31 tests)
- [x] `uv run ruff check plugins/nemo-deployments`
- [x] `uv run ty check plugins/nemo-deployments`
- [x] Pre-commit hooks pass (including `no-nmp-common-in-plugins`)
- [ ] Manual smoke: `nemo services run` → CRUD against `/apis/deployments/v1/...` (follow-on if desired)

## Release Notes

* **New Features**
  * Added the NeMo Deployments plugin for end-to-end deployment lifecycle management (configs, deployments, and volumes).
  * Introduced v2 REST API endpoints for full CRUD plus controller-only status updates.
  * Added support for multiple backend executors and backend-driven create/read/delete and logging.
  * Implemented prerequisite dependency validation (cycle detection) and referential integrity checks for safe deletes.
  * Added pagination, sorting, and filtering (including deployment status filtering).

* **Tests**
  * Added unit tests covering API behavior, entity defaults/validation, reference resolution, and backend registry logic.

* **Documentation**
  * Added plugin README with high-level lifecycle overview.

* **Chores**
  * Enabled and wired the plugin into the workspace build and test discovery.


plugins/nemo-deployments/README.md
```diff
@@ -0,0 +1,33 @@
+# NeMo Deployments Plugin
+
+Substrate-agnostic deployment lifecycle for the NeMo Platform. This plugin provides
+entity schemas, CRUD APIs, a `DeploymentBackend` ABC, and an executor registry.
+
+**Scope (this ticket):** scaffold only — entity types, v1 CRUD routes, backend contract,
```
benmccown: Nit: We can probs drop everything from this `Scope` line through to the end of the file, save for the `Tests` section.

After the work is done and backends are ready, I'm sure an agent can cook up a good README that's helpful and doesn't have transient info in it.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
```diff
@@ -0,0 +1,93 @@
...
+    backoff_limit: int = 6
+    prerequisites: list[Prerequisite] = Field(default_factory=list)
+    drift_recovery: DriftRecoveryPolicy | None = None
+    labels: dict[str, str] = Field(default_factory=dict)
+    backend_config: DeploymentBackendConfig = Field(default_factory=DeploymentBackendConfig)
+
+
+class CreateDeploymentRequest(BaseModel):
+    name: str
+    deployment_config_name: str
```
benmccown: You might have lifted this one to one from the RFC, but I'm wondering aloud, should we just go with `deployment_config` so that we could in theory support `workspace/name` _or_ `name` (assume same workspace)? Maybe there's a world where we want to support deploying a deployment config from workspace A into workspace B? Just keeps our options open.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
```diff
@@ -0,0 +1,93 @@
...
+    backend_config: VolumeBackendConfig = Field(default_factory=VolumeBackendConfig)
+
+
+class UpdateDeploymentStatusRequest(BaseModel):
+    status: DeploymentStatus
+    status_message: str = ""
+    endpoints: list[Endpoint] = Field(default_factory=list)
+    exit_code: int | None = None
+    error_details: dict[str, Any] | None = None
+    status_history: list[StatusEvent] | None = None
```
benmccown: I'm not sure we need `status_history` on the API request? The entity itself should keep a `status_history` field for us, but it shouldn't be on the requester to preserve/maintain that history field.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
```diff
@@ -0,0 +1,93 @@
...
+    name: str
+    size: str = "1Gi"
+    access_modes: list[AccessMode] = Field(default_factory=lambda: ["ReadWriteOnce"])
+    backend_config: VolumeBackendConfig = Field(default_factory=VolumeBackendConfig)
+
+
+class UpdateDeploymentStatusRequest(BaseModel):
+    status: DeploymentStatus
+    status_message: str = ""
+    endpoints: list[Endpoint] = Field(default_factory=list)
```
benmccown: Does `endpoints` need to exist on this request object? Wouldn't it be something on the top level `Deployment` entity?

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -0,0 +1,245 @@
...
+    containers: list[Container] = Field(default_factory=list)
+    init_containers: list[Container] = Field(default_factory=list, alias="initContainers")
+    volume_mounts: list[VolumeMount] = Field(default_factory=list, alias="volumeMounts")
+    config_files: list[ConfigFile] = Field(default_factory=list, alias="configFiles")
+    restart_policy: RestartPolicy = Field(default="Always", alias="restartPolicy")
+    backoff_limit: int = Field(default=6, alias="backoffLimit")
+    prerequisites: list[Prerequisite] = Field(default_factory=list)
+    drift_recovery: DriftRecoveryPolicy = Field(default_factory=DriftRecoveryPolicy, alias="driftRecovery")
+    labels: dict[str, str] = Field(default_factory=dict)
+    backend_config: DeploymentBackendConfig = Field(default_factory=DeploymentBackendConfig, alias="backendConfig")
```
benmccown, line 210: This is straight from the RFC, but now I'm _that guy_ and thinking about one alternative out loud:

Wdyt about `executor_config`? I hadn't thought about the "Executor (instance) vs. Service Backend (class)" disambiguation much when writing the RFC. The class could become `DeploymentBackendConfig -> DeploymentExecutorConfig`. I feel like it's potentially clearer?

No strong feelings here, we can keep it as is as well.

plugins/nemo-deployments/src/nemo_deployments_plugin/service.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def get_routers(self) -> list[RouterSpec]:
+        from nemo_deployments_plugin.api.v1 import (
+            deployment_configs,
+            deployments,
+            status,
+            volumes,
+        )
+
+        prefix = "/v1/workspaces/{workspace}"
```
benmccown: This (and all API routes anywhere in the PR) should be `/v2/` and not `/v1/`.

plugins/nemo-deployments/src/nemo_deployments_plugin/types.py
```diff
@@ -0,0 +1,31 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
benmccown, line 1: Nit: I think these types could all just go in `schemas.py` but it doesn't really matter, tomatoe/tomatoe.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
```diff
@@ -0,0 +1,41 @@
...
+
+    executors: list[ExecutorConfigEntry] = Field(
+        default_factory=list,
+        description="Named executor instances. May be empty at scaffold time.",
+    )
+    default_executor: str | None = Field(
+        default=None,
+        description="Fallback executor when Deployment.executor is unset.",
+    )
+    port_range_start: int = Field(default=9000, description="Default Docker port range start.")
```
benmccown: Shouldn't the port range start/end go in the docker service/executor config? Cause k8s executors won't use these, right?

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
```diff
@@ -0,0 +1,93 @@
...
+    Volume,
+    VolumeBackendConfig,
+    VolumeMount,
+    VolumeStatus,
+)
+from nemo_platform_plugin.schema import NemoFilter, NemoListResponse
+from pydantic import BaseModel, Field
+
+
+class CreateDeploymentConfigRequest(BaseModel):
```
benmccown, line 31: Doesn't have to be in _this_ PR, but we should probs think through `Update` APIs at some point (which fields are mutable, maybe implement a naive form of entity versioning copying the established pattern from models service).

plugins/nemo-deployments/src/nemo_deployments_plugin/api/v2/deployment_configs.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@router.delete("/deployment-configs/{name}", status_code=204, tags=["Deployment Configs"])
+async def delete_deployment_config(
+    workspace: str,
+    name: str,
+    entity_client: NemoEntitiesClient = Depends(get_entity_client),
+) -> None:
+    try:
+        await entity_client.delete(DeploymentConfig, name=name, workspace=workspace)
```
benmccown, line 149: Could we ensure that there are no existing Deployments that reference the config and 4xx with a helpful message if so? Same goes for Volumes (if they're referenced by deployments).

benmccown (changes requested):
Mostly minor stuff that an agent should be able to address without too much issue. Nice work so far.

plugins/nemo-deployments/README.md
tylersbray: Done in 32f18f5 — trimmed to intro + Tests only. Agree transient scaffold docs are not worth maintaining; we can regenerate a fuller README once backends land.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
tylersbray: Keeping `deployment_config_name` for this merge to stay aligned with the RFC and existing entity/filter field names. The cross-workspace `workspace/name` ref idea is good — I would rather capture it as an RFC amendment when we have a concrete use case (deploy config from workspace A into B), since it adds parsing/validation surface. Happy to revisit if that scenario is near-term.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
tylersbray: Agree — removed `status_history` from `UpdateDeploymentStatusRequest` in 32f18f5. The reconciler (758) will append to `Deployment.status_history` server-side on each transition rather than expecting the caller to round-trip history.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
tylersbray: Keeping `endpoints` on the status PUT for now. This route is controller-only (service principal); it is the write path for projecting `BackendStatusUpdate` from the reconciler into the entity — same as status, exit_code, and error_details. End users never call it directly. Open to splitting later if we want a narrower contract.

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
tylersbray: `executor_config` / `DeploymentExecutorConfig` is clearer disambiguation — agree in principle. Since you said no strong feelings and we are RFC-aligned, I would defer the rename to avoid churn across entities, JSON aliases, and 758. Can open a naming cleanup ticket if we want consistency with `Deployment.executor` + `BACKEND_CLASSES`.

plugins/nemo-deployments/src/nemo_deployments_plugin/service.py
tylersbray: Done in 32f18f5 — renamed `api/v1` → `api/v2`, prefix is now `/v2/workspaces/{workspace}`, matching agents/unsloth convention. Will rebase 758 onto this.

plugins/nemo-deployments/src/nemo_deployments_plugin/types.py
tylersbray: Keeping the split intentionally: `types.py` holds shared literals + `Endpoint` with no `NemoEntity` imports, which avoids the import cycle we hit with `backends/base.py` (CodeQL flagged it when everything was in one module). `schema.py` stays API request/filter models. Happy to merge later if the layering stabilizes, but not much win right now.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
tylersbray: Done in 32f18f5 — removed plugin-level `port_range_*` from `DeploymentsConfig`. Docker port range remains on `DockerDeploymentConfig` in entities and in per-executor config dicts (756). K8s executors will not see it.

plugins/nemo-deployments/src/nemo_deployments_plugin/schema.py
tylersbray: Agree — worth a dedicated pass once we know which fields are mutable (config is immutable-by-design today). Suggest a follow-up ticket post-758, potentially borrowing models-service versioning patterns. Out of scope for 755.

plugins/nemo-deployments/src/nemo_deployments_plugin/api/v2/deployment_configs.py
tylersbray: Done in 32f18f5 — config delete returns 409 when deployments reference it; volume delete returns 409 when deployment-config volume mounts reference it. Helpers in `references.py` + unit tests.

plugins/nemo-deployments/src/nemo_deployments_plugin/api/v2/deployment_configs.py
```diff
@@ -0,0 +1,152 @@
...
+        joined = ", ".join(referencing)
+        raise HTTPException(
+            status_code=409,
+            detail=(
+                f"DeploymentConfig '{name}' is referenced by deployment(s): {joined}. Delete those deployments first."
+            ),
+        )
+
+    try:
+        await entity_client.delete(DeploymentConfig, name=name, workspace=workspace)
```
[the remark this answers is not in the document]
tylersbray: Entity store does not expose a conditional delete (or equivalent compare-and-swap) API today, so we cannot make the referential check and delete fully atomic in plugin code alone. The pre-delete list check in `references.py` is best-effort to prevent accidental deletes and return a helpful 409 — it closes the common case but not the concurrent-create race you describe.

I added a comment at the delete handler noting this limitation. A durable fix would be entity-store support for referential integrity or conditional delete; happy to track that as a platform follow-up if useful.

benmccown:
Changes look good so far. Thoughts on the 3 remaining?

---

Here's the final complete picture:

**Fully addressed (5/9 actionable items):**
- ✅ v1 → v2 routes
- ✅ README trimmed to just the Tests section — exactly what you asked for
- ✅ Referential check on `DeploymentConfig` delete (409 with deployment names)
- ✅ Referential check on `Volume` delete (409 listing blocking configs, wired into `volumes.py`)
- ✅ `port_range_start/end` — need to verify, let me check `config.py` is updated

**Not addressed (3/9 actionable items):**
- ❌ `deployment_config_name` → `deployment_config` on `CreateDeploymentRequest` — field unchanged
- ❌ `status_history` shouldn't be on `UpdateDeploymentStatusRequest` — still there
- ❌ `endpoints` shouldn't be on `UpdateDeploymentStatusRequest` — still there

tylersbray:
Thanks for the follow-up — quick status on the three:

**`status_history` on `UpdateDeploymentStatusRequest`** — already done in `32f18f59` (removed from the request schema and status handler). Reconciler (758) will append server-side.

**`endpoints` on `UpdateDeploymentStatusRequest`** — agreed; removed in latest push. Endpoints stay on the `Deployment` entity; the reconciler can set them via entity client when projecting `BackendStatusUpdate`, rather than through the status PUT body.

**`deployment_config_name` → `deployment_config`** — done on both `CreateDeploymentRequest` and the `Deployment` entity (plus `DeploymentFilter`). Accepts bare `name` (same workspace) or `workspace/name` on create.

Also rebased the branch onto current `main` (linear history, no merge commit).

benmccown (approved):

PR #286, opened 2026-06-11
fix(customizer): lone-root fallback accepts .json in addition to .jsonl

aray12:
## Summary

- `discover_dataset_files` fallback (`preparation.py`) used `glob("*.jsonl")`, missing `.json` files even though every other discovery path (patterns, subdir walk) treated both extensions equally. A lone `data.json` at fileset root with no train/val naming pattern would raise `DatasetFormatError`.
- Replaced the glob with `iterdir()` + suffix check for `(".jsonl", ".json")` to match the behavior of the other paths.
- Mirrored the fix in Studio's `partitionDatasetFiles` so the UI discovery stays in sync with customizer.

## Test plan

- [x] 3 new Python unit tests: lone `.json` → training, mixed `.jsonl`+`.json` → both claimed, multiple `.json` → all claimed
- [x] Studio spec updated: flipped "does NOT park .json" test and added mixed `.jsonl`+`.json` case
- [x] All 38 Python tests pass (`uv run pytest services/customizer/tests/tasks/training/test_datasets.py`)
- [x] All 13 Studio tests pass (`pnpm --filter nemo-studio-ui test src/hooks/useDatasetFileDiscovery/index.spec.ts`)

Closes ASTD-109

## Release Notes

* **Bug Fixes**
  * Enhanced dataset file discovery to include both `.json` and `.jsonl` files in fallback scenarios when configured training/validation patterns don't match any files.
  * Improved fallback behavior: single unmatched files are automatically designated as training data; multiple files are all treated as training data.

* **Tests**
  * Added comprehensive test coverage for new dataset discovery fallback behavior with mixed file types.

### DIFF
--- a/services/customizer/src/nmp/customizer/tasks/training/datasets/preparation.py
+++ b/services/customizer/src/nmp/customizer/tasks/training/datasets/preparation.py
@@ -321,19 +321,21 @@ def discover_dataset_files(dataset_path: Path) -> tuple[list[Path], list[Path]]:
     # Discover validation files
     val_files = _discover_files_by_patterns(dataset_path, VAL_PATTERNS, VAL_DIRS)
 
-    # Fallback: if no files found with patterns, check for any .jsonl files
+    # Fallback: if no files found with patterns, check for any .jsonl/.json files
     if not train_files and not val_files:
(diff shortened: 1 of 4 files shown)


albcui (commented):
Just making sure that the intent is to allow `.json` files as a fallback, but the content in the file must be in json lines format.

PR #302, opened 2026-06-12
fix(jobs): resolve default task image from platform config on K8s

matthewgrossman:
## Summary

Fixes [AIRCORE-790](https://linear.app/nvidia/issue/AIRCORE-790/job-execution-fails-on-k8s-task-pod-tries-to-pull-placeholderunused): job e2e tests fail on K8s because task pods try to pull `placeholder:unused`.

- Make `ContainerSpec.image` optional — jobs can omit it when they just need to run a command
- Add `default_task_image` to execution profile config as an explicit override
- When neither is set, fall back to `{platform.image_registry}/nmp-cpu-tasks:{platform.image_tag}` (same pattern as automodel/unsloth task images)
- K8s and Docker backends resolve the image at schedule time via `resolve_task_image()`
- E2e tests no longer use `placeholder:unused` — they omit the image and let the platform resolve it
- Add `docs/k8s-dev-guide.md` with updated build/deploy instructions using nemo-platform's `docker-bake.hcl`

## Test plan

- [x] Unit tests: 318 passed (`services/core/jobs/tests/`)
- [x] Integration tests: 5 passed (`services/core/jobs/tests/integration/`)
- [x] E2e on minikube: 7 passed, 1 failed (pre-existing persistent storage issue), 5 skipped
- [x] Validated from clean `minikube delete` — full setup, build, deploy, test cycle
- [x] Task pods use `my-registry/nmp-cpu-tasks:local` (confirmed via pod events)

## Summary by CodeRabbit

* **New Features**
  * Container image is now optional for job steps; when omitted, the execution profile’s new default_task_image is used (falls back to the platform CPU tasks image when unset).

* **Documentation**
  * Configuration reference updated to document default_task_image and its fallback behavior.

* **Tests**
  * E2E job definitions updated and unit tests added to verify image resolution behavior.


docs/set-up/config-reference.mdx
mckornfield, line 1: wonder if this file needs to be marked as generated...

mckornfield (approved):

services/core/jobs/src/nmp/core/jobs/controllers/backends/base.py
```diff
@@ -106,6 +111,27 @@ def validate_env_no_reserved_names(self) -> JobExecutionProfileConfig:
...
+    3. Platform CPU tasks image derived from platform.image_registry / image_tag
+    """
+    if container_image:
+        return container_image
+    if default_task_image:
+        return default_task_image
+
+    from nemo_platform_plugin.jobs.image import get_qualified_image
+
+    return get_qualified_image(_DEFAULT_TASK_IMAGE_NAME)
```
mckornfield:
```suggestion
    from nemo_platform_plugin.jobs.image import get_qualified_image

    return container_image or default_task_image or get_qualified_image(_DEFAULT_TASK_IMAGE_NAME)
```
if the import isn't that expensive

docs/set-up/config-reference.mdx
matthewgrossman: ya its a good question, we should probably do that. putting on the todo listtttt

services/core/jobs/src/nmp/core/jobs/controllers/backends/base.py
matthewgrossman: added!

PR #305, opened 2026-06-12
feat(models): vllm k8s support

benmccown:
# vLLM deployments on the Kubernetes backend (no operator) + reconciler refactor

## Summary

Adds first-class **vLLM** support to the models-controller Kubernetes backend by
emitting **native Kubernetes objects directly** (PVC / weight-puller Job /
Deployment / Service) — no `k8s-nim-operator` required for the vLLM path. As part
of this, the previously monolithic `K8sNimOperatorServiceBackend` is split into a
small reconciler hierarchy so the operator path (NIM) and the direct-emission
path (vLLM) live in separate, focused units behind a common interface.

The NIM-on-operator path is preserved and behavior-neutral throughout.

Linear: AIRCORE-694

## Motivation

The k8s backend previously only spoke `NIMService` / `NIMCache` CRs and delegated
all reconciliation to the in-cluster `k8s-nim-operator`. To run open models via
vLLM we need to stand up the serving stack ourselves. Rather than bolt vLLM logic
onto the operator backend, the engine paths are separated cleanly.

### vLLM on k8s (native objects, no operator)
- New `vllm_k8s_compiler.py`: engine-agnostic compilers for the PVC, weight-puller
  Job, Deployment, and Service.
- Staged, controller-gated rollout driven from the status path:
  - **P0 (create):** emit the PVC + weight-puller Job only (no serving objects yet,
    so the controller can gate on weight readiness).
  - **P3 (status, once the puller Job succeeds):** delete the completed puller Job
    to release its ReadWriteOnce volume, then emit the serving Deployment + Service
    with ownerReferences so a later delete cascades the PVC/Service.
- vLLM pods run as the image's `vllm` user (uid 2000 / gid 0) to avoid the torch
  `getpwuid` crash; the puller writes weights as the same uid/gid.
- In-cluster weight pull from the Files service via a cluster-routable HF endpoint
  (`apis/files/v2/hf`), not the local-service `localhost` URL the puller can't reach.
- Re-pull policy on update: weights are only re-pulled when the model source
  (name/revision) changes; otherwise the Deployment is patched in place and the
  owned PVC/Job survive.
- LoRA support: a cache-init container + adapters sidecar wired into the Deployment
  when LoRA is enabled.
- `generic` engine is explicitly rejected on the k8s backend (clear error).

### Reconciler refactor
- New `reconcilers/` package:
  - `base.py` — `BaseReconciler` ABC, the `ResolvedDeployment` input dataclass, and
    the shared, engine-agnostic Kubernetes status helpers (pod-log fetch, crash-loop
    detection, pod-status drill-down, pending-timeout/crash-loop error builders) +
    the idempotent, 404-tolerant single-object delete.
  - `nim_operator.py` — `NimOperatorReconciler`: emits `NIMService` / `NIMCache` CRs
    and projects status from the operator-reported `NIMService.status`.
  - `k8s.py` — `K8sReconciler`: the direct-emission vLLM reconciler described above.
- `K8sNimOperatorServiceBackend` is now a thin coordinator: it owns the
  `nemo_platform` SDK and API-object work, resolves everything a reconciler needs
  into a `ResolvedDeployment`, selects the reconciler by `config.engine`, and
  delegates. Delete asks **both** reconcilers (delete has no engine context) and
  aggregates results; list unions both paths for orphan reconciliation.

### Behavior changes worth calling out
- **No more NIM-operator CRD validation at startup.** The backend no longer fails
  fast when the `k8s-nim-operator` CRDs are absent, so a cluster without the
  operator can run vLLM-only deployments. Missing NIM CRDs now surface lazily only
  if a NIM deployment is actually created (delete/list tolerate their absence).
- **Status requires a config.** `get_model_deployment_status` takes the engine from
  the config (same selection as create/update). When the controller can't supply a
  config for a cycle (e.g. a transient config-fetch failure), the backend returns
  `UNKNOWN` instead of probing the cluster to guess the engine; the controller
  retries on the next poll and escalates to `ERROR` after its existing retry budget.

## Testing

- `services/core/models` unit suite: **passing** (added coverage for the vLLM
  compiler, the k8s backend phases/teardown/list/delete aggregation, the
  config-less `UNKNOWN` path, and the new config fields).
- `ruff` (lint + format), `ty`, and `pre-commit` all pass on the changed files.
- The vLLM path was validated end-to-end on a live cluster: a real Qwen3-1.7B vLLM
  deployment pulled weights, reached READY, and served inference (non-streaming and
  streaming) through the inference gateway.

# Review guide

This PR does two things at once: it **adds the vLLM-on-k8s path** and **refactors
the k8s backend into reconcilers**. Reading the files in dependency order (shared
pieces → reconcilers → coordinator → tests) makes both much easier to follow than
reading the raw diff top-to-bottom.

> Note on the diff: review against the branch's merge-base. A raw `origin/main..HEAD`
> diff also shows an unrelated prompts/prompt-service removal — that's already on
> `main`, not this branch's work.

### 1. Engine plumbing (shared, small) — start here
- `controllers/backends/engine.py` (+55) — engine discriminants (`nim` / `vllm` /
  `generic`), `config_engine()`, and `resolve_health_path()`. This is the
  vocabulary the rest of the PR dispatches on.
- `controllers/backends/vllm_compiler.py` (moved from `docker/`, ~178 lines) — the
  backend-agnostic vLLM compiler (image, `vllm serve` args, env). Relocated so both
  the docker and k8s backends share it; the diff is mostly the move + import fixups.
  Compare with `docker/creation_reconciler.py` to confirm the docker path is
  unchanged behaviorally.

### 2. The vLLM k8s object compiler
- `controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py` (new, 428) — pure
  functions that build the **PVC, weight-puller Job, Deployment, Service**. No
  control flow, no API calls — easiest place to verify the actual k8s objects
  (volumes, securityContext uid/gid, `/dev/shm`, probes, ownerRefs, labels).
  Paired tests: `tests/unit/controllers/backends/test_vllm_k8s_compiler.py` (+307).

### 3. The reconciler split (the heart of the refactor)
Read base → the two implementations → the coordinator.

- `.../reconcilers/base.py` (new, 448) — `BaseReconciler` ABC, the
  **`ResolvedDeployment`** input dataclass (what the coordinator pre-resolves and
  hands down), and the shared engine-agnostic status helpers (pod-log fetch,
  crash-loop detection, pod-status drill-down, pending-timeout/crash-loop error
  builders) + the idempotent, 404-tolerant single-object delete.
- `.../reconcilers/nim_operator.py` (new, 466) — `NimOperatorReconciler`. This is
  the **previous operator logic, moved verbatim** (NIMService/NIMCache CRUD, status
  from `NIMService.status`). The cheapest way to review: diff its method bodies
  against the old `backend.py` and confirm they're unchanged apart from taking a
  `ResolvedDeployment`.
- `.../reconcilers/k8s.py` (new, 600) — `K8sReconciler`, the new vLLM logic and the
  **most important file to review carefully** (see "Where to focus" below).
- `.../k8s_nim_operator/backend.py` (919 deletions, now 377) — the coordinator. The
  big deletion is code moving into the reconcilers. New responsibilities:
  `_resolve()` (build `ResolvedDeployment`), `_select_reconciler(engine)`,
  delegation in create/update/status, **delete that calls both reconcilers and
  aggregates**, and **list that unions both paths**.

### 4. Config + tests
- `.../k8s_nim_operator/config.py` (+57) — new vLLM-related fields (image/tag,
  service account, uid/gid defaults, `/dev/shm` size limit). Tests:
  `tests/unit/controllers/test_backend_config_fields.py`.
- `tests/unit/controllers/test_k8s_nim_operator_backend.py` (+681) — the bulk of the
  test churn. Note the helpers near the top: `_sync_reconcilers()` (propagates the
  per-test mock clients onto the reconcilers the backend built at init),
  `_StatusHelperReconciler` (exercises the `BaseReconciler` status helpers directly),
  and `_nim_config()` / `_vllm_config()`.

## Where to focus your attention (the subtle bits)

These are the parts most worth scrutiny — they encode hard-won operational behavior:

1. **The staged P0 → P3 rollout in `k8s.py`** (`create` + `get_status` +
   `_create_vllm_serving_objects`). Create emits **only** the PVC + puller Job;
   the serving Deployment/Service are created later from the status path once the
   puller Job succeeds. This is intentional weight-readiness gating.

2. **RWO volume hand-off** (`_delete_puller_job` + `_create_vllm_serving_objects`).
   The completed puller Job is deleted *before* the serving Deployment is created so
   its pod releases the ReadWriteOnce volume (avoids a Multi-Attach error if the
   server schedules on another node). Note the "still terminating → defer" return,
   and the `Job-absent + PVC-present → resume P3 (not LOST)` branch in `get_status`.

3. **`ownerReferences`** — PVC and Service are owned by the Deployment so a single
   Deployment delete cascades the rest.

4. **Update re-pull policy** (`update` + `_existing_model_source`) — weights re-pull
   only when the model-source annotation changes; unchanged source patches in place.

5. **Engine-agnostic teardown** (`backend.delete_model_deployment` →
   both reconcilers → aggregate). Delete has no engine context (also used for orphan
   reconciliation), so it attempts both paths, is 404-tolerant per-resource, and
   surfaces real failures as ERROR rather than falsely reporting DELETED. The
   forbidden-CR-doesn't-block-vLLM-cleanup case is covered by tests.

6. **uid/gid 2000/0** for vLLM pods + puller (`config.py` defaults, used in
   `vllm_k8s_compiler.py`). Avoids the torch `getpwuid` crash; there's a FUTURE note
   in the compiler flagging that a NIM raw-object path must pass its own uid/gid.

## Behavior changes to sanity-check (not just refactor)

- **No startup CRD validation.** `_validate_nim_operator_crds` is gone — a cluster
  without `k8s-nim-operator` can run vLLM-only. Confirm you're comfortable that
  missing NIM CRDs now surface lazily (on NIM create) rather than at boot.
- **Status without a config → `UNKNOWN`.** `get_model_deployment_status` no longer
  probes the cluster to guess the engine when `config is None`; it returns `UNKNOWN`
  and lets the controller retry/escalate. See the short branch at the top of that
  method and `deployment_reconciler.py`'s existing `_handle_unknown_status`.

## Release Notes

* **New Features**
  * Added Kubernetes raw-object emission for vLLM with staged weights pulling (PVC + pull Job) before creating serving resources.
  * Added engine-aware lifecycle dispatch for Kubernetes NIM Operator backends, including LoRA sidecar support where applicable.

* **Bug Fixes**
  * Improved deployment creation/status flows by using full deployment context consistently across backends.
  * Enhanced Kubernetes NIM Operator delete/status handling with better pending-timeout and pod crash diagnostics.

* **Documentation**
  * Updated NIM Operator config defaults (vLLM security context, images, service account, shared memory) and BusyBox image reference.
  * Updated the default Hugging Face model puller image to a dynamically qualified value.

### DIFF
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
@@ -1199,26 +1199,24 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
         annotations = (job.metadata.annotations or {}) if job.metadata else {}
         return annotations.get(vk8s.MODEL_SOURCE_ANNOTATION)
 
-    def _delete_vllm_resources(self, resource_name: str) -> None:
-        """Delete the directly-emitted vLLM objects by name (idempotent)."""
+    def _delete_vllm_resources(self, resource_name: str) -> list[str]:
+        """Delete the directly-emitted vLLM objects by name (idempotent).
+
+        Returns a list of concise error strings for any real (non-404) failures;
+        empty when everything was deleted or already absent.
+        """
         deleters = [
-            (self._apps_v1.delete_namespaced_deployment, resource_name, "Deployment"),
-            (self._core_v1.delete_namespaced_service, resource_name, "Service"),
-            (self._batch_v1.delete_namespaced_job, vk8s.pull_job_name(resource_name), "puller Job"),
-            (
-                self._core_v1.delete_namespaced_persistent_volume_claim,
-                vk8s.pvc_name(resource_name),
-                "PVC",
-            ),
+            (self._apps_v1.delete_namespaced_deployment, "Deployment", resource_name),
+            (self._core_v1.delete_namespaced_service, "Service", resource_name),
+            (self._batch_v1.delete_namespaced_job, "puller Job", vk8s.pull_job_name(resource_name)),
+            (self._core_v1.delete_namespaced_persistent_volume_claim, "PVC", vk8s.pvc_name(resource_name)),
         ]
-        for delete_fn, obj_name, kind in deleters:
-            try:
-                delete_fn(name=obj_name, namespace=self._k8s_namespace)
-                logger.info(f"Deleted {kind} {obj_name} in {self._k8s_namespace}")
-            except k8s_client.exceptions.ApiException as e:
-                if e.status == 404:
-                    continue
-                logger.warning(f"Error deleting {kind} {obj_name}: {e}")
+        errors: list[str] = []
+        for delete_fn, kind, obj_name in deleters:
+            err = self._delete_one(delete_fn, kind, obj_name)
+            if err:
+                errors.append(err)
+        return errors
 
     async def update_model_deployment(
         self, deployment: ModelDeployment, config: ModelDeploymentConfig, model_entity: Optional[ModelEntity] = None
@@ -1411,61 +1409,92 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
                 host_url=None,
             )
 
+    def _delete_one(self, delete_fn, kind: str, obj_name: str) -> Optional[str]:
+        """Delete a single namespaced object by name, tolerating "already gone".
+
+        Teardown deletes every resource type by name (no engine detection): this is
+        idempotent and self-heals partial-deletion states. A 404 (object absent) is
+        success. Any other failure is logged concisely (no stack trace) and
+        returned as a short error string so the caller can aggregate and surface it
+        (we must NOT mark a deployment DELETED if cluster resources may remain).
+        """
+        try:
+            delete_fn(name=obj_name, namespace=self._k8s_namespace)
+            logger.info(f"Deleted {kind} {self._k8s_namespace}/{obj_name}")
+            return None
+        except (k8s_client.exceptions.ApiException, k8s_dynamic_exceptions.NotFoundError) as e:
+            # NotFound (typed status 404 or dynamic NotFoundError) -> already gone.
+            if isinstance(e, k8s_dynamic_exceptions.NotFoundError) or getattr(e, "status", None) == 404:
+                logger.debug(f"{kind} {obj_name} not found, already deleted")
+                return None
+            return self._classify_delete_error(e, kind, obj_name)
+        except k8s_dynamic_exceptions.ForbiddenError as e:
+            return self._classify_delete_error(e, kind, obj_name)
+
+    @staticmethod
+    def _classify_delete_error(e: Exception, kind: str, obj_name: str) -> str:
+        """Concise, human-readable delete failure (no stack trace) for aggregation."""
+        status = getattr(e, "status", None)
+        is_forbidden = status == 403 or isinstance(e, k8s_dynamic_exceptions.ForbiddenError)
+        if is_forbidden:
+            # With the models ServiceAccount RBAC in place this should not happen;
+            # if it does, the SA is missing delete on this resource type.
+            msg = f"forbidden to delete {kind} {obj_name} (ServiceAccount lacks RBAC)"
+            logger.error(msg)
+            return msg
+        msg = f"error deleting {kind} {obj_name}: {status or type(e).__name__}"
+        logger.warning(msg)
+        return msg
+
     def _delete_resources_by_model_deployment_id(self, workspace: str, name: str) -> DeploymentStatusUpdate:
-        """Delete NIMService and NIMCache for the given model deployment (by workspace/name)."""
+        """Delete every resource a deployment could own, by name (engine-agnostic).
+
+        Delete has only workspace/name (no config/engine -- it is also called for
+        orphan reconciliation), so we attempt deletion of BOTH the operator path
+        (NIMService/NIMCache CRs) and the directly-emitted vLLM path (Deployment/
+        Service/Job/PVC). Each delete is independent and 404-tolerant; one
+        resource's failure never aborts the others. Real (non-404) failures are
+        aggregated and surfaced as ERROR so we never report DELETED while cluster
+        resources may remain.
+        """
         nimservice_name = get_deployment_resource_name(workspace, name)
         nimcache_name = get_nimcache_resource_name(workspace, name)
-        try:
-            nimservice_api = self._dynamic_client.resources.get(
-                api_version=NIMSERVICE_API_VERSION,
-                kind="NIMService",
-            )
+        errors: list[str] = []
 
+        # Operator path: NIMService / NIMCache CRs (via the dynamic client).
+        for api_version, kind, cr_name in (
+            (NIMSERVICE_API_VERSION, "NIMService", nimservice_name),
+            (NIMCACHE_API_VERSION, "NIMCache", nimcache_name),
+        ):
             try:
-                nimservice_api.delete(
-                    name=nimservice_name,
-                    namespace=self._k8s_namespace,
-                )
-                logger.info(f"Successfully deleted NIMService {self._k8s_namespace}/{nimservice_name}")
-            except k8s_dynamic_exceptions.NotFoundError:
-                logger.info(f"NIMService {nimservice_name} not found, may have been already deleted")
-
-            # Try to delete associated NIMCache if it exists
-            try:
-                nimcache_api = self._dynamic_client.resources.get(
-                    api_version=NIMCACHE_API_VERSION,
-                    kind="NIMCache",
-                )
-                nimcache_api.delete(
-                    name=nimcache_name,
-                    namespace=self._k8s_namespace,
-                )
-                logger.info(f"Successfully deleted NIMCache {self._k8s_namespace}/{nimcache_name}")
-            except k8s_dynamic_exceptions.NotFoundError:
-                logger.debug(f"No NIMCache found for {nimcache_name}, skipping cleanup")
+                cr_api = self._dynamic_client.resources.get(api_version=api_version, kind=kind)
             except Exception as e:
-                logger.warning(f"Error deleting NIMCache {nimcache_name}: {e}")
-
-            # Also tear down any directly-emitted vLLM objects (idempotent; delete
-            # has no config/engine, so we clean up both paths by name). Deleting the
-            # Deployment cascades its owned PVC/Job/Service, but the puller objects
-            # may exist before the Deployment (phases P0-P2), so delete explicitly.
-            self._delete_vllm_resources(nimservice_name)
-
-            return DeploymentStatusUpdate(
-                status="DELETED",
-                status_message="Deployment deletion initiated successfully",
-                host_url=None,
+                errors.append(f"error resolving {kind} API: {e}")
+                continue
+            err = self._delete_one(
+                lambda name, namespace, _api=cr_api: _api.delete(name=name, namespace=namespace),
+                kind,
+                cr_name,
             )
+            if err:
+                errors.append(err)
 
-        except Exception as e:
-            logger.exception(f"Failed to delete NIMService {nimservice_name}")
+        # Directly-emitted vLLM path: Deployment / Service / puller Job / PVC.
+        errors.extend(self._delete_vllm_resources(nimservice_name))
+
+        if errors:
+            summary = "; ".join(errors)
             return DeploymentStatusUpdate(
                 status="ERROR",
-                status_message=f"Failed to delete deployment {nimservice_name} due to a service backend error",
-                error_details={"error": str(e), "error_type": type(e).__name__},
+                status_message=f"Failed to fully delete deployment {workspace}/{name}: {summary}",
+                error_details={"errors": errors},
                 host_url=None,
             )
+        return DeploymentStatusUpdate(
+            status="DELETED",
+            status_message="Deployment deletion initiated successfully",
+            host_url=None,
+        )
 
     async def delete_model_deployment(self, workspace: str, name: str) -> DeploymentStatusUpdate:
         """Delete a NIM Operator model deployment by workspace and name (model deployment ID)."""
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
@@ -813,6 +813,9 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
                 namespace=self._k8s_namespace,
                 service_account_name=self._backend_config.service_account_name,
                 image_pull_secret=self._backend_config.huggingface_model_puller_image_pull_secret,
+                # Engine-specific uid/gid: vLLM uses 2000/0 (its image's user). A
+                # future NIM raw-object path must pass NIM's own uid/gid here, not
+                # these -- see the FUTURE note in vllm_k8s_compiler.py.
                 user_id=self._backend_config.default_vllm_user_id,
                 group_id=self._backend_config.default_vllm_group_id,
                 model_source=source_tag,
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
@@ -813,8 +813,8 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
                 namespace=self._k8s_namespace,
                 service_account_name=self._backend_config.service_account_name,
                 image_pull_secret=self._backend_config.huggingface_model_puller_image_pull_secret,
-                user_id=self._backend_config.default_user_id,
-                group_id=self._backend_config.default_group_id,
+                user_id=self._backend_config.default_vllm_user_id,
+                group_id=self._backend_config.default_vllm_group_id,
                 model_source=source_tag,
             )
 
@@ -986,8 +986,8 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
             gpu=view.gpu,
             namespace=self._k8s_namespace,
             service_account_name=self._backend_config.service_account_name,
-            user_id=self._backend_config.default_user_id,
-            group_id=self._backend_config.default_group_id,
+            user_id=self._backend_config.default_vllm_user_id,
+            group_id=self._backend_config.default_vllm_group_id,
             shared_memory_size_limit=self._backend_config.default_shared_memory_size_limit,
             startup_grace_seconds=startup_grace,
             init_containers=init_containers,
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
@@ -22,7 +22,6 @@ from urllib.parse import urljoin
 from kubernetes import client as k8s_client
 from kubernetes import config as k8s_config
 from kubernetes.dynamic import DynamicClient
-from kubernetes.dynamic import exceptions as k8s_dynamic_exceptions
 from nemo_platform.types.inference.model_deployment import ModelDeployment
 from nemo_platform.types.inference.model_deployment_config import ModelDeploymentConfig
 from nemo_platform.types.models.model_entity import ModelEntity
@@ -44,8 +43,6 @@ from nmp.core.models.controllers.backends.k8s_nim_operator.config import K8sNimO
 from nmp.core.models.controllers.backends.k8s_nim_operator.reconcilers.base import ResolvedDeployment
 from nmp.core.models.controllers.backends.k8s_nim_operator.reconcilers.k8s import K8sReconciler
 from nmp.core.models.controllers.backends.k8s_nim_operator.reconcilers.nim_operator import (
-    NIMCACHE_API_VERSION,
-    NIMSERVICE_API_VERSION,
     NimOperatorReconciler,
 )
 
@@ -91,8 +88,6 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
         self._k8s_namespace = self._get_current_namespace()
         logger.info(f"Models controller will deploy models to namespace: {self._k8s_namespace}")
 
-        self._validate_nim_operator_crds()
-
         self._nim_reconciler = NimOperatorReconciler(
             k8s_client_=self._k8s_client,
             dynamic_client=self._dynamic_client,
@@ -130,48 +125,6 @@ class K8sNimOperatorServiceBackend(ServiceBackend):
 
         logger.warning("Could not determine k8s namespace, using 'default'")
         return "default"
-
-    def _validate_nim_operator_crds(self) -> None:
-        """
-        Validate that NIM Operator APIs are available via API discovery.
-
-        Raises:
-            RuntimeError: If required APIs are not found. This will prevent the backend
-                        from initializing and cause the controller to fail fast.
-        """
-        # Validate NIMService API is available
-        try:
-            self._dynamic_client.resources.get(
-                api_version=NIMSERVICE_API_VERSION,
-                kind="NIMService",
-            )
-            logger.info(f"Validated NIMService API is available: {NIMSERVICE_API_VERSION} NIMService")
-        except k8s_dynamic_exceptions.ResourceNotFoundError as e:
-            logger.error(f"NIMService CRD not found: {e}")
-            raise RuntimeError(
-                f"NIMService API ({NIMSERVICE_API_VERSION}) not found. "
-                f"The k8s-nim-operator must be installed before starting this backend."
-            ) from e
-        except Exception as e:
-            logger.exception("Unexpected error validating NIMService API")
-            raise RuntimeError(f"Failed to validate NIMService API ({NIMSERVICE_API_VERSION}): {e}") from e
-
-        # Validate NIMCache API is available
-        try:
-            self._dynamic_client.resources.get(
-                api_version=NIMCACHE_API_VERSION,
-                kind="NIMCache",
-            )
-            logger.info(f"Validated NIMCache API is available: {NIMCACHE_API_VERSION} NIMCache")
-        except k8s_dynamic_exceptions.ResourceNotFoundError as e:
-            logger.error(f"NIMCache CRD not found: {e}")
-            raise RuntimeError(
-                f"NIMCache API ({NIMCACHE_API_VERSION}) not found. "
-                f"The k8s-nim-operator must be installed before starting this backend."
-            ) from e
-        except Exception as e:
-            logger.exception("Unexpected error validating NIMCache API")
-            raise RuntimeError(f"Failed to validate NIMCache API ({NIMCACHE_API_VERSION}): {e}") from e
 
     # ------------------------------------------------------------------
     # Name + weight-source resolution (API-object work owned by the backend)
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
@@ -6,6 +6,7 @@
 import os
 from logging import getLogger
 from typing import Any, Dict, Optional
+from urllib.parse import urljoin
 
 from kubernetes import client as k8s_client
 from kubernetes import config as k8s_config
@@ -41,7 +42,6 @@ from nmp.core.models.controllers.backends.engine import (
 from nmp.core.models.controllers.backends.k8s_nim_operator import vllm_k8s_compiler as vk8s
 from nmp.core.models.controllers.backends.k8s_nim_operator.config import K8sNimOperatorConfig
 from nmp.core.models.controllers.backends.k8s_nim_operator.nimservice_compiler import (
-    _get_files_hf_url,
     compile_nimcache,
     compile_nimservice,
 )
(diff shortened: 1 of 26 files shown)


services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
```diff
@@ -0,0 +1,448 @@
...
+* ``NimOperatorReconciler`` -- emits ``NIMService`` / ``NIMCache`` CRs and lets the
+  in-cluster k8s-nim-operator do the actual reconciliation; status is propagated
+  upward from the operator-created resources.
+* ``K8sReconciler`` -- emits native Kubernetes objects directly (PVC / Job /
+  Deployment / Service) and drives a staged rollout itself, advancing the
+  deployment one phase at a time as it is polled via ``get_status``.
+
+The shared, engine-agnostic Kubernetes status helpers (pod log fetch, crash-loop
+detection, pod-status drill-down, pending-timeout/crash-loop error builders) live
+on :class:`BaseReconciler` because both reconcilers project status from the same
```
mckornfield: personally, and this is just to get this out of the way, I dislike base methods. I like interfaces but prefer composition over it, so wonder if the base things should just be moved off onto a `status_reconciler` or something that captures what these lower level things do, and just compose them in here. But 🤷 maybe I'm just being a doof

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
```diff
@@ -0,0 +1,448 @@
...
+        """Delete a single namespaced object by name, tolerating "already gone".
+
+        Teardown deletes every resource type by name (no engine detection): this is
+        idempotent and self-heals partial-deletion states. A 404 (object absent) is
+        success. Any other failure is logged concisely (no stack trace) and
+        returned as a short error string so the caller can aggregate and surface it
+        (we must NOT mark a deployment DELETED if cluster resources may remain).
+        """
+        try:
+            delete_fn(name=obj_name, namespace=self._k8s_namespace)
```
mckornfield: like personally I'd rather have this be its own class that's composed in here, and we supply a variant so we're calling `variant.delete`

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
```diff
@@ -0,0 +1,448 @@
...
+                            "pod": pod_name,
+                            "restart_count": restart_count,
+                            "max_restarts": max_restarts,
+                            "waiting_reason": waiting_reason,
+                        },
+                    )
+                    return self._build_crash_loop_error(resource_name, pod_name, restart_count)
+                else:
+                    logger.debug(
+                        "Pod has restarts above threshold but is not in waiting state",
```
mckornfield: nit- maybe also print or extra add the threshold?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
```diff
@@ -0,0 +1,448 @@
...
+                status_msg = f"Waiting for k8s deployment - pod status is {pod_status}"
+                status_msg = self._with_restart_info(status_msg, restart_count)
+                return DeploymentStatusUpdate(
+                    status="PENDING",
+                    status_message=status_msg,
+                    host_url=None,
+                )
+
+            recent_event = max(
+                events.items, key=lambda e: e.last_timestamp or e.event_time or e.metadata.creation_timestamp
```
mckornfield: lol gotta check em all

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
```diff
@@ -0,0 +1,448 @@
...
+
+            reason = recent_event.reason
+            message = recent_event.message
+
+            for search_string, return_message in POD_EVENT_TO_MESSAGE_MAP.items():
+                if search_string in message.lower():
+                    status_msg = self._with_restart_info(return_message, restart_count)
+                    return DeploymentStatusUpdate(status="PENDING", status_message=status_msg, host_url=None)
+
+            if len(message) > 200:
```
mckornfield: is this just for not spamming the logs? or is there a real limit

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+from nemo_platform.types.inference.model_deployment import ModelDeployment
+from nemo_platform.types.models.model_entity import ModelEntity
+from nmp.common.config import get_platform_config
+from nmp.core.models.app import get_deployment_resource_name
+from nmp.core.models.app.constants import MODEL_MANAGED_BY_LABEL, MODEL_MANAGED_BY_MODELS_CONTROLLER
+from nmp.core.models.controllers.backends import vllm_compiler
+from nmp.core.models.controllers.backends.backends import DeploymentStatusUpdate
+from nmp.core.models.controllers.backends.common import DeploymentConfigView
+from nmp.core.models.controllers.backends.engine import ENGINE_VLLM, resolve_health_path
+from nmp.core.models.controllers.backends.k8s_nim_operator import vllm_k8s_compiler as vk8s
```
mckornfield: nit. I was like `wtf is vk8s` so I think just `vllm_compiler` or `vllm_k8s_compiler` (without renaming) is fine

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+            return self._project_deployment_readiness(resource_name)
+
+        # No Deployment yet: we're still in the pull phase. Consult the puller Job.
+        job_name = vk8s.pull_job_name(resource_name)
+        try:
+            job = self._batch_v1.read_namespaced_job(name=job_name, namespace=self._k8s_namespace)
+        except k8s_client.exceptions.ApiException as e:
+            if e.status != 404:
+                raise
+            # Job absent. This is either (a) the transient P3 window after we
```
mckornfield: nit- put (a) and (b) on their own lines for the comment

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+        """Resolve the puller's model repo (``namespace/name``) and a source tag.
+
+        The source tag (``namespace/name@revision``) is stamped on the PVC + Job so
+        the update path can detect a weight-source change and decide to re-pull.
+        """
+        namespace = resolved.model_namespace
+        name = resolved.model_name
+        revision = resolved.model_revision
+        if not namespace or not name:
+            raise ValueError(f"Cannot resolve model source for vLLM deployment: namespace={namespace}, name={name}")
```
mckornfield: nit, put the values in single quotes, so it's clear if one is blank

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+                return False
+            raise
+
+    def _create_or_skip(self, create_fn, body, kind: str) -> None:
+        """Create a namespaced object, tolerating 409 Conflict (already exists)."""
+        try:
+            create_fn(namespace=self._k8s_namespace, body=body)
+            logger.info(f"Created {kind} {body.metadata.name} in {self._k8s_namespace}")
+        except k8s_client.exceptions.ApiException as e:
+            if e.status == 409:
```
mckornfield, line 358: wish there was an api based way to just be like "hey if it exists just return a 2xx"

mckornfield: and do we ever want to "patch/recreate" what's there?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+            workspace=deployment.workspace,
+            name=deployment.name,
+            engine=engine,
+            image=f"{image_name}:{image_tag}",
+            args=args,
+            health_path=health_path,
+            env=env,
+            gpu=view.gpu,
+            namespace=self._k8s_namespace,
+            service_account_name=self._backend_config.service_account_name,
```
mckornfield, line 427: wonder if it would be nice if we could pass a backend_config into here and pull these attrs off

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -0,0 +1,600 @@
...
+                k8s_client.V1VolumeMount(name="model-store", mount_path=vk8s.MODEL_STORE_PATH, read_only=True),
+                k8s_client.V1VolumeMount(name="scratch", mount_path=vk8s.SCRATCH_PATH),
+            ],
+        )
+        # imagePullSecrets are pod-level; the compiler sets them from the puller
+        # secret, but the sidecar image comes from the platform registry. Attach
+        # via the sidecar's own spec is not possible (pod-level only), so rely on
+        # the pod's service account / pull secret. (Platform pull secrets are
+        # applied at the chart level for the models SA.)
+        _ = image_pull_secrets  # documented: pod-level pull secrets handled by SA
```
mckornfield: why are we doing this though specifically? just to show it's not used or?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
```diff
@@ -0,0 +1,466 @@
...
+    # Reconciler interface
+    # ------------------------------------------------------------------
+
+    async def create(self, resolved: ResolvedDeployment) -> DeploymentStatusUpdate:
+        deployment = resolved.deployment
+        config = resolved.config
+        model_entity = resolved.model_entity
+
+        logger.info(
+            f"Creating NIMService: {deployment.workspace}/{deployment.name} (version: {deployment.entity_version})"
```
mckornfield: should we use structured logging for these and below or nah?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
```diff
@@ -0,0 +1,466 @@
...
+                logger.error(f"Failed to create/update NIMCache for Files service model: {e}")
+                return DeploymentStatusUpdate(
+                    status="ERROR",
+                    status_message=f"Failed to create/update NIMCache for Files service model: {str(e)}",
+                    error_details={"error": str(e), "error_type": type(e).__name__},
+                    host_url=None,
+                )
+        else:
+            logger.debug(
+                f"No Files service model detected for deployment update {deployment.workspace}/{deployment.name}"
```
mckornfield: maybe a nit, for these files service conditionals that end with an "else", could we just return early or something? just to help with the indents.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
```diff
@@ -0,0 +1,466 @@
...
+                f"The resource may have been manually deleted or removed during namespace cleanup."
+            )
+            return DeploymentStatusUpdate(
+                status="LOST",
+                status_message="NIMService not found in cluster. Resource may have been deleted externally.",
+                host_url=None,
+            )
+
+        nim_status = nimservice.get("status", {})
+        state = nim_status.get("state", "").lower()
```
mckornfield: could nim_status.get("val") be None?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
```diff
@@ -523,303 +176,144 @@ def _resolve_model_source(
...
+            model_namespace=model_namespace,
+            model_name=model_name,
+            model_revision=model_revision,
+            files_hf_url=self._remote_files_hf_url(),
+            huggingface_model_puller=self._huggingface_model_puller,
+        )
 
-            nimservice_dict = nimservice.model_dump(exclude_none=True, by_alias=True)
+    def _select_reconciler(self, engine: str):
+        """Select the reconciler for an engine, rejecting the unsupported one."""
```
mckornfield: is `None` the rejection?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+def compile_puller_job(
+    *,
+    resource_name: str,
+    workspace: str,
+    name: str,
+    engine: str,
+    image: str,
+    args: list[str],
```
mckornfield: maybe `container_args` ?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
```diff
@@ -0,0 +1,428 @@
...
+    selector_labels = {"app": resource_name}
+    pod_labels = {
+        **selector_labels,
+        **common_labels(workspace, name, engine),
+    }
+    if extra_labels:
+        pod_labels.update(extra_labels)
+    # health path is recorded as an annotation, not a label: the value (e.g.
+    # "/health") contains '/', which is invalid in a k8s label value.
+    health_annotations = {HEALTH_PATH_ANNOTATION: health_path}
```
mckornfield: why though?

services/core/models/src/nmp/core/models/controllers/deployment_reconciler.py
```diff
@@ -198,8 +198,12 @@ async def reconcile_deployments(self, deployment_contexts: list[ModelContext]) -
                             existing_provider=ctx.model_provider,
                         )
                     case "PENDING" | "READY" | "UNKNOWN":
-                        # Check status and handle drift/backend issues
-                        status_update = await backend.get_model_deployment_status(deployment)
+                        # Check status and handle drift/backend issues. Pass the
+                        # config + entity so backends that advance creation in the
+                        # status path (k8s vLLM) can compile the serving objects.
+                        status_update = await backend.get_model_deployment_status(
+                            deployment, ctx.model_deployment_config, ctx.model_entity
```
mckornfield: nit- should we just pass `ctx` through?

mckornfield (commented):
well that was fun

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
benmccown: feedback implementation response, built with claude: Not logs — `status_message` is persisted as the deployment's status and surfaced in the UI/CLI status history, so we cap the recent-event message to keep history entries readable. It's not a protocol limit. Pulled the magic 200/197 into a named `MAX_EVENT_MESSAGE_CHARS` constant on `StatusProjector` to make that explicit.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: feedback implementation response, built with claude: Agreed — there's no native upsert in the k8s API, hence the 409-tolerant `_create_or_skip`. We deliberately don't blind-patch existing objects from the status path; the `update()` path owns changes: an unchanged model source patches the Deployment in place, and a changed source re-pulls (deletes + recreates PVC/Job/Deployment). So 'recreate what's there' happens only on a real source change, intentionally.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: feedback implementation response, built with claude: Went back and forth on this and decided to keep the params explicit. The compiler is deliberately decoupled from `K8sNimOperatorConfig` so it stays a pure, engine-agnostic emission layer (see the FUTURE/NIM note at the top of the file). The footgun is uid/gid: the config holds two pairs (`default_vllm_*` vs the NIM defaults), and the compiler must not know which to read or it stops being vLLM-vs-NIM-agnostic. SA/pull-secret aren't engine-specific, but splitting 'pass config for some, explicit for others' is muddier than keeping all params explicit + keyword-only (which already prevents misordering). Keeping as-is.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: feedback implementation response, built with claude: Right — it was documenting that the sidecar's pull secrets are intentionally not set per-container (imagePullSecrets are pod-level and handled by the models ServiceAccount / chart). It read as confusing, so I removed the `_ = image_pull_secrets` dance entirely and folded the explanation into a comment; the value is no longer computed.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
benmccown: feedback implementation response, built with claude: Done — converted the nim_operator logs to structured `extra={}`. Heads-up for anyone copying the pattern: don't use `extra={"name": ...}` — `name` is a reserved `LogRecord` attribute and raises under plain stdlib logging; used `resource_name`/`deployment_name` instead.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
benmccown: feedback implementation response, built with claude: Yes — `_select_reconciler` returns `None` for the `generic` engine, and callers map that to `_unsupported_engine(...)`. vLLM -> `K8sReconciler`, everything else -> `NimOperatorReconciler` (NIM is the default). Added a docstring spelling out that `None` contract.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
benmccown: I went ahead (with Claude's help) and refactored methods out of the base ABC and into component classes in `status_projector.py` and `resource_deleter.py`. LMK what you think.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
benmccown: I went ahead (with Claude's help) and refactored methods out of the base ABC and into component classes in `status_projector.py` and `resource_deleter.py`. LMK what you think.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/base.py
benmccown: done!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: agreed, `vk8s` is a terrible module name, didn't catch this in my own initial review. Renamed to `vllm_k8s_compiler`.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: done!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: done!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: It wasn't just claude waffling on this one, I went back and forth. I started to refactor it, but realized then the compiler would need to be "engine aware" for vLLM vs. NIM (vs. "generic"). So I left it as is. Wdyt?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
benmccown: Fixed this up, thanks for the feedback.

---

1. **Extracted** the whole branch into the `_ensure_nimcache` helper, so `create`/`update` just call `nimcache_name, error = await self._ensure_nimcache(...)` (`nim_operator.py:100`, `:176`) instead of nesting the logic inline.

2. **Flattened with guard clauses / early returns** inside the helper — no `else` ladders:
   - Not a Files-service model → `return None, None` (`:325`)
   - Missing namespace/name → `return None, DeploymentStatusUpdate(ERROR...)` (`:339`)
   - Then the happy path (`compile_nimcache` + create) runs unindented at the function's top level (`:350`+), with only the `try/except` for the create call.

So the previous `if files_service: ... else: ...` indentation is gone — each precondition bails early, and the main logic is no longer buried under conditionals. Exactly what he asked for.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
benmccown: Yes, thank you for catching.

---

 It absolutely could — the operator may not have populated `status` or `status.state` yet (absent **or** explicitly null). We hardened it on two levels (`nim_operator.py:433-434`):

```python
# ``status`` / ``status.state`` may be absent or explicitly null while the
# operator is still populating them -- coerce to "" so .lower() is safe.
nim_status = nimservice.get("status") or {}
state = (nim_status.get("state") or "").lower()
```

- `nimservice.get("status") or {}` — guards against `status` being absent/null, so `nim_status` is always a dict.
- `(nim_status.get("state") or "")` — guards against `state` being absent/null, so `.lower()` never gets a `None`.

A `None` state now coerces to `""`, which falls through the `match` to the default/unknown case rather than raising. There's also an explanatory comment documenting exactly the scenario Matt flagged (operator still populating). So his concern is resolved and the rationale is captured inline.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
benmccown: Updated to `container_args`!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
benmccown: I removed this annotation.

I think it was an artifact of the docker service backend's vLLM implementation using a to persist the health check, and the platform itself being responsible for monitoring the container's health instead of using built-in docker health checks. So claude blindly copied the annotation when we were porting the docker vLLM implementation over to k8s.

As far as why it's like that for docker, I'll include claude's analysis below for context. But to be clear, we dropped the nonsensical annotation for k8s.

---

This gives me enough to reason concretely instead of guessing. Caveat up front: **the rationale isn't written down anywhere I've found** — so this is inference from the code's constraints, not a documented decision. But it's a fairly defensible "weird for a good reason" call.

## The core problem: a Docker `HEALTHCHECK` only reports *inside* the container

Docker's native healthcheck runs the probe command **from inside the model container's own filesystem/network namespace**, and surfaces the result on that container's `State.Health`. For it to work, two things have to be true:

1. **The probe tool must exist in the model image.** The healthcheck `test` runs in the container — so you need `curl`/`wget`/a shell, or a binary, present in the vLLM/NIM image. You don't control those images. NIM and `vllm/vllm-openai` are upstream; betting that every engine image ships a usable probe tool (and keeping that working across image versions) is fragile. An external httpx probe needs nothing in the image.

2. **You still have to read it back out.** Even with a native healthcheck, the platform would have to call `docker inspect` and parse `State.Health.Status` every poll — so you don't actually escape the "platform reads container state" round-trip. You'd just trade an HTTP GET for an inspect call, *and* take on dependency #1.

## The networking modes make external probing the natural fit anyway

The `host_url` logic (`creation_reconciler.py:279`) shows the platform reaches model containers **three different ways** depending on `models_docker_networking_mode`:

- `local` → `http://localhost:{host_port}` (port-forwarded to the host)
- `dond` → `http://{container_name}:8000` (shared Docker network, by name)
- `dind` → `http://{host_service_name}:{host_port}` (port-forward to a DinD service)

The platform already has to compute a reachable URL to **route inference traffic** to the container. Once you've built that URL, probing `host_url + health_path` is the same machinery the gateway uses to actually talk to the model — so readiness is verified against *the exact path inference will traverse*. A native Docker healthcheck verifies "the server answers itself from inside its own namespace," which is necessary but not sufficient: it wouldn't catch a port-forwarding/`dind`/network misconfig that breaks the path the platform actually uses. External probing tests the real thing.

## Plus engine-specific path + override, uniform with k8s

The path is engine-specific (`/v1/health/ready` for NIM, `/health` for vLLM) and user-overridable (`executor_config.health_check_path`), resolved by the same `resolve_health_path()` the k8s path uses. Doing it platform-side keeps one readiness model — "the controller resolves a path and asserts 200 on it" — across both docker and k8s, rather than docker uniquely delegating to Docker's healthcheck semantics (start-period, retries, interval) that you'd then have to translate back into the platform's READY/PENDING state machine.

## So: weird, but defensible

Net: it's not that Docker can't health-check — it's that a native healthcheck (a) requires a probe tool inside images you don't own, (b) still forces a read-back via inspect, (c) only proves in-namespace reachability, not the platform's actual route, and (d) wouldn't share the readiness model with k8s. External probing sidesteps all four.

The one thing I'd flag as a genuine downside of the chosen approach: if the platform and the model container *aren't* mutually reachable (a network misconfig), external probing can report "not ready" for something that's actually healthy — whereas a native healthcheck wouldn't care about platform→container reachability. But since that same reachability is required for inference to work at all, failing readiness on it is arguably correct.

services/core/models/src/nmp/core/models/controllers/deployment_reconciler.py
benmccown: went ahead and made this change!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/resource_deleter.py
```diff
@@ -0,0 +1,61 @@
...
+            return None
+        except (k8s_client.exceptions.ApiException, k8s_dynamic_exceptions.NotFoundError) as e:
+            # NotFound (typed status 404 or dynamic NotFoundError) -> already gone.
+            if isinstance(e, k8s_dynamic_exceptions.NotFoundError) or getattr(e, "status", None) == 404:
+                logger.debug(f"{kind} {obj_name} not found, already deleted")
+                return None
+            return self._classify_delete_error(e, kind, obj_name)
+        except k8s_dynamic_exceptions.ForbiddenError as e:
+            return self._classify_delete_error(e, kind, obj_name)
+
```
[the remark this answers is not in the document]
benmccown: feedback implementation response, built with claude: Good catch — fixed in bde648d2dc. The `delete_one()` docstring promised any non-404 failure is returned as a short error string for the caller to aggregate, but only `ApiException`/`NotFoundError`/`ForbiddenError` were caught, so any other exception (transport/connection error, dynamic API error, …) would escape and abort the per-resource delete loop in `_delete_vllm_resources` partway — leaving later resources unattempted and risking a false DELETED.

Replaced the `ForbiddenError`-specific handler with a catch-all `except Exception` that routes everything through `_classify_delete_error` (which already distinguishes 403 from other failures). The 404→success path is unchanged. Added unit tests for `delete_one` covering typed 404 → None, dynamic NotFoundError → None, 403 → classified string, and an unexpected `ConnectionError` → classified (not raised) to lock in the aggregation contract.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
mckornfield: yeo that's fine. I just want my monadic and dyadic functions lol

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/status_projector.py
```diff
@@ -0,0 +1,320 @@
...
+        resource_name: str,
+        pod_name: str,
+        restart_count: int,
+    ) -> DeploymentStatusUpdate:
+        """Build ERROR status update for a crash loop."""
+        error_stack = self.fetch_pod_logs(pod_name)
+        status_msg = (
+            f"Deployment entered crash loop after {restart_count} container restarts "
+            f"(max: {self._backend_config.max_restart_count}).\n\n"
+            f"Inspect the NIM pod logs with:\n"
```
mckornfield: per this comment, maybe "inspect the Model pod logs"?

mckornfield (approved):
fancy pr is fancy

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/status_projector.py
benmccown: fixed!

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/status_projector.py
```diff
@@ -0,0 +1,320 @@
...
+            f"  kubectl logs -n {self._k8s_namespace} {pod_name}"
```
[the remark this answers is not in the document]
benmccown: fixed!

PR #308, opened 2026-06-12
fix(studio): Consolidate Studio console

steramae-nvidia:
We wrap the console to ensure that anything logged to console is also passed to the optional telemetry service. There are still some exceptions but nearly all statements should now be wrapped.

## Summary by CodeRabbit

* **Chores**
  * Standardized logging across the Studio UI by replacing console.* calls with the site logger to improve structured error/warn reporting without changing user-facing behavior.
  * Enforced no-console linting for the Studio package with scoped exceptions for CLI and mock files.
  * Enhanced the site logger to emit richer diagnostics and normalize error values for more consistent troubleshooting.

* **Tests**
  * Updated unit tests to align with the new logging signatures and behaviors.


aray12 (approved):

web/packages/studio/src/api/datasets/useDatasetFileContent.ts
```diff
@@ -7,6 +7,7 @@ import type { EntityIdentifier } from '@studio/api/common/types';
 import { getDatasetFileContentQueryKey } from '@studio/api/datasets/invalidateDatasetCaches';
 import { PLATFORM_BASE_URL } from '@studio/constants/environment';
 import { isBinaryExtension } from '@studio/util/binaryFile';
+import { websiteLogger } from '@studio/util/logger';
```
marcusds: nit: I don't love this func name, can we do something simpler?

marcusds (approved):

web/packages/studio/src/api/datasets/useDatasetFileContent.ts
steramae-nvidia: Sure, I can just use `logger`

PR #315, opened 2026-06-12
feat(deployments): reconcile controller and prerequisite DAG (AIRCORE-758)

tylersbray:
## Summary

Adds the deployments plugin reconcile loop on top of #280 (AIRCORE-755): a background `DeploymentsController` that drives Deployment and Volume state machines against registered executor backends, with prerequisite DAG gating, volume-mount readiness, drift recovery, and deployment orphan cleanup.

- **Ticket:** [AIRCORE-758](https://linear.app/nvidia/issue/AIRCORE-758/deployments-plugin-deployment-reconciler-prerequisite-dag)
- **Stacks on:** #280 (`755-deployments-plugin-scaffold-the-plugin-entities-api-backend/tbray`)
- **83 unit tests** in `plugins/nemo-deployments/tests/unit` (integration E2E skipped pending AIRCORE-756)

## Breaking changes

- **`Prerequisite.condition` field** — new enum on `DeploymentConfig.prerequisites` entries: `ready` (prerequisite `Deployment.status == READY`) or `succeeded` (default: `SUCCEEDED` with `exit_code == 0`). Existing configs without the field get `succeeded` by default.
- **`nemo.controllers` entry point** — registering `DeploymentsController` starts a background reconcile loop when the deployments plugin service runs. Empty executor registry causes new deployments to project `FAILED` (no silent no-op).
- **`desired_state=STOPPED` semantics** — reconciler deletes the deployment entity (no terminal `STOPPED` status persisted). Matches RFC intent; callers should not expect a long-lived stopped deployment row.
- **Per-config drift backoff overrides** — optional `DeploymentConfig.driftRecovery.maxAttempts`, `baseDelaySeconds`, and `maxDelaySeconds` override controller defaults when set (including `maxAttempts=0` to disable recovery).

## Noteworthy net-new behavior

- **`DeploymentsController`** — paginated cross-workspace listing (no 100-item cap), volumes reconciled before deployments, split health flags for deployment vs volume list failures, orphan cleanup gated when deployment list fails.
- **Prerequisite DAG** — deployment create gated on prerequisite satisfaction; terminal prerequisites (e.g. puller `SUCCEEDED`) fetched from entity store by `deployment_config_name` even when absent from the non-terminal list.
- **Volume mount gating** — deployment create blocked until referenced volumes reach `BOUND`; failed mounts can project `FAILED`.
- **Drift recovery** — `LOST` Always-policy deployments recreated with exponential backoff; per-config policy overrides; failed recreate stays `LOST` (preserves backoff path).
- **Orphan cleanup** — periodic substrate sweep via `list_managed_deployment_names`; skipped when entity list is unhealthy.
- **Volume reconciler (partial)** — `PENDING` → create volume; `BOUND` → read status. Delete/`RELEASED` deferred.

## Deferred to follow-on PRs

Tracked in plugin README and piggybacked onto the [AIRCORE-756 plan](https://linear.app/nvidia/issue/AIRCORE-756):

| Item | Target |
|------|--------|
| `DockerDeploymentBackend` / K8s backend | AIRCORE-756 / 757 |
| Unskip `test_reconcile_docker.py` (puller→server E2E) | After 756 |
| Volume delete → `RELEASED` + volume orphan cleanup | Needs `list_managed_volume_names` on backend ABC |
| Per-volume executor routing | Needs `Volume.executor` field (755 schema) |
| `create_deployment` idempotency validation | Real backend in 756 |
| Prerequisite naming convention enforcement | Optional hardening |

## Test plan

- [x] `uv run pytest plugins/nemo-deployments/tests/unit -q` (83 passed)
- [x] Pre-commit (ruff, ty, uv lock)
- [ ] Manual: start platform with deployments plugin + empty registry → confirm controller starts and unhealthy deployments list is handled gracefully
- [ ] E2E after 756 lands: puller (OnFailure) + server (Always + prerequisite) on shared volume

## Summary by CodeRabbit

* **New Features**
  * Background reconciliation controller for deployments and volumes (discoverable/registered); per-config drift-recovery overrides.
  * Prerequisite checks with "ready" and "succeeded" modes; volume-mount readiness gating and periodic orphan backend cleanup.

* **Configuration**
  * New controller configuration section (interval, drift-recovery backoff/limits, orphan-cleanup cadence); live reconciliation requires at least one executor backend.

* **Documentation**
  * README expanded with controller behaviors, prerequisites, deferred items, and next steps.

* **Tests**
  * New unit and integration tests covering controller, reconciliation, drift recovery, listing, prerequisites, and volumes.

### DIFF
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -12,6 +12,10 @@ gates deployment create on mounted volumes reaching `BOUND`, and writes status v
 entity client (including endpoints and status history). Orphan substrate cleanup runs on
 a configurable interval and is skipped when the deployment list is unhealthy.
 
+The controller exposes `is_healthy`, which is `False` when either the deployment-list or
+volume-list query fails. Internally these are tracked separately so operators can tell
+which list query failed without losing that signal behind a single boolean.
+
 Per-config drift backoff overrides live on `DeploymentConfig.driftRecovery`; unset fields
 fall back to `DeploymentsConfig.controller`.
 
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -1,6 +1,6 @@
 # NeMo Deployments Plugin
 
-Substrate-agnostic deployment lifecycle for the NeMo Platform: entity schemas,
+Backend-agnostic deployment lifecycle for the NeMo Platform: entity schemas,
 CRUD APIs, a `DeploymentBackend` ABC, an executor registry, and a background
 reconcile controller (`DeploymentsController`).
 
@@ -9,8 +9,8 @@ reconcile controller (`DeploymentsController`).
 Register `DeploymentsController` via the `nemo.controllers` entry point. The controller
 paginates non-terminal deployment/volume lists, reconciles volumes before deployments,
 gates deployment create on mounted volumes reaching `BOUND`, and writes status via the
-entity client (including endpoints and status history). Orphan substrate cleanup runs on
-a configurable interval and is skipped when the deployment list is unhealthy.
+entity client (including endpoints and status history). Orphan backend resource cleanup
+runs on a configurable interval and is skipped when the deployment list is unhealthy.
 
 The controller exposes `is_healthy`, which is `False` when either the deployment-list or
 volume-list query fails. Internally these are tracked separately so operators can tell
@@ -19,33 +19,6 @@ which list query failed without losing that signal behind a single boolean.
 Per-config drift backoff overrides live on `DeploymentConfig.driftRecovery`; unset fields
 fall back to `DeploymentsConfig.controller`.
 
-## Deferred (follow-on tickets)
-
-| Item | Why deferred |
-|------|----------------|
-| Volume delete → `RELEASED` | No `DELETING` state or `list_managed_volume_names` on the backend ABC yet |
-| Volume orphan cleanup | Requires backend support to list substrate volumes without entities |
-| Docker/K8s E2E | AIRCORE-756/757 — `BACKEND_CLASSES` empty until backends register |
-| Per-volume executor routing | No `Volume.executor` field in 755; volumes use `default_executor` |
-
-## API base path
-
-`/apis/deployments/v2/workspaces/{workspace}/...`
-
-Cross-workspace bulk queries use the entity-store sentinel workspace ``-``:
-
-``GET /apis/deployments/v2/workspaces/-/deployments?status_in=pending,starting``
-
-## Tests
-
-```bash
-uv sync
-uv run pytest plugins/nemo-deployments/tests/unit -v
-```
-
-## Next steps
-
-- **[AIRCORE-756](https://linear.app/nvidia/issue/AIRCORE-756):** Docker `DeploymentBackend`
-- **[AIRCORE-757](https://linear.app/nvidia/issue/AIRCORE-757):** Kubernetes `DeploymentBackend`
-- **[AIRCORE-759](https://linear.app/nvidia/issue/AIRCORE-759):** Models/agents adoption
-- **755 scaffold:** entity CRUD and executor registry ([PR #280](https://github.com/NVIDIA-NeMo/nemo-platform/pull/280))
+Prerequisites are currently declared on `DeploymentConfig` and resolve to a single
+deployment per config name in a workspace. Multiple deployments sharing one config is
+unsupported until prerequisites move to the `Deployment` entity (follow-up work).
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -1,7 +1,36 @@
 # NeMo Deployments Plugin
 
 Substrate-agnostic deployment lifecycle for the NeMo Platform: entity schemas,
-CRUD APIs, a `DeploymentBackend` ABC, and an executor registry.
+CRUD APIs, a `DeploymentBackend` ABC, an executor registry, and a background
+reconcile controller (`DeploymentsController`).
+
+## Controller
+
+Register `DeploymentsController` via the `nemo.controllers` entry point. The controller
+paginates non-terminal deployment/volume lists, reconciles volumes before deployments,
+gates deployment create on mounted volumes reaching `BOUND`, and writes status via the
+entity client (including endpoints and status history). Orphan substrate cleanup runs on
+a configurable interval and is skipped when the deployment list is unhealthy.
+
+Per-config drift backoff overrides live on `DeploymentConfig.driftRecovery`; unset fields
+fall back to `DeploymentsConfig.controller`.
+
+## Deferred (follow-on tickets)
+
+| Item | Why deferred |
+|------|----------------|
+| Volume delete → `RELEASED` | No `DELETING` state or `list_managed_volume_names` on the backend ABC yet |
+| Volume orphan cleanup | Requires backend support to list substrate volumes without entities |
+| Docker/K8s E2E | AIRCORE-756/757 — `BACKEND_CLASSES` empty until backends register |
+| Per-volume executor routing | No `Volume.executor` field in 755; volumes use `default_executor` |
+
+## API base path
+
+`/apis/deployments/v2/workspaces/{workspace}/...`
+
+Cross-workspace bulk queries use the entity-store sentinel workspace ``-``:
+
+``GET /apis/deployments/v2/workspaces/-/deployments?status_in=pending,starting``
 
 ## Tests
 
@@ -9,3 +38,10 @@ CRUD APIs, a `DeploymentBackend` ABC, and an executor registry.
 uv sync
 uv run pytest plugins/nemo-deployments/tests/unit -v
 ```
+
+## Next steps
+
+- **[AIRCORE-756](https://linear.app/nvidia/issue/AIRCORE-756):** Docker `DeploymentBackend`
+- **[AIRCORE-757](https://linear.app/nvidia/issue/AIRCORE-757):** Kubernetes `DeploymentBackend`
+- **[AIRCORE-759](https://linear.app/nvidia/issue/AIRCORE-759):** Models/agents adoption
+- **755 scaffold:** entity CRUD and executor registry ([PR #280](https://github.com/NVIDIA-NeMo/nemo-platform/pull/280))
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
@@ -22,22 +22,22 @@ class ControllerConfig(BaseModel):
 
     interval_seconds: int = Field(default=5, gt=0, description="Reconciliation loop interval in seconds.")
     drift_recovery_max_attempts: int = Field(default=5, ge=0, description="Max drift recovery attempts before FAILED.")
-    drift_recovery_base_delay_seconds: int = Field(
-        default=5, ge=0, description="Base delay for drift recovery backoff."
+    drift_recovery_initial_delay_seconds: int = Field(
+        default=5, ge=0, description="Initial delay for drift recovery backoff."
     )
     drift_recovery_max_delay_seconds: int = Field(
         default=300, ge=0, description="Max delay cap for drift recovery backoff."
     )
-    orphan_cleanup_every_n_cycles: int = Field(
-        default=6,
+    orphan_cleanup_interval_seconds: int = Field(
+        default=30,
         ge=0,
-        description="Run orphan substrate cleanup every N reconcile cycles (0 disables).",
+        description="Run orphaned backend resource cleanup after this many seconds (0 disables).",
     )
 
     @model_validator(mode="after")
     def _validate_backoff(self) -> ControllerConfig:
-        if self.drift_recovery_base_delay_seconds > self.drift_recovery_max_delay_seconds:
-            raise ValueError("drift_recovery_base_delay_seconds must not exceed drift_recovery_max_delay_seconds")
+        if self.drift_recovery_initial_delay_seconds > self.drift_recovery_max_delay_seconds:
+            raise ValueError("drift_recovery_initial_delay_seconds must not exceed drift_recovery_max_delay_seconds")
         return self
 
 
@@ -51,9 +51,6 @@ class DeploymentsConfig(NemoConfig):
     )
     default_executor: str | None = Field(
         default=None,
-        description="Fallback executor when Deployment.executor is unset.",
+        description="Default executor when Deployment.executor is unset.",
     )
-    controller: ControllerConfig = Field(
-        default_factory=ControllerConfig,
-        description="Deployment reconciler controller settings.",
-    )
+    controller: ControllerConfig = Field(default_factory=ControllerConfig)
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
@@ -8,13 +8,37 @@ from __future__ import annotations
 from typing import Any, ClassVar
 
 from nemo_platform_plugin.config import NemoConfig
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 
 class ExecutorConfigEntry(BaseModel):
     name: str = Field(description="Unique executor name used by Deployment.executor.")
     backend: str = Field(description="Backend type key registered in BACKEND_CLASSES.")
     config: dict[str, Any] = Field(default_factory=dict)
+
+
+class ControllerConfig(BaseModel):
+    """Configuration for the deployments reconcile controller."""
+
+    interval_seconds: int = Field(default=5, gt=0, description="Reconciliation loop interval in seconds.")
+    drift_recovery_max_attempts: int = Field(default=5, ge=0, description="Max drift recovery attempts before FAILED.")
+    drift_recovery_base_delay_seconds: int = Field(
+        default=5, ge=0, description="Base delay for drift recovery backoff."
+    )
+    drift_recovery_max_delay_seconds: int = Field(
+        default=300, ge=0, description="Max delay cap for drift recovery backoff."
+    )
+    orphan_cleanup_every_n_cycles: int = Field(
+        default=6,
+        ge=0,
+        description="Run orphan substrate cleanup every N reconcile cycles (0 disables).",
+    )
+
+    @model_validator(mode="after")
+    def _validate_backoff(self) -> ControllerConfig:
+        if self.drift_recovery_base_delay_seconds > self.drift_recovery_max_delay_seconds:
+            raise ValueError("drift_recovery_base_delay_seconds must not exceed drift_recovery_max_delay_seconds")
+        return self
 
 
 class DeploymentsConfig(NemoConfig):
@@ -28,4 +52,8 @@ class DeploymentsConfig(NemoConfig):
     default_executor: str | None = Field(
         default=None,
         description="Fallback executor when Deployment.executor is unset.",
+    )
+    controller: ControllerConfig = Field(
+        default_factory=ControllerConfig,
+        description="Deployment reconciler controller settings.",
     )
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
@@ -8,17 +8,21 @@ from __future__ import annotations
 import logging
 from typing import ClassVar
 
-from nemo_deployments_plugin.backends.registry import ExecutorRegistry
+from nemo_deployments_plugin.backends.registry import ExecutorRegistry, ExecutorSpec
 from nemo_deployments_plugin.config import ControllerConfig, DeploymentsConfig
 from nemo_deployments_plugin.entities import Deployment, DeploymentConfig, Volume
 from nemo_deployments_plugin.reconciler.deployment_reconciler import DeploymentReconciler
-from nemo_deployments_plugin.reconciler.listing import get_deployment_for_config_name, list_all_pages
+from nemo_deployments_plugin.reconciler.entity_client import get_deployment_for_config_name, list_all_pages
 from nemo_deployments_plugin.reconciler.orphan_cleanup import reconcile_orphans
+from nemo_deployments_plugin.reconciler.volume_mounts import collect_volume_mount_names
 from nemo_deployments_plugin.reconciler.volume_reconciler import VolumeReconciler
 from nemo_deployments_plugin.types import NON_TERMINAL_DEPLOYMENT_STATUSES, NON_TERMINAL_VOLUME_STATUSES
+from nemo_deployments_plugin.validation import prerequisite_names
+from nemo_platform.resources.entities import AsyncEntitiesResource
 from nemo_platform_plugin.controller import NemoController
-from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError
+from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError, NemoEntityNotFoundError
 from nemo_platform_plugin.filter_ops import ComparisonOperation, FilterOperator
+from nemo_platform_plugin.sdk_provider import get_async_platform_sdk
 
 logger = logging.getLogger(__name__)
 
@@ -36,7 +40,7 @@ class DeploymentsController(NemoController):
         self._deployment_reconciler: DeploymentReconciler | None = None
         self._volume_reconciler: VolumeReconciler | None = None
         self._interval_seconds: float = 5.0
-        self._cycle_count: int = 0
+        self._orphan_cleanup_elapsed_seconds: float = 0.0
         self._deployments_list_ok: bool = True
         self._volumes_list_ok: bool = True
 
@@ -61,11 +65,6 @@ class DeploymentsController(NemoController):
         return self._controller_config
 
     async def on_startup(self) -> None:
-        from nemo_deployments_plugin.backends.registry import ExecutorRegistry, ExecutorSpec
-        from nemo_platform.resources.entities import AsyncEntitiesResource
-        from nemo_platform_plugin.entity_client import NemoEntitiesClient
-        from nemo_platform_plugin.sdk_provider import get_async_platform_sdk
-
         config = DeploymentsConfig.get()
         self._controller_config = config.controller
         self._interval_seconds = float(config.controller.interval_seconds)
@@ -95,10 +94,12 @@ class DeploymentsController(NemoController):
         logger.info("DeploymentsController shut down.")
 
     async def reconcile(self) -> None:
-        self._cycle_count += 1
-        assert self._deployment_reconciler is not None
-        assert self._volume_reconciler is not None
-        assert self._registry is not None
+        if self._deployment_reconciler is None:
+            raise RuntimeError("DeploymentsController.reconcile() called before on_startup()")
+        if self._volume_reconciler is None:
+            raise RuntimeError("DeploymentsController.reconcile() called before on_startup()")
+        if self._registry is None:
+            raise RuntimeError("DeploymentsController.reconcile() called before on_startup()")
 
         deployments = await self._list_deployments()
         volumes = await self._list_volumes()
@@ -131,10 +132,16 @@ class DeploymentsController(NemoController):
             except Exception:
                 logger.exception("Error reconciling deployment %s/%s", deployment.workspace, deployment.name)
 
-        orphan_every = self.controller_config.orphan_cleanup_every_n_cycles
-        if orphan_every > 0 and self._cycle_count % orphan_every == 0 and self._deployments_list_ok:
+        self._orphan_cleanup_elapsed_seconds += self._interval_seconds
+        orphan_interval = self.controller_config.orphan_cleanup_interval_seconds
+        if (
+            orphan_interval > 0
+            and self._orphan_cleanup_elapsed_seconds >= orphan_interval
+            and self._deployments_list_ok
+        ):
             known_ids = {f"{d.workspace}/{d.name}" for d in deployments}
             await reconcile_orphans(self._registry.all_backends(), known_ids)
+            self._orphan_cleanup_elapsed_seconds = 0.0
 
     async def list_objects(self) -> list:
         raise NotImplementedError("DeploymentsController uses reconcile() override")
@@ -204,8 +211,6 @@ class DeploymentsController(NemoController):
         configs: dict[tuple[str, str], DeploymentConfig],
         volumes_by_name: dict[tuple[str, str], Volume],
     ) -> None:
-        from nemo_deployments_plugin.reconciler.volume_mounts import collect_volume_mount_names
-
         for (workspace, _config_name), config in configs.items():
             for mount_name in collect_volume_mount_names(config):
                 key = (workspace, mount_name)
@@ -231,8 +236,6 @@ class DeploymentsController(NemoController):
         by_config: dict[tuple[str, str], Deployment],
     ) -> None:
         """Load prerequisite deployments from entity store (including terminal states)."""
-        from nemo_deployments_plugin.validation import prerequisite_names
-
         for (workspace, _config_name), config in configs.items():
             for prereq_config_name in prerequisite_names(config.prerequisites):
                 key_by_config = (workspace, prereq_config_name)
@@ -244,9 +247,14 @@ class DeploymentsController(NemoController):
                         workspace=workspace,
                         config_name=prereq_config_name,
                     )
-                    if dep is None:
-                        raise LookupError(prereq_config_name)
-                except Exception:
+                except NemoEntityNotFoundError:
+                    logger.debug(
+                        "Prerequisite deployment '%s' not yet available in workspace '%s'",
+                        prereq_config_name,
+                        workspace,
+                    )
+                    continue
+                if dep is None:
                     logger.debug(
                         "Prerequisite deployment '%s' not yet available in workspace '%s'",
                         prereq_config_name,
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
@@ -1,0 +1,284 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deployments reconcile controller — drives Deployment and Volume state machines."""
+
+from __future__ import annotations
+
+import logging
+from typing import ClassVar
+
+from nemo_deployments_plugin.backends.registry import ExecutorRegistry
+from nemo_deployments_plugin.config import ControllerConfig, DeploymentsConfig
+from nemo_deployments_plugin.entities import Deployment, DeploymentConfig, Volume
+from nemo_deployments_plugin.reconciler.deployment_reconciler import DeploymentReconciler
+from nemo_deployments_plugin.reconciler.listing import get_deployment_for_config_name, list_all_pages
+from nemo_deployments_plugin.reconciler.orphan_cleanup import reconcile_orphans
+from nemo_deployments_plugin.reconciler.volume_reconciler import VolumeReconciler
+from nemo_deployments_plugin.types import NON_TERMINAL_DEPLOYMENT_STATUSES, NON_TERMINAL_VOLUME_STATUSES
+from nemo_platform_plugin.controller import NemoController
+from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError
+from nemo_platform_plugin.filter_ops import ComparisonOperation, FilterOperator
+
+logger = logging.getLogger(__name__)
+
+
+class DeploymentsController(NemoController):
+    """Reconciles deployments and volumes against registered executor backends."""
+
+    name = "deployments"
+    dependencies: ClassVar[list[str]] = ["entities"]
+
+    def __init__(self) -> None:
+        self._entities: NemoEntitiesClient | None = None
+        self._registry: ExecutorRegistry | None = None
+        self._controller_config: ControllerConfig | None = None
+        self._deployment_reconciler: DeploymentReconciler | None = None
+        self._volume_reconciler: VolumeReconciler | None = None
+        self._interval_seconds: float = 5.0
+        self._cycle_count: int = 0
+        self._deployments_list_ok: bool = True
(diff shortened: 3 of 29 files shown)


tylersbray:
@coderabbitai full review

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,336 @@
...
+            await self._project_status(
+                deployment,
+                BackendStatusUpdate(status="DELETING", status_message="Stopping deployment"),
+            )
+
+        if backend is not None:
+            try:
+                await backend.delete_deployment(deployment.workspace, deployment.name)
+            except Exception:
+                logger.warning("Backend delete failed for %s", dep_id, exc_info=True)
```
benmccown: Could we consider an approach that tolerates failures when deleting backend resources?

For Models service, we leave the status `DELETING` until backend deletion comes back clean.

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -178,13 +179,45 @@ class VolumeBackendConfig(BaseModel):
...
+            raise ValueError("base_delay_seconds must not exceed max_delay_seconds")
+        return self
 
 
 class Prerequisite(BaseModel):
     deployment_name: str = Field(
         description=(
-            "Name of another DeploymentConfig in the same workspace that must reach "
-            "a ready terminal state before this config's deployment may start."
+            "Name of another DeploymentConfig in the same workspace whose corresponding "
```
benmccown, line 212: I'm realizing a small problem I didn't catch in the RFC.

The "prereqs" dependency graph is about `Deployment` entities. You can deploy a `DeploymentConfig` more than once. So tying the DAG to `DeploymentConfig` isn't going to work for us, particularly because the `_find_prerequisite_deployment` method will have issues when DeploymentConfigs are deployed more than once. I'm thinking we move `prerequisites` to `Deployment`. Claude summary of the concern below. Let me know what you think.

---

**Original problem:** `DeploymentConfig.prerequisites` references configs by name, but multiple `Deployment` entities can share the same `DeploymentConfig`. The controller builds a `by_config` map keyed on `(workspace, config_name)` to resolve prerequisite lookups, but when two deployments share a config, only one wins in the map — the loser is silently dropped. This makes prerequisite evaluation non-deterministic: whether a dependent deployment proceeds, waits, or fails depends on which of the competing deployments happens to land in the map, which varies with entity list order across reconcile cycles.

**Design conclusion:** Prerequisites are lifecycle gates on running instances, not on templates. `DeploymentConfig` is a reusable, stateless template — instance-level dependencies don't belong there. Moving `prerequisites` onto the `Deployment` entity and changing `Prerequisite.deployment_name` to reference a specific `Deployment` name (not a config name) makes the reference semantically correct and eliminates the ambiguity entirely. The `by_name` map is already a clean 1:1 index, so prerequisite lookups become a simple name lookup with no collision case.

This also simplifies the code: `_find_prerequisite_deployment`'s dual `by_config`/`by_name` fallback logic, `_ensure_prerequisite_refs_loaded`'s config-name resolution, and `listing.get_deployment_for_config_name` all become unnecessary and can be removed.

The one caller-visible consequence is that the deployment creating a dependency must know the prerequisite deployment's name at creation time. Given that auto-deploying the prerequisite graph is explicitly not intended behavior, this is the right tradeoff — it forces explicit wiring rather than implicit graph execution.

plugins/nemo-deployments/README.md
```diff
@@ -1,11 +1,51 @@
...
 ## Tests
 
 ```bash
 uv sync
 uv run pytest plugins/nemo-deployments/tests/unit -v
 ```
+
+## Next steps
+
+- **[AIRCORE-756](https://linear.app/nvidia/issue/AIRCORE-756):** Docker `DeploymentBackend`
```
benmccown: nit again, I don't think any of the README additions are worth keeping except for the `Controller` section.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
```diff
@@ -17,6 +17,30 @@ class ExecutorConfigEntry(BaseModel):
...
+
+    interval_seconds: int = Field(default=5, gt=0, description="Reconciliation loop interval in seconds.")
+    drift_recovery_max_attempts: int = Field(default=5, ge=0, description="Max drift recovery attempts before FAILED.")
+    drift_recovery_base_delay_seconds: int = Field(
+        default=5, ge=0, description="Base delay for drift recovery backoff."
+    )
+    drift_recovery_max_delay_seconds: int = Field(
+        default=300, ge=0, description="Max delay cap for drift recovery backoff."
+    )
+    orphan_cleanup_every_n_cycles: int = Field(
```
benmccown: nit: this seems a bit unintuitive to me. Could we go with `orphan_cleanup_interval_seconds` and have the controller do the math for when it runs? For example, if the regular `interval_seconds == 5` and I set `orphan_cleanup_interval_seconds = 28` I'd assume the orphan cleanup would happen on the next interval after `28` which would be `30`.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
```diff
@@ -17,6 +17,30 @@ class ExecutorConfigEntry(BaseModel):
     config: dict[str, Any] = Field(default_factory=dict)
 
 
+class ControllerConfig(BaseModel):
+    """Configuration for the deployments reconcile controller."""
+
+    interval_seconds: int = Field(default=5, gt=0, description="Reconciliation loop interval in seconds.")
+    drift_recovery_max_attempts: int = Field(default=5, ge=0, description="Max drift recovery attempts before FAILED.")
+    drift_recovery_base_delay_seconds: int = Field(
```
benmccown: super nit: I think the more common term for this in backoff/retry libs, instead of `base_delay`, is `initial_delay`. Could we go with `drift_recovery_initial_delay_seconds`?

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
```diff
@@ -17,6 +17,30 @@ class ExecutorConfigEntry(BaseModel):
...
+        default=5, ge=0, description="Base delay for drift recovery backoff."
+    )
+    drift_recovery_max_delay_seconds: int = Field(
+        default=300, ge=0, description="Max delay cap for drift recovery backoff."
+    )
+    orphan_cleanup_every_n_cycles: int = Field(
+        default=6,
+        ge=0,
+        description="Run orphan substrate cleanup every N reconcile cycles (0 disables).",
```
benmccown: nit that's my fault: the term `"substrate"` got baked into the RFC more than I would have liked. In my experience, it's not a super common term, and I think a better fit is "service backend resources", so I've been trying to avoid using "substrate" when I can. It's just personal preference, so if you disagree I'm fine leaving it, but wanted to flag for consideration.

So example description, reworded, might be "Run orphaned backend resource cleanup every N reconcile cycles (0 disables)."

up to you :)

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,284 @@
...
+from nemo_deployments_plugin.reconciler.volume_reconciler import VolumeReconciler
+from nemo_deployments_plugin.types import NON_TERMINAL_DEPLOYMENT_STATUSES, NON_TERMINAL_VOLUME_STATUSES
+from nemo_platform_plugin.controller import NemoController
+from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError
+from nemo_platform_plugin.filter_ops import ComparisonOperation, FilterOperator
+
+logger = logging.getLogger(__name__)
+
+
+class DeploymentsController(NemoController):
```
benmccown, line 30: For consideration in this PR or as a follow up:

**No stop/cancellation signal during reconcile**
The models controller checks a stop signal at multiple points mid-reconcile (`lines 263, 427, 441, 454, 490`) so it can bail early on shutdown without waiting to process every entity. `DeploymentsController` has no equivalent — on shutdown it would run through all deployments and volumes before stopping. For small entity counts this is fine, but at scale a reconcile cycle could block shutdown noticeably.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,284 @@
...
+                    exc_info=True,
+                )
+        return configs
+
+    async def _ensure_volume_refs_loaded(
+        self,
+        configs: dict[tuple[str, str], DeploymentConfig],
+        volumes_by_name: dict[tuple[str, str], Volume],
+    ) -> None:
+        from nemo_deployments_plugin.reconciler.volume_mounts import collect_volume_mount_names
```
benmccown: Nit: are the inline imports here and in `_ensure_prerequisite_refs_loaded` loaded inline for an intentional reason, or could they be lifted?

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,284 @@
...
+        logger.info("DeploymentsController started.")
+
+    async def on_shutdown(self) -> None:
+        if self._registry is not None:
+            self._registry.shutdown_all()
+        logger.info("DeploymentsController shut down.")
+
+    async def reconcile(self) -> None:
+        self._cycle_count += 1
+        assert self._deployment_reconciler is not None
```
benmccown: I think we should probably just do a conditional check and raise a runtime error like we do for entities around line 53, instead of assertions? Might as well be consistent.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,284 @@
...
+                )
+            except NemoEntityConflictError:
+                logger.debug("Optimistic lock conflict on deployment '%s' — retry next cycle.", deployment.name)
+            except Exception:
+                logger.exception("Error reconciling deployment %s/%s", deployment.workspace, deployment.name)
+
+        orphan_every = self.controller_config.orphan_cleanup_every_n_cycles
+        if orphan_every > 0 and self._cycle_count % orphan_every == 0 and self._deployments_list_ok:
+            known_ids = {f"{d.workspace}/{d.name}" for d in deployments}
+            await reconcile_orphans(self._registry.all_backends(), known_ids)
```
benmccown, line 143: I think, since `known_ids` is built from deployments in non-terminal states, that "finite" (job style) deployments in `SUCCEEDED` state would have their backend resources deleted as part of orphan reconciliation.

I think that's _probably_ fine? We do want to garbage collect them at some point. The only question is whether there should be a TTL or something before we reap/cleanup. There's also the failure state (`FAILED` I think?) which would have the same thing happen, and it might be hard to debug why the underlying k8s job failed or something. A TTL before cleanup happens might be particularly useful for that case.

If you want we could probably leave as is and address in follow up work.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,331 @@
...
+                self._config_cache[config_key] = config
+            except NemoEntityNotFoundError:
+                await self._project_failure(
+                    deployment, f"DeploymentConfig '{deployment.deployment_config_name}' not found"
+                )
+                return
+
+        if deployment.desired_state == "STOPPED" or deployment.status == "DELETING":
+            await self._reconcile_delete(deployment, config)
+            return
```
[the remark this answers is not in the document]
benmccown: I think this is potentially one that's worth addressing.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,336 @@
...
+                policy.max_delay_seconds
+                if policy.max_delay_seconds is not None
+                else ctrl.drift_recovery_max_delay_seconds
+            ),
+        )
+
+    def resolve_backend(self, deployment: Deployment) -> DeploymentBackend:
+        return self._registry.resolve(deployment.executor)
+
+    async def _resolve_backend_or_fail(self, deployment: Deployment) -> DeploymentBackend | None:
```
benmccown, line 66: Nit: The difference between `_resolve_backend_or_fail` and `_try_resolve_backend` wasn't super clear to me without asking an agent (response below). Maybe a docstring/comment could explain, or we could make the method names clearer?

---

`_try_resolve_backend` is used only in `_reconcile_delete` — when deleting, a missing executor isn't fatal. The code logs a warning and proceeds to delete the entity anyway (backend-less deletion). That's intentional: you don't want a misconfigured executor to block cleanup.

`_resolve_backend_or_fail` is used in the main `reconcile_one` path — when creating or polling a deployment, a missing executor is a real failure that should mark the deployment `FAILED`.

So the two methods represent genuinely different semantics: "best-effort, proceed without backend" vs "required, fail the deployment if missing." The naming could be clearer — `_try_resolve_backend` doesn't communicate why you'd want a `None` return — but the split itself is justified.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,336 @@
...
+        try:
+            return self.resolve_backend(deployment)
+        except ExecutorNotFoundError as exc:
+            await self._project_failure(deployment, f"No executor available: {exc}")
```
benmccown: Nit: Maybe it's just me, LLMs seem to like the word `project` for some reason (I think it's slipped through in my own code and RFCs) but _if you're having an agent do any refactoring already_ (don't want it to be a dedicated task on its own) maybe `update_deployment_status`/`update_deployment_status_failure`/`update_deployment_status_error` would be clearer? Not a hill I'll die on, I just think it's a weird verb :).

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/entity_client.py
```diff
@@ -0,0 +1,74 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
benmccown, line 1: Nit: filename feels strange to me. Maybe `entity_client.py`, `entity_client_helpers.py`, or even just `utils.py`?

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/orphan_cleanup.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import annotations
+
+import logging
+
+from nemo_deployments_plugin.backends.base import DeploymentBackend
+
+logger = logging.getLogger(__name__)
+
+
+async def reconcile_orphans(
```
benmccown, line 15: Nit: Feels like this could just live in `deployment_reconciler.py` but /shrug

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_mounts.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
benmccown, line 1: Nit: I think this file is entirely about gating deployment creation on volume readiness, but reading it at first I assumed it was used by `volume_reconciler.py`. Maybe this could move into `prerequisite.py` and we could just rename it `dependencies.py` or something. Not a must-do.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class VolumeReconciler:
+    """Reconciles Volume entities with substrate volume resources."""
+
+    def __init__(self, entities: NemoEntitiesClient, registry: ExecutorRegistry) -> None:
+        self._entities = entities
+        self._registry = registry
+
+    def resolve_backend(self, volume: Volume) -> DeploymentBackend:
```
benmccown: Is this dead code? Doesn't look like it's called?

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -0,0 +1,93 @@
...
+
+from nemo_deployments_plugin.backends.base import DeploymentBackend, VolumeStatusUpdate
+from nemo_deployments_plugin.backends.registry import ExecutorNotFoundError, ExecutorRegistry
+from nemo_deployments_plugin.entities import Volume
+from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError
+
+logger = logging.getLogger(__name__)
+
+
+class VolumeReconciler:
```
benmccown, line 18: Either now or in follow up, we need to add deletion handling (follow up is fine by me).

benmccown (commented):
Quite a few nits that you can take or leave. Mostly just focused on code quality and edge cases. Core patterns LGTM.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
benmccown: Basically, if deleting a backend resource transiently fails for some reason, leave the deployment in `DELETING` so that we try again on the next controller loop.

plugins/nemo-deployments/README.md
```diff
@@ -1,16 +1,39 @@
...
+
+## Controller
+
+Register `DeploymentsController` via the `nemo.controllers` entry point. The controller:
+
+- Paginates non-terminal deployment/volume lists (no 100-item cap)
+- Reconciles volumes before deployments (puller→server ordering)
+- Gates deployment create on mounted volumes reaching `BOUND`
+- Writes status via the service-principal entity client
+- Reports `is_healthy=False` when entity list queries fail
```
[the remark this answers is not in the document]
tylersbray: Addressed in d036a9233a: README Controller section now documents separate deployment-list vs volume-list health signals behind `is_healthy`.

plugins/nemo-deployments/README.md
```diff
@@ -23,7 +46,7 @@ Cross-workspace bulk queries use the entity-store sentinel workspace ``-``:
 ## Next steps
 
 - **756 / 757:** Docker and Kubernetes `DeploymentBackend` implementations
-- **758:** Reconcile controller wiring status writes and backend lifecycle
+- **759:** Models/agents adoption projecting from plugin `Deployment` status
```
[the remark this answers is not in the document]
tylersbray: Addressed per @benmccown's feedback in d036a9233a: trimmed README to the Controller section only (Next steps/Tests/Deferred sections removed).

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
```diff
@@ -17,6 +17,19 @@ class ExecutorConfigEntry(BaseModel):
...
+
+    interval_seconds: int = Field(default=5, description="Reconciliation loop interval in seconds.")
+    drift_recovery_max_attempts: int = Field(default=5, description="Max drift recovery attempts before FAILED.")
+    drift_recovery_base_delay_seconds: int = Field(default=5, description="Base delay for drift recovery backoff.")
+    drift_recovery_max_delay_seconds: int = Field(default=300, description="Max delay cap for drift recovery backoff.")
+    orphan_cleanup_every_n_cycles: int = Field(
+        default=6,
+        description="Run orphan substrate cleanup every N reconcile cycles.",
+    )
+
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489/d036a9233a: `ControllerConfig` uses `Field(gt=0)` on `interval_seconds` and a model validator for backoff ordering.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,274 @@
...
+
+class DeploymentsController(NemoController):
+    """Reconciles deployments and volumes against registered executor backends."""
+
+    name = "deployments"
+    dependencies: ClassVar[list[str]] = ["entities"]
+
+    def __init__(self) -> None:
+        self._entities: NemoEntitiesClient | None = None
+        self._registry = None
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: `self._registry: ExecutorRegistry | None = None` at controller.py:34.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
```diff
@@ -0,0 +1,274 @@
...
+                    logger.debug(
+                        "Prerequisite deployment '%s' not yet available in workspace '%s'",
+                        prereq_config_name,
+                        workspace,
+                    )
+                    continue
+                by_name[(workspace, dep.name)] = dep
+                by_config.setdefault((workspace, dep.deployment_config_name), dep)
+                if dep.name != prereq_config_name:
+                    by_config.setdefault(key_by_config, dep)
```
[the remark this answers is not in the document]
tylersbray: Partially addressed: `_index_deployments` logs a warning when multiple deployments share a config name. Full fix (prerequisites on `Deployment`) is tracked as follow-up work — see reply on @benmccown's entities.py thread.

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -178,13 +179,32 @@ class VolumeBackendConfig(BaseModel):
...
+        description="Override controller drift_recovery_max_attempts when set.",
+    )
+    base_delay_seconds: int | None = Field(
+        default=None,
+        description="Override controller drift_recovery_base_delay_seconds when set.",
+    )
+    max_delay_seconds: int | None = Field(
+        default=None,
+        description="Override controller drift_recovery_max_delay_seconds when set.",
+    )
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: `DriftRecoveryPolicy` model validator ensures `initial_delay_seconds <= max_delay_seconds` (field renamed in d036a9233a).

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,331 @@
...
+from nemo_deployments_plugin.reconciler.drift_recovery import DriftRecoveryCache, DriftRecoveryLimits, RecoveryAction
+from nemo_deployments_plugin.reconciler.prerequisite import PrerequisiteResult, prerequisites_met
+from nemo_deployments_plugin.reconciler.volume_mounts import VolumeMountResult, volume_mounts_ready
+from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError, NemoEntityNotFoundError
+
+logger = logging.getLogger(__name__)
+
+
+def deployment_id(deployment: Deployment) -> str:
+    return f"{deployment.workspace}/{deployment.name}"
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489/d036a9233a: `_reconcile_delete` calls `self._drift_cache.remove(dep_id)`; unit test `test_delete_clears_drift_recovery_cache` added.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
```diff
@@ -0,0 +1,331 @@
...
+    deployments_by_name: dict[tuple[str, str], Deployment],
+    deployment: Deployment,
+) -> bool:
+    if result.blocking_prerequisite is None:
+        return "failed" in result.reason.lower()
+    workspace = deployment.workspace
+    target = deployments_by_name.get((workspace, result.blocking_prerequisite))
+    if target is None:
+        target = deployments_by_config.get((workspace, result.blocking_prerequisite))
+    return target is not None and target.status == "FAILED"
```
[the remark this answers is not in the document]
tylersbray: Out of scope for this PR: prerequisites are same-workspace DeploymentConfig names today. Cross-workspace resolution will be addressed when prerequisites move to `Deployment` (follow-up ticket).

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -0,0 +1,89 @@
...
+                backend_config=backend_config,
+            )
+            await self._project_status(volume, update)
+            logger.info("Volume %s/%s created: %s", volume.workspace, volume.name, update.status)
+        except Exception as exc:
+            logger.exception("Failed to create volume %s/%s", volume.workspace, volume.name)
+            await self._project_status(
+                volume,
+                VolumeStatusUpdate(status="FAILED", status_message=f"Failed to create volume: {exc}"),
+            )
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: `NemoEntityConflictError` is re-raised in volume create/read paths; controller retries on next cycle.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/orphan_cleanup.py
```diff
@@ -0,0 +1,38 @@
...
+            logger.warning("Failed to list managed deployments for orphan cleanup", exc_info=True)
+            continue
+
+        orphans = set(backend_names) - known_deployment_ids
+        for deployment_id in orphans:
+            parts = deployment_id.split("/", 1)
+            if len(parts) != 2:
+                logger.warning("Invalid deployment id from backend: %r, skipping", deployment_id)
+                continue
+            workspace, name = parts
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: `orphan_cleanup.py` validates non-empty workspace/name after split.

plugins/nemo-deployments/tests/unit/reconciler/test_controller.py
```diff
@@ -0,0 +1,278 @@
...
+
+@pytest.mark.asyncio
+async def test_controller_unhealthy_after_list_failure() -> None:
+    ctrl = DeploymentsController()
+    ctrl._entities = AsyncMock()
+    ctrl._entities.list.side_effect = RuntimeError("entity store down")
+
+    deployments = await ctrl._list_deployments()
+    assert deployments == []
+    assert ctrl.is_healthy is False
```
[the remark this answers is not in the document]
tylersbray: Stale suggestion: `is_healthy` is a `@property` (not a method), and tests correctly use `ctrl.is_healthy` without calling it.

plugins/nemo-deployments/tests/unit/reconciler/test_deployment_reconciler.py
```diff
@@ -0,0 +1,448 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
tylersbray: Addressed in d036a9233a: removed `from __future__ import annotations` from `test_deployment_reconciler.py`.

plugins/nemo-deployments/tests/unit/reconciler/test_deployment_reconciler.py
```diff
@@ -0,0 +1,448 @@
...
+        dep, deployments_by_config={}, deployments_by_name={}, volumes_by_name=NO_VOLUMES
+    )
+    assert dep.status == "LOST"
+    assert "Will retry" in dep.status_message
+
+    await deployment_reconciler.reconcile_one(
+        dep, deployments_by_config={}, deployments_by_name={}, volumes_by_name=NO_VOLUMES
+    )
+    assert dep.status == "LOST"
+    assert dep.status != "FAILED"
```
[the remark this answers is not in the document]
tylersbray: Already covered: `test_drift_recovery_create_failure_stays_lost_and_backoffs` asserts `len(mock_backend.create_calls) == 1` after the second reconcile.

plugins/nemo-deployments/tests/unit/reconciler/test_prerequisite.py
```diff
@@ -0,0 +1,64 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from helpers import make_deployment, make_deployment_config
```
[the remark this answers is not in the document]
tylersbray: Keeping `from helpers import ...` — `pyproject.toml` adds `tests/unit` to `pythonpath`; relative imports fail because test dirs are not packages.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/drift_recovery.py
```diff
@@ -0,0 +1,72 @@
...
+    def should_recover(self, deployment_id: str, limits: DriftRecoveryLimits) -> RecoveryAction:
+        state = self._states.get(deployment_id)
+        if state is None:
+            return RecoveryAction.PROCEED
+        if state.attempts >= limits.max_attempts:
+            return RecoveryAction.EXHAUSTED
+        if state.last_attempt_at:
+            backoff_seconds = min(
+                limits.base_delay_seconds * (2**state.attempts),
+                limits.max_delay_seconds,
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: backoff uses `2 ** max(state.attempts - 1, 0)` in `drift_recovery.py`.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/listing.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+    if deployments:
+        return deployments[0]
+
+    try:
+        dep = await entities.get(Deployment, name=config_name, workspace=workspace)
+    except NemoEntityNotFoundError:
+        return None
+    except Exception:
+        return None
```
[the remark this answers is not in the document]
tylersbray: Addressed: `entity_client.get_deployment_for_config_name` only catches `NemoEntityNotFoundError`; other errors propagate. File renamed from `listing.py` to `entity_client.py` in d036a9233a.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/prerequisite.py
```diff
@@ -0,0 +1,78 @@
...
+    deployments_by_config: dict[tuple[str, str], Deployment],
+    deployments_by_name: dict[tuple[str, str], Deployment],
+) -> Deployment | None:
+    """Resolve the Deployment entity for a prerequisite DeploymentConfig name."""
+    workspace = deployment.workspace
+    key_by_name = (workspace, prerequisite.deployment_name)
+    if key_by_name in deployments_by_name:
+        return deployments_by_name[key_by_name]
+    key_by_config = (workspace, prerequisite.deployment_name)
+    return deployments_by_config.get(key_by_config)
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: `prerequisite.py` consults `deployments_by_config` first, then name index with config-name match.

plugins/nemo-deployments/tests/integration/test_reconcile_docker.py
```diff
@@ -0,0 +1,20 @@
...
+Requires AIRCORE-756 DockerDeploymentBackend to be registered in BACKEND_CLASSES.
+Until then, these tests are skipped — unit tests with MockDeploymentBackend provide coverage.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+pytestmark = pytest.mark.skip(reason="Requires DockerDeploymentBackend (AIRCORE-756)")
+
```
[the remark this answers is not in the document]
tylersbray: Addressed in fc03c2e489: integration tests use `pytest.mark.skipif("docker" not in BACKEND_CLASSES, ...)`.

plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
tylersbray: Addressed in d036a9233a: removed `from __future__ import annotations` from `test_volume_reconciler.py`.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
tylersbray: Addressed in d036a9233a: backend delete failures now leave the deployment in `DELETING` and return without deleting the entity, so the next reconcile loop retries.

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
tylersbray: Agree with the design — moving prerequisites to `Deployment` (referencing deployment names) is the right long-term fix. Deferring to a follow-up ticket; README Controller section notes the current config-scoped MVP limitation.

plugins/nemo-deployments/README.md
tylersbray: Addressed in d036a9233a: README trimmed to intro + Controller section only.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
tylersbray: Addressed in d036a9233a: renamed to `orphan_cleanup_interval_seconds`; controller tracks elapsed seconds and runs cleanup on the first reconcile tick at/after the interval.

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
tylersbray: Addressed in d036a9233a: renamed to `drift_recovery_initial_delay_seconds` (controller) and `initial_delay_seconds` (policy override).

plugins/nemo-deployments/src/nemo_deployments_plugin/config.py
tylersbray: Addressed in d036a9233a: user-facing strings now say backend resource(s) instead of substrate.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
tylersbray: Deferring to follow-up — entity counts are small today and shutdown blocking is acceptable for the initial controller. Will add stop-signal checks when we scale up.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
tylersbray: Addressed in d036a9233a: inline imports hoisted to module top in `controller.py`.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
tylersbray: Addressed in d036a9233a: `reconcile()` raises `RuntimeError` instead of `assert` when reconcilers/registry are unset.

plugins/nemo-deployments/src/nemo_deployments_plugin/controller.py
tylersbray: Acknowledged — terminal deployments are excluded from `known_ids` so backend resources can be reaped. Agree a TTL before cleanup would help debugging FAILED/SUCCEEDED jobs; deferring to follow-up.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
tylersbray: Addressed in d036a9233a: added docstrings to `_resolve_backend_or_fail` and `_try_resolve_backend` explaining fail-vs-best-effort semantics.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/deployment_reconciler.py
tylersbray: Addressed in d036a9233a: renamed `_project_*` to `_update_deployment_status*` in deployment and volume reconcilers.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/entity_client.py
tylersbray: Addressed in d036a9233a: renamed `listing.py` → `entity_client.py` (and test file accordingly).

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/orphan_cleanup.py
tylersbray: Keeping `orphan_cleanup.py` as a separate module for readability — logic is invoked from the controller and tested independently.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_mounts.py
tylersbray: Keeping `volume_mounts.py` for now — docstring clarifies it gates deployment create on volume readiness (not used by volume reconciler).

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
tylersbray: Addressed in d036a9233a: removed unused `VolumeReconciler.resolve_backend()` dead code.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
tylersbray: Deferring volume entity deletion handling to follow-up — volume delete/`RELEASED` requires backend ABC changes not in 755 scope.

benmccown (approved):
Thanks for addressing feedback!

PR #316, opened 2026-06-12
feat(studio): add column visibility and filters to test cases table

walston:
Closes FP-239

https://github.com/user-attachments/assets/2dae4d3f-dc61-499a-acb1-40cade79e99b

# Summary

The test cases table (\`ExperimentSessionsDataView\`) had no toolbar controls — no way to hide columns, search, or filter. This adds all three, lifting the pattern from \`ExperimentGroupDataView\`.

**Column visibility** — an \`EditColumnsMenu\` dropdown in the toolbar lets users toggle any column on or off.

**Case ID search** — a search bar wired to the \`test_case_id\` API filter field. Matches are exact (the API doesn't support substring matching on this field).

**Status filter** — a single-select dropdown with the four known status values (Completed, Error, Cancelled, Unknown). Labels match what the \`StatusBadge\` displays. Because the API only accepts a single status string, filtering is applied client-side immediately against the current page while the debounced API request catches up — avoiding the stale-data flicker that would otherwise appear.

Also fixes a pre-existing type error in \`ExperimentGroupDataView\` and \`ExperimentDetailMetrics\` where \`agent_names\`/\`agent_versions\` (arrays) were referenced after the SDK renamed them to \`agent_name\`/\`agent_version\` (strings).

# Test plan
- [ ] Navigate to an experiment detail page → confirm a toolbar appears above the test cases table with a search bar, a Filter button, and a Columns button
- [ ] Click **Columns** → confirm all columns are listed and toggling one hides/shows it in the table
- [ ] Type a full \`test_case_id\` value in the search bar → confirm only matching rows are returned
- [ ] Click **Filter** → select a status (e.g. Completed) → confirm the table filters immediately to matching rows without a visible loading flash
- [ ] Clear the status filter → confirm all rows return

## Summary by CodeRabbit

* **New Features**
  * Added status-based filtering for experiment sessions.
  * Introduced column visibility controls to customize the data view.
  * Added search capability for experiment sessions.
  
* **Improvements**
  * Enhanced UI responsiveness with immediate client-side filtering.
  * Improved empty-state messaging when filters are active.

### DIFF
--- a/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/index.tsx
+++ b/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/index.tsx
@@ -132,13 +132,15 @@ export const ExperimentGroupDataView: FC<ExperimentGroupDataViewProps> = ({
           );
         },
       }),
-      accessor('agent_name', {
-        header: 'Agent Name',
+      accessor((original) => original.agent_names?.join(', '), {
+        id: 'agent_names',
+        header: 'Agent Names',
         enableSorting: false,
         cell: ({ getValue }) => <Text>{getValue<string>() || '-'}</Text>,
       }),
-      accessor('agent_version', {
-        header: 'Agent Version',
+      accessor((original) => original.agent_versions?.join(', '), {
+        id: 'agent_versions',
+        header: 'Agent Versions',
         enableSorting: false,
         cell: ({ getValue }) => <Text>{getValue<string>() || '-'}</Text>,
       }),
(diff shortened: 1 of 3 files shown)


rrhyne:
Looks great! A few tweaks:

* Change the search input placeholder text to "Search case..." if that's all it searches. 
* Change the empty state when no test case matches a filter... this isn't there are no test cases in this experiment, add some, this is "No test cases matched your filters". As a result, remove the command area for adding test cases, and change the header copy to "No matching test cases" and sub text "Change your filters and try again." If you can get a link in there add an href for "Clear filters" to the end.

BrianNewsom (approved):

PR #318, opened 2026-06-12
ci(release): declare publishable container images for release

a2bondar:
## Summary

Makes `release/assets.yaml` the **single source of truth** for which container images are eligible for release publishing, and records the selected images as typed entries in the release bundle manifest -- converging containers onto the existing release flow.

- New `container:` list in `release/assets.yaml` (mirrors the `sdk:` list) declaring the publishable images: `nmp-automodel-tasks`, `nmp-automodel-training`, `nmp-unsloth-training`, `auditor-tasks`, `safe-synthesizer-tasks`.
- `release-bundle.yaml` `plan-assets` validates the catalog (non-empty, unique ids) and selects SDKs **and** containers per `release_scope`, appending `{"type": "container", "id": ...}` entries to the selected artifacts.
- `write_release_bundle_metadata.py` records them in `release-manifest.json` as metadata-only entries (type, id, version = release label; no path or checksum -- image bits are built and staged by the release consumer from its dev registry at the bundle's source SHA). Unit tests extended (19 passing).
- `RELEASING.md` documents the eligibility contract and the `release_scope` table.

`release_scope` (default `all`): `all` = every SDK + every container; `sdks` = SDKs only; `containers` = containers only; `custom` = exactly the comma-separated `sdk_ids` + `container_ids` (either may be empty). Eligibility is version-pinned -- re-staging an old tag publishes the container set declared at that commit.

Consumer side (NVIDIA-NeMo/Platform-Deploy#95): `release-bundle-received` automatically stages the selected containers after the SDK publish, reading this list at the release ref. Public NGC publishing is out of band via the NGC catalog process (`ngc-publishing-configs`); this repo only declares eligibility.

## Merge order

Either order is safe: without the consumer PR, the manifest's container entries are recorded but unconsumed (the existing SDK publish plan filters `type == "sdk"`); without this PR, the consumer's staging step no-ops with "no containers declared". The full flow activates when both land.

Refs: [AIREINF-216](https://linear.app/nvidia/issue/AIREINF-216)

### DIFF
--- a/.github/scripts/write_release_bundle_metadata.py
+++ b/.github/scripts/write_release_bundle_metadata.py
@@ -1,7 +1,12 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Write release bundle metadata for downloaded SDK wheel artifacts."""
+"""Write release bundle metadata for downloaded SDK wheel artifacts.
+
+Container artifacts are metadata-only manifest entries: the image bits are
+built and staged by Platform-Deploy from its dev registry at the bundle's
+source SHA, so container entries carry no path or checksum.
+"""
 
 import argparse
 import hashlib
@@ -22,10 +27,12 @@ class BundleMetadataError(Exception):
     """Raised when the release bundle metadata cannot be written safely."""
 
 
-def safe_sdk_id(sdk_id: str) -> str:
-    if not re.fullmatch(r"[A-Za-z0-9._-]+", sdk_id) or sdk_id in {".", ".."}:
-        raise BundleMetadataError(f"selected SDK id must be a safe single path segment: {sdk_id}")
-    return sdk_id
+def safe_artifact_id(artifact_type: str, artifact_id: str) -> str:
+    if not re.fullmatch(r"[A-Za-z0-9._-]+", artifact_id) or artifact_id in {".", ".."}:
+        raise BundleMetadataError(
+            f"selected {artifact_type} id must be a safe single path segment: {artifact_id}"
+        )
+    return artifact_id
 
 
 def parse_release_date_json(value: str) -> str | None:
@@ -43,7 +50,8 @@ def artifact_ref(artifact_type: object, artifact_id: object) -> str:
     return f"{artifact_type}:{artifact_id}"
 
 
-def parse_selected_sdk_ids(value: str) -> list[str]:
+def parse_selected_artifact_ids(value: str) -> dict[str, list[str]]:
+    """Parse selected_artifacts_json into ids grouped by artifact type."""
     try:
         parsed = json.loads(value)
     except json.JSONDecodeError as error:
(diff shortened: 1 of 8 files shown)


.github/workflows/release-bundle.yaml
```diff
@@ -302,8 +302,24 @@ jobs:
             exit 1
           fi
 
+          # Containers ride along on every release regardless of release_scope
+          # (scope governs SDK selection only): the container catalog is the
+          # eligibility list, and image bits are not built here. Entries are
+          # recorded in the manifest as typed artifacts; Platform-Deploy stages
```
svvarom: Should we drop the references to the internal repo in this PR?

.github/workflows/release-bundle.yaml
```diff
@@ -302,8 +302,24 @@ jobs:
             exit 1
           fi
 
+          # Containers ride along on every release regardless of release_scope
```
svvarom: If we do this, do we lose support for release‑specific items - for e.g., if we only want to release a single container or just the SDKs?

svvarom: Like if we wanted to do a patch release for a single container, iirc that wasn’t possible in the nmp days, that's why I tried to add `release_scope`

release/assets.yaml
```diff
@@ -1,3 +1,15 @@
...
+# source of truth: Platform-Deploy's Release Deploy Artifacts and Release
+# Promote Public workflows read it from this repository at the release ref.
+# Each id must match an image name pushed to the dev registry tagged by this
+# repository's commit SHA, and should have a catalog metadata entry in
+# Platform-Deploy release/nemo-assets-config.yaml.
+container:
+  - id: nmp-automodel-base
+  - id: nmp-automodel-tasks
+  - id: nmp-automodel-training
+  - id: nmp-unsloth-training
```
crookedstorm, line 13: we also need nmp-api, nmp-core, nmp-cpu-tasks. These are the bare minimum required by customizer for local work. If we are pushing a chart and a deployable service, we need those three as well.

.github/workflows/release-bundle.yaml
crookedstorm: +1

.github/workflows/release-bundle.yaml
a2bondar: Addressed — `release_scope` now selects containers too (full parity with SDKs), so the subsetting you built `release_scope`/`sdk_ids` for works for containers as well:

| scope | includes |
| --- | --- |
| `all` (new default) | every SDK + every container |
| `sdks` | SDK-only (no containers) |
| `containers` | container-only (no SDKs) |
| `custom` | `sdk_ids` + `container_ids` subset (single-container / arbitrary) |

Added a `container_ids` input mirroring `sdk_ids` across the bundle and the three cadence wrappers. Container-only and `custom` subsets are supported end to end: the >=1-SDK rule in `write_release_bundle_metadata.py` is dropped, `build-sdks` is skipped and `assemble`/the release upload handle the no-wheels case. The consumer stages exactly the selected set (NVIDIA-NeMo/Platform-Deploy#95). So a single-container patch release (`release_scope: custom`, `container_ids: nmp-automodel-tasks`) and SDK-only releases both work.

release/assets.yaml
a2bondar: Done — added `nmp-api`, `nmp-core`, `nmp-cpu-tasks` to the `container:` list. They are built per-merge by Platform-Deploy `docker.yaml` (`docker-cpu` target, pushed to the dev registry by commit SHA) and already have overview + catalog-metadata entries, so they stage and promote alongside the automodel images.

crookedstorm: Oof...actually, apparently that's wrong. The container list does not include api, core or cpu-tasks yet. It does include
- safe-synthesizer-tasks
- auditor-tasks

Are we publishing nmp-automodel-base? I thought that was just a build stage

.github/workflows/release-bundle.yaml
a2bondar: Scrubbed the internal release-consumer repo name from the public-repo additions (the plan-assets comment in `release-bundle.yaml`, the `release/assets.yaml` header comment, and the RELEASING.md eligibility section). A few pre-existing mentions remain elsewhere in RELEASING.md (predating this PR) — can clean those in a separate pass if you want.

release/assets.yaml
a2bondar: Trimmed the `container:` list to the three automodel runtimes (`nmp-automodel-tasks`, `nmp-automodel-training`, `nmp-unsloth-training`). Dropped `nmp-automodel-base` (builder image; main excluded it) and `nmp-api`/`nmp-core`/`nmp-cpu-tasks` per your follow-up — matches main’s `available: true` set.

crookedstorm: We must have safe-synthesizer-tasks and auditor-tasks as well.

.github/scripts/write_release_bundle_metadata.py
```diff
@@ -1,7 +1,12 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Write release bundle metadata for downloaded SDK wheel artifacts."""
+"""Write release bundle metadata for downloaded SDK wheel artifacts.
+
+Container artifacts are metadata-only manifest entries: the image bits are
+built and staged by Platform-Deploy from its dev registry at the bundle's
```
svvarom: I think we can also rmv the internal repo name from here

.github/workflows/release-bundle.yaml
```diff
@@ -411,6 +454,9 @@ jobs:
   assemble-release-bundle:
     name: Assemble release bundle
     needs: [plan-release, build-sdks]
```
svvarom: Should this wait for `reserve-release-tag` too? If tag creation fails, I don’t think we should still build and dispatch the release bundle.

release/assets.yaml
a2bondar: added!

.github/workflows/release-bundle.yaml
a2bondar: Fixed in `442e2e300`. `assemble-release-bundle` now `needs` (and waits on) `reserve-release-tag` and requires `needs.reserve-release-tag.result == 'success'`. Previously it only needed `[plan-release, build-sdks]` and tolerated a skipped `build-sdks` (the container-only case), but `build-sdks` is also skipped when `reserve-release-tag` fails (its dependency failed), so a failed tag reservation would still assemble and dispatch the bundle. `dispatch-release-bundle` cascades from `assemble` (no `always()`), so it halts too. `reserve-release-tag` always runs (nightly exits 0 early), so nightly and container-only releases still proceed - only an actual tag-creation failure now blocks build/dispatch.

.github/workflows/release-bundle.yaml
a2bondar: Addressed by the `release_scope` work on this PR. It now supports `all` / `sdks` / `containers` / `custom`; `custom` takes comma-separated `sdk_ids` + `container_ids` (either may be empty), so a single-container or arbitrary-subset release is `release_scope: custom` with e.g. `container_ids: nmp-automodel-tasks` and no `sdk_ids`. Release-specific selection is preserved for both SDKs and containers (see the RELEASING.md "Container image eligibility" table).

.github/scripts/write_release_bundle_metadata.py
a2bondar: Done in `956729111` - the docstring now says "built and staged by the release consumer" rather than naming the internal repo. Confirmed no other internal-repo references remain in this PR's workflow or script files.

svvarom (approved):

PR #319, opened 2026-06-12
feat: prepare helm chart for kubernetes deployment options

crookedstorm:
## Summary by CodeRabbit

* **New Features**
  * Full Helm chart for the platform, including ingress/HTTPRoute/OpenShift Route, optional Envoy proxy, PostgreSQL options, and Kyverno-based multi-node GPU networking.
  * Added an NCCL multi-node allreduce test suite (orchestrated for CI validation).
* **Documentation**
  * Expanded Helm README and richer post-install instructions, including encryption-key guidance.
* **Bug Fixes**
  * Improved API environment secret rendering and upgrade-time encryption-key safeguards; enhanced image-pull secret handling.
* **Chores**
  * CI now detects Helm changes and runs dedicated Helm lint/verification jobs; lint/build tooling and Helm packaging defaults updated.


svvarom (approved):

k8s/helm/files/nccl-test/entrypoint.sh
```diff
@@ -0,0 +1,147 @@
...
+NUM_NODES="${NUM_NODES:-1}"
+NPROC_PER_NODE="${NPROC_PER_NODE:-1}"
+NODE_RANK="${NODE_RANK:-0}"
+
+if [ "${NODE_RANK}" = "0" ]; then
+  echo "NODE_RANK 0 (leader node): skipping wait for LEADER_ADDR DNS ($LEADER_ADDR)"
+else
+  echo "Waiting for $LEADER_ADDR to be available"
+  while true; do
+    if getent hosts "$LEADER_ADDR" >/dev/null 2>&1; then
```
svvarom, line 100: Posting this from Codex, so please take it with a grain of salt: I think this can hang when `LEADER_ADDR` is a pod IP. The orchestrator passes the leader pod IP to followers, then this loop waits on `getent hosts "$LEADER_ADDR"`, which depends on DNS resolving that IP.

Some clusters do not provide reverse DNS for pod IPs, so followers could loop here before torch starts even if the IP is reachable. Could we skip this DNS wait for IP literals, or use a bounded TCP check against `$LEADER_ADDR:$MASTER_PORT` instead?

svvarom: Codex wrote this and marked it as a P1

crookedstorm: The nccl test is an optional thing in general in case someone really wants to try an allreduce in every node in their cluster without using other examples of doing this on the web or using any other form of telemetry to get similar info. I'm ok with it.

crookedstorm: I will take a look before I merge though :)

k8s/helm/templates/core/controller-role.yaml
```diff
@@ -0,0 +1,67 @@
...
+  verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources: ["persistentvolumeclaims"]
+  verbs: ["get", "list", "create", "delete"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "create", "delete"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["create", "delete"]
```
mquong-ca, line 33: Do we care that it has the role to create or delete any secret and not scoped to specific secrets that matters to the chart?

crookedstorm: That's essential to their secrets controller in general. It cannot be scoped because of how it works.

crookedstorm: It can create arbitrary secrets and inject them into pods.

k8s/helm/helm-docs-template/nemo-helm-readme.md.gotmpl
```diff
@@ -0,0 +1,25 @@
+# {{ template "chart.description" . }}
+
+{{ template "chart.typeBadge" . }}
+
+Documentation can be found at: https://docs.nvidia.com/nemo/microservices/.
```
svvarom: Is this link correct? Do we want to change it to https://docs.nvidia.com/nemo-platform?

crookedstorm: Good catch! Definitely.

k8s/helm/templates/ngc-api-secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ngc-api
+  labels:
+    {{- include "nemo-platform.labels" . | nindent 4 }}
+type: Opaque
+data:
+  NGC_API_KEY: {{ .Values.ngcAPIKey | b64enc }}
```
svvarom, line 11: Do we want to add a guard at the top to check whether `ngcAPIKey` is set to `YOUR-NGC-API-KEY` or is empty? We might want to fail fast instead of deploying the platform with invalid credentials

matthewgrossman (commented):
Two thoughts from exploring the PR w/ an agent:

1. The README has good guidance about restoring the key on upgrade, but `NOTES.txt` is what users see right after `helm install`. A one-liner like "Back up the `<fullname>-api-env` Secret — if lost, existing encrypted platform secrets become unrecoverable" could be nice

2. (moreso a question) The generator Job defaults to `docker.io/library/python:3.12-slim`. Is this "normal"/okay for our charts these days, as opposed to using a mirror? I know its configurable, so we should in our own ToT, but wondering if using the docker.io image versus something else is standard (versus using I guess our nmp-api image?)

k8s/helm/templates/ngc-api-secret.yaml
svvarom: Punting this for later - https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExOHQ5Y2o2OHhjaWZoemFwOWZiNWN6cjN5Y20wZW94MmNubXcxaThlNCZlcD12MV9naWZzX3NlYXJjaCZjdD1n/S5bhjdodaDGqjXkYA6/giphy.gif

svvarom (approved):
LGTM

matthewgrossman (approved):
ty!

crookedstorm:
> Two thoughts from exploring the PR w/ an agent:
> 
> 1. The README has good guidance about restoring the key on upgrade, but `NOTES.txt` is what users see right after `helm install`. A one-liner like "Back up the `<fullname>-api-env` Secret — if lost, existing encrypted platform secrets become unrecoverable" could be nice
> 2. (moreso a question) The generator Job defaults to `docker.io/library/python:3.12-slim`. Is this "normal"/okay for our charts these days, as opposed to using a mirror? I know its configurable, so we should in our own ToT, but wondering if using the docker.io image versus something else is standard (versus using I guess our nmp-api image?)

We should never default to anything mirrored or internal anymore. All default containers should be public (or aspirationally public such as the ones this repo will create.  I'll poke the NOTES.txt :)

PR #327, opened 2026-06-12
fix(studio): Align on a single test suffix pattern

steramae-nvidia:
long live `.test.ts`, death to `.spec.ts`

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for search utility functions with additional test cases for parameter handling and merging.
  * Standardized test file naming convention to `.test.*` pattern across the codebase.

* **Documentation**
  * Updated testing guidelines to reflect new test file naming conventions.

* **Chores**
  * Updated ESLint configuration to enforce standardized test filename patterns.


web/eslint-plugins/test-naming.js
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/** @type {import('eslint').Rule.RuleModule} */
+const noSpecSuffix = {
```
aray12: can we just use the official plugin we already have installed

https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/consistent-test-filename.md

web/packages/studio/e2e-tests/projects/datasets.test.ts
```diff
@@ -46,7 +46,7 @@ const test = baseTest.extend<TestFixtures>({
   },
   testProject: async ({ request }, runFixture) => {
     const projectDisplayName = generateTestResourceName('project');
-    const projectDescription = `Project created by datasets.spec.ts E2E test on ${CURRENT_YYYY_MM_DD}`;
+    const projectDescription = `Project created by datasets.test.ts E2E test on ${CURRENT_YYYY_MM_DD}`;
```
aray12, line 49: do you think it might make sense to keep `.spec.ts` for e2e tests and reserve `.test.ts` for unit tests?

marcusds (approved):

aray12 (approved):
a couple of thoughts but thanks for doing this. I have been wanting to do the same

web/eslint-plugins/test-naming.js
steramae-nvidia: oh cool, didn't know about this. I'll update

web/packages/studio/e2e-tests/projects/datasets.test.ts
steramae-nvidia: Is there a reason for making that other than distinction of test type? I think keeping e2e in the e2e folder should suffice for that

aray12: Potentially different lint rules or whatever we want. The thought was basically just that it is two different techs that consume them (playwright vs vitest). Not blocking but just an idea

steramae-nvidia: I think its still more confusing to keep two different test suffixes so let's see how it shapes up just as a .test suffix within the e2e folder.

PR #328, opened 2026-06-15
fix: bearer token validate when auth enabled, resolve nat_runner

mckornfield:
## Release Notes

* **Bug Fixes**
  * Improved bearer token authentication handling when authentication is disabled to prevent configuration errors.
  * Fixed default model placeholder resolution in agent configurations to use host context defaults.

### DIFF
--- a/packages/nmp_common/src/nmp/common/auth/middleware.py
+++ b/packages/nmp_common/src/nmp/common/auth/middleware.py
@@ -12,6 +12,7 @@ from nmp.common.config import AuthConfig, get_auth_config
 from nmp.common.observability.context import get_app_ctx
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.responses import JSONResponse
+from starlette.types import ASGIApp
 
 from .client import AuthClient
 from .dependencies import auth_client_context
@@ -19,6 +20,17 @@ from .exceptions import InvalidPrincipalHeader, InvalidScopeFormatError
 from .models import Principal
 
 logger = logging.getLogger(__name__)
+
+
+class PrincipalExtractionError(RuntimeError):
+    """Raised when principal extraction returns an impossible success state."""
+
+
+def _require_principal(principal: Principal | None) -> Principal:
+    """Return a principal from a successful extraction result or fail loudly."""
+    if principal is None:
+        raise PrincipalExtractionError("principal extraction succeeded without returning a principal")
+    return principal
 
 
 def _describe_pdp_failure(exc: BaseException) -> str:
@@ -92,7 +104,7 @@ class AuthorizationMiddleware(BaseHTTPMiddleware):
 
     def __init__(
         self,
-        app,
+        app: ASGIApp,
         service_name: Optional[str] = None,
         http_client: Optional[httpx.AsyncClient] = None,
     ):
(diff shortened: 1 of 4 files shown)


packages/nmp_common/tests/auth/test_middleware.py
```diff
@@ -85,7 +86,7 @@ async def hf_download_endpoint(workspace: str, name: str, revision: str, path: s
     # Override config
     Configuration.set_override(auth_config)
 
-    app.add_middleware(AuthorizationMiddleware, service_name="test-service")
+    app.add_middleware(cast(Any, AuthorizationMiddleware), service_name="test-service")
```
mikeknep: Why the casts? Some `ty` issue?

packages/nmp_common/src/nmp/common/auth/middleware.py
```diff
@@ -309,6 +308,7 @@ async def _handle_service_principal_request(
         principal, error_response = self._principal_from_headers(headers_dict)
         if error_response is not None:
             return error_response
+        assert principal is not None
```
mikeknep: Using `assert` here seems a little odd to me, feels like raising a well-named error would be better?

packages/nmp_common/src/nmp/common/auth/middleware.py
```diff
@@ -239,15 +239,14 @@ async def dispatch(self, request: Request, call_next: Callable) -> Response:
 
         # Try to extract principal from Authorization: Bearer header (native OIDC or unsigned JWT)
         auth_header = headers_dict.get("authorization", "")
-        if auth_header.lower().startswith("bearer "):
+        if auth_header.lower().startswith("bearer ") and self.config.enabled:
```
mikeknep: nit: maybe lift this block (from just below the diff)...
```python
if not self.config.enabled:
    return await self._handle_auth_disabled_request(request, call_next)
```
...to instead be right _above_ this diff. Allows you to drop the `and ...` you added here, while also keeping the removal of the `if self.config.enabled` check for OIDC not configured -> 401

anastasia-nesterenko (commented):
could it be useful to add a test for enabled=True + oidc.enabled=False + bearer token → 401 ?

packages/nmp_common/tests/auth/test_middleware.py
mckornfield: yeah, if I had to guess. it might just be the coding agent as well not having the right context

mckornfield:
supposedly an existing test `test_bearer_token_oidc_not_configured in packages/nmp_common/tests/auth/test_middleware.py:153.` covers that case anastasia, https://github.com/NVIDIA-NeMo/nemo-platform/pull/328/changes#diff-d5507ae58ac1922cd8021d5edafdd2c0c2ab8505c5f563eaebb4971b26e159d3R155 but I stuck another assertion there

packages/nmp_common/src/nmp/common/auth/middleware.py
```diff
@@ -309,6 +320,7 @@ async def _handle_service_principal_request(
         principal, error_response = self._principal_from_headers(headers_dict)
         if error_response is not None:
             return error_response
+        principal = _require_principal(principal)
```
mikeknep, line 323: It looks like `self._principal_from_headers` should never return `(None, None)` and therefore this shouldn't ever get hit, but 🤷

mikeknep (approved):

PR #331, opened 2026-06-15
fix: make opa local setup more resilient

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Embedded authentication PDP can now automatically build policy artifacts from source checkouts when missing or outdated.
  * Added `embedded_pdp_auto_build_wasm` configuration option (defaults to enabled) to control auto-build behavior.

* **Documentation**
  * Updated policy engine and deployment documentation with embedded PDP build/runtime details.
  * Added configuration reference for new auto-build setting.

* **Chores**
  * Removed manual OPA installation step from GitHub Actions workflow.


docs/auth/authorization/policy-engine.mdx
```diff
@@ -24,16 +24,23 @@ The embedded PDP uses a WASM-compiled Rego policy engine built into the auth ser
 
 ### How It Works
 
+- Rego policy files are compiled to `policy.wasm` with `make build-policy`
```
drazvan: Maybe move these details to a separate section in the doc focused on the embedded_pdp? Maybe documented next to `embedded_pdp_cpu_limit` and `embedded_pdp_memory_limit_mb`. 
Initially, this doc page was written to be user-facing, so the reference to `script/build_policy_wasm.sh` wouldn't quite fit here. But not 100% sure.

packages/nmp_platform_runner/src/nmp/platform_runner/run.py
```diff
@@ -52,6 +53,25 @@ def _database_display(db_url: str) -> str:
     return db_type
 
 
+def _is_policy_wasm_error(error: Exception) -> bool:
+    return (
+        error.__class__.__name__ == "PolicyWasmError"
+        and error.__class__.__module__ == "nmp.core.auth.app.embedded_pdp.policy_wasm"
```
drazvan: This creates a coupling which feels hack-ish. Is there a more generic way to achieve this? I assume we want to emphasize this type of error.

services/core/auth/src/nmp/core/auth/app/embedded_pdp/policy_wasm.py
```diff
@@ -0,0 +1,196 @@
...
+    """Return True when policy.wasm is missing or older than source policy files."""
+    if not wasm_path.exists():
+        return True
+
+    sources = _source_files(policy_dir, repo_root)
+    if not sources:
+        return False
+
+    wasm_mtime = wasm_path.stat().st_mtime_ns
+    return any(source.stat().st_mtime_ns > wasm_mtime for source in sources)
```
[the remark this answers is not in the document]
drazvan: Agree.

drazvan (commented):
Overall looks good. The comment from code rabbit regarding the check if the wasm bundle is stale, relying on mtime is valid. Should we maybe include a suffix with a hash in the name of the bundle? This will require some changes in how the wasm bundle is loaded as well. Alternatively, we could drop the "rebuild on stale" requirement and only build if missing, which solves at least one friction point.

services/core/auth/src/nmp/core/auth/app/embedded_pdp/policy_wasm.py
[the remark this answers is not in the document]
mckornfield: I believe I chucked most of the rebuild stuff based on time, this just builds it if it isn't built at this point

docs/auth/authorization/policy-engine.mdx
mckornfield: moved

packages/nmp_platform_runner/src/nmp/platform_runner/run.py
mckornfield: yeah removed this

drazvan:
LGTM :+1:

drazvan (approved):

svvarom (approved):
LGTM

PR #332, opened 2026-06-15
refactor(authz): unify plugin HTTP authz under NemoService

maxdubrinsky:
## Summary

Plugin HTTP authorization is currently split between two mechanisms: the `nemo.authz` / `get_authz_contribution()` surface and the service routers. This change makes `NemoService` the single source of plugin auth policy. A plugin attaches authorization rules directly to its route handlers, and the platform derives the policy from the mounted routes when it builds the OPA bundle. There is no separate permission list to keep in sync.

The model is fail-closed. Any route the platform cannot confidently authorize becomes an explicit deny, so a missing or malformed rule cannot fall through to an allow.

## How it works

- A plugin declares permissions as typed objects (each carries its own id and a human-readable description) and puts a `@path_rule(...)` on each route handler listing the caller kinds and permissions that route requires.
- At bundle build the platform walks the final mounted route set and derives the permission catalog and per-endpoint policy from those rules. Standard job routes (create, list, get, cancel, logs, results) are stamped automatically by the job-route factory, so a plugin passes its namespace once instead of hand-listing every endpoint.
- Bundle build then validates the result and fails closed if a service references an undeclared permission, declares one outside its own namespace, emits a malformed id, or leaves any route with no rule. How much gets denied is configurable via `auth.on_invalid_plugin` (default: deny only the offending routes).

## Behavior changes worth checking during review

- Caller kind is enforced per route. A route can require a human principal, a service principal, or both. Humans can no longer reach service-only routes and services cannot reach human-only routes.
- Unknown or unruled routes are denied instead of allowed. A route mounted before its authz is in the bundle no longer falls through to the authenticated-user or service bypass.
- List routes that have no workspace in their path (for example, list workspaces) now require their permission in the `system` workspace rather than only authentication. This is not a regression on a default deployment: the platform seeds a wildcard Viewer binding in `system` for all authenticated users, so everyone already holds it. It only changes behavior for deployments that do not have that grant.
- All seven plugin services and the two customization contributors are re-annotated to the new model. No endpoints change. This is an authz-metadata refactor, not an API change.

## Performance

The new rules run on every request and pushed policy evaluation cost up enough to matter for the embedded engine. The per-request endpoint and workspace lookups are now computed once per evaluation instead of repeated at each call site, which brings cost back well under budget.

## How to verify

- Policy tests (238 pass): `opa test services/core/auth/src/nmp/core/auth/app/policies services/core/auth/src/nmp/core/auth/app/policy_tests services/core/auth/src/nmp/core/auth/assets/static-authz.yaml`. The `static-authz.yaml` argument is required; without it the inference-gateway cases fail because they read endpoint data from that file.
- Python: each plugin has a `test_authz.py` asserting it derives a valid policy with every route ruled and every permission inside its own namespace.
- Optional, not in CI: `e2e/authz_oidc` runs the policy against a live platform using real signed JWTs from a test OIDC issuer (`pytest e2e/authz_oidc --run-e2e`).

## Commits

1. `refactor`: the model, bundle validation, and the plugin migrations.
2. `perf`: the policy-engine evaluation-cost fix.
3. `test`: the end-to-end OIDC harness.

## Out of scope

Authentication and machine identity are unchanged. Plugin-defined roles, plugin-defined scopes, and IAM/UI surfacing are not part of this change.

## Summary by CodeRabbit

* **New Features**
  * Route-level authorization can now be declared directly on endpoints, including explicit caller-kind plus permission and scope requirements.
  * Added tighter authorization coverage across platform plugins and generated job/function routes.

* **Bug Fixes & Improvements**
  * Stronger fail-closed handling for invalid plugin authorization, including quarantine/fencing for unenumerated routes.
  * Workspace listing now requires system-level permissions.

* **Configuration**
  * Added `auth.on_invalid_plugin` to control fail-mode (`deny_route`, `quarantine`, `hard_fail`).
  * Added `auth.platform_admin_exempt_from_service_only` for PlatformAdmin exemptions.

* **Documentation & Testing**
  * Updated configuration reference and refreshed the OIDC authorization end-to-end test harness/reports.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
@@ -92,8 +92,10 @@ def _collapse_rules(
     Returns ``(permissions, callers)`` for the representative rule.
     """
     # Only callers are OR'd across rules (unioned below); their permissions must match, since
-    # the wire format holds one permissions list per (path, method). Reject a mismatch rather
-    # than silently picking one rule's.
+    # the wire format holds one permissions list per (path, method). Compare by Permission.id —
+    # the ``service.resource.action`` wire key joined from each Permission's structured parts —
+    # not by object, so a description-only difference isn't read as a distinct set. Reject a
+    # mismatch rather than silently picking one rule's.
     distinct_permission_sets = {frozenset(p.id for p in rule.permissions) for rule in rules}
     if len(distinct_permission_sets) > 1:
         raise ValueError(
@@ -317,7 +319,7 @@ def _derive_service_contribution(service: NemoService) -> tuple[AuthzContributio
     # cross-namespace can reach the merged policy.
     owner = service.name
     malformed = sorted(pid for pid in catalog if not is_valid_permission_id(pid))
-    out_of_namespace = sorted(pid for pid in catalog if pid.split(".", 1)[0] != owner)
+    out_of_namespace = sorted(p.id for p in catalog.values() if p.service != owner)
     if malformed:
         errors.append(f"malformed permission id(s) (fail-closed): {malformed}")
     if out_of_namespace:
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
@@ -40,6 +40,7 @@ from nemo_platform_plugin.authz import (
     PathRule,
     Permission,
     get_path_rules,
+    get_path_scope,
 )
 from nemo_platform_plugin.authz_format import is_valid_permission_id
 from nemo_platform_plugin.service import NemoService
@@ -75,20 +76,24 @@ def _wire_callers(rules: list[PathRule]) -> list[str] | None:
 
 def _collapse_rules(
     rules: list[PathRule], *, path: str, method: str, service: str
-) -> tuple[list[Permission], list[str] | None, list[str] | None]:
+) -> tuple[list[Permission], list[str] | None]:
     """Collapse the (OR'd) ``PathRule``\\ s on one ``(path, method)`` into one binding.
 
     v1 supports OR across rules only in the **caller** dimension: caller kinds are unioned,
-    but ``permissions`` and ``scopes`` must agree across rules. The single-slot wire format
-    (one AND'd ``permissions`` list per method) and the Rego permission check cannot
-    represent an OR of *distinct* permission sets, so that case is rejected loudly rather
-    than silently mis-authorized.
+    but ``permissions`` must agree across rules. The single-slot wire format (one AND'd
+    ``permissions`` list per method) and the Rego permission check cannot represent an OR of
+    *distinct* permission sets, so that case is rejected loudly rather than silently
+    mis-authorized.
 
-    Returns ``(permissions, scopes, callers)`` for the representative rule.
+    Scope is not a rule dimension — it is declared once per route with ``@AuthzScope.read`` /
+    ``.write`` and read separately via :func:`~nemo_platform_plugin.authz.get_path_scope`, so
+    there is nothing to reconcile across rules here.
+
+    Returns ``(permissions, callers)`` for the representative rule.
     """
-    # Only callers are OR'd across rules (unioned below); their permissions and scopes must
-    # match, since the wire format holds one permissions/scopes list per (path, method). Reject
-    # a mismatch rather than silently picking one rule's.
+    # Only callers are OR'd across rules (unioned below); their permissions must match, since
+    # the wire format holds one permissions list per (path, method). Reject a mismatch rather
+    # than silently picking one rule's.
     distinct_permission_sets = {frozenset(p.id for p in rule.permissions) for rule in rules}
     if len(distinct_permission_sets) > 1:
         raise ValueError(
@@ -97,19 +102,9 @@ def _collapse_rules(
             f"distinct permission sets — use one rule with shared permissions, or a single "
             f"rule listing multiple callers."
         )
-    distinct_scope_sets = {None if rule.scopes is None else frozenset(rule.scopes) for rule in rules}
-    if len(distinct_scope_sets) > 1:
-        raise ValueError(
-            f"{service}: {method.upper()} {path} has @path_rule rules with differing scopes — "
-            f"all rules on one endpoint must declare the same scopes."
-        )
 
     representative = rules[0]
-    return (
-        list(representative.permissions),
-        list(representative.scopes) if representative.scopes is not None else None,
-        _wire_callers(rules),
-    )
+    return list(representative.permissions), _wire_callers(rules)
 
 
 @dataclass
@@ -268,7 +263,7 @@ def _derive_service_contribution(service: NemoService) -> tuple[AuthzContributio
             errors.append(f"{route.path} ({', '.join(methods) or 'no methods'}) has no @path_rule")
         else:
             try:
-                permissions, scopes, callers = _collapse_rules(
+                permissions, callers = _collapse_rules(
                     rules, path=route.path, method=methods[0] if methods else "", service=service.name
                 )
             except (ValueError, AttributeError, TypeError) as exc:
@@ -281,6 +276,9 @@ def _derive_service_contribution(service: NemoService) -> tuple[AuthzContributio
             else:
                 for perm in permissions:
                     _register_permission(catalog, perm, warnings)
+                # Scope is declared on the route by @AuthzScope.read/.write, independent of the
+                # permission rules, and read straight off the (identity-preserved) endpoint.
+                scopes = get_path_scope(route.endpoint)
                 binding = AuthzEndpointMethod(
                     permissions=[perm.id for perm in permissions], scopes=scopes, callers=callers
                 )
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
@@ -311,12 +311,33 @@ def _derive_service_contribution(service: NemoService) -> tuple[AuthzContributio
     for perm in extra:
         _register_permission(catalog, perm, warnings)
 
+    # Extra role grants: permissions the suffix heuristic (.list/.read → Viewer+Editor, else
+    # Editor) wouldn't assign to a role — e.g. the gateway .invoke permission a Viewer needs.
+    # Register each in the catalog (so it carries a registry entry + description) and thread
+    # role -> [id] into the contribution; merge_authz_contributions unions these with the
+    # suffix-derived defaults. A broken hook must NOT abort derivation (that would drop the
+    # route bindings and let them fall through the service: bypass), so it's recorded, not raised.
+    role_grants: dict[str, list[str]] = {}
+    try:
+        extra_roles = service.extra_role_permissions()
+    except Exception as exc:
+        extra_roles = {}
+        errors.append(f"extra_role_permissions() raised {exc!r}")
+    for role_name, perms in extra_roles.items():
+        for perm in perms:
+            _register_permission(catalog, perm, warnings)
+        granted = sorted({perm.id for perm in perms})
+        if granted:
+            role_grants[role_name] = granted
+
     # Pass 2: validate the catalog. A malformed permission id would 500 the bundle's
     # ``validate_static_authz_data`` if it reached the wire; a permission whose first segment
     # isn't the service's own name is namespace squatting (it would silently widen the
     # Viewer/Editor role grants for another service's namespace). Either is a fail-closed
     # error: deny every route and contribute no permissions, so nothing malformed or
     # cross-namespace can reach the merged policy.
+    # Role-granted permissions are registered in the catalog above, so the ownership fence below
+    # covers them too: a plugin cannot grant a role a permission outside its own namespace.
     owner = service.name
     malformed = sorted(pid for pid in catalog if not is_valid_permission_id(pid))
     out_of_namespace = sorted(p.id for p in catalog.values() if p.service != owner)
@@ -333,7 +354,11 @@ def _derive_service_contribution(service: NemoService) -> tuple[AuthzContributio
         for path, methods in bindings.items()
     }
     permissions = {perm.id: perm.description for perm in catalog.values()}
-    return AuthzContribution(permissions=permissions, endpoints=endpoints), errors, warnings
+    return (
+        AuthzContribution(permissions=permissions, endpoints=endpoints, role_permissions=role_grants),
+        errors,
+        warnings,
+    )
 
 
 _plugin_authz_cache: list[PluginAuthzResult] | None = None
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
@@ -1,154 +1,396 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Discover plugin authorization contributions for policy merge."""
+"""Derive plugin authorization contributions from the ``NemoService`` route surface.
+
+Plugins attach :func:`~nemo_platform_plugin.authz.path_rule` rules to route handlers,
+referencing :class:`~nemo_platform_plugin.authz.Permission` constants. This module
+instantiates each discovered ``NemoService``, walks its mounted routes — computing the
+same ``/apis/<name>/<prefix>`` paths the platform mounts at runtime — reads the
+function-attached :class:`~nemo_platform_plugin.authz.PathRule`\\ s, and builds the
+wire-format :class:`~nemo_platform_plugin.authz.AuthzContribution` consumed by the OPA
+bundle builder and ``auth-tools sync-plugins``.
+
+The permission catalog (ids + descriptions) and the service namespace are derived
+*entirely from the routes* (plus the optional :meth:`NemoService.extra_permissions`
+hatch). There is no separately-declared permission list: the permission is the object
+referenced on the route, and it carries its own description.
+
+Path composition mirrors production: the platform runner mounts each service app at
+``/apis/<service.name>`` and the service app includes each ``RouterSpec`` router at its
+``prefix`` (see ``nmp.platform_runner.server`` and ``nmp.common.service.base``). We
+re-create that composition with a throwaway router so FastAPI computes the final paths —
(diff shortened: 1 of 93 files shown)


packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,378 @@
...
+    owner = service.name
+    malformed = sorted(pid for pid in catalog if not is_valid_permission_id(pid))
+    out_of_namespace = sorted(pid for pid in catalog if pid.split(".", 1)[0] != owner)
+    if malformed:
+        errors.append(f"malformed permission id(s) (fail-closed): {malformed}")
+    if out_of_namespace:
+        errors.append(f"permission id(s) outside the service namespace {owner!r} (fail-closed): {out_of_namespace}")
+    if malformed or out_of_namespace:
+        denied = {path: {method: _deny_binding() for method in methods} for path, methods in bindings.items()}
+        return AuthzContribution(permissions={}, endpoints=denied), errors, warnings
```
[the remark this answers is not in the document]
maxdubrinsky: Not actioning. A non-`Permission` does raise here, but it's caught per-plugin in `discover_plugin_authz` and the entire `/apis/<name>` namespace gets fenced — strictly fail-closed, and *broader* than the per-route deny suggested here (which would not cover a runner-mounted-but-derivation-unseen route). The current behavior is the safer outcome, so leaving it.

services/core/auth/src/nmp/core/auth/app/policy_tests/caller_kind_test.rego
```diff
@@ -0,0 +1,308 @@
+package authz_test
```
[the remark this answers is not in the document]
maxdubrinsky: Not actioning. Regal isn't wired into our CI or pre-commit, and 18 sibling files in this directory already use `package authz_test`. Renaming just this one would break that established convention rather than fix anything enforced.

services/core/auth/tests/test_embedded_pdp.py
```diff
@@ -792,3 +795,36 @@ def test_audit_workspace_endpoints_have_audit_scopes(self, static_authz_data):
...
+        static_authz_data["authz"].setdefault("config", {})["denied_plugin_prefixes"] = ["/apis/brokenplugin"]
+        set_policy_data(static_authz_data)
+        cases = [
+            ("/apis/brokenplugin/sub/route", False),  # subpath fenced (the bug: this was allowed)
+            ("/apis/brokenplugin", False),  # bare prefix fenced
+            ("/apis/brokenplugin-extra/x", True),  # sibling prefix not collaterally fenced
+        ]
+        for path, expect in cases:
+            result = evaluate("allow", {"principal_id": "service:probe", "method": "GET", "path": path})
+            assert result["allowed"] is expect, f"GET {path} as service:probe: expected allowed={expect}"
```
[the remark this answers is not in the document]
maxdubrinsky: Not a bug — this assertion is intentional. The test verifies the `/apis/brokenplugin` fence does **not** collaterally match the sibling prefix `/apis/brokenplugin-extra`; the `True` is the pre-existing service-principal reach on a non-fenced path, not a fail-closed violation. Confirmed the test passes as written.

packages/nemo_platform_plugin/src/nemo_platform_plugin/functions/routes.py
```diff
@@ -103,6 +103,10 @@ def add_function_routes(
     function_cls: type[NemoFunction],
     *,
     heartbeat_interval_seconds: float = HEARTBEAT_INTERVAL_SECONDS,
+    permission_namespace: str | None = None,
+    api_area: str | None = None,
+    permission_id: str | None = None,
+    permission_description: str | None = None,
```
mikeknep: nit: It might be worth bundling these into a single, optional object? e.g. `authz_details: RouteAuthzDetails | None = None` (hopefully can find a better name). Within that object, permission namespace and api area can both be required strings, effectively pushing the "both set or both omitted" to the type-checker—either they are both set and passed via the object, or the object is not provided at all. We could probably add some methods on that object too to handle resolving permission_id and/or perhaps even creating the `Permission` object (lower in the diff in this file)

maxdubrinsky: Yeah that makes sense, would also open up making it reusable between function and job routes. Will noodle and report back with a change.

maxdubrinsky: Added an `AuthzScope` class that you pass in to both generators, reduces the number of places you need to define the scope.

mikeknep: Nice, the `AuthzScope` change looks solid to me, thanks. I'd defer to @ironcommit for overall approval since I haven't followed auth details too closely.

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,378 @@
...
+    if len(perm_sets) > 1:
+        raise ValueError(
+            f"{service}: {method.upper()} {path} has @path_rule rules with differing "
+            f"permissions ({[sorted(p) for p in perm_sets]}). v1 cannot represent an OR of "
+            f"distinct permission sets — use one rule with shared permissions, or a single "
+            f"rule listing multiple callers."
+        )
+    scope_sets = {None if rule.scopes is None else frozenset(rule.scopes) for rule in rules}
+    if len(scope_sets) > 1:
+        raise ValueError(
```
ironcommit: I wonder if we want to declare the scope for the endpoint as a separate decorator, and ensure there's only one, and then we can make this error impossible.  Does that make sense?

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,378 @@
...
+
+    v1 supports OR across rules only in the **caller** dimension: caller kinds are unioned,
+    but ``permissions`` and ``scopes`` must agree across rules. The single-slot wire format
+    (one AND'd ``permissions`` list per method) and the Rego permission check cannot
+    represent an OR of *distinct* permission sets, so that case is rejected loudly rather
+    than silently mis-authorized.
+
+    Returns ``(permissions, scopes, callers)`` for the representative rule.
+    """
+    perm_sets = {frozenset(p.id for p in rule.permissions) for rule in rules}
```
ironcommit: I'm expecting us to group by caller here, but what it's actually doing is grouping by id.  Is that right?

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,378 @@
...
+    if len(perm_sets) > 1:
+        raise ValueError(
+            f"{service}: {method.upper()} {path} has @path_rule rules with differing "
+            f"permissions ({[sorted(p) for p in perm_sets]}). v1 cannot represent an OR of "
+            f"distinct permission sets — use one rule with shared permissions, or a single "
+            f"rule listing multiple callers."
+        )
+    scope_sets = {None if rule.scopes is None else frozenset(rule.scopes) for rule in rules}
```
ironcommit: same here, expecting to group by caller

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,378 @@
...
+    ``discover()`` swallows load failures and excludes the plugin entirely — exactly the
+    silent drop this fail-closed path must avoid.
+
+    Only an **all-clean** derivation is cached. A degraded result (e.g. a transient first-build
+    import error) is never pinned for the process lifetime — that would 403 the plugin's
+    namespace until restart — so the next call re-derives until the failure clears. Call
+    ``clear_plugin_authz_cache()`` (and ``discover_entry_points.cache_clear()``) in tests after
+    changing the installed plugin set.
+    """
+    global _plugin_authz_cache
```
ironcommit: maybe the cache can be fetched via a function?  That way the function doesn't need to know how the cache works.

ironcommit: this could be moved into a wrapper function for example.

services/core/auth/src/nmp/core/auth/app/policies/authz.rego
```diff
@@ -253,6 +314,71 @@ deny_request if {
...
+}
+
+# True only when a PlatformAdmin caller is present AND the config knob exempts platform
+# admins from service-only enforcement. Read defensively: an absent config key is treated
+# as false (default deny for human platform admins on service-only routes).
+default platform_admin_exempt := false
+
+platform_admin_exempt if {
+	platform_admin_in_system
+	data.authz.config.platform_admin_exempt_from_service_only == true
```
ironcommit: I get that we are being defensive here, but do we really need to make a distinction between service and non-service routes at all?  Shouldn't a PlatformAdmin just have access to everything always regardless if it's principal or service principal?  I don't totally understand why we need this distinction at all, and I would err on the side of just keeing things simple.

services/core/auth/src/nmp/core/auth/app/bundle.py
```diff
@@ -83,21 +83,113 @@ async def get_opa_bundle_with_etag(entities_client: Optional[EntityClient] = Non
...
+        if result.problems:
+            degraded[result.key] = result.problems
+            logger.error(
+                "Plugin %r contributed invalid authz (%d problem(s)); on_invalid_plugin=%s: %s",
+                result.key,
+                len(result.problems),
+                on_invalid,
+                "; ".join(result.problems),
+            )
+            if on_invalid == "hard_fail":
```
ironcommit, line 150: this "hard_fail" the default?  I'm expecting a fast fail on misconfiguration.

services/core/auth/src/nmp/core/auth/app/bundle.py
```diff
@@ -83,21 +83,113 @@ async def get_opa_bundle_with_etag(entities_client: Optional[EntityClient] = Non
...
+                raise RuntimeError(
+                    f"Plugin {result.key!r} contributed invalid authz and "
+                    f"authz.on_invalid_plugin=hard_fail: {'; '.join(result.problems)}"
+                )
+            # Fence the plugin's whole /apis/<name> namespace (deny-all) whenever per-route
+            # coverage can't be trusted:
+            #   * no route enumerated at all (load/derivation failure) — the runner may still
+            #     mount this plugin via a separate instantiation; OR
+            #   * quarantine — _quarantine_contribution only rewrites the routes derivation SAW,
```
ironcommit, line 159: not sure what SAW means here.  It's also not clear to me what the usecase is here for quarantine.    There might be a purpose for it, it's just not clear from the code yet.

services/core/auth/src/nmp/core/auth/config.py
```diff
@@ -58,6 +58,24 @@ class AuthServiceConfig(SharedAuthConfig):
         description="Maximum linear memory (MB) the embedded PDP WASM runtime can consume.",
     )
 
+    # Plugin HTTP authz fail-mode: what to do when a plugin contributes invalid authz
+    # (an unruled route, or a rule referencing an undeclared / out-of-namespace permission).
+    # The offending routes are always emitted as explicit denies; this controls the blast
+    # radius. deny_route: deny only the bad routes. quarantine: deny the whole plugin.
+    # hard_fail: refuse to build the OPA bundle.
+    on_invalid_plugin: Literal["deny_route", "quarantine", "hard_fail"] = Field(
```
ironcommit, line 69: I would argue for hard_fail as the default for misconfiguration.

ironcommit: Maybe consider just removing the config field all together, if everything is just fast fail on configuration error, then this simplifies.  I don't see a use case for this.

services/core/auth/src/nmp/core/auth/config.py
```diff
@@ -58,6 +58,24 @@ class AuthServiceConfig(SharedAuthConfig):
...
+        default="deny_route",
+        description="Fail-mode for a plugin that contributes invalid HTTP authz.",
+    )
+
+    # When true, a human PlatformAdmin is allowed on plugin routes restricted to
+    # SERVICE_PRINCIPAL callers. Default false (deny). Read by the Rego policy via
+    # data.authz.config.platform_admin_exempt_from_service_only.
+    platform_admin_exempt_from_service_only: bool = Field(
+        default=False,
+        description="Allow a human PlatformAdmin on SERVICE_PRINCIPAL-only plugin routes.",
```
ironcommit: I'm not sure if "human" is correct here.  this is really a question of if "principal" can access a route that is defined for "service_principal" only.  We don't really know if it's human or not, we only know "principal/service_pricipal".  I think principal means an OIDC user, and service principal roughly maps to either an API key, or other non-oidc access.  So the question is, should an OIDC user be able to access an endpoint that is restricted to service keys only, and I think the answer should be always no.  Happy to consider a use case where the answer might be yes.

ironcommit (commented):
Amazing work, super happy with how it came to gether.  Just a few questions and clarifications.

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
maxdubrinsky: Good call, it has been patched

services/core/auth/src/nmp/core/auth/app/policies/authz.rego
maxdubrinsky: Yeah I think this was me being preemptively configurable, it's also not super clear what access is expected for PlatformAdmins. To that end I'll just drop this config and grant admins access to all routes.

services/core/auth/src/nmp/core/auth/config.py
maxdubrinsky: This param came up several times as this got built, and what I landed on was configurability but setting the default to "hard_fail" probably makes the most sense. The scenario I had in my head was:

- user installs a plugin from a 3rd party
- that plugin doesn't implement auth and/or isn't meant to be used with auth
- "hard_fail" being the only option means that user either has to chose auth or the plugin

Happy to chat through this, but I do think this is a knob that we should consider having.

services/core/auth/src/nmp/core/auth/config.py
maxdubrinsky: Removed for now, can always add back later.

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,443 @@
...
+    (one AND'd ``permissions`` list per method) and the Rego permission check cannot
+    represent an OR of *distinct* permission sets, so that case is rejected loudly rather
+    than silently mis-authorized.
+
+    Returns ``(permissions, scopes, callers)`` for the representative rule.
+    """
+    # Only callers are OR'd across rules (unioned below); their permissions and scopes must
+    # match, since the wire format holds one permissions/scopes list per (path, method). Reject
+    # a mismatch rather than silently picking one rule's.
+    distinct_permission_sets = {frozenset(p.id for p in rule.permissions) for rule in rules}
```
ironcommit, line 99: I couldn't find the code that declares p.id, which I assume is the composite key, but it's not clear in this code that it's the case.

services/core/auth/src/nmp/core/auth/config.py
ironcommit: sounds good.

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
```diff
@@ -1,154 +1,468 @@
...
 
+    _cache_plugin_authz(results)
+    return results
 
-@cache
-def discover_authz_contributions() -> list[AuthzContribution]:
-    """Collect authz contributions from entry points and plugin surfaces.
 
-    Sources (in order):
+def clear_plugin_authz_cache() -> None:
```
ironcommit, line 453: if this is only called in tests, should we move it into test code?

plugins/example-plugin/src/nemo_example_plugin/middleware_service.py
```diff
@@ -86,6 +89,11 @@ def build_middleware_config_router() -> APIRouter:
         status_code=status.HTTP_201_CREATED,
         summary="Create ExampleMiddlewareConfig",
     )
+    @scope.write
```
ironcommit, line 92: I'm expecting the scope decorator to be something like `@scope("example:write")`, which includes the data domain being accessed as well as the modifier.  At a minimum, we need the domain scope defined likely per plugin.

plugins/nemo-deployments/src/nemo_deployments_plugin/api/v2/status.py
```diff
@@ -6,15 +6,20 @@
...
 from nemo_deployments_plugin.schema import UpdateDeploymentStatusRequest, UpdateVolumeStatusRequest
+from nemo_platform_plugin.authz import CallerKind, path_rule
 from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityConflictError, NemoEntityNotFoundError
 
 router = APIRouter()
 
 
 @router.put("/deployments/{name}/status", response_model=Deployment, tags=["Deployment Status"])
+@scope.write
+@path_rule(callers=[CallerKind.SERVICE_PRINCIPAL], permissions=[DeploymentPerms.STATUS_UPDATE])
```
ironcommit, line 22: why is this one service_principal and the other's "principal"?  I don't see why they would be different.

ironcommit (approved):
Just a few commands, but looking ready to go.

packages/nemo_platform_plugin/src/nemo_platform_plugin/authz_discovery.py
maxdubrinsky: Since it's referencing a global variable it'd be a bit awkward to move it out of here, I should be able to clean up some of its usage though.

plugins/example-plugin/src/nemo_example_plugin/middleware_service.py
maxdubrinsky: We do have scope-per-domain at this point, defined ideally in a plugin-specific `authz.py`. This generates the decorators for `read` and `write` scopes (including the associated `platform` scope). Do you think there will be plugin-specific scopes beyond `read` and `write` in the near term?

plugins/nemo-deployments/src/nemo_deployments_plugin/api/v2/status.py
maxdubrinsky: This is what we did for previous `PUT /status` endpoints, not for something (even principals to do). If we want to change this I'm not opposed though, not sure if we really need to restrict this.

PR #344, opened 2026-06-15
fix: adjust imagepullsecret handling

crookedstorm:
imagepullsecrets are not going to be required in the end, but they should always remain an option

## Summary by CodeRabbit

* **Configuration Changes**
  * Refactored image pull secret configuration: Removed legacy parameters and introduced a new `imagePullSecrets` configuration for specifying existing Kubernetes image pull secrets.
  * Updated NGC API key behavior to focus on NGC API authentication rather than automatic image pull secret generation.
  * Users must migrate existing configurations to the new image pull secret specification.


k8s/helm/values.yaml
svvarom, line 165: Do we need to propagate the new top-level `imagePullSecrets` value to the bundled `k8s-nim-operator` subchart?

```yaml
k8s-nim-operator:
  operator:
    image:
      pullSecrets:
        - private-registry
```

crookedstorm: We never did before?

crookedstorm: Nothing used by nim operator is private

k8s/helm/templates/tests/nccl-test.yaml
```diff
@@ -175,8 +176,10 @@ spec:
       restartPolicy: Never
       terminationGracePeriodSeconds: 0
       serviceAccountName: {{ printf "%s-sa" $ncclTestBase | trunc 63 }}
+{{- if $imagePullSecrets }}
       imagePullSecrets:
-        {{ include "nemo-common.imagepullsecrets" . | nindent 8 | trim }}
+{{- $imagePullSecrets | nindent 8 }}
+{{- end }}
```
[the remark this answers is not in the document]
crookedstorm: https://helm.sh/docs/chart_template_guide/control_structures/#ifelse an empty array is falsey.

svvarom (approved):

mquong-ca (commented):
lgtm

PR #348, opened 2026-06-16
feat(studio): row pinning for ExperimentGroupDataView

nakolean:
https://github.com/user-attachments/assets/8daf3c58-e1ae-435f-be65-42d6ae836f30

## Summary by CodeRabbit

**New Features**
* Added experiment pinning in the Experiment Group table. A new “Pinned” column lets users pin or unpin experiments.
* Pinned experiments are shown first while keeping the existing sort order for both pinned and unpinned items.
* Pin preferences persist across sessions, so your pinned selection is retained for the same workspace and group.

**Bug Fixes**
* None reported.


walston (approved):

shanaiabuggy (approved):
🚀

PR #349, opened 2026-06-16
feat(experiments): Add BE for new Experiment pinned_at field

shanaiabuggy:
## Summary by CodeRabbit

* **New Features**
  * Added experiment pinning support, exposing a nullable `pinned_at` timestamp to distinguish pinned vs unpinned experiments.
  * Introduced dedicated pin/unpin endpoints to update pinned state.
  * Updated the experiments list endpoint to support filtering by pinned state (`filter[is_pinned]=true/false`) and sorting by `pinned_at` (including `-pinned_at`).
* **Documentation**
  * Updated the API contract with the new pin/unpin endpoints, `filter[is_pinned]`, and `pinned_at` sort options.
* **Tests**
  * Added integration coverage for pin/unpin behavior, pinned-state filtering, and `pinned_at` sorting.


services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -59,6 +60,16 @@ def _sanitize_for_log(value: str) -> str:
 EXPERIMENTS_TAG = "Experiments"
 
 SortField = Literal["-created_at", "created_at", "-updated_at", "updated_at", "-name", "name"]
+ExperimentSortField = Literal[
```
shanaiabuggy, line 63: how are these diff from SorfField above

asutermo (approved):

services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
shanaiabuggy: `SortField` is used for Experiment **Groups**. I am renaming it to `ExperimentGroupSortField` to be more descriptive. We now have `ExperimentSortField` and `ExperimentGroupSortField`.

cc: @asutermo

PR #354, opened 2026-06-16
fix: helm chart KEK secret management must accept ready-made secret

crookedstorm:
When switching to the auto-generated secret, we also need to allow for a presented secret to work. Otherwise, you cannot restore your backup easily. `envFromSecret` might work and all, but that's a very different use case and practice than creating a secret from a vault and presenting it.

## Summary by CodeRabbit

* **New Features**
  * Added support for referencing an externally managed Kubernetes Secret for the default secrets encryption key via Helm values (`existingSecret` name/key).
  * Added optional Hugging Face token support for Docker builds and embedding model downloads when configured.
* **Documentation**
  * Expanded Helm README/values and install notes to document encryption-key selection rules, including mutual exclusivity.
* **Bug Fixes**
  * Improved render-time validation and conditional wiring to prevent conflicting environment/encryption-key configurations.

### DIFF
--- a/k8s/helm/README.md
+++ b/k8s/helm/README.md
@@ -10,9 +10,15 @@ The platform secrets service reads `NMP_SECRETS_DEFAULT_ENCRYPTION_KEY` from the
 API env Secret. The value must be base64-encoded and decode to at least 32 bytes.
 
 Set `secrets.defaultEncryptionKey.value` to provide your own key. When that value
-is empty and `envFromSecret` is not set, the chart runs a pre-install hook that
-creates `<fullname>-api-env` with a per-install random key. The hook is
-install-only and refuses to patch or rotate an existing Secret.
+is empty and neither `envFromSecret` nor
+`secrets.defaultEncryptionKey.existingSecret.name` is set, the chart runs a
+pre-install hook that creates `<fullname>-api-env` with a per-install random key.
+The hook is install-only and refuses to patch or rotate an existing Secret.
+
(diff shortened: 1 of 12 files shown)


k8s/helm/README.md
```diff
@@ -10,9 +10,15 @@ The platform secrets service reads `NMP_SECRETS_DEFAULT_ENCRYPTION_KEY` from the
...
-creates `<fullname>-api-env` with a per-install random key. The hook is
-install-only and refuses to patch or rotate an existing Secret.
+is empty and neither `envFromSecret` nor
+`secrets.defaultEncryptionKey.existingSecret.name` is set, the chart runs a
+pre-install hook that creates `<fullname>-api-env` with a per-install random key.
+The hook is install-only and refuses to patch or rotate an existing Secret.
+
+Set `secrets.defaultEncryptionKey.existingSecret.name` to use an existing Secret
+for only the secrets service encryption key. After Kubernetes decodes the Secret
+data, the value loaded from `secrets.defaultEncryptionKey.existingSecret.key`
```
svvarom, line 20: Maybe we could mention that the default is `NMP_SECRETS_DEFAULT_ENCRYPTION_KEY`

svvarom (approved):
Makes sense to me, feel free to ignore my nit

k8s/helm/README.md
crookedstorm: I'm not sure that matters? If you create an existing secret, you should specify the key really.

PR #365, opened 2026-06-16
fix: correct customizer values in the helm chart

crookedstorm:
Changes to the structure of customizer require changes in the helm chart to keep it functioning. This seems to fix things in terms of the e2e tests, i think.

## Summary by CodeRabbit

* **New Features**
  * Training execution profiles now configurable for AutoModel and Unsloth services with default settings.
  * Safe synthesizer container configuration added with default job execution mode and image specifications.


svvarom (approved):
The Helm changes LG, assuming the values are what we need.

PR #367, opened 2026-06-16
chore: add a job to clear the PR cache on merge/close

crookedstorm:
We have a cache problem. See https://github.com/NVIDIA-NeMo/nemo-platform/actions/caches

I'll do a manual cleanup, but this seems easy enough to fix.

## Summary by CodeRabbit

* **Chores**
  * Added an automated GitHub Actions cache cleanup workflow that removes stale caches after pull requests are closed.
  * Supports scheduled and manual runs to keep caches tidy, including cleanup of older merge-queue related caches outside pull request events.
  * Updated CodeQL runs to disable dependency caching.

### DIFF
--- a/.github/workflows/cache-cleanup.yaml
+++ b/.github/workflows/cache-cleanup.yaml
@@ -1,16 +1,47 @@
-name: Cleanup Cache on PR Close
+name: Cleanup GitHub Actions Caches
+
 on:
   pull_request:
     types: [closed]
+  schedule:
+    - cron: "17 3 * * *"
+  workflow_dispatch:
 
 jobs:
-  cleanup:
+  cleanup-pr:
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     permissions:
       actions: write
     steps:
       - name: Delete PR Caches
-        run: |
-          gh cache delete --all --ref refs/pull/${{ github.event.pull_request.number }}/merge
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/cache-cleanup.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - name: Delete PR Caches
+        run: |
+          gh cache delete --all --ref refs/pull/${{ github.event.pull_request.number }}/merge
```
svvarom: `gh` won’t infer the repo from `GITHUB_REPOSITORY` without checkout, so I think we need to pass `--repo ${{ github.repository }}` or set `GH_REPO: ${{ github.repository }}`.

crookedstorm: Good point!

svvarom: Source: https://cli.github.com/manual/gh_help_environment and https://cli.github.com/manual/gh_cache_delete

crookedstorm:
After manual pruning, of caches not used in 4 days, we went from 35 pages to 5 pages of caches :)

.github/workflows/cache-cleanup.yaml
crookedstorm: Should be fixed now.

svvarom (approved):

PR #369, opened 2026-06-17
feat(evaluator): add agent-eval engine (AgentEvaluator)

arpitsardhana:
Add the standalone agent-eval orchestration engine on top of the domain model: AgentEvaluator (generation + scoring), run-bundle persistence, and an HTML dashboard.

Adapted to the refactored domain model: trial/score naming, status enums, AggregatedMetricResult-based summary, and from_scores aggregation. Online inference seams (inference fns, clients, default headers) are injected on the AgentEvaluator instance rather than AgentEvalRunConfig. Also maps candidate evidence through build_metric_input so metrics can consume trial evidence.

## Release Notes

* **New Features**
  * Added an agent evaluation orchestrator to validate inputs, run scoring concurrently, and return a complete evaluation result bundle.
  * Added HTML dashboard rendering and file writing for compact evaluation summaries, metric rollups, and per-trial score tables.
  * Added run persistence to save benchmark/summary JSON, JSONL task/trial/score data, and a run manifest to disk.
  * Extended evaluation sample handling to include `evidence` in candidate payloads.
* **Tests**
  * Added dashboard and evaluator test suites covering scoring flows, failure modes, persistence, and live trial generation.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
@@ -1,0 +1,542 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Standalone agent evaluation orchestration."""
+
+from __future__ import annotations
+
+import asyncio
+from collections import defaultdict
+from collections.abc import Awaitable, Callable, Sequence
+from datetime import UTC, datetime
+from logging import getLogger
+from pathlib import Path
+from typing import Any, cast
+from urllib.parse import urlparse
+
+import httpx
+import nemo_evaluator_sdk.inference as inference
+from nemo_evaluator_sdk.agent_eval.dashboard import write_dashboard
+from nemo_evaluator_sdk.agent_eval.persistence import persist_run
+from nemo_evaluator_sdk.agent_eval.results import AgentEvalResult, AgentEvalSummary
+from nemo_evaluator_sdk.agent_eval.scores import (
+    AgentEvalDiagnostic,
+    AgentEvalDiagnosticSeverity,
+    AgentEvalScoreStatus,
+    AgentEvalTaskScore,
+)
+from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig, AgentEvalTask
+from nemo_evaluator_sdk.agent_eval.trials import (
+    AgentEvalTarget,
+    AgentEvalTrial,
+    AgentEvalTrialStatus,
+    AgentOutput,
+    AgentTaskRunner,
+)
+from nemo_evaluator_sdk.agent_inference import (
+    AgentInferenceFn,
+    make_agent_inference_request,
+    new_agent_inference_client,
+)
+from nemo_evaluator_sdk.execution.metric_execution import generate_online_sample, run_sync
+from nemo_evaluator_sdk.execution.samples import build_metric_input
+from nemo_evaluator_sdk.inference import InferenceFn
+from nemo_evaluator_sdk.metrics.protocol import Metric, validate_metric_result
+from nemo_evaluator_sdk.metrics.utils import metric_type_name
+from nemo_evaluator_sdk.values import Agent, Model, RunConfig, RunConfigOnline, RunConfigOnlineModel
+from nemo_evaluator_sdk.values.evidence import CandidateEvidence, EvidenceDescriptor
+from openai import AsyncOpenAI
+
+log = getLogger(__name__)
+
+
+class AgentEvaluator:
+    """Run stored-trial or live-target agent evaluations.
+
+    The online inference seam (an optional ``inference_fn``, transport ``client``, and
+    ``default_headers``) is injected on the evaluator instance rather than the run config,
+    because these are runtime transport concerns rather than declarative run settings. A
+    single ``inference_fn``/``client`` pair serves both model and agent targets; leave them
+    unset to let the evaluator build a default client for the resolved target type.
+    """
+
+    def __init__(
+        self,
+        *,
+        inference_fn: InferenceFn | AgentInferenceFn | None = None,
+        client: AsyncOpenAI | httpx.AsyncClient | None = None,
+        default_headers: dict[str, str] | None = None,
+    ) -> None:
+        self.inference_fn = inference_fn
+        self.client = client
+        self.default_headers = default_headers
+
+    async def run(
+        self,
+        *,
+        tasks: Sequence[AgentEvalTask],
+        trials: Sequence[AgentEvalTrial] | None = None,
+        target: AgentEvalTarget | None = None,
+        config: AgentEvalRunConfig | None = None,
+    ) -> AgentEvalResult:
+        """Evaluate imported trials or generate live trials before scoring.
+
+        Exactly one of ``trials`` or ``target`` must be provided.
+        """
+        resolved_config = config or AgentEvalRunConfig()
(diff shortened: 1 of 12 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,520 @@
...
+
+    Online inference seams (inference functions, clients, and default headers) are
+    injected on the evaluator instance rather than the run config, because they are
+    runtime transport concerns rather than declarative run settings.
+    """
+
+    def __init__(
+        self,
+        *,
+        model_inference_fn: Any | None = None,
```
arpitsardhana: This is an open question, how we want to pursue AgentEvaluator params.
these params enable to interact with arbitrary HTTP endpoint, agent with chat/completion compliant behaviour and custom runtimes.

SandyChapman: Couple of thoughts that I've been considering. Not sure it's 100% relevant, but we can likely discuss a bit:

I'm not sure I like the split for model/agent here. I've been wondering if we can build a single abstraction here. For example, calling an Model endpoint should just be a more specific version of an agent endpoint; one where the input is text only and the JSON payload is structured in messages format. The challenge is for some tasks, the model simply won't accept the right input (for example: local files) without injecting their content. This isn't all that different than multi-modal models vs text-only though, so maybe we just skip those tasks if someone needs it. Tasks may need to define a set of `required_capabilities` or something that allow the runtime to skip trying to complete a task that isn't possible.

Example where it might work:
Task: Evaluate the expression sqrt((4 * 8) / 20)^3

This is something where both a "math-centric" agent and a regular model could try to accomplish the task, but we'd need a consistent way to pass this to the model (adapting the task input into messages format) or the agent (passing directly as the instruction).

 I think in the short term, we should drop `model_inference_fn` and `model_client` from this evaluator and just name the params `client` and `inference_fn` to support future extensions. We could eventually provide some out-of-the-box inference_fn for both model and agent support.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,520 @@
...
+                try:
+                    return await _score_metric(
+                        run_id=run_id,
+                        task=task,
+                        trial=trial,
+                        metric=metric,
+                        row_index=row_index,
+                    )
+                except Exception as exc:
+                    if config.fail_fast:
```
SandyChapman, line 190: Can we add a log statement here? log.error or log.warn?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,520 @@
...
+            raise ValueError("at least one task is required")
+
+        run_id = resolved_config.run_id or _new_run_id()
+        runtime_config = resolved_config.model_copy(update={"run_id": run_id})
+        trial_list = (
+            list(trials)
+            if trials is not None
+            else await self._generate_trials(
+                tasks=task_list,
+                target=cast(AgentEvalTarget, target),
```
SandyChapman: We shouldn't need a cast here. Indicates something wrong with out typing. It could be the type narrowing in this statement isn't working well for the typechecker:
```python
        if (trials is not None) == (target is not None):
            raise ValueError("provide exactly one of trials or target")
```
probably should just rewrite to be something like:
```python
if trials is not None:
  trials_list = list(trials)
elif target is None:
  raise ValueError("provide exactly one of trials or target")
else:
  # target is not None and the typechecker should know this.
```
We can likely add some `@overload`s as well to better communicate that either trials OR target must be passed.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,520 @@
...
+        self,
+        *,
+        tasks: list[AgentEvalTask],
+        target: AgentEvalTarget,
+        config: AgentEvalRunConfig,
+    ) -> list[AgentEvalTrial]:
+        if isinstance(target, AgentTaskRunner):
+            return list(await target.run_tasks(tasks, config=config))
+
+        generation_target = cast(Model | Agent, target)
```
SandyChapman: Similarly here. If we're just going to cast to Model | Agent, we shouldn't be using `AgentEvalTarget` as the param type. I think we _do_ want `AgentEvalTarget` and probably should just be raising a `NotImplementedError` for the cases we're not handling yet.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,520 @@
...
+        params = _resolve_live_params(config, generation_target)
+        prompt_template = config.prompt_template or _default_prompt_template(generation_target)
+        semaphore = asyncio.Semaphore(params.parallelism)
+
+        client: Any | None = None
```
SandyChapman: `client` should probably have a better type than `Any`

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: Cool! I have reduced it to inference function and client. We can evolve it as need arises

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: done

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: Done. Avoiding overload as of now to avoid bit of boilerplate

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: Removed the casting. Thanks for catching

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: Gave it a concrete type

SandyChapman (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -0,0 +1,542 @@
...
+) -> AgentEvalResult:
+    path = Path(output_dir)
+    dashboard_path = None
+    if write_html:
+        dashboard_path = write_dashboard(result.model_copy(update={"output_dir": path}), path / "report.html")
+    return persist_run(result.model_copy(update={"output_dir": path, "dashboard_path": dashboard_path}), path)
+
+
+def _new_run_id() -> str:
+    return f"agent-eval-{datetime.now(UTC).strftime('%Y%m%d%H%M%S')}"
```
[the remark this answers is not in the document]
arpitsardhana: This is non-issue

PR #377, opened 2026-06-17
feat(jobs): auth e2e testing and http_client DI

ironcommit:
Allow callers of get_platform_sdk, get_task_sdk, and PlatformSDKProvider to pass an explicit sync httpx.Client, enabling task code to be tested in-process without spawning a real HTTP server.

Add integration tests in services/core/jobs that verify the full NMP_PRINCIPAL → on-behalf-of auth propagation path: a task using get_task_sdk can access secrets as the job creator, and is correctly denied when the creator lacks permission. Register standard pytest markers in the jobs pyproject.toml to support these test categories.

## Release Notes

* **New Features**
  * SDK factory and provider now support optional dependency-injected synchronous HTTP clients for both platform and task SDK creation.

* **Tests**
  * Added an integration test covering task-side authentication propagation and delegated access behavior.
  * Added a unit test verifying explicit HTTP client injection is preserved.
  * Introduced custom pytest markers for clearer test categorization.

* **Documentation**
  * Updated inline guidance on the HTTP client injection and migration path.

### DIFF
--- a/services/core/jobs/pyproject.toml
+++ b/services/core/jobs/pyproject.toml
@@ -60,3 +60,12 @@ dev = [
 [tool.pytest.ini_options]
 # testpaths = ["tests"]  # Commented to prevent namespace collision with root pytest
 asyncio_mode = "auto"
+markers = [
+    "integration: Service integration tests - test individual service interfaces and interactions (uses ASGI, mocks external services via SDK)",
+    "e2e: End-to-end tests - test complete customer workflows on deployed infrastructure (Helm/Docker Compose)",
+    "regression: Regression tests - test individual functional microservices for baseline functionality",
+    "canary: Canary tests - test deployed integration environments like top of tree",
+    "slow: Tests that take a long time to run",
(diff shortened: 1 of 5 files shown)


services/core/jobs/pyproject.toml
```diff
@@ -60,3 +60,12 @@ dev = [
 [tool.pytest.ini_options]
 # testpaths = ["tests"]  # Commented to prevent namespace collision with root pytest
 asyncio_mode = "auto"
+markers = [
```
mckornfield, line 63: do nested markers matter? I guess if you run pytest in this dir?

mckornfield (approved):
kind of a nit on the PR title, this is an integration test, not e2e, yeah?

mckornfield:
also might be worth including in the pr title the http client DI you added

services/core/jobs/pyproject.toml
ironcommit: I just added this to get rid of an annoying warning.  I'm not sure if it's strictly needed.

PR #379, opened 2026-06-17
feat(evaluator): persist metrics as entities with CRUD, SDK, and job refs

SandyChapman:
## Summary
Adds metric persistence to the evaluator plugin: store reusable metrics on the platform and reference them from eval jobs. Restores the metrics surface removed in #230, aligned to its `workspace/name` conventions.

## What's included
- **Entity + storage** — `MetricBundleEntity` (`entity_type` `metric_bundle`) in the entity store; the executable cloudpickle bundle is uploaded to the **Files service** (one fileset per metric), with the entity holding only `bundle_ref` + integrity digest.
- **CRUD API** (name-in-path, matching the removed service) — `POST` / `GET` / `DELETE /apis/evaluator/v2/workspaces/{workspace}/metrics/{name}` + workspace list. Metrics are **immutable** (no update).
- **SDK** — `client.evaluator.metrics` (`create` / `retrieve` / `list` / `delete`).
- **Job integration** — `EvaluateInputSpec.metrics` accepts inline `MetricInline` and/or `MetricRef` (`workspace/name`); refs are resolved from the entity store + Files during server-side spec resolution (`to_spec`).
- **OpenAPI** — explicit `CloudpickleMetricPayload` schema (discriminated on `kind`); regenerated `plugins/nemo-evaluator/openapi/openapi.yaml`.

## Testing
- **237 unit tests** (entity JSON round-trip, Files storage, service CRUD, HTTP routes, SDK resource, ref resolution).
- **End-to-end against running services** (entities, files, jobs + controller, evaluator):
  - metric CRUD + bundle round-trip through the real Files service;
  - a **submit-with-ref eval job** — the server resolved the `MetricRef` at submit time (canonical spec inlined the metric), the job ran to completion, and aggregate scores were returned.
- `ruff`, `ty`, and `pyright` clean.

## Notes / follow-ups
- Cleanup failure during create/delete is **leak-only** (an orphaned `metric-bundle.*` fileset, logged) — never a corrupted/broken metric. A GC sweep is tracked as a follow-up.
- cloudpickle payloads are shared executable code within a workspace (workspace-scoped + permission-gated under `evaluator.metrics.*`).
- Not in this PR: the unrelated core `files`-API SDK drift surfaced when regenerating (tracked separately).

## Release Notes

* **New Features**
  * Added stored metrics CRUD APIs for listing, creating, retrieving, and deleting metrics within workspaces.
  * Updated evaluation metric contracts to support inline metrics plus references, with reference resolution for execution.
  * Introduced new metric payload and querying schemas (including typed payload handling) and enhanced list filtering/sorting/pagination.
  * Added synchronous and asynchronous SDK resources for stored metrics.

* **Tests**
  * Added unit and end-to-end coverage for metrics endpoints, SDK resources, metric references, storage persistence, and entity serialization round-trips.

### DIFF
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -419,6 +419,189 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/metrics:
+    get:
+      tags:
+      - Evaluator Plugin Metrics Routes
+      summary: List Metrics By Workspace
+      description: List stored metrics for a specific workspace.
+      operationId: list_metrics_apis_evaluator_v2_workspaces__workspace__metrics_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: page
+        in: query
+        required: false
+        schema:
+          type: integer
+          minimum: 1
+          description: Page number.
+          default: 1
+          title: Page
+        description: Page number.
+      - name: page_size
+        in: query
+        required: false
+        schema:
+          type: integer
+          maximum: 1000
+          minimum: 1
+          description: Page size.
+          default: 100
+          title: Page Size
+        description: Page size.
+      - name: sort
+        in: query
+        required: false
+        schema:
+          allOf:
+          - $ref: '#/components/schemas/MetricSort'
+          description: The field to sort by. To sort in decreasing order, use `-`
+            in front of the field name.
+          default: created_at
+        description: The field to sort by. To sort in decreasing order, use `-` in
+          front of the field name.
+      - in: query
+        name: filter
+        style: deepObject
+        required: false
+        explode: true
+        schema:
+          $ref: '#/components/schemas/MetricFilter'
+        description: Filter metrics by workspace, name, metric_type, description,
+          created_at, and updated_at.
+      responses:
+        '200':
+          description: Return stored metrics for a workspace
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MetricsPage'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/metrics/{name}:
+    post:
+      tags:
+      - Evaluator Plugin Metrics Routes
+      summary: Create Metric
+      description: Store a new metric, addressed by workspace/name.
+      operationId: create_metric_apis_evaluator_v2_workspaces__workspace__metrics__name__post
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          maxLength: 255
+          pattern: ^[\w\-\.]+$
+          title: Name
+      - name: project
+        in: query
+        required: false
+        schema:
+          description: Optional project to associate with the metric.
+          title: Project
+          type: string
+        description: Optional project to associate with the metric.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MetricInline'
+      responses:
+        '201':
+          description: Store a new metric
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Metric'
+        '409':
+          description: Metric already exists
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+    get:
+      tags:
+      - Evaluator Plugin Metrics Routes
+      summary: Get Metric
+      description: Get a stored metric by workspace and name.
+      operationId: get_metric_apis_evaluator_v2_workspaces__workspace__metrics__name__get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Return stored metric details
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Metric'
+        '404':
+          description: Metric not found
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+    delete:
+      tags:
+      - Evaluator Plugin Metrics Routes
+      summary: Delete Metric
+      description: Delete a stored metric by workspace and name.
+      operationId: delete_metric_apis_evaluator_v2_workspaces__workspace__metrics__name__delete
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '204':
+          description: Delete a stored metric
+        '404':
+          description: Metric not found
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
 components:
   schemas:
     Agent:
@@ -491,6 +674,54 @@ components:
       - value_json_schema
       title: BundledMetricOutputSpec
       description: JSON-safe projection of a runtime metric output spec.
+    CloudpickleMetricPayload:
+      properties:
+        kind:
+          type: string
+          const: cloudpickle
+          title: Kind
+          description: Payload format discriminator.
+        python_version:
+          type: string
+          title: Python Version
+          description: Python version the metric was pickled with (must match at execution).
+        cloudpickle_version:
+          type: string
+          title: Cloudpickle Version
+          description: cloudpickle version used to serialize the metric.
+        pickle_protocol:
+          type: integer
+          title: Pickle Protocol
+          description: Pickle protocol used.
+        blob:
+          type: string
+          format: base64url
+          title: Blob
+          description: Base64-encoded cloudpickled metric object.
+        digest:
+          title: Digest
+          description: SHA-256 digest of the payload bytes. Informational; recomputed
+            server-side.
+          type: string
+      additionalProperties: false
+      type: object
+      required:
+      - kind
+      - python_version
+      - cloudpickle_version
+      - pickle_protocol
+      - blob
+      title: CloudpickleMetricPayload
+      description: 'Wire schema for a cloudpickle-serialized metric payload.
+
+
+        Mirrors the runtime ``CloudpickleMetricPayload`` so the API contract is
+
+        explicit in the OpenAPI spec. The runtime bundle model serializes payloads
+
+        polymorphically (typed as an abstract base), which renders as an opaque
+
+        object in the spec; this concrete DTO documents the actual fields.'
     DatetimeFilter:
       additionalProperties: false
       properties:
@@ -508,13 +739,6 @@ components:
       type: object
     EvaluateInputSpec:
       properties:
-        metrics:
-          items:
-            $ref: '#/components/schemas/MetricBundleInput'
-          type: array
-          minItems: 1
-          title: Metrics
-          description: Bundled metric entities to evaluate.
         dataset:
           anyOf:
           - items:
@@ -551,11 +775,21 @@ components:
           - $ref: '#/components/schemas/FieldMapping'
           description: Optional mapping from canonical evaluator fields to dataset
             columns.
+        metrics:
+          items:
+            anyOf:
+            - $ref: '#/components/schemas/MetricInline'
+            - $ref: '#/components/schemas/MetricRef'
+          type: array
+          minItems: 1
+          title: Metrics
+          description: Metrics to evaluate, given as inline metrics and/or references
+            to stored metrics.
       additionalProperties: false
       type: object
       required:
-      - metrics
       - dataset
+      - metrics
       title: EvaluateInputSpec
       description: Submitter-facing SDK evaluation input for the evaluator plugin
         job.
@@ -699,13 +933,6 @@ components:
       title: EvaluateJobsSortField
     EvaluateSpec:
       properties:
-        metrics:
-          items:
-            $ref: '#/components/schemas/MetricBundleOutput'
-          type: array
-          minItems: 1
-          title: Metrics
-          description: Bundled metric entities to evaluate.
         dataset:
           anyOf:
           - items:
@@ -742,13 +969,21 @@ components:
           - $ref: '#/components/schemas/FieldMapping'
           description: Optional mapping from canonical evaluator fields to dataset
             columns.
+        metrics:
+          items:
+            $ref: '#/components/schemas/MetricInline'
+          type: array
+          minItems: 1
+          title: Metrics
+          description: Inline metrics with all references resolved.
       additionalProperties: false
       type: object
       required:
-      - metrics
       - dataset
+      - metrics
       title: EvaluateSpec
-      description: Canonical SDK evaluation spec with platform model references resolved.
+      description: Canonical SDK evaluation spec with platform model and metric references
+        resolved.
     FieldMapping:
       properties:
         input:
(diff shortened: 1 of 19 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
```diff
@@ -0,0 +1,211 @@
...
+    async def create_metric(
+        self,
+        name: str,
+        metric: MetricInline,
+        *,
+        workspace: str,
+        project: str | None = None,
+    ) -> Metric:
+        """Store a new metric (addressed by workspace/name): upload its bundle, then index it."""
+        logger.debug("Creating metric", extra={"workspace": workspace, "metric_name": name})
```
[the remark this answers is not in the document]
SandyChapman: Fixed: user-derived values (`workspace`/`name`/`bundle_ref`) are now sanitized via `_sanitize_for_log` in all service log calls, matching the route layer. The code-scanning alerts auto-closed on the re-scan of the amended commit.

plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
```diff
@@ -0,0 +1,211 @@
...
+    async def delete_metric(self, workspace: str, name: str) -> bool:
+        """Delete a stored metric and its backing bundle. Returns False if not found."""
+        try:
+            entity = await self.entity_client.get(MetricBundleEntity, workspace=workspace, name=name)
+        except EntityNotFoundError:
+            return False
+
+        await self.entity_client.delete(MetricBundleEntity, name, workspace=workspace)
+        await self._discard_bundle(entity.bundle_ref)
+        logger.info("Metric deleted", extra={"workspace": workspace, "metric_name": name})
```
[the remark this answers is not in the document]
SandyChapman: Fixed: user-derived values (`workspace`/`name`/`bundle_ref`) are now sanitized via `_sanitize_for_log` in all service log calls, matching the route layer. The code-scanning alerts auto-closed on the re-scan of the amended commit.

plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
```diff
@@ -0,0 +1,211 @@
...
+        single metric version, so deleting it can only remove bytes nothing
+        else points at. A failure here leaks that one fileset (logged) but never
+        affects a live metric.
+        """
+        try:
+            await delete_bundle_by_ref(self.sdk, bundle_ref)
+        except Exception:
+            logger.warning(
+                "Failed to delete unreferenced metric bundle fileset; storage may be leaked",
+                extra={"bundle_ref": bundle_ref},
```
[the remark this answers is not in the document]
SandyChapman: Fixed: user-derived values (`workspace`/`name`/`bundle_ref`) are now sanitized via `_sanitize_for_log` in all service log calls, matching the route layer. The code-scanning alerts auto-closed on the re-scan of the amended commit.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -419,6 +419,183 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Metric'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
```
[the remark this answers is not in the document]
SandyChapman: Fixed: declared `409 Conflict` in the create route decorator's `responses=`; the regenerated OpenAPI documents it.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -419,6 +419,183 @@ paths:
...
```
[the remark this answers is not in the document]
SandyChapman: Fixed: declared `404` in the GET route decorator's `responses=`; regenerated spec documents it.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -419,6 +419,183 @@ paths:
...
+          title: Name
+      responses:
+        '204':
+          description: Delete a stored metric
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
```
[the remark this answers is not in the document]
SandyChapman: Fixed: declared `404` in the DELETE route decorator's `responses=`; regenerated spec documents it.

plugins/nemo-evaluator/src/nemo_evaluator/metric_storage.py
```diff
@@ -0,0 +1,106 @@
...
+        workspace=workspace,
+        description=f"Stored metric bundle for {workspace}/{name}.",
+    )
+    body = bundle.model_dump_json().encode("utf-8")
+    await sdk.files._upload_file(
+        path=BUNDLE_FILENAME,
+        body=body,
+        workspace=workspace,
+        name=fileset,
+    )
```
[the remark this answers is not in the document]
SandyChapman: Partially addressed: `load_bundle` now wraps a corrupt/unreadable stored bundle as `MetricBundleStorageError` (+ regression test). Declined the blanket SDK-exception wrap — raw `NotFoundError`/transport errors are more diagnostic than a generic wrapper, the service already handles them (rollback + mapped HTTP status), and only the parse/digest paths are part of the module's typed-error contract.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/metric_resources.py
```diff
@@ -0,0 +1,169 @@
...
+            headers=self._headers(),
+        )
+        response.raise_for_status()
+        return Page[Metric].model_validate(response.json())
+
+    def delete(self, name: str, *, workspace: str | None = None) -> None:
+        """Delete a stored metric and its backing bundle."""
+        response = self._http_client.delete(self._item_url(name, workspace), headers=self._headers())
+        response.raise_for_status()
+
```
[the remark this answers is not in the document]
SandyChapman: Fixed: added `timeout=self._platform.timeout` to `retrieve`/`list`/`delete` on both the sync and async resources.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -937,35 +1250,50 @@ components:
           type: string
           minLength: 1
           title: Metric Type
+          description: Runtime metric type name.
         metadata:
-          $ref: '#/components/schemas/MetricMetadata'
+          allOf:
+          - $ref: '#/components/schemas/MetricMetadata'
+          description: User-facing metric metadata.
```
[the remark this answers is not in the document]
SandyChapman: No action needed — representation-only change: `metadata` now renders as `allOf: [$ref: MetricMetadata]` vs an inline object, which is semantically identical and produces the **same wire JSON**. No client migration required (only strict schema codegen would notice).

plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
```diff
@@ -0,0 +1,223 @@
...
+            entity = await self.entity_client.get(MetricBundleEntity, workspace=workspace, name=name)
+        except EntityNotFoundError:
+            return False
+
+        await self.entity_client.delete(MetricBundleEntity, name, workspace=workspace)
+        await self._discard_bundle(entity.bundle_ref)
+        logger.info(
+            "Metric deleted", extra={"workspace": _sanitize_for_log(workspace), "metric_name": _sanitize_for_log(name)}
+        )
+        return True
```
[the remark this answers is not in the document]
SandyChapman: Fixed as suggested: the `delete` call is now wrapped in `try/except EntityNotFoundError` (returns `False`), preserving the documented contract under a concurrent-delete race; added a regression test.

plugins/nemo-evaluator/src/nemo_evaluator/metric_refs.py
```diff
@@ -0,0 +1,97 @@
...
+    root: str = Field(
+        description="Reference to a stored metric (format: workspace/metric-name, or metric-name in the job workspace).",
+    )
+
+
+def parse_metric_ref(root: str, default_workspace: str) -> tuple[str, str]:
+    """Split a metric reference into ``(workspace, name)``."""
+    workspace, separator, name = root.partition("/")
+    if separator != "/":
+        if not root:
```
arpitsardhana: Nit: If there is no root, it should be instant error right? Ideally this check should be at start ie
```python
If not root:
    return ValueError("MetricRef must not be empty")
workspace, separator, name = root.partition("/")
if workspace and separator and name:
     return workspace, name
if workspace or name:
     return default_workspace, root
if separator:
     return ValueError("MetricRef must be in format 'workspace/name')
``` 
This is based on my understanding of what we are trying to achieve

plugins/nemo-evaluator/src/nemo_evaluator/metric_storage.py
```diff
@@ -0,0 +1,111 @@
...
+
+#: Filename of the serialized bundle stored within each metric's fileset.
+BUNDLE_FILENAME = "bundle.json"
+
+#: Prefix for the per-upload filesets backing stored metrics.
+FILESET_PREFIX = "metric-bundle"
+
+
+class MetricBundleStorageError(RuntimeError):
+    """Raised when a stored metric bundle cannot be written, read, or verified."""
```
arpitsardhana, line 36: Is it deliberately meant to be empty?

plugins/nemo-evaluator/src/nemo_evaluator/metric_storage.py
```diff
@@ -0,0 +1,111 @@
...
+
+def parse_bundle_ref(bundle_ref: str) -> tuple[str, str, str]:
+    """Split a ``workspace/fileset#path`` reference into its parts."""
+    if "#" not in bundle_ref:
+        raise MetricBundleStorageError(f"invalid metric bundle reference (missing fragment): {bundle_ref!r}")
+    location, path = bundle_ref.split("#", 1)
+    if "/" not in location:
+        raise MetricBundleStorageError(f"invalid metric bundle reference (missing workspace): {bundle_ref!r}")
+    workspace, fileset = location.split("/", 1)
+    if not workspace or not fileset or not path:
```
arpitsardhana, line 59: So even if it is default workspace, it will still need to be specified?

plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
```diff
@@ -0,0 +1,227 @@
...
+        try:
+            await self.entity_client.get(MetricBundleEntity, name=name, workspace=workspace)
+            raise ValueError(f"Metric with name '{name}' already exists in workspace '{workspace}'")
+        except EntityNotFoundError:
+            pass
+
+        # Convert the wire DTO into the runtime bundle (JSON round-trip keeps the
+        # base64 payload handling consistent), then store/index it.
+        runtime_bundle = RuntimeMetricBundle.model_validate_json(metric.model_dump_json())
+        bundle_ref = await store_bundle(self.sdk, workspace, name, runtime_bundle)
```
arpitsardhana, line 129: store_bundle also call bunch of network call, it should be under try/except loop to account for errors

plugins/nemo-evaluator/src/nemo_evaluator/service.py
```diff
@@ -21,11 +22,36 @@
...
+            f"{permission_prefix}.create": f"Create {permission_prefix}",
+            f"{permission_prefix}.list": f"List {permission_prefix}",
+            f"{permission_prefix}.read": f"Read {permission_prefix}",
+            f"{permission_prefix}.delete": f"Delete {permission_prefix}",
+        },
+        endpoints={
+            base: {
+                "get": AuthzEndpointMethod(permissions=[f"{permission_prefix}.list"], scopes=read_scopes),
+            },
+            f"{base}/{{name}}": {
```
arpitsardhana, line 41: are these permissions granted only to service user or to actual user as well who triggerred the workflow?

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
```diff
@@ -0,0 +1,112 @@
...
+    metric_type: str = Field(min_length=1, description="Runtime metric type name.")
+    metadata: MetricMetadata = Field(default_factory=MetricMetadata, description="User-facing metric metadata.")
+    outputs: list[BundledMetricOutputSpec] = Field(min_length=1, description="The metric's output contracts.")
+    secrets: dict[str, SecretRef] = Field(
+        default_factory=dict, description="Secret references required to execute the metric."
+    )
+    payload: MetricPayload = Field(description="Format-specific serialized metric.")
+
+
+class Metric(BaseModel):
```
arpitsardhana, line 69: We also have a protocol in SDK with same name right
Even though SDK and plugins are two different packages with two different paths, does it make sense to have slightly more descriptive name here? like MetricRepr? or StoredMetric

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -937,35 +1250,50 @@ components:
...
-      type: object
-      title: MetricBundlePayload
-      description: Base class for concrete Pydantic metric bundle payloads.
+      title: MetricInline
+      description: 'Wire schema for an executable metric submitted to the platform.
+
+
+        Field-compatible with the runtime
+
+        ``nemo_evaluator.shared.metric_bundles.bundles.MetricBundle`` so it
```
arpitsardhana: Referencing code paths in openapi spec may not scale, since code path are subject to change and re-org

arpitsardhana (approved):
Minor nits and suggestions

plugins/nemo-evaluator/src/nemo_evaluator/metric_refs.py
SandyChapman: Good catch on the empty case. Rather than guarding inside the parser, I moved validation onto the field itself — `MetricRef.root` now carries `pattern=r"^[\w\-.]+(/[\w\-.]+)?$"`, so empty/malformed refs (`""`, `"ws/"`, `"/name"`, `"ws/a/b"`, `"bad name"`) are rejected at validation time with a Pydantic `ValidationError` before any parsing runs. `parse_metric_ref` is now a pure splitter. Done in 48cf854.

plugins/nemo-evaluator/src/nemo_evaluator/metric_storage.py
SandyChapman: Yes, deliberate. A dedicated `MetricBundleStorageError` lets callers catch storage-specific failures (a bundle that can't be written, read, or verified) precisely, rather than reaching for a broad `except RuntimeError`/`Exception`. The type itself is the signal, so it needs no extra fields.

plugins/nemo-evaluator/src/nemo_evaluator/metric_storage.py
SandyChapman: Yes — but the caller never specifies it. `bundle_ref` is constructed entirely server-side (the only place is `store_bundle`, which returns `f"{workspace}/{fileset}#{BUNDLE_FILENAME}"`), where `workspace` is the request's path workspace threaded through `MetricService.create_metric(workspace=…)` → `store_bundle(…, workspace, …)`. It's only ever read back off the `Metric` response, never accepted on a request body. So the concrete workspace — including `default` — is always already embedded; `parse_bundle_ref` just splits it back out and never has to default.

plugins/nemo-evaluator/src/nemo_evaluator/api/service/metric_service.py
SandyChapman: Agreed. `store_bundle` now wraps both network calls: a fileset-create failure raises a typed `MetricBundleStorageError`, and an upload failure rolls back the just-created (empty) fileset before raising — so a blip mid-upload doesn't leak storage. Added a regression test (`test_store_cleans_up_fileset_on_upload_failure`). Done in 48cf854.

plugins/nemo-evaluator/src/nemo_evaluator/service.py
SandyChapman: They're two distinct identities. The `evaluator.metrics.*` permissions/scopes declared here gate the **triggering user** — the platform auth middleware checks the incoming request's token against them before the route handler runs. The **service principal** is a separate identity used only for the downstream entity-store/Files calls the handler makes on the user's behalf. So these authorize the user who initiated the request, not the service account.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
SandyChapman: Good point — there is a `Metric` in the SDK protocol (`nemo_evaluator_sdk.metrics.protocol`). We went back and forth on this. The decision was to keep the public/contract type named `Metric` (the canonical name SDK users get via `client.evaluator.metrics`), and treat the in-process clash as a Python import concern — the two spots that touch both alias the protocol one as `RuntimeMetric`. We did consider `StoredMetric`/`MetricRepr`; if you'd prefer a descriptive *internal* class name while keeping `Metric` on the wire, the spec generator's `rename_schema_references` supports that (platform does it for `PageResponse`→`DeploymentsPage`). Leaving this open for your call.

plugins/nemo-evaluator/openapi/openapi.yaml
SandyChapman: Agreed — reworded the `MetricInline` docstring to describe the type by its fields and role rather than pointing at the runtime `MetricBundle` module path, so the generated spec description no longer references a code path. Done in 48cf854.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
arpitsardhana: Lets leave it as is here and come back to it if it becomes bit of import pain

PR #383, opened 2026-06-20
feat(evaluator): promote generic agentic runtimes into the Agent-Eval SDK

arpitsardhana:
## Summary

Promotes a small set of **generic** agent-evaluation primitives into the Agent-Eval SDK, and ships a runnable `run_agent_eval` example that drives the full **build env → run agent → verify → score → gate → persist** flow end to end. The SDK stays deliberately lean — it owns *generating trials and scoring them*; CI policy (the pass/fail gate), the pipeline wrapper, the Harbor-style verifier, the run layout, and all Docker/build glue live in the example.

### Agent-Eval SDK (`nemo_evaluator_sdk/agent_eval/`)

Three focused additions, no speculative surface:

- **`runtimes/environment.py`** — a pluggable agent-execution seam that sits below `AgentTaskRunner`: `AgentEnvironmentProvider` / `AgentEnvironmentHandle` protocols, an `AbstractEnvironmentHandle` that routes `run_agent` / `run_verifier` through a single `run(spec, role)`, the `EnvRunSpec` / `EnvCommandResult` value types, and a Docker reference implementation (`DockerEnvironmentProvider` / `DockerEnvironmentHandle`, stdlib-`subprocess` `docker run` with secret redaction).
- **`metrics.py`** — reusable scorers plus the measurement contract: `AgentPhaseSuccessMetric` (reads the agent-phase outcome) and `EvidencePresenceMetric` (scores over `candidate.evidence` on disk rather than a stamped reward — the SDK's metric-over-evidence value-add), and `TrialMeasurements`, the typed projection of the token/runtime/reward keys producers write onto `trial.metadata`.
- **`trials.py`** (extended) — adds the `AgentTrialSerde` protocol (the offline counterpart to `AgentTaskRunner`, so prior runs can be re-scored), and the runtime-agnostic trial-shaping helpers `resolve_trial_status` and `standard_evidence_descriptors` (the documented evidence-key builder).
- Vendored SDK copies (`beta/evaluator/agent_eval/`) regenerated to match.

### `run_agent_eval` example (`nemo_evaluator_sdk/examples/run_agent_eval/`)

A self-contained example that drives agent-eval tasks through the SDK with two paths:

- **Toy path** (default) — `mini_agent.py` runs as a host subprocess via `workflow_runtime.py` (an `AgentTaskRunner` + `TrialJsonSerde` + example tasks/metric), so it runs end to end with no external infrastructure.
- **Real path** (`--agentic-task <name>`) — runs an actual `tests/agentic-use` task through Docker with two backends:
  - **workflow** (`platform_runtime.py`) — task-local `nat run` of `workflow.yml`, with BUILD (image) / AGENT (`nat run`) / VERIFY (pytest) phases.
  - **aut** (`aut_runtime.py`) — create → seed inference providers → deploy → health-check → invoke a deployed agent-under-test.
- Example-local CI/backend glue: `build_spec.py` (`environment.yaml` → `BuildPlan`, with a Dockerfile escape hatch), `layout.py` (run-dir scaffold), `verify.py` (Harbor `reward.txt` convention), `gating.py` (`evaluate_gate` / `GateThresholds` → `gate.json`), `pipeline.py` (online run + offline rescore + gate wrapper), and `usage.py` (token/runtime extraction onto each trial).
- `run_agent_eval.py` is the CLI harness (online run, offline rescore, gate); ships a task-capable `aut_agent.workspace.example.yml` matched to `workspace-basic-mcp` (text-based ReAct so it works across whichever backend the gateway routes to).

### Tests (`tests/agent_eval/` mirrors the SDK package 1:1)

- `test_environment.py`, `test_metrics.py`, and `test_trials.py` (incl. the trial-shaping helpers) cover the new SDK code; `test_import_hygiene.py` keeps `agent_eval/` free of NeMo-Platform imports.
- Tests that exercised example-local code rather than the SDK were removed from the SDK suite, including the now example-coupled `test_profbench.py`.

## Test plan

- `uv run pytest packages/nemo_evaluator_sdk/tests/agent_eval/` → green (55 passed)
- `uv run ruff check` / `ruff format --check` / `uv run ty check` on the new code → clean
- `make vendor` → SDK mirror in sync (no drift)
- `run_agent_eval` toy path (default `mini_agent`, no external infra) end to end, plus offline rescore of the same bundle:

  ```text
  tasks: 2  trials: 2
  agent_phase_success.agent_phase_success: 2/2 true
  evidence_presence.evidence_present: 2/2 true
  output_contains.output_contains: 2/2 true
  gate.json → gate_passed: true, pass_rate=1.000
  ```

- `run_agent_eval` real path (AUT backend) verified end to end on `workspace-basic-mcp`:

  ```text
  agent_phase_success: 1/1 true
  agentic_use_verifier_reward.verifier_reward: mean=1.000
  runtime_sec: 87.9
  gate.json → gate_passed: true, pass_rate=1.000
  ```


arpitsardhana:
@coderabbitai full review

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trial_artifacts.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Standard keys: ``initial_state`` (task input filesystem, when staged),
+    ``trace`` (trajectory, ATIF-normalized when available), ``logs`` (agent log
+    dir), ``final_state`` (workspace), and ``verifier_logs`` (only when present).
+    Callers may add their own extension keys to the returned mapping.
+    """
+    descriptors: dict[str, EvidenceDescriptor] = {}
+
+    if initial_state_ref:
+        descriptors["initial_state"] = EvidenceDescriptor(
```
SandyChapman: We probably want to provide some constants for these "well-known" descriptor keys. Might still be useful to leave it as a string for custom fields for now, but I suspect once we have some active use cases, we'll want a `Literal` or similar for this dict.

packages/nemo_evaluator_sdk/examples/run_agent_eval/gating.py
```diff
@@ -0,0 +1,456 @@
...
+    summary = payload.get("summary") if isinstance(payload.get("summary"), dict) else payload
+    _validate_baseline_summary(summary, source)
+    return summary
+
+
+def summarize_run(
+    result: AgentEvalResult,
+    *,
+    reward_outputs: tuple[str, ...] = DEFAULT_REWARD_OUTPUTS,
+) -> dict[str, Any]:
```
SandyChapman, line 102: Would be good to provide at least a dataclass as return type here.

packages/nemo_evaluator_sdk/examples/run_agent_eval/gating.py
SandyChapman, line 1: Can you expand a bit on the purpose gating functionality here? If this is supposed to result in pass/fail semantics, I'm not sure it should belong in the SDK. Evaluator's responsibility is to generate trials and run scorers/metrics over them and not to make decisions on their outcomes (this is an external concern). If we want to build something like a pass/fail for CI, IMO, we should consider this an external tool (maybe a new plugin built on top of Evaluator).

packages/nemo_evaluator_sdk/examples/run_agent_eval/pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+
+        result = await AgentEvaluator().run(
+            tasks=prepared,
+            target=target,
+            config=self._run_config(output_dir=output_dir, run_id=run_id, benchmark=benchmark),
+        )
+        self._maybe_write_gate(result)
+        return result
+
+    async def score_trials(
```
SandyChapman, line 80: I'm a bit confused about the need for `AgentEvalPipeline`. Why aren't we just putting `prepare_task` into `AgentEvaluator.run` (or one level lower into the Runner) if it's a necessary precondition? Additionally, the `AgentEvaluator` handling both with/without trials was to have the same interface and this class just exposes two separate functions instead.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/measurements.py
```diff
@@ -0,0 +1,99 @@
...
+    total_tokens: int | None = None
+    cache_creation_tokens: int | None = None
+    cache_read_tokens: int | None = None
+    runtime_sec: float | None = None
+    reward: float | None = None
+    passed: bool | None = None
+    provenance: dict[str, Any] = Field(default_factory=dict)
+
+    @classmethod
+    def from_metadata(cls, metadata: Mapping[str, Any] | None) -> TrialMeasurements:
```
SandyChapman: `metadata` should be strongly typed IMO. Where does it get created? Can we parse into a structured type at the source?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/verify.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generic verifier-phase mechanic: collect a reward + stamp trial metadata.
+
+Backend-agnostic core. The caller runs its verifier through an environment handle,
```
SandyChapman: This doesn't strike me as backend-agnostic. It looks very specific to the harbor verifier outputs. Additionally, is this used anywhere?

packages/nemo_evaluator_sdk/examples/run_agent_eval/layout.py
```diff
@@ -0,0 +1,63 @@
...
+    An explicit ``output_dir`` must be made absolute: run-dir subpaths are used as
+    Docker bind-mount sources, and Docker treats a relative ``-v`` source as a
+    (slash-free) named volume rather than a host directory.
+    """
+    if output_dir is not None:
+        return Path(output_dir).resolve()
+    return default_factory()
+
+
+def prepare_run_layout(
```
SandyChapman, line 40: Is this planned to keep long term? Ideally we wouldn't need to materialize to the filesystem and could just pass the details to the runner.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/build_spec.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def execute_build_plan(plan: BuildPlan) -> None:
+    """Build the Docker image described by ``plan``."""
+    from nemo_evaluator_sdk.agent_eval.runtimes.docker import build_dockerfile
+
+    build_dockerfile(plan.dockerfile, plan.context_dir, plan.image_tag)
+
+
+_DOCKERFILE_UNSAFE_CHARS = set('\n\r"`$;&|<>')
```
SandyChapman: This blocklist includes `<` and `>`, which rejects the most common pip version pins — E.g. `pytest>=8`, `numpy<2`, `pandas>=1.5,<2`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/gating.py
```diff
@@ -0,0 +1,456 @@
...
+from nemo_evaluator_sdk.agent_eval.measurements import TrialMeasurements
+from nemo_evaluator_sdk.agent_eval.results import AgentEvalResult
+from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEvalTaskScore
+from pydantic import BaseModel
+
+# Metric outputs, in priority order, that represent a task's pass/reward signal.
+DEFAULT_REWARD_OUTPUTS: tuple[str, ...] = ("verifier_reward", "agent_phase_success")
+
+# Provenance fields collapsed into a single run-level summary.
+_PROVENANCE_FIELDS: tuple[str, ...] = (
```
SandyChapman: Two things on provenance:

1. **Nothing produces it.** No runtime or build step writes a `provenance` key onto `trial.metadata` — `build_trial_from_artifacts`, the workflow/generalized-agent drivers all omit it. So `summarize_run` always aggregates an empty list, `_aggregate_provenance([])` returns `{"available": false}` with every field null, and `commit_sha_consistent_within_run` / `commit_sha_matches_baseline` always hit their "not recorded; skipping" branch. The provenance block in `gate.json` is present but hollow, and persisted trials carry none. If the goal is human-reviewable provenance in the output (reasonable on its own), we still need a producer to stamp it (commit SHA, base-image digest, branch, ...).

2. The current tests cover only the consumer, by hand-injecting `metadata["provenance"]`, and only `commit_sha`/`commit_short` — the rest of `_PROVENANCE_FIELDS` (`agentic_base_image_digest`, `branch`, `remote_url`, `pinned*`) is never exercised.

Suggested test — a producer→output round-trip that proves provenance actually reaches the summary/`gate.json`, covering the documented fields rather than a hand-built dict:

```python
def test_provenance_roundtrips_from_producer_into_summary():
    # run the real runtime path that is supposed to record provenance
    trials = run_tasks_that_stamp_provenance(...)
    summary = summarize_run(result_with(trials))
    assert summary["provenance"]["available"] is True
    assert summary["provenance"]["commit_sha"]
    assert summary["provenance"]["agentic_base_image_digest"]  # documented field, not just commit_sha
```

Until a producer is wired this stays red — which is the point: it pins the contract that provenance must actually be populated to be useful in the output.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/generalized_agent.py
```diff
@@ -0,0 +1,277 @@
...
+        return {}
+
+    def final_output(self, artifacts: RunArtifacts, stdout_text: str) -> str:
+        """Final answer text: prefer the written final-output file, else stdout."""
+        if artifacts.final_output_path.exists():
+            return artifacts.final_output_path.read_text(encoding="utf-8")
+        return stdout_text
+
+
+class GeneralizedAgentDriver:
```
SandyChapman: Is there a consumer for this coming in a follow-up? As it stands, `GeneralizedAgentDriver` / `GeneralizedAgentSpec` / `ClaudeCodeSpec` / `CursorAgentSpec` are referenced only by their own unit test — nothing in the SDK or the `run_agent_eval` example uses them (the example drives `WorkflowAgentRuntime` / `NatWorkflowRuntime` / `NatAutRuntime` instead).

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trial_artifacts.py
arpitsardhana: added module level constants

packages/nemo_evaluator_sdk/examples/run_agent_eval/gating.py
arpitsardhana: gating moved to examples, so this may not be needed.

packages/nemo_evaluator_sdk/examples/run_agent_eval/gating.py
arpitsardhana: I moved out gating to examples.
Initial intent was to provide as much as lever for full agent orchestration and execution. I do agree, we can keep it outside

packages/nemo_evaluator_sdk/examples/run_agent_eval/pipeline.py
arpitsardhana: AgentEvalPipeline purpose is to provide agent/harness orchestration framework. I moved it out to examples

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/verify.py
arpitsardhana: Moved it to examples and reworded the docstring

packages/nemo_evaluator_sdk/examples/run_agent_eval/layout.py
arpitsardhana: Moved it into example along with verify.py

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/build_spec.py
arpitsardhana: Thanks for catching. fixed it

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/gating.py
arpitsardhana: Removed the provenance part

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/generalized_agent.py
arpitsardhana: there are no direct consumer. Assumption was when we replace nat runner, it will be handy to have claude, cursor as available agent harness.
I removed them as of now, We can be bring it back in followup when we start replacing nat runner

packages/nemo_evaluator_sdk/examples/run_agent_eval/build_spec.py
```diff
@@ -0,0 +1,211 @@
...
+def execute_build_plan(plan: BuildPlan) -> None:
+    """Build the Docker image described by ``plan``."""
+    from nemo_evaluator_sdk.agent_eval.runtimes.docker import build_dockerfile
+
+    build_dockerfile(plan.dockerfile, plan.context_dir, plan.image_tag)
+
+
+# Note: '<'/'>' are intentionally allowed — pip version pins (``numpy<2``,
+# ``pytest>=8``) use them and a Dockerfile ``RUN`` arg doesn't shell-redirect.
+_DOCKERFILE_UNSAFE_CHARS = set('\n\r"`$;&|')
```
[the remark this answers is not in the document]
arpitsardhana: Addressed

SandyChapman (approved):
lgtm! thanks for the edits

PR #384, opened 2026-06-22
ci: cache the embed model for the k8s smoke tests

crookedstorm:
Down loading the qdrant/all-MiniLM model for every PR is causing 429s. Using the cache makes sense for this, and it likely needs exactly one cache, period, not something that is keyed to a particular PR or commit.

## Summary by CodeRabbit

* **Chores**
  * Improved the Docker CPU smoke build pipeline by introducing FastEmbed model snapshot caching with automatic restore on cache hit and save on cache miss.
  * Reduced build times by preloading a shared cache and only downloading the ONNX model when it’s not already present.
  * Simplified the build/publish flow by separating image loading vs pushing based on the publish setting, and avoiding token wiring in those steps.
  * Removed the Hugging Face token from the local kind CPU smoke setup.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -243,10 +243,7 @@ jobs:
         run: make docker-push TARGET=docker-cpu
 
       - name: Save fastembed model cache
-        if: >
-          steps.fastembed-cache.outputs.cache-hit != 'true' &&
-          github.ref == 'refs/heads/main' &&
-          (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+        if: steps.fastembed-cache.outputs.cache-hit != 'true'
         uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         continue-on-error: true
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -195,7 +195,7 @@ jobs:
 
       - name: Restore fastembed model cache
         id: fastembed-cache
-        uses: actions/cache/restore@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: ${{ runner.temp }}/fastembed-cache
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -197,17 +235,21 @@ jobs:
...
         run: make docker-load TARGET=docker-cpu
 
       - name: Build and publish CPU images
         if: env.PUBLISH_IMAGES == 'true'
         shell: bash
-        env:
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: make docker-push TARGET=docker-cpu
 
+      - name: Save fastembed model cache
```
svvarom, line 245: This creates a cache per PR. Do we instead want to create the cache on `main` and have PRs reuse it, so we avoid duplicate caches per PR?

crookedstorm: It should not. Check out the key used. It's static.

crookedstorm: Checking things a bit.

crookedstorm: Yeah, a PR against main should use the cache on main with this key. Right now it is tied to this PR, but that should not be true once this runs on main. Importantly, it has to run on main at least once.

svvarom: I thought so too, but Codex is pretty adamant about this issue:

> The key is static, but GitHub also scopes caches by ref. Per the Actions cache docs, caches created by pull_request runs are stored under refs/pull/.../merge and can only be restored by reruns of that PR, not by main or sibling PRs. So I think we need main to create/save the shared cache, and PRs should restore from that.

Source: https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache

crookedstorm: Only on cache miss....which prevents breakage. see https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching

svvarom: Oh, I didn’t see the next two messages -  I was replying to the first one

svvarom (approved):

.github/workflows/ci.yaml
crookedstorm: Your link also confirms that codex is wrong lol.

PR #385, opened 2026-06-22
fix(studio): agent rule for KUI Select API, CSS patches existing bad usages

nv-odrulea:
fixes ASTD-249/bug-dropdowns-have-transparent-options-across-the-site

After the KUI v1 upgrade, some Select dropdowns render with a transparent background. The issue is NOT universal — it only hits call sites that put SelectItems directly inside SelectContent without a SelectListbox wrapper. Playground/ModelCompare selects appear unaffected because their SelectContent children include elements with their own
background (Block with bg-surface, or an explicit bg-surface-raised workaround).

### Root Cause

KUI v0: `SelectContent` rendered via Ariakit → MenuRoot → nv-menu-root class.

`nv-menu-root` carries `background-color: var(--background-color-surface-raised)`

KUI v1: `SelectContent` is now a bare `<div popover="auto" class="nv-select-content">`

`generated/styles.css:6511` explicitly resets the browser popover default:

```css
.nv-select-content[popover] {
  background: 0 0;   /* transparent */
  border: 0;
  ...
}
```

Background now lives on SelectListbox → MenuRoot (nv-menu-root). Sites that skip `SelectListbox` and go straight to `SelectItems` have no background element at all.

Confirmed broken: `SchemaSelectControl` (DatasetSchema right pane).
Confirmed working: `Playground/ModelCompare` selects (inner Block with `bg-surface` covers it).

Affected call sites (no explicit bg, no inner bg fallback)

```
- web/packages/studio/src/routes/FilesetDetailRoute/DatasetSchemaEditor/SchemaSelectControl/index.tsx
- web/packages/studio/src/components/customizer/CustomizationFilesetSelect/index.tsx
- web/packages/studio/src/components/filesets/AddToFolderModal/index.tsx
- web/packages/studio/src/components/filesets/FilesetFileExplorer/UploadToFolderModal/index.tsx
- web/packages/studio/src/components/evaluation/Configurations/form/LLMJudgeInput.tsx
- web/packages/studio/src/components/evaluation/Configurations/form/StringCheckInput.tsx
- web/packages/studio/src/components/DatasetFileSelect/index.tsx
- web/packages/studio/src/routes/InferenceProvidersListRoute/CreateInferenceProviderSidePanel/InferenceModelProviderSelect.tsx
```

### Fix

One CSS rule in `web/packages/studio/src/index.css`

Following the pattern of existing KUI hack fixes already in that file (nv-tooltip-content, nv-switch-track, nv-checkbox-input), add:

```
/* hack fix: KUI v1 SelectContent uses native popover which resets background;
   restore surface styles so all dropdowns render opaque. */
.nv-select-content[popover] {
  background-color: var(--background-color-surface-raised);
  border: 1px solid var(--border-color-base);
  border-radius: var(--radius-md);
  box-shadow: var(--shadow-md);
  overflow: hidden;
}
```

Why this wins:
- 1 file, ~7 lines
- Equal specificity to the background: 0 0 rule, placed AFTER `@import './generated/styles.css'` → source order ensures our rule wins
- Global: fixes all current and future SelectContent usages automatically
- No component restructuring, no wrapper, no per-call-site changes
- Existing explicit workarounds (EvaluationModelSelect, ControlledSearchableSelect) become
redundant but harmless

## Release Notes

* **Bug Fixes**
  * Improved dropdown/popover rendering for searchable selects and evaluation model selection, standardizing list rendering for consistent background, borders, rounded corners, shadows, and overflow behavior.
* **Documentation**
  * Added clearer KUI v1 Select guidance, including a required pattern for wrapping dropdown items to ensure correct styling, plus an exception for custom `SelectContent` content.

### DIFF
--- a/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
+++ b/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
@@ -263,7 +263,7 @@ export const ControlledSearchableSelect = ({
               />
             </Block>
             {/* eslint-disable-next-line no-restricted-syntax */}
-            <Stack className="overflow-auto w-full" style={{ maxHeight }} role="listbox">
+            <Stack className="overflow-auto w-full" style={{ maxHeight }}>
               {isLoading && filteredOptions.length === 0 ? (
                 <Flex align="center" justify="center" className="py-4">
                   <Spinner aria-label="Loading options" size="small" />
--- a/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
+++ b/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
@@ -241,7 +241,7 @@ export const ControlledSearchableSelect = ({
           status={status || (error ? 'error' : undefined)}
           {...selectProps}
         />
-        <SelectContent className="w-(--radix-popper-anchor-width) bg-surface-raised border border-base rounded-md shadow-md overflow-hidden">
+        <SelectContent className="w-(--radix-popper-anchor-width)">
           <Block className="p-2 w-full sticky top-0 bg-surface z-10">
             <TextInput
               ref={searchInputRef}
--- a/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
+++ b/web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
@@ -9,6 +9,7 @@ import {
   FormField,
   SelectContent,
   SelectItem,
+  SelectListbox,
   SelectProps,
   SelectRoot,
   SelectTrigger,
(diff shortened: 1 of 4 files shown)


web/packages/common/src/components/form/ControlledSearchableSelect/index.tsx
```diff
@@ -241,7 +241,7 @@ export const ControlledSearchableSelect = ({
           status={status || (error ? 'error' : undefined)}
           {...selectProps}
         />
-        <SelectContent className="w-(--radix-popper-anchor-width) bg-surface-raised border border-base rounded-md shadow-md overflow-hidden">
+        <SelectContent className="w-(--radix-popper-anchor-width)">
```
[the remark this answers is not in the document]
nv-odrulea: Finding is invalid. Skip.

@nemo/common has exactly two consumers: studio and storybook. Storybook explicitly imports ../../studio/src/index.css. Both consumers load studio's stylesheet — there is no cross-package CSS dependency problem in practice.

The theoretical concern (common consumed by a non-studio app without studio's CSS) doesn't apply to this codebase. No fix needed.

web/packages/studio/AGENTS.md
```diff
@@ -28,3 +28,29 @@
   - `@studio/` → `packages/studio/src/`
   - `@e2e-tests/` → `packages/studio/e2e-tests/`
 - Other local packages are imported via pnpm workspaces
+
+### KUI v1 Select: always wrap items in SelectListbox
```
steramae-nvidia, line 32: Why do the components you changed not follow this layout? Do other call sties need to be updated to match the required patten?

nv-odrulea: I asked the same question.
Response was that in both of these call sites, the exception applied:

```
Exception: call sites that fill SelectContent with custom children containing their own background (e.g. a sticky Block with bg-surface) are fine as-is.
```

However, investigating this one more time, I got a different result, seems this was an oversight.  Although both had their own custom children, they still had incomplete protection from the transparency issue, and do indeed need to follow the new pattern.

steramae-nvidia (approved):

PR #388, opened 2026-06-22
fix: remove tokens from local file on logout

mckornfield:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed logout to clear OAuth credentials only for the selected context, leaving other contexts unchanged.
  * Improved OAuth credential handling so explicitly clearing tokens via config overrides works correctly.

* **Improvements**
  * Enhanced device authorization flow handling for OAuth endpoints.
  * Refined OAuth scope parsing to handle different token claim formats more reliably.

* **Tests**
  * Added coverage for logout clearing behavior and config write behavior when tokens are explicitly set to null.


anastasia-nesterenko (approved):
👏

PR #391, opened 2026-06-22
fix(studio): replace ScrollTable with KUI table, fix scrolling

nv-odrulea:
`ScrollTable` wasn't the best fit for the files list in the Dataset file picker modal.  One reason is that it didn't bring sticky table headers with it, leaving us to either use DataView (overkill) or roll our own. So we roll our own basic implementation of KUI table here, which is a fairly simple use case and appropriate.

As a bonus, the styling looks better since the rows do not have padding beyond their separators. We also gain sticky table headers.

To fix sticky table header bottom row from scrolling out of view, we needed to apply a known CSS hack involving `box-shadow`:

> border-bottom on a table element (tr, th, thead) is laid out by the browser's table algorithm, not the element's own paint layer — sticky doesn't take it with the element. box-shadow bypasses the table layout model and is always painted with the sticky element. This is standard practice; MDN and CSS Working Group guidance both recommend box-shadow for sticky table header separators.

Screenshot 2026-06-22 at 12 52 07 PM

## Summary by CodeRabbit

* **New Features**
  * Improved file selection in the upload list, supporting both single-select (radio) and multi-select (checkbox) modes with clearer per-row controls.
* **Style**
  * Redesigned the file upload modal layout for a simpler, cleaner presentation.
  * Enhanced the file table presentation with sticky headers and clearer visual separators.
* **Bug Fixes**
  * Improved upload table rendering reliability in automated tests by stabilizing layout measurements.

### DIFF
--- a/web/packages/common/src/components/UploadModal/SimpleFilesTable.tsx
+++ b/web/packages/common/src/components/UploadModal/SimpleFilesTable.tsx
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import * as DataView from '@nemo/common/src/components/DataView/internal';
 import { useUploadModalContext } from '@nemo/common/src/components/UploadModal/Context/useUploadModalContext';
 import { useInlinePickerSlot } from '@nemo/common/src/components/UploadModal/InlinePickerSlot';
 import { UploadFile } from '@nemo/common/src/components/UploadModal/types';
@@ -8,21 +9,23 @@ import { formatFileSize } from '@nemo/common/src/components/UploadModal/utils';
 import {
   Button,
   Checkbox,
-  Text,
   Flex,
-  Stack,
-  RadioGroupRoot,
-  RadioGroupItem,
   RadioGroupInput,
-  TableBody,
-  TableDataCell,
-  TableHead,
-  TableHeaderCell,
-  TableRoot,
-  TableRow,
+  RadioGroupItem,
+  RadioGroupRoot,
+  Stack,
+  Text,
 } from '@nvidia/foundations-react-core';
 import { CircleAlert } from 'lucide-react';
-import { useCallback, useMemo, useRef } from 'react';
+import { type ComponentProps, useCallback, useMemo, useRef } from 'react';
+
+type FileRow = {
+  id: string;
+  name: string;
+  size: number;
+  isDisabled: boolean;
+  uploadFile: UploadFile;
+};
 
 export const SimpleFilesTable = () => {
   const [state, dispatch] = useUploadModalContext();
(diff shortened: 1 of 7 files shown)


web/packages/common/src/components/UploadModal/SimpleFilesTable.tsx
```diff
@@ -108,59 +105,63 @@ export const SimpleFilesTable = () => {
...
-          ],
-        };
+        return { id: uploadFile.id, name, size, isDisabled, uploadFile };
       }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [visibleFiles, selectedFiles, toggleFileSelection, allowMultipleFileSelection, invalidFileMode]
+    [visibleFiles, invalidFileMode, allowedExtensions]
+  );
+
+  const table = (
```
steramae-nvidia: Rather than rolling up our own table can we just use a DataView? I bet you could get the sticky table header without needing the box-shadow fix

nv-odrulea: Sure, I'll try it. My coding assistant led me to believe that DataView would be overkill, and I think ScrollTable kind of was for this use case. I'll give it a try and see if it can simplify the changes.

nv-odrulea: @steramae-nvidia I'm going to push back on this idea, mainly because I agree with the LLM that DataView is overkill in this use case. Here's some rationale:

---------------------

1. Radio select has no DataView equivalent. RowSelectionCell is checkbox-only. allowMultipleFileSelection=false uses radio buttons — we'd have to drop that entirely or hand-roll it alongside DataView anyway.
2. DataView is a full-page list component. StudioDataView brings a search toolbar, bulk actions, filters, pagination — all dead weight inside a modal with 5–20 files.
3. Scroll bound is simpler than virtualization. DataView scrolling uses VirtualizedTableContent with a virtualizer. Our max-h-[45dvh] wrapper div is ~6x less code for the same visual result at this scale.

We could technically use the bare DataView.Root + DataView.TableContent primitives (skipping StudioDataView), but we'd still need makeColumns, useDataViewState, and column definitions boilerplate for 2 columns — more lines than what we have now, with worse radio support.

Current hand-rolled table is the right call for an embedded modal picker.

---------------------

In case you might be concerned that we risk diverging from the styles in DataView, I double checked that this implementtion uses the same primitives as DataView:

---------------------

Same primitives. DataView.TableContent uses TableRoot, TableHead, TableRow, TableBody, TableDataCell, TableHeaderCell — all from @nvidia/foundations-react-core. SimpleFilesTable imports the exact same set.

Both render identical KUI markup → identical CSS classes → identical styles. No divergence risk.

The one difference: DataView uses stickyTableHeader prop (which applies sticky top-0 z-10 to TableHead) — same pattern we use in SimpleFilesTable. Everything is driven by the same KUI tokens.

web/packages/common/src/components/UploadModal/index.tsx
```diff
@@ -70,34 +69,30 @@ const UploadModalContent: FC<UploadModalProps> = ({
...
-            </ModalFooter>
-          </Stack>
+          <ModalHeading {...attributes?.ModalHeading}>{title}</ModalHeading>
+          <ModalMain {...attributes?.ModalMain}>
+            <UploadPickerBody
+              workspace={workspace}
+              includeDataset={includeDataset}
+              includeTabs={includeTabs}
+            />
+          </ModalMain>
```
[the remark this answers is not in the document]
nv-odrulea: done

web/packages/common/src/components/UploadModal/SimpleFilesTable.tsx
nv-odrulea: ok this is done now.

steramae-nvidia (approved):

PR #393, opened 2026-06-22
fix(e2e): fix local minikube scripts and switch CLI health checks to /status

matthewgrossman:
## Summary

Two fixes:

### 1. CLI and e2e health checks: `/health/ready` → `/status`

The `nemo` CLI (`services start`, `setup`, `quickstart`) and e2e test harness were polling `/health/ready` to check if the platform is up. This endpoint is not exposed through ingress — by design, `/health` is for K8s pod probes, not external access. Switched to `/status`, which is already in the ingress allowlist and returns richer application-level health info.

**Changed files:** CLI source (`packages/nemo_platform_ext/.../quickstart/cli.py`, `services/cli.py`, `setup.py`), vendored SDK copies, `e2e/conftest.py`, `e2e/k8s/scripts/run_auth_e2e.sh`

### 2. Fix `local_build_and_upgrade.sh`

The script was broken in several ways:
- Referenced nonexistent `e2e/k8s/values/local.yaml`
- Didn't set `api.image.repository` or `pullPolicy` — pods tried to pull from `nvcr.io` instead of using local images
- Used the GPU setup script (no port mapping) — `localhost:30080` didn't work
- Duplicated Helm install logic from `install_helm_e2e.sh`

Now delegates to `install_helm_e2e.sh` for the Helm install, uses the CPU setup script with port mapping, and supports configurable registry via `NMP_REGISTRY` env var. Also added `NMP_E2E_PULL_POLICY` to `install_helm_e2e.sh` so local builds can set `pullPolicy=Never`.

## Test plan

- [x] `minikube delete` → `bash e2e/k8s/scripts/local_build_and_upgrade.sh` succeeds end-to-end
- [x] `curl localhost:30080/status` returns 200
- [x] Full e2e suite against minikube: 54 passed, 2 failed (pre-existing), 8 skipped

Closes AIRCORE-786
Closes AIRCORE-791


e2e/k8s/scripts/install_nmp_auth_e2e.sh
```diff
@@ -8,7 +11,8 @@ REPO_ROOT="$(git -C "${SCRIPT_DIR}" rev-parse --show-toplevel)"
 
 export NAMESPACE="${NAMESPACE:-${KUBE_NAMESPACE:-default}}"
 export HELM_RELEASE_NAME="${HELM_RELEASE_NAME:-nemo-platform}"
-export HELM_VALUES="${HELM_VALUES:-${REPO_ROOT}/e2e/k8s/values/minikube-auth.yaml}"
+export HELM_VALUES="${HELM_VALUES:-${REPO_ROOT}/e2e/k8s/values/minikube.yaml}"
+export HELM_EXTRA_ARGS="${HELM_EXTRA_ARGS:-} -f ${REPO_ROOT}/e2e/k8s/values/minikube-auth.yaml"
```
mckornfield: is there a place where we use HELM_VALUES but not HELM_EXTRA_ARGS ?

e2e/k8s/scripts/local_build_and_upgrade.sh
```diff
@@ -1,61 +1,65 @@
...
-# Build the images with a local tag and then load them
-# Use epoch seconds (date +%s) so each run gets a unique tag and upgrades pick up new images
+# Use epoch seconds so each run gets a unique tag and upgrades pick up new images
 IMAGE_TAG="${IMAGE_TAG:-local-$(date +%s)}"
 
 # Detect platform for build (match host arch)
 BUILD_ARCH="${BUILD_ARCH:-linux/$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')}"
 GIT_SHA=$(git -C "${REPO_ROOT}" rev-parse HEAD)
 
+NMP_REGISTRY="${NMP_REGISTRY:-docker.io/my-registry}"
```
mckornfield, line 37: is `docker.io` the default host? I think so but wanted to check, bc I know we don't set a host

e2e/k8s/values/minikube.yaml
```diff
@@ -1,7 +1,23 @@
 # Minikube values for local development
 # This is a standalone values file - use it in place of ./default.yaml
 #
-# Usage: helm upgrade -i nemo-platform k8s/helm -f e2e/k8s/values/minikube.yaml
+# Usage (local build):
+#   helm upgrade -i nemo-platform k8s/helm -f e2e/k8s/values/minikube.yaml \
```
mckornfield: nit- want to buy more helm overlay files instead of these style commands

e2e/k8s/values/minikube.yaml
```diff
@@ -30,6 +46,13 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-send-timeout: "600"
     nginx.ingress.kubernetes.io/proxy-connect-timeout: "600"
     nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
+  hosts:
+    - name: ""
+      paths:
+        - path: /
```
mckornfield: this means we'll let in all traffic to `/` right? assuming we want that for minikube but not necessarily for "the real deal"?

mckornfield (approved):
some minor things

e2e/k8s/scripts/local_build_and_upgrade.sh
crookedstorm: It has to be something in k8s 1.34+. docker.io is what blank means in the docker app, so it's not unreasonable.

k8s/helm/values.yaml
```diff
@@ -410,6 +410,10 @@ ingress:
           pathType: Exact
           service: '{{ include "nemo-platform.ingressBackendService" . }}'
           port: '{{ include "nemo-platform.ingressBackendPort" . }}'
+        - path: /health
```
mckornfield: yis, looks important lol

mckornfield (commented):
some minor things

e2e/k8s/values/minikube-auth-portforward.yaml
```diff
@@ -1,42 +1,18 @@
...
-
-postgresql:
-  persistence:
-    storageClass: standard
-
-core:
-  storage:
-    storageClass: standard
-    volumePermissionsImage: busybox
+# Disables ingress and envoy proxy for machines where registry.k8s.io is
```
crookedstorm: registry.k8s.io is blocked somewhere??

The main reason I've used port fowarding is on my local machine, and I port forward the ingress controller, since it doesn't expose a typical nodeport on a Mac or on Windows.

e2e/k8s/values/minikube-auth-portforward.yaml
```diff
@@ -1,42 +1,18 @@
...
+# instead.
 
 envoyProxy:
   enabled: false
 
 ingress:
   enabled: false
-
-platformConfig:
-  auth:
```
crookedstorm, line 30: Is auth a default now? If not, disabling it sounds like this file needs a rename.

e2e/k8s/values/minikube-auth.yaml
```diff
@@ -1,46 +1,24 @@
...
 
-postgresql:
-  persistence:
-    storageClass: standard
-
-core:
-  storage:
-    storageClass: standard
-    volumePermissionsImage: busybox
-
```
crookedstorm, line 24: This is going to break the e2e tests I finally got working in the other repo, I think...

e2e/k8s/scripts/local_build_and_upgrade.sh
crookedstorm: For this it shouldn't really matter what the host is. It just has to be something.

crookedstorm (approved):
Looks good to me, for what it's worth.

e2e/k8s/scripts/install_nmp_auth_e2e.sh
matthewgrossman: Yes — `local_build_and_upgrade.sh` sets `HELM_VALUES=minikube.yaml` without any `HELM_EXTRA_ARGS` (no auth overlay needed for the default local dev path). They're independent env vars in `install_helm_e2e.sh`.

e2e/k8s/values/minikube.yaml
matthewgrossman: Helm values files can't interpolate env vars, so the `--set` overrides are necessary for dynamic values like registry and tag. We could create per-registry overlay files (`minikube-local.yaml`, `minikube-ghcr.yaml`) but you'd still need `--set` for the tag, and it means maintaining N files for N registries. The `--set` approach is the standard Helm pattern for this. The comments in the file are just documenting the usage — the script (`local_build_and_upgrade.sh`) handles all of this for the common case.

e2e/k8s/values/minikube.yaml
matthewgrossman: Resolved — this ingress override was removed. Instead of exposing `/health` through ingress, the CLI and e2e harness now use `/status` which is already in the chart's ingress allowlist.

k8s/helm/values.yaml
matthewgrossman: Reverted — we're using `/status` instead, so no chart changes needed.

e2e/k8s/values/minikube-auth-portforward.yaml
matthewgrossman: This file is reverted to main — the overlay refactor was dropped from this PR.

e2e/k8s/values/minikube-auth-portforward.yaml
matthewgrossman: Reverted — this file is back to its original standalone form on main.

e2e/k8s/values/minikube-auth.yaml
matthewgrossman: Reverted — the overlay refactor was dropped from this PR entirely. `minikube-auth.yaml` is unchanged from main.

e2e/k8s/values/minikube.yaml
matthewgrossman: didn't end up changing this in this PR actually, so removing

PR #398, opened 2026-06-23
feat(jobs): add auth-aware E2E tests and job diagnostics infrastructure

ironcommit:
## Summary

Add auth-aware E2E tests that validate workspace isolation and principal propagation under an auth-enabled platform config. Introduce a reusable `diagnostics` module for structured job/step/task state collection on errors, and fix a workspace-scoping bug in `list_tasks`.

### Auth-aware E2E test suite (`e2e/test_jobs_auth.py`)
- **Principal propagation**: verifies a user with workspace Editor role can submit a job and read its output fileset
- **Workspace isolation**: verifies a user cannot access filesets in a workspace they lack permissions for (expects 403)
- **Admin cross-workspace listing**: verifies an admin can list jobs across all workspaces

### Per-module platform config support (`e2e/conftest.py`)
- Refactor E2E fixtures to support multiple running-services instances keyed by config hash, enabling per-test-module platform configs (e.g., auth-enabled) to coexist in a single session
- Add `pytest.mark.e2e_config(...)` marker for declaring ordered config file layers and inline dict overlays per test module
- Replace the session-scoped `_services` fixture with a module-scoped `_services_instance` backed by a session-scoped pool
- Remove the `nemo_run` fixture; replace with `run_nemo_local` from `nmp.testing`

### Job diagnostics module (`services/core/jobs/.../controllers/diagnostics.py`)
- `collect_job_diagnostics()`: fetches job, step, task, status, and log data via the SDK and returns a structured dict
- `log_job_diagnostics_if_debug()`: conditionally emits diagnostics at DEBUG level
- Wired into the reconciler and scheduler to automatically log diagnostics when steps transition to ERROR or encounter unexpected exceptions

### Bug fix: workspace-scoped task listing
- Pass `workspace` through `dispatcher.list_tasks()` to `store.list()` so task queries for non-default workspaces return correct results
- Added integration test reproducing the bug (`test_list_tasks_uses_step_workspace_for_non_default_jobs`)

### Type safety improvements in scheduler/reconciler
- Use typed `PlatformJobStepsListFilterParam` and `SDKPlatformJobStatus` instead of raw dicts/strings
- Pass `.value` for enum status fields to match SDK expectations
- Add `assert step.step_spec is not None` guard before accessing executor

### Other
- New `e2e/configs/local-subprocess.yaml` E2E config (runtime: none, subprocess executor)
- `run_nemo_local` now accepts `base_url` and `workspace` keyword arguments
- Registered `e2e_config` as a custom pytest marker in `pytest.ini`
- Refactored `test_task_auth_runtime.py` to use a Protocol-based task stub returning the secret value directly instead of capturing stdout

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed job task listing to respect workspace boundaries end-to-end by scoping task retrieval to the current workspace.
  * Improved local subprocess job synchronization when subprocess metadata arrives late (grace period for recent pending steps).
* **New Features**
  * Added `include_job_logs_in_diagnostics` to optionally include raw job logs in controller diagnostics.
  * Added local subprocess-based E2E configuration with per-test config layering and pooled service startup by effective config.
  * Added authenticated E2E job coverage (job principal propagation, forbidden workspace access, admin listing across workspaces).
* **Tests**
  * Expanded diagnostics, scheduler/reconciler, dispatcher scoping, and subprocess backend test coverage.


e2e/configs/local-subprocess.yaml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  # Local E2E-only debugging aid. This may leak secrets or PII from job output,
+  # so it must remain disabled in non-test configs.
+  include_job_logs_in_diagnostics: true
+  executors:
+    - provider: subprocess
+      profile: default
+      backend: subprocess
+      config:
+        working_directory: /tmp/nmp-subprocess-jobs
```
mckornfield: we want to use `/tmp` and not a local/.gitignored dir?

e2e/conftest.py
```diff
@@ -87,36 +140,250 @@ def ngc_secret(sdk: NeMoPlatform, workspace: str, ngc_api_key: str) -> Iterator[
...
+    url: str
+    log_path: Path | None
+    proc: subprocess.Popen[Any] | None
+    config_path: Path | None
+    auth_enabled: bool = False
+    key: _ServicesPoolKey | None = None
+
+
+@dataclass(frozen=True)
+class _ModuleConfigState:
```
mckornfield: nit on making these all private. I think it just looks yucky, and don't suspect anything will try to import from e2e? or is there a reason to do it? (I don't think we use the `from e2e.conftest import *` anywhere, do we?

e2e/test_jobs.py
```diff
@@ -17,7 +17,10 @@
 
 JOB_SOURCE = "e2e-test-jobs"
 
-pytestmark = [pytest.mark.timeout(600)]
+pytestmark = [
+    pytest.mark.timeout(600),
+    pytest.mark.e2e_config("e2e/configs/local-subprocess.yaml"),
```
mckornfield, line 22: should this just be the default? or you prefer the trace-ability?

services/core/jobs/src/nmp/core/jobs/api/v2/jobs/endpoints.py
```diff
@@ -651,7 +651,7 @@ async def list_job_step_tasks(
         if not step_entity:
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Job step not found")
 
-        return PlatformJobListTaskResponse(data=await dispatcher.list_tasks(step_entity.id))
+        return PlatformJobListTaskResponse(data=await dispatcher.list_tasks(step_entity.id, workspace=workspace))
```
mckornfield, line 654: this was the fix you were discussing in the desc yeah?

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
```diff
@@ -281,6 +282,24 @@ def sync(self, step: PlatformJobStepWithContext) -> JobUpdate:
...
+            if task_fallback is not None:
+                return task_fallback
+            if step.status == PlatformJobStatus.PENDING and not self._pending_step_missing_metadata_is_stale(step):
+                # Stopgap only: the subprocess backend keeps execution metadata in controller
+                # memory, while step/task state is persisted in the jobs database. Those two
+                # sources of truth are not fully synchronized today because subprocess was not
+                # originally designed around durable jobs-backed execution state. That means a
+                # step can already be visible in the database as pending before this backend has
+                # registered local subprocess metadata for it. Keep the step pending briefly
+                # instead of failing it. The real fix is to move subprocess onto properly
```
mckornfield, line 295: should we make a ticket for that? or will subprocess go bye bye?

services/core/jobs/src/nmp/core/jobs/controllers/reconciler.py
```diff
@@ -129,16 +149,17 @@ def step(self):
...
                     logger.exception("Could not complete cleanup steps for backend", exc_info=True)
 
-    def get_steps_for_reconciliation(self, statuses: list[str]) -> list[PlatformJobStepWithContext]:
+    def get_steps_for_reconciliation(self, statuses: list[SDKPlatformJobStatus]) -> list[PlatformJobStepWithContext]:
         """
         Return the list of steps to reconcile.
         """
         # Iterate through all pages to get all steps
         steps = []
+        filter_params: PlatformJobStepsListFilterParam = {"status": statuses}
```
mckornfield, line 158: is this just a cleanup, or for typing, or so that it can be mutated?

mckornfield (approved):
mostly q

e2e/conftest.py
ironcommit: I agree that it's ugly, but python doesn't provide public/private. I could extract it into it's own module, and then export from there.  It would cleanup the conftest a lot.

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
ironcommit: I have a separate spec for this, which I hope to get to soon.

e2e/conftest.py
ironcommit: Extracted to keep conftest focussed.

e2e/test_jobs.py
ironcommit: This just overrides the default.  I think it would be good to have explicit configs for the e2e tests, but that would be a follow up.

services/core/jobs/src/nmp/core/jobs/controllers/reconciler.py
ironcommit: just for typing

e2e/configs/local-subprocess.yaml
ironcommit: updated.

PR #399, opened 2026-06-23
feat(deployments): implement DockerDeploymentBackend (AIRCORE-756)

tylersbray:
## Summary

Implements `DockerDeploymentBackend` for `nemo-deployments-plugin` ([AIRCORE-756](https://linear.app/nvidia/issue/AIRCORE-756)).

Rebased onto `main` after [#469](https://github.com/NVIDIA-NeMo/nemo-platform/pull/469) merged (prerequisites on `Deployment`, volume `DELETING`, orphan grace, shutdown cancellation). Branch history was reset to cherry-pick only the Docker backend commits — no duplicated reconciler code from the old #315 stack.

* **v1 scope:** single-container deployments; rejects `init_containers` and multi-container specs
* **Entity contract:** reconciler passes `config_name` only; backend loads full `DeploymentConfig` from the entity store via `NemoEntitiesClient`
* **Lifecycle:** docker pull/run, inspect, stop/rm, volume create/delete, port allocation from entity `DockerDeploymentConfig`
* **Probes:** exec and http_get readiness supported; grpc deferred
* **GPU:** plugin-local `DockerGPUPool` (consolidation to `nemo_platform_plugin` tracked separately in AIRCORE-759)
* **Idempotency:** workspace, name, `managed-by`, and `deployment-config` label matching

## Test plan

- [x] `pytest plugins/nemo-deployments/tests/unit` — 138 passed
- [x] `pytest plugins/nemo-deployments/tests/integration` — 4 passed (Docker daemon required)
- [x] `test_puller_server_prerequisite_chain` — uses deployment-scoped prerequisites (`puller` → `server`)
- [x] CodeRabbit review fixes from `d1270d0` included (port allocation, probe host resolution, label constants, test hardening)

## Release Notes

* **New Features**
  * Added Docker deployment backend registered by default, with GPU allocation pooling, automatic host port selection, container readiness probes, managed log retrieval (with truncation), and Docker volume create/read/delete support.
* **Bug Fixes**
  * Improved deployment idempotency and “deployment lost” handling.
  * Updated `status_in` filtering to use the stored nested `data.status` field.
* **Dependencies**
  * Added `httpx>=0.27`; introduced Docker extra pinned to `docker>=7.0`.
* **Tests**
  * Expanded Docker unit/integration coverage, including new GPU/labels/ports/probe/status tests and adjusted integration import paths.


tylersbray:
@coderabbitai full review

tylersbray:
## Smoke testing (local, post-rebase)

Ran manual smoke against a local platform on `localhost:8080` with the Docker executor configured in platform YAML (`deployments.default_executor` / `executors[].backend: docker`). **GPU not required.**

### Tier 1 — automated

```bash
uv sync --extra docker
uv run pytest plugins/nemo-deployments/tests -q
```

- **142 unit tests** passed (138 before the `data.status` filter fix)
- **4 Docker integration tests** passed against a real daemon:
  - `test_volume_lifecycle`
  - `test_never_deployment_succeeds`
  - `test_lost_detection_for_always` (with `drift_recovery.action: ignore`)
  - `test_puller_server_prerequisite_chain` (deployment-scoped prerequisites per #469)

### Tier 2 — platform REST E2E (manual)

Workspace `smoke`, scenarios exercised via the deployments v2 API:

| Scenario | Config | Expected outcome | Result |
|----------|--------|------------------|--------|
| **A** | `alpine:3.20`, `restart_policy: Never`, `echo smoke` | `PENDING` → `SUCCEEDED`, `exit_code: 0` | pass |
| **B** | `nginx:alpine`, host ports 9050–9060, `restart_policy: Always`, `drift_recovery: ignore` | `READY`, curl endpoint, `docker rm` container → `LOST` | pass |
| **C** | Volume `weights` + puller (`OnFailure`) → server (`Always`) with `prerequisites: [{deployment_name: puller, condition: succeeded}]` | puller `SUCCEEDED`, server `READY`/`STARTING` | pass |

### Issues found and fixed during smoke

1. **`data.status` filter bug** (`3901ed80`) — controller and `status_in` list API filtered on `status` instead of `data.status`, causing entity list queries to 500 and leaving `DELETING` deployments stuck. Fixed in this branch.
2. **Async delete ordering** — deployment DELETE is async (`DELETING` → entity removed by reconciler); config DELETE must wait for deployment to disappear.
3. **Platform config location** — executor config belongs in platform service YAML (`NMP_CONFIG_FILE_PATH` / `packages/nmp_platform_runner/.../local.yaml`), not `~/.config/nmp/config.yaml` (CLI kubeconfig).
4. **Full platform start** — `nemo services run --controller-group all` alone starts controllers without API services; use `nemo services run` (default) or `--service-group all --controller-group all`.

### Not run

- GPU / `nvidia.com/gpu` limits (optional Tier 3; needs Container Toolkit + GPU host)

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -0,0 +1,515 @@
...
+            ) from exc
+
+        self._docker = docker
+        self._docker_errors = docker_errors
+        self._executor_config = DockerExecutorConfig.model_validate(self._config)
+        self._entities = NemoEntitiesClient(AsyncEntitiesResource(self._sdk))
+        self._gpu_pool = get_shared_gpu_pool()
+        self._client = self._create_client()
+
+    def _create_client(self) -> Any:
```
benmccown: > **Suggestion: use `TYPE_CHECKING` to properly type the docker SDK objects throughout this module**
>
> The `Any` annotations here aren't defensible on the grounds that `docker` is optional — by the time any of these methods run, `init()` has already imported and validated the SDK. The loose typing is just an artifact of the lazy import pattern, and it hides real type information from the checker.
>
> Proposed fix in `backend.py`: add a `TYPE_CHECKING` guard at the top and annotate the class-level attributes:
>
> ```python
> from typing import TYPE_CHECKING
> if TYPE_CHECKING:
>     import docker
>     import docker.models.containers
>
> class DockerDeploymentBackend(DeploymentBackend):
>     _client: docker.DockerClient
>     _docker_errors: types.ModuleType  # or a Protocol if preferred
> ```
>
> This would let the following signatures be tightened:
>
> **`backend.py`**
> - `_create_client(self) -> Any` → `-> docker.DockerClient`
> - `_container_matches_deployment(self, container: Any, ...)` → `container: docker.models.containers.Container`
> - `_extract_host_ports(self, container: Any)` → same
> - `_endpoints_from_container_ports(self, container: Any, ...)` → same
> - `_build_endpoints(self, container_spec: Any, ...)` → `container_spec: Container` (the nemo entity type, already imported)
>
> **`gpu.py`**
> - `discover_managed_gpu_allocations(client: Any)` → `client: docker.DockerClient` (same `TYPE_CHECKING` guard)
>
> **`ports.py`** (already imports `docker` unconditionally at the top)
> - `collect_used_host_ports(containers: list[Any])` → `list[docker.models.containers.Container]`
>
> `parse_gpu_device_ids(device_requests: list[Any] | None)` in `gpu.py` is legitimately `Any` — it handles both raw dicts and SDK objects defensively, so that one's fine as-is.
>
> Note that `volumes.py` and `probes.py` already import `docker` unconditionally and are well-typed — no changes needed there.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -0,0 +1,515 @@
...
+
+    async def read_status(self, *, workspace: str, name: str) -> BackendStatusUpdate:
+        c_name = container_name(workspace, name)
+        try:
+            container = await asyncio.to_thread(self._client.containers.get, c_name)
+            await asyncio.to_thread(container.reload)
+        except self._docker_errors.NotFound:
+            restart_policy = await self._resolve_restart_policy(workspace, name)
+            return missing_container_status(restart_policy, container_name=c_name)
+        except Exception as exc:
```
benmccown, line 258: I asked claude to compare against models service's docker backend, and it flagged this:

> Transient Docker API errors (connection timeouts, `APIError`, `ReadTimeout`) are caught by the bare `except Exception` here and returned as `FAILED`, which is a terminal status. A momentary Docker daemon blip would permanently fail a healthy deployment.
>
> `DockerServiceBackend.get_model_deployment_status()` handles this by catching `APIError`, `ReadTimeout`, `Urllib3ReadTimeoutError`, and `RequestsConnectionError` explicitly and returning `UNKNOWN` (a recoverable status). Worth doing the same here — catch the specific Docker/requests exception types first and return a transient status before falling through to the generic `FAILED`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -0,0 +1,515 @@
...
+        self._executor_config = DockerExecutorConfig.model_validate(self._config)
+        self._entities = NemoEntitiesClient(AsyncEntitiesResource(self._sdk))
+        self._gpu_pool = get_shared_gpu_pool()
+        self._client = self._create_client()
+
+    def _create_client(self) -> Any:
+        kwargs: dict[str, Any] = {"timeout": self._executor_config.docker_timeout}
+        if self._executor_config.docker_host:
+            kwargs["base_url"] = self._executor_config.docker_host
+        return self._docker.from_env(**kwargs)
```
benmccown: I asked claude to compare against models service's docker backend, and it flagged this:

> `DockerServiceBackend` also sets `self._client.api.timeout = timeout` after calling `docker.from_env()`. The high-level SDK timeout and the low-level `api.timeout` (urllib3 socket timeout) are independent — without the second assignment, long-running Docker operations can silently hang past the configured timeout value. Worth adding:
> ```python
> client = self._docker.from_env(**kwargs)
> client.api.timeout = self._executor_config.docker_timeout
> return client
> ```

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -0,0 +1,515 @@
...
+                status=status,
+                status_message=message,
+                exit_code=exit_code,
+                endpoints=endpoints,
+            )
+
+        if state == "removing":
+            return BackendStatusUpdate(status="DELETING", status_message=f"Container removing (ID: {container_id})")
+
+        return BackendStatusUpdate(status="STARTING", status_message=f"Container state: {state}")
```
benmccown, line 345: I asked claude to compare against models service's docker backend, and it flagged this:

> The new backend has no pending/startup timeout — a container stuck in `STARTING` indefinitely will poll forever. `DockerServiceBackend` tracks wall-clock elapsed time and transitions to `ERROR` after a configurable threshold. This may be intentional if the reconciler is expected to own that concern, but worth explicitly confirming and documenting if so.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/config.py
```diff
@@ -0,0 +1,16 @@
...
+
+from pydantic import BaseModel, Field
+
+
+class DockerExecutorConfig(BaseModel):
+    """Knobs for a named docker executor instance (not entity backend_config)."""
+
+    docker_host: str | None = Field(default=None, description="Override DOCKER_HOST for this executor.")
+    docker_timeout: int = Field(default=60, ge=1)
+    pull_images: bool = Field(default=True, description="Pull container images before run when missing locally.")
```
benmccown, line 20: **Comment 1** — place on `config.py` line 16 (`docker_timeout: int = Field(default=60, ge=1)`):

> The models backend (`DockerBackendConfig`) defaults `docker_timeout` to 600s (10 minutes) because image pulls can take significantly longer than 60s for large images. This backend also pulls images synchronously via `asyncio.to_thread` in `create_deployment`, so a large image will silently time out the Docker client at 60s. Worth bumping the default or at minimum documenting why 60s is sufficient here.

---

**Comment 2** — place on `config.py` line 16 as well (or as a follow-up in the same thread):

> `port_range_start`/`port_range_end` currently live in `DockerDeploymentConfig` (the entity-level `backend_config.docker` struct). Host port allocation is an executor/host concern — it depends on what ports are available on the machine running the backend, not on what a specific deployment wants to expose. These should live in `DockerExecutorConfig` alongside `docker_host` and `docker_timeout`, with `DockerDeploymentConfig` restricted to container-level concerns (what ports the container listens on, network name, etc.).

---

**Comment 3** — place on `config.py` line 17 (`pull_images: bool`), or as a general file-level comment:

> The models backend has a `DockerNetworkingMode` enum covering `local`, `dond`, and `dind` topologies, which maps to how the host URL is constructed for probes and endpoints. The new backend handles this implicitly via `docker_host` + `NMP_LOOPBACK_ADDRESS`, which works but is less operator-discoverable. Worth considering whether to add a `networking_mode` field here (even if only `local` is supported in v1) so the config surface is self-documenting as more topologies are added.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
```diff
@@ -0,0 +1,83 @@
...
+
+from __future__ import annotations
+
+import hashlib
+import re
+
+from nemo_deployments_plugin.constants import MANAGED_BY_LABEL
+
+MANAGED_BY_KEY = "managed-by"
+DEPLOYMENT_WORKSPACE_LABEL = "nmp.nvidia.com/deployment-workspace"
```
benmccown: Maybe we should migrate all of these to just be `nemo.nvidia.com/...` instead of `nmp`? Only reason models service had that set (and didn't migrate to `nemo.nvidia.com`) was because it uses labels for discovering things and I didn't want to brick any already running deployments.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/gpu.py
```diff
@@ -0,0 +1,219 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class GPUAllocationError(Exception):
+    """Raised when GPU allocation fails due to insufficient resources."""
+
+
+@dataclass
+class GPUPoolStatus:
```
benmccown: nit/optional (not sure if the comment is off base or not):


**`GPUPoolStatus` dataclass is defined but never used**

`GPUPoolStatus` at lines 25-31 has no callers anywhere in the diff — no `status()` method on `DockerGPUPool`, no API endpoint, nothing. It looks like it was scaffolded for a future observability endpoint but got left in. Dead code in a new file is worth cleaning up before merge, or alternatively wiring it up to an actual `status()` method on `DockerGPUPool` if that was the intent.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
```diff
@@ -0,0 +1,83 @@
...
+DEPLOYMENT_NAME_LABEL = "nmp.nvidia.com/deployment-name"
+RESTART_POLICY_LABEL = "nmp.nvidia.com/restart-policy"
+CONFIG_NAME_LABEL = "nmp.nvidia.com/deployment-config"
+VOLUME_WORKSPACE_LABEL = "nmp.nvidia.com/volume-workspace"
+VOLUME_NAME_LABEL = "nmp.nvidia.com/volume-name"
+
+
+def k8s_safe_name(base_name: str, *, max_length: int = 63, suffix: str = "") -> str:
```
benmccown, line 22: I think I found a bug that already existed and was just ported over, but probs something worth capturing as a follow up linear issue to hit at some point. Edge case seems pretty unlikely to occur to me, in the short term.

---

> **Potential bug: `k8s_safe_name` can produce collisions in the common (non-truncated) case**
>
> `container_name` and `docker_volume_name` both call `k8s_safe_name` with a dash-joined concatenation of workspace and resource name (e.g. `f"dep-{workspace}-{deployment_name}"`). Because the separator between workspace and name is the same character (`-`) that gets normalized from any other special character, two different workspace/name pairs can produce identical inputs to `k8s_safe_name` — for example, `container_name("foo", "bar-baz")` and `container_name("foo-bar", "baz")` both feed `"dep-foo-bar-baz"` into the normalizer.
>
> The hash suffix exists to protect against exactly this, but it is only appended when truncation occurs. In the common case where the name fits within 63 characters, the hash is omitted and collisions are undetected.
>
> This likely exists in the models service already and is being carried forward here. Two possible fixes (trade-offs left to the implementer):
> 1. Always append the hash suffix unconditionally, not just on truncation.
> 2. Detect the naming conflict at runtime (Docker returns an error or a container with different labels already exists) and fall back to the hash-suffixed form — though this has its own complexity around recovery and idempotency.
>
> Suggest filing a follow-up issue rather than blocking this PR, since the same pattern is used in the existing models backend.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/probes.py
```diff
@@ -0,0 +1,151 @@
...
+    named_ports: dict[str, int] | None = None,
+) -> tuple[bool, str]:
+    """Return (ready, reason). When no probe is configured, running implies ready."""
+    if probe is None:
+        return True, "no readiness probe configured"
+
+    if probe.exec_action is not None and probe.exec_action.command:
+        return await _check_exec_probe(container, probe)
+
+    if probe.http_get is not None and host_url is not None:
```
benmccown: > When `host_url` is `None` (which happens when the container has no published ports), both the `http_get` and `tcp_socket` branches are silently skipped and the function falls through to `return True, "probe type not implemented; treating as ready"`. A deployment with an HTTP or TCP readiness probe but no host URL will be incorrectly reported as ready, and the message is misleading — it says "not implemented" when the real cause is "no host URL available". These branches should return `False` with a clear reason instead, e.g. `return False, "http probe requires a mapped host port — no host_url available"`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/probes.py
```diff
@@ -0,0 +1,151 @@
...
+    if target_port is None:
+        return False, f"tcp probe port not mapped: {port_value}"
+    timeout = probe.timeout_seconds
+
+    def _connect() -> None:
+        with socket.create_connection((host, target_port), timeout=timeout):
+            return
+
+    try:
+        await asyncio.wait_for(asyncio.to_thread(_connect), timeout=timeout)
```
benmccown, line 148: (claude flagged, looks like a minor nit)

> `socket.create_connection` is already called with `timeout=timeout`, and `asyncio.wait_for` wraps it with the same value again. The outer `wait_for` is redundant — the socket will time out first and raise `OSError`, which is caught by the bare `except Exception`. The same double-timeout pattern exists in `_check_exec_probe`. Pick one: either rely on the inner timeout and drop `wait_for`, or pass `timeout=None` to the blocking call and let `wait_for` own it. Mixing both with the same value works today but will cause confusion if the values ever diverge.

benmccown (approved):
Plenty of feedback but it's all nits or small items that you can make a judgment call on. I'm fine extracting work into a follow up PR as well.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/config.py
tylersbray: Tracked as a follow-up rather than blocking this PR:

- **Comment 2 (port range on executor config):** [AIRCORE-858](https://linear.app/nvidia/issue/AIRCORE-858/deployments-docker-move-port-range-to-dockerexecutorconfig)
- **Comment 3 (`networking_mode` enum):** [AIRCORE-859](https://linear.app/nvidia/issue/AIRCORE-859/deployments-docker-add-networking-mode-config-enum-localdonddind)

(Comment 1 — `docker_timeout` default — was addressed in 6410228d1c: default bumped to 600s.)

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
tylersbray: Filed follow-up as suggested: [AIRCORE-860 — Fix k8s_safe_name collision when name fits within 63 chars](https://linear.app/nvidia/issue/AIRCORE-860/fix-k8s-safe-name-collision-when-name-fits-within-63-chars)

Child of AIRCORE-756; shared with the models docker naming pattern.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
tylersbray: Agreed this should live outside v1 scope for now. Filed follow-up: [AIRCORE-861 — Deployments Docker: starting/pending timeout for STARTING state](https://linear.app/nvidia/issue/AIRCORE-861/deployments-docker-startingpending-timeout-for-starting-state)

v1 intentionally leaves stuck `STARTING` deployments polling until readiness passes or the container exits; the ticket covers adding elapsed-time timeout aligned with the models backend pattern.

PR #406, opened 2026-06-23
docs(evaluator): fix submit() packager, FilesetRef import, and leaked titles

SandyChapman:
## Summary

The Evaluator docs had drifted from the current `nemo_evaluator` plugin SDK contract, so several copy-paste examples failed before reaching execution. This PR realigns the runnable snippets and fixes two unrelated broken page titles found along the way.

### What changed
- **`submit()` requires a metric bundle packager.** Added `metric_bundle_packager=CloudpickleMetricBundlePackager()` (and its import) to every runnable `evaluator.submit(...)` example across `index`, `sdk-resources`, and the `metrics/*` + `tutorials/*` pages. `run()` examples are untouched — `run()` does not take the packager.
- **FilesetRef import fix.** The LLM Judge tutorial imported `FilesetRef` from the non-existent `nemo_evaluator_sdk.values`; it now imports from `nemo_evaluator.sdk` (an `ImportError` at the top of the tutorial before).
- **ModelRef documented.** `metrics/model-configuration.mdx` now states that local `run()` requires an inline `Model`, while remote `submit()` also accepts a `ModelRef` (`workspace/model-name`).
- **Stale validation script replaced.** `test_doc_examples.py` no longer calls the old `/v2/...` REST endpoints; it now runs offline contract checks for the SDK import paths and the `submit()` packager requirement.
- **Leaked title marker.** Two tutorial pages rendered `@nemo-nb: hide` as the page title (a `nemo_nb` cell marker that leaked into frontmatter `title`). Set real titles and removed the duplicate body H1, matching every other evaluator page.

### Not changed (investigated, no action needed)
- **Doc links / slugs.** The reported "stale slugs" are correct — Fern derives slugs from page titles, so `agentic-metrics`, `rag-metrics`, `bring-your-own-metric`, etc. resolve. `make docs-broken-links` passes.

## Testing
- `make docs-check` → 0 errors, 194 MDX files parse cleanly
- `make docs-broken-links` → all checks passed
- `ruff` + `ty` clean; new contract test passes 7/7
- Both fixed tutorial titles confirmed via the Fern dev-server render (correct `<title>`/`<h1>`, no `@nemo-nb`)

## Summary by CodeRabbit

* **Documentation**
  * Updated remote/durable job examples across evaluator docs and tutorials to consistently include `metric_bundle_packager=CloudpickleMetricBundlePackager()` when submitting jobs.
  * Expanded model configuration guidance to better explain local inline models vs remote `ModelRef` usage.
  * Refreshed tutorial code snippets for judge evaluations to include the metric bundle packager in remote submissions.
* **Tests**
  * Converted evaluator documentation checks from HTTP-based validation to offline contract verification, adding assertions around importability and submit-only `metric_bundle_packager` behavior.

### DIFF
--- a/docs/evaluator/metrics/model-configuration.mdx
+++ b/docs/evaluator/metrics/model-configuration.mdx
@@ -204,15 +204,16 @@ result = evaluator.run(metric=metric, dataset=dataset, target=model)
 
 ### `ModelRef` (supported by `evaluator.submit(...)`)
 
-Durable remote `evaluator.submit(...)` jobs additionally accept a `ModelRef` target. A `ModelRef` names a platform model entity (`workspace/model-name`) and is resolved by the evaluator backend when the job runs, so you do not have to resolve the endpoint yourself. Use this for platform-managed model routing:
+Durable remote `evaluator.submit(...)` jobs additionally accept a `ModelRef` target. A `ModelRef` names a platform model entity (`workspace/model-name`) and is resolved by the evaluator backend when the job runs, so you do not have to resolve the endpoint yourself. Use this for platform-managed model routing. A `ModelRef` target generates outputs online, so it requires an online run config (`RunConfigOnlineModel`):
 
 ```python
-from nemo_evaluator_sdk import ModelRef
+from nemo_evaluator_sdk import ModelRef, RunConfigOnlineModel
 from nemo_evaluator.shared.metric_bundles.cloudpickle import CloudpickleMetricBundlePackager
 
 job = evaluator.submit(
     metric=metric,
     dataset=dataset,
+    config=RunConfigOnlineModel(),
     target=ModelRef(root="default/my-model"),
     metric_bundle_packager=CloudpickleMetricBundlePackager(),
 )
--- a/docs/evaluator/metrics/model-configuration.mdx
+++ b/docs/evaluator/metrics/model-configuration.mdx
@@ -183,7 +183,42 @@ Use plain `RunConfig` for offline evaluations where the dataset already contains
 
 ## Model References
 
-The plugin SDK examples on this page use inline `Model` objects. If your deployment resolves platform model entities into model endpoint details, perform that lookup before constructing the `Model`, then pass the resulting inline model to the metric or request.
+You can supply the evaluation target two ways. Which one is valid depends on whether you run the evaluation locally or submit it as a durable platform job.
+
+### Inline `Model` (required for `evaluator.run(...)`)
+
+`evaluator.run(...)` executes in your local Python process, so it needs the resolved endpoint details inline. Always pass an inline `Model` as the `target` (or as a judge/embeddings field on the metric). If your deployment stores platform model entities, resolve the entity into endpoint details before constructing the `Model`:
+
+```python
+from nemo_evaluator_sdk import Model
+
+model_entity = client.models.retrieve("my-model", workspace="default")
+model = Model(
+    url=client.models.get_model_entity_route_openai_url(model_entity),
+    name="my-model",
+    api_key_secret="NVIDIA_API_KEY",
+)
+
+result = evaluator.run(metric=metric, dataset=dataset, target=model)
+```
+
+### `ModelRef` (supported by `evaluator.submit(...)`)
+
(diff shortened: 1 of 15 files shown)


gabwow (approved):

docs/evaluator/test_doc_examples.py
```diff
@@ -2,167 +2,125 @@
...
+    """Build an Evaluator resource without contacting any service.
+
+    Client construction and the ``submit`` argument guard are both offline; the
+    guard runs before any executor/HTTP work.
+    """
+    client = NeMoPlatform(base_url="http://localhost:8080", workspace="default")
+    return client.evaluator
+
+
+def test_submit_exposes_metric_bundle_packager_but_run_does_not() -> None:
```
arpitsardhana: seems like too weird and too long function name?

docs/evaluator/metrics/rag.mdx
```diff
@@ -61,7 +61,13 @@ evaluator: Evaluator = client.evaluator  # this object is an Evaluator resource
 Use `evaluator.run(metric=metric, dataset=dataset)` for a local synchronous evaluation. Use `evaluator.submit(metric=metric, dataset=dataset)` when you need a durable remote job:
 
 ```python
-job = evaluator.submit(metric=metric, dataset=dataset)
+from nemo_evaluator.shared.metric_bundles.cloudpickle import CloudpickleMetricBundlePackager
+
+job = evaluator.submit(
+    metric=metric,
+    dataset=dataset,
+    metric_bundle_packager=CloudpickleMetricBundlePackager(),
```
arpitsardhana, line 69: Are we deliberately requiring passing BundleManager even for metrics which "Service" is aware of?

arpitsardhana (approved):

docs/evaluator/metrics/model-configuration.mdx
```diff
@@ -183,7 +183,42 @@ Use plain `RunConfig` for offline evaluations where the dataset already contains
...
+```python
+from nemo_evaluator_sdk import ModelRef
+from nemo_evaluator.shared.metric_bundles.cloudpickle import CloudpickleMetricBundlePackager
+
+job = evaluator.submit(
+    metric=metric,
+    dataset=dataset,
+    target=ModelRef(root="default/my-model"),
+    metric_bundle_packager=CloudpickleMetricBundlePackager(),
+)
```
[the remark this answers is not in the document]
SandyChapman: Good catch on the substance — `_submit_params` raises `TypeError("ModelRef target requires RunConfigOnlineModel")`, so this example would have thrown. Fixed in 1d156b8.

Two corrections to the suggested diff, for the record: the public `evaluator.submit()` keyword is `config=` (not `params=`), and `RunConfigOnlineModel` is imported from `nemo_evaluator_sdk` (matching every other doc page), so the applied fix is:

```python
from nemo_evaluator_sdk import ModelRef, RunConfigOnlineModel
from nemo_evaluator.shared.metric_bundles.cloudpickle import CloudpickleMetricBundlePackager

job = evaluator.submit(
    metric=metric,
    dataset=dataset,
    config=RunConfigOnlineModel(),
    target=ModelRef(root="default/my-model"),
    metric_bundle_packager=CloudpickleMetricBundlePackager(),
)
```

docs/evaluator/test_doc_examples.py
SandyChapman: Renamed to `test_packager_param_is_submit_only` in 1d156b8.

docs/evaluator/metrics/rag.mdx
SandyChapman: Yes — today `Evaluator.submit()` requires it unconditionally: the guard raises before execution, `bundle_metric` doesn't special-case metrics the service already recognizes (built-in RAGAS types included), and `CloudpickleMetricBundlePackager` is the only packager. So the docs have to pass it or the example throws, which is why it's here.

Agreed it's messy, though — requiring an explicit cloudpickle bundler even for service-recognized metrics that need no local-code serialization isn't great. We should follow up on the SDK design (e.g. default/skip bundling for service-known metrics) so docs and users don't have to pass it where it adds nothing. Keeping it in the docs for now since it matches current behavior.

PR #408, opened 2026-06-23
ci: use action-semantic-pull-requests directly to update as needed

crookedstorm:
The version used in the shared template uses node 16, which will be shut down this week on github runners. Beyond that, there are some simplifications over the upstream template that seem helpful.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow for pull request validation with stricter title requirements, including a maximum character limit of 100 characters.


svvarom (approved):
LGTM

PR #413, opened 2026-06-23
feat: Support more Data Designer seed sources

mikeknep:
Adds support for `DirectorySeedSource` and `FileContentsSeedSource` seed datasets in remote contexts. 

Bumps DD to 0.8.0 and Anonymizer to 0.3.1

Also does some refactoring of seed-related validation logic that was kind of sprawling.

## Summary by CodeRabbit

* **New Features**
  * Added support for directory and file-contents seed sources, including paths under NeMo Platform filesets.
  * Preview and create flows now accept filesystem-backed seeds more consistently, whether local or remote.

* **Bug Fixes**
  * Improved validation for seed paths and fileset references, with clearer errors for missing files, empty paths, or invalid locations.
  * Local and remote seed handling now better distinguishes between local directories and fileset-backed paths.

* **Tests**
  * Expanded coverage for local/remote seed validation, preview behavior, and filesystem routing.


packages/data_designer_nemo/src/data_designer_nemo/seed.py
```diff
@@ -2,52 +2,167 @@
...
+        logger.exception(
+            "Error listing fileset path",
+            extra={"fileset_name": fileset_name, "workspace": workspace, "fragment": fragment},
+        )
+        raise NDDInternalError(
+            f"An unexpected error occurred while listing path {fragment!r} in fileset {fully_qualified_fileset_name!r}: {e}"
+        ) from e
+
+    if not response.data:
+        raise NDDInvalidConfigError(f"Path {fragment!r} not found in fileset {fully_qualified_fileset_name!r}")
```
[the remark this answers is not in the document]
mikeknep: I updated the error message for this case in 657f31fede. When `not response.data`, there is no valid seed data to use, so it is correct to consider this invalid.

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
```diff
@@ -0,0 +1,138 @@
...
+    SeedReaderFileSystemContext,
+)
+from data_designer_nemo.sdk_translation import async_to_sync_sdk
+from fsspec.implementations.dirfs import DirFileSystem
+from nemo_platform import AsyncNeMoPlatform, NeMoPlatform
+from nemo_platform.filesets import FilesetFileSystem, FilesetPathError, build_fileset_ref, parse_fileset_ref
+
+
+class _FilesetDirFileSystem(DirFileSystem):
+    """DirFileSystem that handles FilesetFileSystem's '#' path separator.
```
matthewgrossman, line 23: I realized that adding this `#` bespoke path separate was going to cause some of this, but I just wish we had somehow better centralized all the parsing :(

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
```diff
@@ -0,0 +1,138 @@
...
+
+        fs = FilesetFileSystem(self._sdk)
+        if fs.exists(root):
+            self._validated_roots.add(root)
+            return
+
+        fileset_root = build_fileset_ref("", workspace=workspace, fileset=fileset)
+        fully_qualified_fileset_name = f"{workspace}/{fileset}"
+        if not fs.exists(fileset_root):
+            raise SeedReaderConfigError(f"🛑 Fileset {fully_qualified_fileset_name!r} not found.")
```
matthewgrossman, line 106: I love the consistent usage of 🛑

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
```diff
@@ -0,0 +1,138 @@
...
+        return build_fileset_ref(fragment, workspace=workspace, fileset=fileset)
+
+    def _parse(self, runtime_path: str) -> tuple[str, str, str]:
+        try:
+            return parse_fileset_ref(runtime_path, workspace_fallback=self._workspace)
+        except FilesetPathError as error:
+            raise SeedReaderError(f"🛑 Invalid fileset seed source path {runtime_path!r}: {error}") from error
+
+
+class HybridFileSystemProvider:
```
matthewgrossman, line 124: I wonder if we just did a better job of using this https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally then you could just use the FilesetFileSystem OOTB

matthewgrossman (approved):
lgtm from a files perspective

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
mikeknep: This object manages the "local / `run` should support both local and remote resources" requirement.
- Running standalone library: `dd.DirectorySeedSource(path="./local/path")`
- Running platform plugin remotely: `dd.DirectorySeedSource(path="my-ws/my-fileset#some/subdir")`
- Running platform plugin locally: **either** of the above should ideally work

We inject a `FileSystemProvider(Protocol)` into the seed readers. In the remote context we provide one that only understands Fileset references (`FilesetFileSystem`); in the local context we provide this hybrid one that can manage both.

I don't think caching files locally would make a difference here; the `HybridFileSystemProvider` allows you to use local files that were never uploaded to the Files service in the first place.

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
mikeknep: To be fair I think the parsing **is** quite centralized as far as the core AbstractFileSystem implementation is concerned. This is me modifying one of the standard **decorators** (see [here](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.dirfs.DirFileSystem)) to work with Filesets. If anything else needs this functionality, we could lift it into the files package.

packages/data_designer_nemo/src/data_designer_nemo/unsupported_features.py
mikeknep, line 1: Extracted to feature-specific modules: `seed.py` and `tool_configs.py`

crookedstorm (approved):
Approving for wheel constraints.

packages/data_designer_nemo/src/data_designer_nemo/seed.py
```diff
@@ -2,56 +2,186 @@
...
+        raise NDDInvalidConfigError(f"Access denied to workspace {workspace!r}") from e
+    except Exception as e:
+        logger.exception("Error retrieving fileset", extra={"fileset_name": fileset_name, "workspace": workspace})
+        raise NDDInternalError(
+            f"An unexpected error occurred while retrieving fileset {fileset_name!r} in workspace {workspace!r}: {e}"
+        ) from e
+
+    canonical_root = build_fileset_ref(fragment, workspace=workspace, fileset=fileset_name)
+    if not fragment:
+        return canonical_root
```
mmogallapalli: Do we care about fileset-root seed paths with no files? For example, if a user sets a directory seed to `default/docs,` fragment is empty, so `validate_seed` returns after confirming the `docs` fileset exists and skips the `list_files`/empty-response check. A non-root path like `default/docs#corpus` does get that check. I don’t see an earlier Files service invariant that prevents empty filesets from existing, not sure if there is something on the DD side that handles this case before. Just wanted to flag incase

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        self._files_client: FilesClient | None = None
+        self._async_files_client: AsyncFilesClient | None = None
+        self._workspace = workspace
+        self._validated_roots = validated_roots or set()
+
+    def create_context(self, *, runtime_path: str) -> SeedReaderFileSystemContext:
+        root = self._canonical_root(runtime_path)
+        fs = self._get_filesystem()
+        rooted_fs = _FilesetDirFileSystem(path=root, fs=fs)
+        return SeedReaderFileSystemContext(fs=rooted_fs, root_path=PurePosixPath(root))
```
mmogallapalli: The custom `_FilesetDirFileSystem._relpath `handles root fileset paths correctly for enumeration, but do we also need to preserve canonical refs in `source_path`? For a root seed like `ws/fs`, `relative_path` becomes `data.parquet`, then `_build_metadata_record` computes `source_path` as `PurePosixPath("ws/fs") / "data.parquet"`, yielding `ws/fs/data.parquet` instead of `ws/fs#data.parquet`. Is that legacy slash form acceptable here, or should root fileset seed rows expose canonical fileset refs?

mmogallapalli (approved):
overall lgtm, just a few edge case questions

packages/data_designer_nemo/src/data_designer_nemo/seed.py
mikeknep: Great catch, thanks. Addressed in 6f4118580

packages/data_designer_nemo/src/data_designer_nemo/fileset_filesystem_provider.py
mikeknep: And another great catch here. Addressed in 652a06f61. I'm not wild about having to create a custom subclass of `PurePosixPath` here, but it does at least nicely mirror the custom subclass of `DirFileSystem` (which we can't avoid)

PR #421, opened 2026-06-23
fix(docs): update IGW target to be consistent with other docs

parkanzky:
* Update docs to point at IGW provider that is created by setup scripts.
* Add auditor seed job to necessary workspaces so that it works according to docs.

Note that running the platform in local mode will not automatically run the seed jobs. To test, I manually ran the full platform seed task as ```nemo-platform run task --task nmp.platform_seed``` , which now correctly discovers `AuditorSeedJob`. This isolates that particular problem. If `nmp.platform_seed` is not executed on startup where expected, that would be a separate bug.

## Summary by CodeRabbit

* **Documentation**
  * Updated auditor examples, target creation/update, the inference gateway worked example, and the local audit tutorial to use the `nvidia-build` inference gateway provider name instead of `build`.
* **Chores**
  * Updated platform seed service package metadata and added `nemo-auditor-plugin` as a new runtime dependency, including the matching workspace source entry.
  * Extended the platform seed service description to include “auth, guardrails, auditor, and plugin seed jobs.”

### DIFF
--- a/docs/auditor/sdk-resources.mdx
+++ b/docs/auditor/sdk-resources.mdx
@@ -119,7 +119,7 @@ target = auditor.targets.create(
     options={
         "nim": {
             "nmp_uri_spec": {
-                "inference_gateway": {"workspace": "default", "provider": "build"},
+                "inference_gateway": {"workspace": "default", "provider": "nvidia-build"},
             },
         },
     },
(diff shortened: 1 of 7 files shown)


gabwow (approved):

parkanzky:
bah working on the lint issues

PR #423, opened 2026-06-23
ci: add test builds for the auditor and safe synthesizer containers

crookedstorm:
Since these containers are involved in up coming releases, we need to ensure they build. These don't push anywhere because they aren't used in tests, and I'm not including them in the required jobs unless they are performant, but I wanted to get them in our CI pipelines here.

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD workflow with improved Docker build infrastructure for testing environments.


svvarom (approved):
LGTM

PR #430, opened 2026-06-24
chore: set telemetry env for safe-synthesizer

mckornfield:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed local runtime execution to correctly set the deployment type via environment variables for the spawned process.

* **Tests**
  * Added a unit test to verify the local command passes the expected environment variable to the runtime subprocess.


kendrickb-nvidia (approved):
There's not a container mode way to run NSS that we need to update yet? Only the local mode?

mckornfield:
correct, we have the container build setup but I'll integrate running it via a container this week/next

PR #432, opened 2026-06-24
feat(evaluator): add ATIF trace and log read handles for evidence

arpitsardhana:
## Summary

Adds the portable trace/log read layer for agent-eval evidence so metrics can score off what the agent actually did (its trace and logs), independent of the producing runtime. Rebased onto `main` after the filesystem handle / `run_verifier` / taskset work landed via #447, so this PR is now scoped to trace and log read handles only.

This PR **ingests** traces in ATIF (Agent Trajectory Interchange Format; RFC 0001). It does **not** normalize foreign formats — producers are responsible for emitting conformant ATIF, and the SDK validates on read and exposes typed handles.

- **Lightweight ATIF read models** (`values/atif.py`): `Trajectory` / `Step` / `ToolCall` / `Metrics` / `FinalMetrics` — a permissive (`extra="ignore"`) read view over just the subset of ATIF the evaluator consumes, rather than vendoring the full reference schema (avoids dependency/licensing bloat and stays forward-compatible).
- **Ingest = validate-on-read** (`values/evidence.py`): `parse_atif` validates a payload as a `Trajectory` (raises `ValidationError` if non-conformant). Trace evidence is persisted in its raw producer form; nothing is normalized at persist time.
- **Read handles**: `CandidateEvidence.trace()` → `TraceHandle` (`trace` / `steps` / `tool_calls` / `token_usage`) and `.logs()` → `LogHandle` (`list_files` / `read_text` / `tail`), both `async` and per-trial cached so sibling metrics share a materialized handle.
- **Well-known keys**: `WellKnownEvidenceKey` Literal (`initial_state` / `trace` / `logs` / `final_state` / `verifier_logs`).
- **Example reference metric** (`examples/run_agent_eval/example_metrics.py`, example-only): `inefficient_retry_loop` scoring off `TraceHandle` (consecutive-streak detection over a canonical tool-call key).
- Vendored SDK mirror synced via `make vendor`.

## Test plan

- [x] `ruff check` + `ruff format --check` clean on changed files
- [x] `ty check` — no new diagnostics in changed files
- [x] `pytest packages/nemo_evaluator_sdk/tests/agent_eval/` → 67 passed (trace/log handle reads, ATIF validation, example retry-loop metric)
- [x] `make vendor` regenerated mirror, no drift
- [ ] CI green

## Notes

- Additions live in `values/evidence.py` + `values/atif.py` (no separate value module) per review preference.
- Ingest-only: no normalization of NAT/OTel/OpenInference. Traces persist raw and are validated lazily by `TraceHandle` on first read.
- `agent_eval/evaluator.py` and `agent_eval/trials.py` are intentionally **not** modified by this PR — trace evidence is persisted as-is and read via `TraceHandle`.

### DIFF
--- a/packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
+++ b/packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
@@ -3,19 +3,23 @@
 
 """Reference metrics-over-evidence for this example (not SDK API).
 
-These show how to score from the SDK's filesystem evidence handle instead of a
-stamped verifier reward:
+These show how to score from the SDK's evidence handles instead of a stamped
+verifier reward:
 
 * :class:`TestsPassMetric` runs a command against ``final_state`` filesystem
   evidence (in a throwaway overlay) and scores on exit 0.
 * :class:`NoTestCheatingMetric` diffs ``initial_state`` against ``final_state``
   and fails if the agent touched protected (e.g. test) paths.
+* :class:`InefficientRetryLoopMetric` reads the normalized ``trace`` and fails
+  when the same tool call repeats past a threshold.
 """
 
 from __future__ import annotations
 
+import json
 from collections.abc import Sequence
 
+from nemo_evaluator_sdk.agent_eval.trials import EVIDENCE_FINAL_STATE, EVIDENCE_INITIAL_STATE, EVIDENCE_TRACE
 from nemo_evaluator_sdk.metrics.protocol import MetricInput, MetricOutput, MetricOutputSpec, MetricResult
 
 
@@ -26,7 +30,7 @@ class TestsPassMetric:
         self,
         command: Sequence[str],
         *,
-        evidence_name: str = "final_state",
+        evidence_name: str = EVIDENCE_FINAL_STATE,
         cwd: str = ".",
         timeout_s: float = 300.0,
     ) -> None:
@@ -60,13 +64,9 @@ class NoTestCheatingMetric:
         *,
         protected: Sequence[str] = ("tests/",),
         change_types: Sequence[str] = ("added", "modified", "deleted"),
-        initial_name: str = "initial_state",
-        final_name: str = "final_state",
     ) -> None:
         self._protected = tuple(protected)
         self._change_types = set(change_types)
-        self._initial_name = initial_name
-        self._final_name = final_name
 
     @property
     def type(self) -> str:
@@ -78,12 +78,56 @@ class NoTestCheatingMetric:
     async def compute_scores(self, input: MetricInput) -> MetricResult:
         evidence = input.candidate.evidence
         clean = True
-        if evidence is not None and evidence.get(self._initial_name) and evidence.get(self._final_name):
-            initial = await evidence.filesystem(self._initial_name)
-            final = await evidence.filesystem(self._final_name)
+        if evidence is not None and evidence.get(EVIDENCE_INITIAL_STATE) and evidence.get(EVIDENCE_FINAL_STATE):
+            initial = await evidence.filesystem(EVIDENCE_INITIAL_STATE)
+            final = await evidence.filesystem(EVIDENCE_FINAL_STATE)
             diff = await initial.diff(final)
             violations = [
                 entry for prefix in self._protected for entry in diff.changed(prefix=prefix, kinds=self._change_types)
             ]
             clean = not violations
         return MetricResult(outputs=[MetricOutput(name="no_test_cheating", value=clean)])
+
+
+class InefficientRetryLoopMetric:
+    """Score ``False`` when the same tool call repeats consecutively past ``threshold`` times."""
+
+    def __init__(self, *, threshold: int = 2) -> None:
+        self._threshold = threshold
+
+    @property
+    def type(self) -> str:
+        return "inefficient_retry_loop"
+
+    def output_spec(self) -> list[MetricOutputSpec]:
+        return [
+            MetricOutputSpec.boolean("efficient_tool_use"),
+            MetricOutputSpec.discrete_score("max_repeated_tool_calls"),
+        ]
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        max_repeats = 0
+        evidence = input.candidate.evidence
+        if evidence is not None and evidence.get(EVIDENCE_TRACE) is not None:
+            calls = await (await evidence.trace(EVIDENCE_TRACE)).tool_calls()
+            # Count the longest run of *consecutive* identical calls (a retry loop), not the
+            # global frequency, so legitimate reuse separated by other work isn't flagged.
+            previous_key: str | None = None
+            current_repeats = 0
+            for call in calls:
+                # Canonicalize for comparison only (sorted keys): semantically identical calls
+                # match regardless of argument insertion order; execution order is untouched.
+                key = json.dumps(
+                    {"function_name": call.function_name, "arguments": call.arguments or {}},
+                    sort_keys=True,
+                    separators=(",", ":"),
+                )
+                current_repeats = current_repeats + 1 if key == previous_key else 1
+                previous_key = key
+                max_repeats = max(max_repeats, current_repeats)
+        return MetricResult(
+            outputs=[
+                MetricOutput(name="efficient_tool_use", value=max_repeats <= self._threshold),
+                MetricOutput(name="max_repeated_tool_calls", value=max_repeats),
+            ]
+        )
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import asyncio
 import difflib
 import hashlib
+import json
 import os
 import shutil
 import signal
@@ -17,6 +18,11 @@ from typing import Any, Literal
 from urllib.parse import urlparse
 
 from pydantic import BaseModel, ConfigDict, Field, JsonValue, PrivateAttr, model_validator
+
+from nemo_evaluator_sdk.values.atif import FinalMetrics, Step, ToolCall, Trajectory
+
+# Well-known evidence keys (mirrored by the ``EVIDENCE_*`` constants in ``agent_eval.trials``).
+WellKnownEvidenceKey = Literal["initial_state", "trace", "logs", "final_state", "verifier_logs"]
 
 
 class FilesystemEntry(BaseModel):
@@ -100,11 +106,7 @@ class LocalFilesystemEvidence:
         return await asyncio.to_thread(path.read_text, encoding=encoding)
 
     async def iter_paths(self, relative_path: str | Path = ".", *, recursive: bool = False) -> list[str]:
-        """List entries (files *and* directories) rooted at ``relative_path``.
-
-        Use this to walk a subtree or test for (non-)emptiness. For a flat,
-        files-only listing matched by a glob pattern, use :meth:`list_files`.
-        """
+        """List entries (files and directories) rooted at ``relative_path``."""
         base = self.path(relative_path)
         return await asyncio.to_thread(self._iter_paths_sync, base, recursive)
 
@@ -120,11 +122,7 @@ class LocalFilesystemEvidence:
         return await asyncio.to_thread(path.read_bytes)
 
     async def list_files(self, pattern: str = "**/*") -> list[str]:
-        """List relative posix paths of files (not directories) matching ``pattern``.
-
-        Complements :meth:`iter_paths`: this is the flat, glob-filtered,
-        files-only view; ``iter_paths`` walks a subtree and includes directories.
-        """
+        """List relative posix paths of files (not directories) matching ``pattern``."""
         return await asyncio.to_thread(self._list_sync, pattern)
 
     def _list_sync(self, pattern: str) -> list[str]:
@@ -135,12 +133,7 @@ class LocalFilesystemEvidence:
         )
 
     async def diff(self, other: LocalFilesystemEvidence) -> FilesystemDiff:
-        """Diff this snapshot (before) against ``other`` (after) by file content hash.
-
-        Cost note: this hashes every file in both trees by reading each fully, so
-        it is O(total bytes). Fine for task-sized evidence; revisit (streamed
-        hashing / size+mtime prefilter) if used on large artifact trees.
-        """
+        """Diff this snapshot (before) against ``other`` (after) by file content hash."""
         return await asyncio.to_thread(self._diff_sync, other)
 
     def _diff_sync(self, other: LocalFilesystemEvidence) -> FilesystemDiff:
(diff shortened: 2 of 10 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -84,6 +213,249 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+
+    def __init__(self, descriptor: EvidenceDescriptor) -> None:
+        self._descriptor = descriptor
+        self._trace: AtifTrace | None = None
+
+    async def trace(self) -> AtifTrace:
+        """Return the normalized trace, reading and parsing on first access."""
+        if self._trace is None:
+            payload = await asyncio.to_thread(self._load_payload)
+            self._trace = normalize_trace(payload, source_format=self._descriptor.format)
```
arpitsardhana: fabric?

Trace handle need to understand source format

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -84,6 +213,249 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+    return _normalize_spans(spans, _OTEL_KEYS)
+
+
+def normalize_openinference_spans(spans: list[dict[str, Any]]) -> AtifTrace:
+    """Normalize OpenInference spans into an :class:`AtifTrace`."""
+    return _normalize_spans(spans, _OPENINFERENCE_KEYS)
+
+
+# (kind_attr, tool_kinds, tool_name_attr, prompt_attr, completion_attr)
+_OTEL_KEYS = (
```
arpitsardhana: Add codex and Claude format

See jsonl based multi -turn conversations

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -320,6 +328,286 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+    """Normalized trace: an ordered event list plus optional aggregate tokens."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    events: list[AtifEvent] = Field(default_factory=list)
+    tokens: AtifTokenUsage | None = None
+    metadata: dict[str, Any] = Field(default_factory=dict)
+
+
+def normalize_nat_trajectory(payload: dict[str, Any]) -> AtifTrace:
```
arpitsardhana: this normalisation is not accurate. need to be fixed

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
```diff
@@ -60,8 +63,8 @@ def __init__(
         *,
         protected: Sequence[str] = ("tests/",),
         change_types: Sequence[str] = ("added", "modified", "deleted"),
-        initial_name: str = "initial_state",
-        final_name: str = "final_state",
+        initial_name: str = EVIDENCE_INITIAL_STATE,
+        final_name: str = EVIDENCE_FINAL_STATE,
```
SandyChapman: Question, do you think these need to be injected here? When would the initial/final states have a different name?

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
```diff
@@ -87,3 +90,38 @@ async def compute_scores(self, input: MetricInput) -> MetricResult:
...
+        return [
+            MetricOutputSpec.boolean("efficient_tool_use"),
+            MetricOutputSpec.discrete_score("max_repeated_tool_calls"),
+        ]
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        max_repeats = 0
+        evidence = input.candidate.evidence
+        if evidence is not None and evidence.get(self._evidence_name) is not None:
+            calls = await (await evidence.trace(self._evidence_name)).tool_calls()
```
SandyChapman: Why would the user choose to get different evidence types in the metric? It seems like this always relies on it being `EVIDENCE_TRACE`, right?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -320,6 +338,307 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+
+    def __init__(self, descriptor: EvidenceDescriptor) -> None:
+        self._descriptor = descriptor
+        self._trajectory: Trajectory | None = None
+
+    async def trace(self) -> Trajectory:
+        """Return the normalized ATIF trajectory, reading and parsing on first access."""
+        if self._trajectory is None:
+            payload = await asyncio.to_thread(self._load_payload)
+            self._trajectory = normalize_trace(payload, source_format=self._descriptor.format)
```
SandyChapman: I don't think I quite understand the process in the PR. We end up normalizing the `EvidenceDescriptor` which converts the payload to atif, but we also call it here to normalize the trajectory. Do we need both? Or am I reading this incorrectly. I'd assume the normalizing happening here is sufficient and we don't need to pre-emptively normalize the evidence descriptor.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -327,7 +327,9 @@ def _trial_from_sample(task: AgentEvalTask, target: Model | Agent, sample: dict[
         # trial stays scorable instead of being dropped as empty output.
         output_text = _reasoning_content_fallback(sample.get("response"))
     if "trajectory" in sample:
-        trace = EvidenceDescriptor(kind="trace", format="json", data=sample["trajectory"])
+        # Normalize to ATIF before the trial is persisted so the stored shape is
+        # source-agnostic (sources in, ATIF out); TraceHandle then reads it uniformly.
+        trace = normalize_trace_descriptor(EvidenceDescriptor(kind="trace", format="json", data=sample["trajectory"]))
```
SandyChapman: I don't think this is the right approach tbh. I think persisting the raw trace and convert on metric usage is the right approach. The problem with converting before persisting is that if there's a bug in our conversion code or we don't handle the source format fully, the conversion may be lossy. At least if we convert on read, users can improve the conversion function without loss.

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
```diff
@@ -26,7 +29,7 @@ def __init__(
         self,
         command: Sequence[str],
         *,
-        evidence_name: str = "final_state",
+        evidence_name: str = EVIDENCE_FINAL_STATE,
```
ngoncharenko, line 33: would it make sense to ad hint for well known literals `Literal = [EVIDENCE_FINAL_STATE, ...] | str`?

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
```diff
@@ -87,3 +90,38 @@ async def compute_scores(self, input: MetricInput) -> MetricResult:
...
+        ]
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        max_repeats = 0
+        evidence = input.candidate.evidence
+        if evidence is not None and evidence.get(self._evidence_name) is not None:
+            calls = await (await evidence.trace(self._evidence_name)).tool_calls()
+            counts: dict[str, int] = {}
+            for call in calls:
+                key = f"{call.function_name}:{sorted((call.arguments or {}).items())}"
```
ngoncharenko: why do we sort call arguments? wouldn't it change the order of args which is important?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
ngoncharenko, line 324: can this be the case when it make tool call and has empty output_text as a result?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
ngoncharenko, line 269: out of scope, but would it make sense to have `kind: str | WellKnownKinds`? Have common kinds pre-defined in WellKnownKinds and str allows for extension

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -320,6 +338,307 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+_DEFAULT_AGENT = Agent(name="unknown", version="0")
+_VALID_SOURCES = {"system", "user", "agent"}
+
+
+def _coerce_source(value: Any) -> Literal["system", "user", "agent"]:
+    """Map a producer step source onto an ATIF source, defaulting unknowns to ``agent``."""
+    return value if value in _VALID_SOURCES else "agent"
+
+
+def normalize_nat_trajectory(payload: dict[str, Any]) -> Trajectory:
```
ngoncharenko: if we expect this to be NeMo Agent Toolkit trajectory, is it possible to use a structured type here instead of arbitrary dict?

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
arpitsardhana: Removed the name injection

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: Earlier, I was trying to do eager conversion. the underlying philosophy was Nemo evaluator. only works or understands ATIF so at all entry points, normalize to Atif and store.
So did the conversation as all points to cover online and offline cases which can use Trace handler conversion function.

I do realise that, it may not be best approach, rather just persist trace as is and have the normalised function available via trace handle

fixed above

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: Agreed, responded in previous comment

Fixed above to have trace only available via trace handle

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
arpitsardhana: Added WellKnownEvidenceKey Literal (initial_state/trace/logs/final_state/verifier_logs) in values/evidence.py

packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
arpitsardhana: Sort build a comparison key to detect identical consecutive calls. the purpose is demonstrate sample usage of trace

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
arpitsardhana: this file is now not in scope of changes since we are not doing eager Atif conversion.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: The normalisation part is removed

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: WellKnownEvidenceKey currently types evidence keys, not descriptor kinds. 
This is out of scope of this MR, can take it as followup

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -215,24 +198,12 @@ async def run_verifier(
...
-        is copied on every call, so verifying large evidence repeatedly is heavy.
-        """
+        """Run ``command`` (no shell) against a throwaway copy of the evidence; not a sandbox (host privileges)."""
         overlay = Path(tempfile.mkdtemp(prefix="evidence-verify-")).resolve()
         try:
             workdir = (overlay / cwd).resolve()
             if workdir != overlay and overlay not in workdir.parents:
                 raise ValueError(f"verifier cwd {cwd!r} resolves outside evidence overlay")
-            # symlinks=True copies links as-is (no host deref); the ignore hook drops
-            # links whose target escapes the evidence root so the verifier can't read them.
```
ngoncharenko, line 235: should we still leave this comment - it might be useful

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -320,6 +284,75 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
 
 
+# ATIF ingest: producers emit conformant ATIF (see values/atif.py, RFC 0001); we validate on read, not normalize.
+def parse_atif(payload: Any) -> Trajectory:
+    """Validate a payload as a canonical ATIF :class:`Trajectory` (raises ``ValidationError`` if non-conformant)."""
+    return payload if isinstance(payload, Trajectory) else Trajectory.model_validate(payload)
+
+
+class TraceHandle:
+    """Lazily validated read handle exposing a trace descriptor as an ATIF :class:`Trajectory`."""
```
ngoncharenko, line 301: is it specific to ATIF or this will be a generic TraceHandle? If specific do ATIF, maybe we should adjust the name to reflect that

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -320,6 +284,75 @@ def _requires_ref_or_data(self) -> EvidenceDescriptor:
...
+        return await self._fs.read_text(name)
+
+    async def tail(self, name: str, lines: int = 50) -> str:
+        """Return the last ``lines`` lines of a log file."""
+        text = await self._fs.read_text(name)
+        return "\n".join(text.splitlines()[-lines:])
+
+
 class CandidateEvidence(BaseModel):
     """Named evidence descriptors attached to an AgentEvalAttempt."""
```
ngoncharenko: out of scope but it would be nice to update: `AgentEvalAttempt` ---> `AgentEvalTrial`

ngoncharenko (approved):

SandyChapman (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: This is a generic TraceHandle. It recognizes ATIF format as of now, want to keep the name agnostic of trace format it handles

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: done

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
arpitsardhana: Added back these comments

PR #440, opened 2026-06-24
feat(jobs): add controller diagnostics and per-module E2E config pooling

ironcommit:
Add a `diagnostics` module to the jobs controller that collects a structured snapshot of job/step/task state for debugging stuck or failing jobs. Log snapshots at DEBUG level in the scheduler and reconciler, with a `include_job_logs_in_diagnostics` config flag (default off) to gate raw log inclusion since logs may contain secrets or PII.

Refactor the E2E harness to support per-module config selection via an `e2e_config()` pytest mark. Config layers (YAML files + inline overlays) are resolved to a canonical hash and pooled, so modules sharing the same effective config reuse one running services instance and it is torn down as soon as the last consumer finishes. Extract all service lifecycle logic from conftest.py into a new `services_pool.py` module.

Add `e2e/configs/local-subprocess.yaml` and a new `test_jobs_auth.py` suite that enables auth and exercises job principal propagation end-to-end using unsigned JWTs. Add unit tests for diagnostics, scheduler priority ordering, subprocess backend, and timestamp contracts.

## Summary by CodeRabbit

* **New Features**
  * Added `include_job_logs_in_diagnostics` (off by default) to optionally include raw job logs in diagnostics snapshots.
  * Introduced pooled local E2E services (with a Docker-less local config) and added auth-enabled E2E job coverage.
  * Added `e2e_config(...)` pytest marker support for selecting layered E2E configurations.

* **Bug Fixes**
  * Improved job step/task workspace scoping, scheduling/reconciliation diagnostics, conflict handling, and subprocess reconciliation behavior when metadata is missing.

* **Tests**
  * Expanded unit, integration, and E2E tests covering diagnostics, auth helpers, timestamp parsing, and the services pool.

### DIFF
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
@@ -57,6 +57,7 @@ SUBPROCESS_WORKDIR_STATUS_KEY = "subprocess_work_dir"
 SUBPROCESS_PID_STATUS_KEY = "pid"
 SUBPROCESS_PGID_STATUS_KEY = "pgid"
 SUBPROCESS_PERSISTENT_STORAGE_STATUS_KEY = "subprocess_persistent_storage_path"
+_MISSING_METADATA_PENDING_GRACE_SECONDS = 5
 SUBPROCESS_INHERITED_ENV_ALLOWLIST = frozenset(
     {
         "PATH",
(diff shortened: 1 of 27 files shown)


services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
```diff
@@ -281,6 +282,24 @@ def sync(self, step: PlatformJobStepWithContext) -> JobUpdate:
...
+            if step.status == PlatformJobStatus.PENDING and not self._pending_step_missing_metadata_is_stale(step):
+                # Stopgap only: the subprocess backend keeps execution metadata in controller
+                # memory, while step/task state is persisted in the jobs database. Those two
+                # sources of truth are not fully synchronized today because subprocess was not
+                # originally designed around durable jobs-backed execution state. That means a
+                # step can already be visible in the database as pending before this backend has
+                # registered local subprocess metadata for it. Keep the step pending briefly
+                # instead of failing it. The real fix is to move subprocess onto properly
+                # serialized, jobs-backed state so reconciliation does not depend on process-local
+                # memory.
```
matthewgrossman, line 297: agreed with this comment; so much of the subprocess backend was a hack, but I don't think making this change would be too big a lift

matthewgrossman (commented):
like the e2e config hashing mechanism!

matthewgrossman (approved):

PR #445, opened 2026-06-24
chore: update base images in case that helps with anything

crookedstorm:
Trying to see if rebuilds on some of these remove concerns in older packages.

## Summary by CodeRabbit

* **Chores**
  * Updated the default build image pins so the base and wheel references now point to newer versions.


svvarom (approved):
LGTM

PR #447, opened 2026-06-24
feat(evaluator): AgentEvalTaskset and filesystem evidence handle

arpitsardhana:
## Summary

First slice of the agent-eval P0 evidence/metrics work, split out for focused review.

- Adds `AgentEvalTaskset` — a named, validated collection of agent-eval tasks (unique task ids).
- Extends `LocalFilesystemEvidence` with `read_bytes`, `list(pattern)`, `diff`, and `run_verifier`.
  - `run_verifier` runs the command (no shell) against a throwaway overlay copy of the evidence, so stored evidence (pytest caches, build artifacts, ...) can never be mutated.
- Adds the `FilesystemEntry` / `FilesystemDiff` / `CommandResult` value types.
- Adds example-only `tests_pass` / `no_test_cheating` metrics that score over filesystem evidence (in `examples/run_agent_eval/example_metrics.py`).
- Vendored SDK mirror synced via `make vendor`.

The remaining ATIF trace + log read handles land separately in the stacked branch `aalgo-258-p0-evidence-metrics`.

## Test plan

- [x] `uv run ruff check` (changed files) — clean
- [x] `uv run --frozen ty check` on `values/evidence.py` — clean
- [x] `uv run --frozen pytest packages/nemo_evaluator_sdk/tests/agent_eval/{test_evidence,test_tasks,test_example_metrics}.py` — 9 passed

## Summary by CodeRabbit

* **New Features**
  * Added filesystem evidence utilities for safe reads, glob-based file listing, snapshot diffing, unified diffs, and running verifier commands in an isolated workspace.
  * Introduced reference metrics to assess test pass status and detect protected-path test cheating.
  * Added taskset support with validated task bundles and a standardized loader interface.
  * Expanded public evidence result types (command outcomes and filesystem change reporting).
* **Bug Fixes**
  * Strengthened traversal and symlink safety to prevent escaping evidence roots during evidence access and verifier runs.
* **Tests**
  * Added coverage for evidence APIs, verifier behavior (including timeouts and isolation), unified diff output, example metrics, and taskset validation.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
@@ -5,9 +5,10 @@
 
 from __future__ import annotations
 
+import importlib
 from enum import Enum
 from pathlib import Path
-from typing import Any
+from typing import Any, Protocol, TypeGuard, runtime_checkable
 
 from nemo_evaluator_sdk.metrics.protocol import Metric
 from nemo_evaluator_sdk.metrics.utils import metric_type_name
@@ -130,29 +131,78 @@ class AgentEvalTask(BaseModel):
         return self
 
 
-class AgentEvalTaskset(BaseModel):
-    """Named collection of tasks; ``id``/``name`` sugar around a task list."""
+class AgentEvalTasksetLoadConfig(BaseModel):
+    """Options passed from callers to a taskset's :meth:`AgentEvalTaskset.load`."""
 
     model_config = ConfigDict(extra="forbid")
 
-    id: str = Field(description="Stable taskset identifier.")
-    name: str | None = Field(default=None, description="Human-readable taskset name.")
-    tasks: list[AgentEvalTask] = Field(description="Tasks in this set; task ids must be unique.")
+    source: str | Path | None = Field(default=None, description="Optional source path/URI the taskset loads from.")
+    limit: int | None = Field(default=None, ge=0, description="Optional cap on the number of tasks to load.")
+    evidence_dir: Path | None = Field(default=None, description="Optional directory holding task evidence inputs.")
 
-    @field_validator("id")
-    @classmethod
-    def _id_must_not_be_empty(cls, value: str) -> str:
-        if not value:
-            raise ValueError("taskset id must not be empty")
-        return value
+
+class AgentEvalTasksetBundle(BaseModel):
+    """SDK-native tasks (and optional metadata) produced by a taskset's ``load()``."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    tasks: list[AgentEvalTask] = Field(default_factory=list, description="Loaded tasks; at least one is required.")
+    metadata: dict[str, Any] = Field(default_factory=dict, description="Free-form taskset metadata for the run.")
 
     @model_validator(mode="after")
-    def _task_ids_unique(self) -> AgentEvalTaskset:
-        ids = [task.id for task in self.tasks]
-        duplicates = sorted({task_id for task_id in ids if ids.count(task_id) > 1})
-        if duplicates:
-            raise ValueError(f"duplicate taskset task ids: {duplicates}")
+    def _require_tasks(self) -> AgentEvalTasksetBundle:
+        if not self.tasks:
+            raise ValueError("taskset bundles require at least one task")
         return self
+
+
+@runtime_checkable
+class AgentEvalTaskset(Protocol):
+    """Experimental protocol for adapting external tasksets into agent-eval.
+
+    A taskset is a named *adapter* that loads SDK-native tasks (optionally from an
+    external source). Per the design, tasksets are resolved upstream into
+    :class:`AgentEvalTask` values via :func:`resolve_agent_eval_taskset`; the
+    evaluator scores those tasks and never consumes a taskset directly.
+    """
+
+    @property
+    def name(self) -> str:
+        """Stable taskset name used in diagnostics, metadata, and user-facing output."""
+        ...
+
+    def load(self, config: AgentEvalTasksetLoadConfig) -> AgentEvalTasksetBundle:
+        """Load this taskset's tasks (and optional metadata) into SDK-native form."""
+        ...
+
+
+def _is_agent_eval_taskset(value: object) -> TypeGuard[AgentEvalTaskset]:
+    return isinstance(getattr(value, "name", None), str) and callable(getattr(value, "load", None))
+
+
+def resolve_agent_eval_taskset(ref: str) -> AgentEvalTaskset:
+    """Resolve a ``module:object`` taskset reference.
+
+    The referenced object may be a taskset instance, a taskset class, or a
+    zero-argument factory returning a taskset instance.
+    """
+    module_name, separator, object_path = ref.partition(":")
+    if not separator or not module_name or not object_path:
+        raise ValueError("taskset references must use module:object syntax")
+
+    resolved: Any = importlib.import_module(module_name)
+    for part in object_path.split("."):
+        resolved = getattr(resolved, part)
+
+    candidate = resolved
+    if isinstance(candidate, type):
+        candidate = candidate()
+    elif not _is_agent_eval_taskset(candidate) and callable(candidate):
+        candidate = candidate()
+
+    if not _is_agent_eval_taskset(candidate):
+        raise TypeError(f"resolved taskset {ref!r} does not implement AgentEvalTaskset")
+    return candidate
 
 
 class AgentEvalRunConfig(BaseModel):
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
@@ -130,6 +130,31 @@ class AgentEvalTask(BaseModel):
         return self
 
 
+class AgentEvalTaskset(BaseModel):
+    """Named collection of tasks; ``id``/``name`` sugar around a task list."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    id: str = Field(description="Stable taskset identifier.")
+    name: str | None = Field(default=None, description="Human-readable taskset name.")
+    tasks: list[AgentEvalTask] = Field(description="Tasks in this set; task ids must be unique.")
+
+    @field_validator("id")
+    @classmethod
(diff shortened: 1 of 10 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+    evidence_dir: Path | None = Field(default=None, description="Optional directory holding task evidence inputs.")
+
+
+class AgentEvalTasksetBundle(BaseModel):
+    """SDK-native tasks (and optional metadata) produced by a taskset's ``load()``."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    tasks: list[AgentEvalTask] = Field(default_factory=list, description="Loaded tasks; at least one is required.")
+    metadata: dict[str, Any] = Field(default_factory=dict, description="Free-form taskset metadata for the run.")
```
SandyChapman, line 147: This is likely a good idea. We should see if there are other fields we want to add to this beyond a bag of str->any. Things like created dates or versions maybe.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+
+    @model_validator(mode="after")
+    def _require_tasks(self) -> AgentEvalTasksetBundle:
+        if not self.tasks:
+            raise ValueError("taskset bundles require at least one task")
+        return self
```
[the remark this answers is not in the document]
SandyChapman: In addition to this comment, we can add `min_length=1` to the `Field` to ensure at least one task.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+    limit: int | None = Field(default=None, ge=0, description="Optional cap on the number of tasks to load.")
+    evidence_dir: Path | None = Field(default=None, description="Optional directory holding task evidence inputs.")
+
+
+class AgentEvalTasksetBundle(BaseModel):
+    """SDK-native tasks (and optional metadata) produced by a taskset's ``load()``."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    tasks: list[AgentEvalTask] = Field(default_factory=list, description="Loaded tasks; at least one is required.")
```
SandyChapman: Something to consider is how we'll want to implement the train/test/validation splits for the taskset. The requirement text was:

> Train / validate  / test splots
> The service must support splitting evaluation data into training, validation, and test sets and running a candidate against a named split.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+    """Options passed from callers to a taskset's :meth:`AgentEvalTaskset.load`."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    source: str | Path | None = Field(default=None, description="Optional source path/URI the taskset loads from.")
+    limit: int | None = Field(default=None, ge=0, description="Optional cap on the number of tasks to load.")
+    evidence_dir: Path | None = Field(default=None, description="Optional directory holding task evidence inputs.")
+
+
+class AgentEvalTasksetBundle(BaseModel):
```
SandyChapman: On naming, I think we should rename `AgentEvalTasksetBundle` to `AgentEvalTaskset` and `AgentEvalTaskset` to `AgentEvalTaskseLoader`. It may also simplify things if we ditched the `AgentEvalTasksetLoadConfig` and just passed `source`, `limit` and `evidence_dir` as params to the `load` function.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+    def load(self, config: AgentEvalTasksetLoadConfig) -> AgentEvalTasksetBundle:
+        """Load this taskset's tasks (and optional metadata) into SDK-native form."""
+        ...
+
+
+def _is_agent_eval_taskset(value: object) -> TypeGuard[AgentEvalTaskset]:
+    return isinstance(getattr(value, "name", None), str) and callable(getattr(value, "load", None))
+
+
+def resolve_agent_eval_taskset(ref: str) -> AgentEvalTaskset:
```
SandyChapman: Why wouldn't a user just import that module directly instead of using this dynamic module loading func?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -6,13 +6,62 @@
...
 from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
 
 
+class FilesystemEntry(BaseModel):
+    """One path that differs between two filesystem snapshots."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    path: str
+    change_type: Literal["added", "modified", "deleted"]
```
SandyChapman, line 28: If the file is modified, do we want to keep the actual diff? One thing you could consider is just using the `diff` CLI tool for these files to generate a patch (we'd still need a copy of both the initial file and the final file to run this).

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +131,80 @@ def _validate_metric_references(self) -> AgentEvalTask:
         return self
 
 
+class AgentEvalTasksetLoadConfig(BaseModel):
+    """Options passed from callers to a taskset's :meth:`AgentEvalTaskset.load`."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    source: str | Path | None = Field(default=None, description="Optional source path/URI the taskset loads from.")
+    limit: int | None = Field(default=None, ge=0, description="Optional cap on the number of tasks to load.")
```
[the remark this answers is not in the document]
arpitsardhana: No longer valid

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
arpitsardhana: Dropped AgentEvalTasksetLoadConfig
Renamed Protocol to Loader and removed bundle suffix

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
arpitsardhana: This needs to be followup once we have more concrete test case.

SandyChapman (approved):
lgtm! thanks for the edits

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -53,6 +113,167 @@ def _iter_paths_sync(self, base: Path, recursive: bool) -> list[str]:
...
+            # symlinks=True copies links as-is (no host deref); the ignore hook drops
+            # links whose target escapes the evidence root so the verifier can't read them.
+            await asyncio.to_thread(
+                shutil.copytree,
+                self._root,
+                overlay,
+                dirs_exist_ok=True,
+                symlinks=True,
+                ignore=self._ignore_escaping_symlinks,
+            )
```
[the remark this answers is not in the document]
arpitsardhana: Fixed in 5c2770560. `_ignore_escaping_symlinks` now also drops symlinks whose `os.readlink` target is absolute, so an in-root absolute link is no longer recreated in the overlay and a verifier write can't reach stored evidence. Covered by `test_absolute_symlinks_are_not_copied_into_verifier_overlay`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -53,6 +113,167 @@ def _iter_paths_sync(self, base: Path, recursive: bool) -> list[str]:
...
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
+        try:
+            stdout, stderr = await asyncio.wait_for(process.communicate(), timeout=timeout_s)
+        except TimeoutError:
+            # wait_for cancels communicate() but leaves the child running; kill it.
+            process.kill()
+            await process.wait()
+            return CommandResult(exit_code=-1, timed_out=True)
```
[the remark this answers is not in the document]
arpitsardhana: Fixed in 5c2770560. The verifier now launches with `start_new_session=True` and the timeout path calls `os.killpg(os.getpgid(pid), SIGKILL)` to reap the whole process group. Regression test `test_verifier_timeout_kills_the_whole_process_tree` spawns a backgrounded grandchild that would write a marker after 1s and asserts it never appears after a 0.3s timeout.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
arpitsardhana: removed it

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/tasks.py
```diff
@@ -130,6 +130,62 @@ def _validate_metric_references(self) -> AgentEvalTask:
...
+        """Stable taskset name used in diagnostics, metadata, and user-facing output."""
+        ...
+
+    def load(
+        self,
+        *,
+        source: str | Path | None = None,
+        limit: int | None = None,
+        evidence_dir: Path | None = None,
+    ) -> AgentEvalTaskset:
```
ngoncharenko, line 179: - it would be nice to allow loading specific list of rows - smth like row_ids = []
- should we have source URI typed?

PR #448, opened 2026-06-24
feat(experiments): sort the experiments list by rollup metrics

shanaiabuggy:
Sort the experiments list by rollup metrics (avg evaluator score, cost, latency, run count), not just entity columns.

Since those metrics live in ClickHouse and entity rows live in Postgres, the list endpoint now does a compute-on-read merge: fetch the group's experiments, hydrate their rollups, then sort/paginate in the app. `sort` accepts `run_count`, `cost_usd.<stat>`, `latency_ms.<stat>`, `evaluators.<name>.<stat>` (stat: mean/median/p90/p95/p99/sum/count), plus the existing entity columns; - prefixes descending, missing metrics sort last.

Bounded to ~1000 experiments per group (logged if exceeded) — the point at which we'd switch to denormalizing metrics for entity-store sorting instead.

## Summary by CodeRabbit

* **New Features**
  * Experiments list sorting now supports additional sort targets via a free-form `sort` expression, including `run_count`, `cost_usd.<stat>`, `latency_ms.<stat>`, and `evaluators.<name>.<stat>` (default: `-created_at`).
* **Bug Fixes**
  * Metric-based sorts now return clearer errors: `400` for unsupported sort expressions, `503` when telemetry rollups are unavailable, and `413` when too many experiments are selected.
  * Experiments with missing metric values reliably appear last with deterministic tie-breaking.
* **Tests**
  * Added unit and integration coverage for metric sorting, missing-data ordering, and the new error conditions.

### DIFF
--- a/services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
+++ b/services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
@@ -62,13 +62,13 @@ EXPERIMENTS_TAG = "Experiments"
 ExperimentGroupSortField = Literal["-created_at", "created_at", "-updated_at", "updated_at", "-name", "name"]
 
 # The experiments list is sorted in the application layer (compute-on-read) so a single request can
-# rank by a ClickHouse rollup metric, not just entity columns. `sort` is therefore a free string,
+# sort by a ClickHouse rollup metric, not just entity columns. `sort` is therefore a free string,
 # validated against these: an entity column, run_count, or a `<metric>.<stat>` rollup path.
 _ENTITY_SORT_FIELDS = frozenset({"name", "created_at", "updated_at", "pinned_at"})
 _METRIC_STATS = frozenset({"sum", "mean", "median", "p90", "p95", "p99", "count"})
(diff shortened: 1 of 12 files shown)


services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -352,24 +357,57 @@ async def list_experiments(
...
+        page_size=_MAX_GROUP_EXPERIMENTS,
     )
     responses = [ExperimentResponse.from_entity(e) for e in result.data]
-    await _hydrate_rollups(workspace=workspace, responses=responses, rollup_repository=rollup_repository)
+    if len(responses) >= _MAX_GROUP_EXPERIMENTS:
+        logger.warning(
+            "Experiment list hit the %d-row ranking cap for this filter; the tail is not ranked. "
+            "Consider denormalizing rollup metrics for entity-store sorting.",
+            _MAX_GROUP_EXPERIMENTS,
+        )
```
callingmedic911, line 398: Consider failing loudly than silently logging partial data.

callingmedic911 (approved):
I think we should fail loudly than work with subset of data. Looks good otherwise!

PR #456, opened 2026-06-25
feat(evaluator): publish_to_intake entrypoint (D9)

SandyChapman:
## What

Adds `publish_to_intake` (D9, [AALGO-290](https://linear.app/nvidia/issue/AALGO-290)) — the explicit, post-run consumer of `AgentEvalResult` that writes a completed agent evaluation to Intake. Builds on the D8 mapping module ([#443](https://github.com/NVIDIA-NeMo/nemo-platform/pull/443)).

```python
report = await publish_to_intake(
    result, platform=client, experiment_id="my-exp",
    workspace=..., agent_name=..., model_name=...,
)
```

Per trial (concurrent, bounded by a semaphore): `POST /ingest/atif` → resolve the root span via the `traces` query-back (design §3.5 option 1) → one `POST /evaluator-results` per metric output. Returns a `PublishReport` (per-trial session/span ids + result counts).

## Design notes

- **No bespoke HTTP client.** All shapes come from the D8 mapping (`*CreateParams` TypedDicts); all wire calls go through the generated SDK `intake` resources (`client.intake.ingest.atif`, `.traces`, `.evaluator_results`). It imports the SDK client types, never the Intake service.
- **References an existing Experiment; never creates one** — the caller pre-creates the group/experiment via the Experiments SDK (`exist_ok=True`). Ingest 400s on an unknown experiment and we surface it.
- **Agent identity is a parameter** (`agent_name`/`agent_version`/`model_name`) because it lives on the run *target*, which `AgentEvalResult` doesn't carry (§3.9 #6).
- Raises on HTTP/validation failure; the local bundle is never touched.

## Testing

- **Unit** (`tests/intake/test_publish.py`, fake async client): round-trip, multi-trial, score-less trial, workspace resolution, span-resolution failure, ingest-failure propagation.
- **Live integration** (`tests/integration/test_publish_to_intake.py`): stands up ClickHouse (Docker) + `auth,entities,intake` via session fixtures, runs the real round-trip, and asserts every field read back through the public Intake API — `experiment_context.{experiment_id,test_case_id}`, `root_span_id`, and each evaluator-result `name`/`data_type`/`value`/`string_value`/`session_id`/`span_id`. Marked `integration` (excluded from the unit suite; skips when Docker is absent). Verified end-to-end locally (~18s).

`ruff`, `ty`, and 28 unit tests green.

## Notes / follow-ups

- Re-publish is **not** idempotent (`/evaluator-results` mints a fresh id per call — ask X1); MVP stance is write-once per result.
- The integration test **spawns the real platform** rather than the in-process testcontainers/ASGI style used elsewhere — works, but heavier; can be aligned later if CI flakes.
- D3/D4/D5 were *not* hard dependencies of D9; whether to keep `publish.py` inline or refactor into adapter layers is a separate, open discussion (tracked, not decided here).

Refs: [AALGO-290](https://linear.app/nvidia/issue/AALGO-290).

## Summary by CodeRabbit

* **New Features**
  * Added an evaluator-to-Intake publish flow that returns a publish report including published trials and skipped items.
  * Added filtering and reporting for non-finite metric outputs and FAILED scoring outputs.
  * Added trace span resolution to correctly associate published rows back to each trial.
* **Bug Fixes**
  * Improved robustness for missing workspace values, unresolvable trace spans, and ingestion/publish partial failures.
* **Tests**
  * Expanded unit tests for multi-trial publishing, empty-score behavior, and skip reasons.
  * Added end-to-end coverage using a live Intake service (with an optional integration test marker).

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
@@ -1,0 +1,154 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Publish a completed agent evaluation to Intake.
+
+``publish_to_intake`` is the explicit, post-run consumer of ``AgentEvalResult``
+(see AALGO-290). It is **not** a side effect of ``AgentEvaluator.run()`` and
+there is no feature flag — optionality is structural: you make the call or you
+don't, and the platform client is a required argument.
+
+It references an **existing** Experiment (created by the caller via the platform
+Experiments SDK) and never creates one. Per Trial it: POSTs the ATIF trajectory,
+resolves the trajectory's root span, then POSTs one evaluator-result per metric
+output. All request shapes come from :mod:`nemo_evaluator.intake.mapping`; the
+HTTP calls go through the generated platform SDK's ``intake`` resources.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections import defaultdict
+
+from nemo_evaluator.intake import mapping
+from nemo_evaluator.sdk import http_utils
+from nemo_evaluator_sdk.agent_eval.results import AgentEvalResult
+from nemo_evaluator_sdk.agent_eval.scores import AgentEvalTaskScore
+from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial
+from nemo_platform import AsyncNeMoPlatform
+from nemo_platform.types.intake.trace_filter_param import TraceFilterParam
+from pydantic import BaseModel, ConfigDict, Field
+
+#: Default ceiling on concurrent per-trial publishes.
+DEFAULT_MAX_CONCURRENCY = 8
+
+
+class PublishError(RuntimeError):
+    """Raised when publishing cannot complete (e.g. a trajectory's span never resolves)."""
+
+
+class PublishedTrial(BaseModel):
+    """Record of one Trial written to Intake."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    trial_id: str = Field(description="Identifier of the published trial.")
+    session_id: str = Field(description="Intake session id minted for the trajectory.")
+    span_id: str = Field(description="Resolved root AGENT span id the scores were attached to.")
+    evaluator_result_count: int = Field(description="Number of evaluator-result rows written for this trial.")
+
+
+class PublishReport(BaseModel):
+    """Summary of a ``publish_to_intake`` run."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    experiment_id: str = Field(description="Experiment the results were published under.")
+    workspace: str = Field(description="Workspace the writes targeted.")
+    run_id: str = Field(description="Source AgentEvalResult run id.")
+    published_trials: list[PublishedTrial] = Field(
+        default_factory=list, description="Per-trial records of what was written."
+    )
+
+    @property
+    def trial_count(self) -> int:
+        """Number of trials published."""
+        return len(self.published_trials)
+
+    @property
+    def evaluator_result_count(self) -> int:
+        """Total evaluator-result rows written across all trials."""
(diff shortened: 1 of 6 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
```diff
@@ -0,0 +1,154 @@
...
+                    written += 1
+
+            return PublishedTrial(
+                trial_id=trial.id,
+                session_id=session_id,
+                span_id=span_id,
+                evaluator_result_count=written,
+            )
+
+    published = await asyncio.gather(*(_publish_trial(trial) for trial in result.trials))
```
arpitsardhana: One downside of gather is if one trial publish fails, other might still continue leading the state in intake to be inconsistent. something like Taskgroup can help

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
```diff
@@ -0,0 +1,154 @@
...
+            body["workspace"] = resolved_workspace
+            await platform.intake.ingest.atif.create(**body)
+
+            session_id = mapping.session_id_for(result.run_id, trial.id)
+            span_id = await _resolve_root_span_id(platform, workspace=resolved_workspace, session_id=session_id)
+
+            written = 0
+            for score in scores_by_trial.get(trial.id, []):
+                for row in mapping.score_to_evaluator_results(score, session_id=session_id, span_id=span_id):
+                    row["workspace"] = resolved_workspace
```
arpitsardhana, line 157: nit: Does failed or NaN score require any special processing before being stored in intake?

arpitsardhana (commented):
There can be potential of intake corruption of some trials of experiment get persisted while other did not

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
SandyChapman: We went with resilient fan-out (`gather(return_exceptions=True)`) rather than `TaskGroup` on purpose. Intake has no rollback/delete, so cancelling siblings when one trial fails wouldn't undo what already landed — it'd just publish *less* and leave more to redo. Partial uploads are acceptable here, so every trial that can land does; failures are collected and raised together as `PublishError` carrying the partial `PublishReport`, with an actionable message pointing at the cached bundle so you can re-run to publish the rest. Real consistency on re-publish belongs in idempotent ingest on the Intake side (cross-team ask X1), not in fail-fast here.

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
SandyChapman: Good catch — yes. A FAILED score isn't a real measurement, and NaN/inf isn't JSON-representable (the platform client's encoder rejects it before it ever hits the wire), so neither can be stored as-is. `score_to_evaluator_results` now returns `(rows, skipped)`: failed and non-finite outputs are skipped and surfaced in `PublishReport.skipped` rather than silently dropped. Intake currently has no way to represent a *failed* metric result (can't distinguish failed from missing), so until it can we skip + report — tracked as cross-team ask X6, with a `TODO(X6)` in the mapper.

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
arpitsardhana: So Intake does not have concept of "atomic commits"? 

Fair point on right place for consistency

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
arpitsardhana: cool

arpitsardhana (approved):

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
SandyChapman: Yeah, I'm going to revisit the intake side and maybe fix this ourselves. They ideally provide a batch insert endpoint for this so we can atomically insert all results (or fail).

PR #463, opened 2026-06-25
fix: nss image, remove ffmpeg, force wandb version

mckornfield:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated container builds to use a newer Python base image and applied a consistent package pin to improve build reliability.
  * Adjusted cleanup steps to remove outdated bundled binaries more safely across environments.

* **Chores**
  * Added an explicit version pin for a key package in multiple build configurations to keep dependency behavior consistent.

### DIFF
--- a/docker/Dockerfile.safe-synthesizer-tasks
+++ b/docker/Dockerfile.safe-synthesizer-tasks
@@ -9,7 +9,7 @@
 # Global ARGs (must be declared before first FROM to be usable in FROM statements)
 # =============================================================================
 ARG PYTHON_VERSION=3.13
-ARG PYTHON_IMAGE=python:${PYTHON_VERSION}-slim-bookworm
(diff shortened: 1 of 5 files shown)


docker/Dockerfile.safe-synthesizer-tasks
```diff
@@ -9,7 +9,7 @@
 # Global ARGs (must be declared before first FROM to be usable in FROM statements)
 # =============================================================================
 ARG PYTHON_VERSION=3.13
-ARG PYTHON_IMAGE=python:${PYTHON_VERSION}-slim-bookworm
+ARG PYTHON_IMAGE=python:${PYTHON_VERSION}-slim-trixie
```
mckornfield, line 12: this I believe should address some random base packages

ironcommit (approved):

PR #470, opened 2026-06-25
feat(e2e): Run e2e tests in kube

matthewgrossman:
## Summary

Validate and adapt the data designer and jobs e2e tests for Kubernetes, fix a launcher log-loss bug, and add a CI job that runs the full e2e suite against a Kind cluster on every PR.

### Data Designer K8s fix
- DD's bridge reads `NEMO_JOB_PERSISTENT_JOB_STORAGE_PATH` from the environment, but on K8s the env var is only injected when the step declares it in `compile()`. DD never declared it, so the bridge crashed with `KeyError` on every K8s run.
- Rather than adding the env var to DD's `compile()` (which would mount a PVC that DD never uses), we made `StoragePaths.persistent` a guarding property that raises a clear `RuntimeError` when accessed without being provisioned. The bridge and dispatcher now tolerate the missing env var by passing `None`.
- This eliminates the footgun: forgetting the env var in `compile()` no longer crashes the pod — it raises a descriptive error only if the job actually tries to use persistent storage.

### Jobs launcher log reliability fix
Two bugs in the Go jobs-launcher caused ~10-20% of fast-exiting K8s job pods to silently lose their OTLP logs:
1. **Pipe read race**: `cmd.Wait()` was called before stdout/stderr reader goroutines finished. Go's `exec.Cmd.Wait()` closes pipes on return, so readers got "file already closed" and missed the output entirely. Fixed by calling `wg.Wait()` before `cmd.Wait()`.
2. **`os.Exit` skipping defers**: The cobra `Run` callback called `os.Exit(exitCode)`, which skipped the deferred `otelShutdown` that flushes the OTLP batch processor. Fixed by stashing the exit code and calling `os.Exit` from `Execute()` after cobra returns.

### E2e test K8s compatibility
- `test_job_passing_data_between_steps`: Added persistent storage env var to step environment (K8s backend requires explicit declaration).
- Container-backend tests (`pause/resume`, `invalid_image`): Changed from unconditional `@pytest.mark.skip` to `@pytest.mark.container_only` — now run on K8s, skip on subprocess.
- `test_job_pause_resume`: Cancel after verifying resume works instead of waiting for `sleep 300` (K8s restarts the timer on pod re-create).
- Smoke health tests: Use `/status` (works on both subprocess and K8s) instead of `/health/ready` (internal-only on K8s).
- Added `subprocess_only` and `container_only` pytest markers with auto-skip logic in conftest.
- Auth tests marked `subprocess_only` (require auth-enabled platform config).
- Artifact download timeout increased from 60s to 300s for K8s pod scheduling.

### CI: `kind-cpu-e2e` job + `setup-kind-cluster` composite action
- New `kind-cpu-e2e` workflow job runs the full `e2e/` directory against a Kind cluster (not just hardcoded files).
- Extracted `setup-kind-cluster` composite action from `kind-cpu-smoke` to deduplicate ~165 lines of cluster setup (kind/kubectl/helm install, cluster creation, gateway verification, image pre-pull, helm deploy, API readiness).
- Added to `ci-status` so it gates PRs.

## Test plan
- [x] 14/14 jobs + data designer tests pass on minikube (1 skipped: additional_volumes)
- [x] 62/66 full e2e suite passes on minikube (4 skipped: 3 auth, 1 additional_volumes)
- [x] Subprocess mode: no regressions (container_only tests properly skipped)
- [x] Launcher log fix: 0/50 log misses on minikube (previously 4-7/30)
- [x] Go launcher unit tests pass
- [x] Python unit tests pass (dispatcher, DD create job)
- [x] `kind-cpu-e2e` CI job passes
- [x] `kind-cpu-smoke` CI job passes
- [x] `python-e2e-test` CI job passes

Closes AIRCORE-844
Closes AIRCORE-845


plugins/nemo-data-designer/src/nemo_data_designer_plugin/jobs/create.py
```diff
@@ -83,7 +85,12 @@ async def compile(
                         ),
                     ),
                     config=spec.model_dump(),
-                    environment=[],
+                    environment=[
+                        EnvironmentVariable(
+                            name=PERSISTENT_JOB_STORAGE_PATH_ENVVAR,
+                            value=DEFAULT_JOB_STORAGE_PATH,
+                        ),
+                    ],
```
mikeknep: TLDR: Should we also update the DD run function [here](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/plugins/nemo-data-designer/src/nemo_data_designer_plugin/jobs/run.py#L76) to use `ctx.storage.persistent`?

---

Currently the Data Designer job does not use the persistent storage, it only uses ephemeral storage (see link above). I made this decision based on the [StoragePaths docstring](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/packages/nemo_platform_plugin/src/nemo_platform_plugin/job_context.py#L53-L58): Data Designer jobs are single-step, so I figured the persistent storage wasn't necessary. (By the way, per Claude, **every** job that currently exists across all plugins is currently single-step...)

So my first thought was, should `ctx.storage.persistent` be optional (and we'd use a `get` in the bridge module that reads the env vars) so that jobs that don't require persistent storage don't needlessly create/mount a PVC?

However, I then thought, actually, should the Data Designer job be using persistent storage instead of ephemeral storage? Perhaps the docstring is wrong and the decision is less about the data _lifecycle_ (i.e. surviving across steps) and more about _scale_... I guess what I'm really asking here is: if I have a DD job generating a large number of records (with potentially large assets like images), can I trust ephemeral storage to handle it? **My instinct is no**, and that actually ephemeral storage is likely intended for known/bounded things like "I need to write a simple config file somewhere before executing some function", whereas **results** (unbounded size) should always be written to persistent storage. 

I asked Claude about this and received this response (edited for brevity, emphasis mine):

> The size concern goes the opposite direction from what you might expect.
>
> - Ephemeral = K8s emptyDir, no size limit set — bounded only by whatever free disk the node has
> - Persistent = a 200Gi shared PVC across all jobs
>
> So ephemeral isn't the constrained one. For a large image dataset, writing to persistent would actually consume a chunk of the shared 200Gi pool that every other concurrent job is also drawing from. Writing to ephemeral and then pushing via ctx.results.save() to a fileset (presumably object storage) is the right pattern — the fileset backend is effectively unbounded, and ephemeral is just a local staging area.
>
 > **The real risk with large image outputs is that the emptyDir could exhaust the node's local disk before the upload completes, but that's a node sizing concern, not a reason to route through the shared PVC.**
>
>  So Data Designer's current approach — ephemeral staging → results.save() to fileset — is actually the correct pattern for large outputs. The docstring should be updated to surface the shared/bounded nature of persistent storage so future plugin authors don't reach for it thinking it's the "big storage" option.

🤨  I'm not convinced. Specifically calling out the part I bolded—the size of the output/result dataset shouldn't dictate the size of the node/pod we request, right?? I'd think we'd rather schedule the job on a very minimal node and write the generated dataset artifacts somewhere larger/scalable along the way. (Also, presumably that 200Gi value is configurable?)

Do the artifacts written by a job to persistent storage get cleared when the whole job (potentially many steps; in practice just one) finishes?

matthewgrossman: resolved via huddle;

tldr we're going to allow for a None persistent storage, and then error out if you try to use it when you didn't mean to

.github/workflows/ci.yaml
```diff
@@ -391,162 +391,126 @@ jobs:
...
         env:
+          _TYPER_FORCE_DISABLE_TERMINAL: "1"
+          E2E_SERVICES_LOG_DIR: ${{ runner.temp }}/e2e-services-logs
           NGC_API_KEY: not-used-for-ghcr-cpu-smoke
-        run: bash "${K8S_E2E_SCRIPTS}/setup_local_kind_cpu.sh"
+        run: |
+          test -n "${NMP_E2E_CLUSTER_URL}"
+          export NMP_BASE_URL="${NMP_E2E_CLUSTER_URL}"
+          uv run --frozen pytest \
+            e2e/test_jobs.py::test_job_using_secret_environment_variable \
```
matthewgrossman, line 420: this was what the existing smoke test ran because we knew it should work. we're still keeping around the smoke test for now

.github/workflows/ci.yaml
```diff
@@ -1488,6 +1451,7 @@ jobs:
       - docker-bake-graph
       - build-cpu-smoke-images
       - kind-cpu-smoke
+      - kind-cpu-e2e
```
crookedstorm: If the cpu-e2e is a required job, that kind of makes the cpu-smoke less useful to ALSO run. Unless you think there is value in proving quickly that the most basic possible option does work. The only thing that does is a quick visual cue, imo. It seems unnecessary though if we introduce a more complete test. Just challenging the idea. We can certainly keep both so there's options to shuffle as needed.

crookedstorm (approved):
LGTM from the standpoint of the github segments with concerns mentioned inline for discussion and thought, not blocking.

mikeknep (approved):
`StoragePaths` changes LGTM!

services/core/jobs/jobs-launcher/cmd/run.go
```diff
@@ -253,23 +259,15 @@ func runExec(args []string, stdinReader io.Reader, loggerProvider *log.LoggerPro
...
+	// Once readers finish, all log records have been submitted to the OTEL
+	// batch processor. The deferred otelShutdown in runExecWithStdin flushes
+	// remaining batches before the process exits.
 	wg.Wait()
 
-	// Force flush the OTEL pipeline to ensure all batched logs are exported
-	// This is especially important for short-lived jobs that fail quickly, where the
-	// batch processor may still have pending logs that haven't been exported yet
-	if loggerProvider != nil {
-		flushCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
ironcommit, line 267: glad we are getting rid of this, thanks for chasing this bug down.

.github/workflows/ci.yaml
matthewgrossman: ya it's a good point, I think we probably dont want this here for now. I'm a bit nervous of how slow they might be, so maybe just keeping around `kind-cpu-smoke` to make sure _something_ can run?

matthewgrossman: removed, ty for the comment

PR #471, opened 2026-06-25
chore: set up dynamic versioning hatch plugin

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Platform/SDK/service/plugin/tooling version metadata is now resolved dynamically at build time using git tag–based rules.
  * Release and build workflows now pass the resolved version through so wheel creation uses the same value.
* **Bug Fixes**
  * Version resolution now fails fast on invalid inputs, and wheel builds are aligned to the resolved version.
  * OpenAPI `info.version` is normalized to `0.0.0`, with consistent runtime version fallbacks across packages.
* **Documentation**
  * Updated release documentation to reflect the stable release process and build-time dynamic versioning.
* **Tests**
  * Extended unit tests to verify tag-based resolution, fallbacks, and CLI output behavior.


.github/workflows/release-bundle.yaml
```diff
@@ -430,6 +430,8 @@ jobs:
           repository: ${{ needs.plan-release.outputs.source_repo }}
           ref: ${{ needs.plan-release.outputs.source_sha }}
           path: source
+          fetch-depth: 0
```
matthewgrossman, line 433: this just speed things up?

packages/data_designer_nemo/pyproject.toml
```diff
@@ -20,16 +20,21 @@ test = [
 ]
 
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "nmp-build-tools"]
```
matthewgrossman: (seeing this repeat down below) did we have to do this for each of the subpackages? Since they aren't ever _actually_ published, wondering if its necessary, and if things exploded without it

.github/workflows/release-bundle.yaml
mckornfield: opposite, it disables a shallow clone lol

packages/data_designer_nemo/pyproject.toml
mckornfield: yeahhh I kept asking if this was really necessary, I guess they'll default to `0.0.0` if we don't do this. there was a version where each one had the uv-version plugin imported and their own versions set, I liked that this was a bit more DRY, but maybe totally unnecessary

mckornfield: I just set them all to 0.0.0 except for a handful

.github/workflows/release-bundle.yaml
matthewgrossman: oh lol

packages/data_designer_nemo/pyproject.toml
```diff
@@ -1,6 +1,5 @@
 [project]
 name = "data-designer-nemo"
-version = "0.1.0"
```
matthewgrossman, line 3: this one is going from 0.1.0 -> 0.0.0, but that shouldn't matter because it ends up being dynamic anyways, right?

matthewgrossman (approved):
overall I like the solution, we should make sure it still works well with the release with some sort of dummy release we can do (rather than waiting until we _cant_ release)

I'm not super familiar with the dynamic mechanism, but it does make sense!

packages/data_designer_nemo/pyproject.toml
mckornfield: yeah... that and since we don't publish this individual thing, I think it's a no-op

mckornfield:
minor thing testing locally, just to show it
```
$ uv run nemo --version
      Built nemo-platform-sdk @ file:///Users/mkornfield/dev/nemo-platform/sdk/python/nemo-platform
Uninstalled 1 package in 60ms
Installed 1 package in 17ms
nemo version 0.2.0.post28.dev0+ca165cb0

~/repo/nemo-platform on dynamic-versioning/mck in cluster nemo-dgxc-k8s-oci-ord-prd3|aire-dev 13:37:00
$ git tag 0.2.1 HEAD

~/repo/nemo-platform on dynamic-versioning/mck in cluster nemo-dgxc-k8s-oci-ord-prd3|aire-dev 13:37:04
$ uv run nemo --version
      Built nemo-platform-sdk @ file:///Users/mkornfield/dev/nemo-platform/sdk/python/nemo-platform
Uninstalled 1 package in 58ms
Installed 1 package in 14ms
nemo version 0.2.1
```

Do you mean running the release pipeline as well, or some sorta dry run?

RELEASING.md
```diff
@@ -20,31 +20,40 @@ Artifacts published on a stable release:
...
+
 ---
 
-## Step 1 — Merge the version bump PR first
+## Step 1 — Choose the source SHA and release version
 
-Before triggering the release, merge a PR that bumps `version.py` to the *next* version (e.g. `0.1.2` → `0.1.3`). You will then release from the commit *just before* that merge — the last commit where `version.py` still reads `0.1.2`. This means:
-- The bump is already on `main` before you touch the release workflow — no race window.
-- Nightlies built after the bump PR merges will immediately produce `0.1.3.dev...` strings.
+Pick the full 40-character commit SHA on `main` that should be released, plus the SemVer core version to publish, for example `0.1.2`. There is no version bump PR to merge first: the stable workflow creates the release tag at `source_sha`, and the wheel build receives the package version from the workflow input.
```
svvarom: > There is no version bump PR to merge first

This bit might be unnecessary

mckornfield: lolol codex getting a little too focused

.github/actions/build-nemo-platform-wheel/action.yaml
```diff
@@ -176,6 +174,7 @@ runs:
         SOURCE_ROOT: ${{ inputs.source-root }}
         OUT_DIR: ${{ inputs.out-dir }}
         PACKAGE: ${{ inputs.package }}
+        UV_DYNAMIC_VERSIONING_BYPASS: ${{ steps.stamp.outputs.wheel-version }}
```
svvarom, line 177: During the release process, selecting a SHA older than when this commit landed will cause a break. I assume that’s okay, since we don’t expect to release something like that?

.github/scripts/stamp_sdk_version.py
```diff
@@ -24,52 +25,40 @@ def safe_sdk_id(sdk_id: str) -> str:
...
+    source_root: Path,
 ) -> str:
     if cadence == "nightly":
-        base_version = read_assignment(shared_sdk_version_path, "platform_sdk_version")
-        if not re.fullmatch(SEMVER_CORE_PATTERN, base_version):
-            raise StampError(f"nightly base SDK version must be SemVer core MAJOR.MINOR.PATCH: {base_version}")
+        base_version = latest_release_core(source_root) or "0.0.0"
         if not re.fullmatch(r"\d{14}", nightly_timestamp):
             raise StampError("nightly timestamp must be YYYYMMDDHHMMSS")
         return f"{base_version}.dev{nightly_timestamp}"
```
svvarom, line 64: Should nightlies use the next version core? `0.1.3.dev`... sorts below `0.1.3rc0`, so pip/uv against pypi.nvidia.com won’t select it as the latest prerelease.

svvarom: But we also never release rc's. So also not sure if this matters

svvarom (approved):
nits, but otherwise LGTM

svvarom:
> nits, but otherwise LGTM

feel free to ignore them

.github/actions/build-nemo-platform-wheel/action.yaml
mckornfield: yeah... if we did hotfixes that would matter, but I think it's fine. Worst case we'd run the pipeline off the older branch? or revert lol

.github/scripts/stamp_sdk_version.py
mckornfield: yeah I'ma send it for now, but I suspect later on we'll be like "wtf is this order"

PR #476, opened 2026-06-25
feat(models): generic engine support

benmccown:
Here's a concise PR description:

---

### What

Adds a new `generic` inference engine alongside `nim` and `vllm`, on both the **docker** and **k8s** service backends. A generic deployment is "declare a container image + tag + health path and run it" — no inference-engine compiler, no model-weight pull. The user's `additional_args` become the container command verbatim and `additional_envs` become its env.

This covers cases like classification NIMs (e.g. NemoGuard JailbreakDetect) and any self-contained server image, without needing engine-specific compile logic.

### Key changes

- **`generic_compiler.py`** (new): trivial backend-agnostic compiler — resolves the image (required, no platform default), and passes args/env through verbatim.
- **Validation**: `engine=generic` now requires `executor_config.image_name` **and** `health_check_path` at create/update time (no platform defaults exist for an arbitrary container). Returns a 400 with a clear message.
- **Docker backend**: branches `generic` in image resolution and env/args compilation; runs the image as a local container with the engine + health-path labels; no model puller; no LoRA sidecar.
- **K8s backend**: routes `generic` (and `vllm`) through the direct-emission `K8sReconciler`, which now branches on engine. The generic path emits a Deployment + Service **immediately** with **no PVC / no puller Job / no `model-store` mount** (it has no weights). `vllm_k8s_compiler.compile_deployment` gained a `mount_model_store` flag to support this. Delete cascades via ownerRef as before.

### Testing

- Unit tests for the generic compiler, docker + k8s backend paths, the no-PVC compile path, and the new validation. (The previous "generic rejected on k8s" test is replaced with positive create/status/LOST coverage.)
- **Manually verified end-to-end in the dev pod on both backends** using `vllm/vllm-openai` as a generic container serving `facebook/opt-125m`:
  - Docker: container runs with args verbatim, correct engine/health-path labels, reaches READY, serves `/v1/completions`.
  - K8s: emits Deployment + Service only (no PVC/Job), pod has no `model-store` mount, `/health` probe, reaches READY, serves inference, deletes cleanly.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for the generic engine deployment path, including generic image/tag resolution and pass-through args/env.
  * Kubernetes generic deployments now support both weightless immediate serving and phased weight pull+serve, with conditional `model-store` mounting.
  * Docker generic creation now supports explicit image/command/env and `health-check-path`, including weight-fileset pull behavior.
* **Bug Fixes**
  * Added stricter generic executor-config validation (required fields, no leading/trailing whitespace) and blocks LoRA for generic deployments.
  * Improved unsupported-engine error details to include the engine value; generic now routes to supported Kubernetes reconciliation.
* **Tests**
  * Expanded unit coverage for generic compilation, Docker/Kubernetes create+status flows, and generic config validation.

### DIFF
--- a/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
+++ b/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
@@ -15,13 +15,39 @@ from nmp.core.models.entities import Model as ModelEntity
 from nmp.core.models.entities import ModelDeployment as ModelDeploymentEntity
 from nmp.core.models.entities import ModelDeploymentConfig as ModelDeploymentConfigEntity
 from nmp.core.models.schemas import (
+    ContainerExecutorConfig,
     CreateModelDeploymentConfigRequest,
+    Engine,
     ModelDeploymentConfig,
     ModelDeploymentStatus,
     UpdateModelDeploymentConfigRequest,
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorConfig) -> None:
+    """Validate engine-specific requirements on the executor config.
+
+    The ``generic`` engine runs an arbitrary container with no inference-engine
+    compiler, so it has no platform-default image and no canonical health
+    endpoint. Both ``image_name`` and ``health_check_path`` must therefore be
+    supplied explicitly; the other engines fall back to their configured
+    defaults when these are unset.
+    """
+    if engine != Engine.GENERIC:
+        return
+    missing: list[str] = []
+    if not (executor_config.image_name and executor_config.image_name.strip()):
+        missing.append("image_name")
+    if not (executor_config.health_check_path and executor_config.health_check_path.strip()):
+        missing.append("health_check_path")
+    if missing:
+        raise ValueError(
+            "The 'generic' engine requires executor_config."
+            + " and executor_config.".join(missing)
+            + " to be set (no platform default exists for a generic container)."
+        )
 
 
 class ReferentialIntegrityError(Exception):
@@ -103,6 +129,8 @@ class ModelDeploymentConfigService:
         existing = await self._get_latest_version(workspace, request.name)
         if existing is not None:
             raise ValueError(f"Deployment config with workspace '{workspace}' and name '{request.name}' already exists")
+
+        _validate_engine_config(request.engine, request.executor_config)
 
         if not request.model_entity_id:
             try:
@@ -254,6 +282,8 @@ class ModelDeploymentConfigService:
         current = await self._get_by_base_name_and_version(workspace, name)
         if not current:
             raise ValueError(f"Deployment config with workspace '{workspace}' and name '{name}' does not exist")
+
+        _validate_engine_config(request.engine, request.executor_config)
 
         new_version = current.entity_version + 1
 
--- a/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
+++ b/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
@@ -34,19 +34,30 @@ def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorCo
     endpoint. Both ``image_name`` and ``health_check_path`` must therefore be
     supplied explicitly; the other engines fall back to their configured
     defaults when these are unset.
+
+    Values are also rejected when they contain surrounding whitespace: an
+    image reference or probe path is used verbatim downstream, where a
+    leading/trailing space would silently produce an invalid value.
     """
     if engine != Engine.GENERIC:
         return
     missing: list[str] = []
-    if not (executor_config.image_name and executor_config.image_name.strip()):
-        missing.append("image_name")
-    if not (executor_config.health_check_path and executor_config.health_check_path.strip()):
-        missing.append("health_check_path")
+    padded: list[str] = []
+    for field in ("image_name", "health_check_path"):
+        value = getattr(executor_config, field)
+        if not (value and value.strip()):
+            missing.append(field)
+        elif value != value.strip():
+            padded.append(field)
     if missing:
         raise ValueError(
             "The 'generic' engine requires executor_config."
             + " and executor_config.".join(missing)
             + " to be set (no platform default exists for a generic container)."
+        )
+    if padded:
+        raise ValueError(
+            "executor_config." + " and executor_config.".join(padded) + " must not have leading or trailing whitespace."
         )
 
 
--- a/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
+++ b/services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
@@ -19,6 +19,7 @@ from nmp.core.models.schemas import (
     CreateModelDeploymentConfigRequest,
     Engine,
     ModelDeploymentConfig,
+    ModelDeploymentConfigModelSpec,
     ModelDeploymentStatus,
     UpdateModelDeploymentConfigRequest,
 )
@@ -26,8 +27,12 @@ from nmp.core.models.schemas import (
 logger = logging.getLogger(__name__)
 
 
-def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorConfig) -> None:
-    """Validate engine-specific requirements on the executor config.
+def _validate_engine_config(
+    engine: Engine,
+    executor_config: ContainerExecutorConfig,
+    model_spec: ModelDeploymentConfigModelSpec,
+) -> None:
+    """Validate engine-specific requirements on the deployment config.
 
     The ``generic`` engine runs an arbitrary container with no inference-engine
     compiler, so it has no platform-default image and no canonical health
@@ -38,6 +43,10 @@ def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorCo
     Values are also rejected when they contain surrounding whitespace: an
     image reference or probe path is used verbatim downstream, where a
     leading/trailing space would silently produce an invalid value.
+
+    LoRA is rejected for ``generic``: there is no engine compiler to wire the
+    adapter sidecar against, so ``lora_enabled`` would otherwise be silently
+    ignored. Reject it up front rather than accept a config that can't be honored.
     """
     if engine != Engine.GENERIC:
         return
@@ -58,6 +67,11 @@ def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorCo
     if padded:
         raise ValueError(
             "executor_config." + " and executor_config.".join(padded) + " must not have leading or trailing whitespace."
+        )
+    if model_spec.lora_enabled:
+        raise ValueError(
+            "The 'generic' engine does not support LoRA (model_spec.lora_enabled); "
+            "there is no engine compiler to wire the adapter sidecar against."
         )
 
 
@@ -141,7 +155,7 @@ class ModelDeploymentConfigService:
         if existing is not None:
             raise ValueError(f"Deployment config with workspace '{workspace}' and name '{request.name}' already exists")
 
-        _validate_engine_config(request.engine, request.executor_config)
+        _validate_engine_config(request.engine, request.executor_config, request.model_spec)
 
         if not request.model_entity_id:
             try:
@@ -294,7 +308,7 @@ class ModelDeploymentConfigService:
         if not current:
             raise ValueError(f"Deployment config with workspace '{workspace}' and name '{name}' does not exist")
 
-        _validate_engine_config(request.engine, request.executor_config)
+        _validate_engine_config(request.engine, request.executor_config, request.model_spec)
 
         new_version = current.entity_version + 1
 
--- a/services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
@@ -503,20 +503,24 @@ class DockerDeploymentCreationReconciler:
         full_image = f"{image_name}:{image_tag}"
         logger.info(f"Using image: {full_image} (engine={engine})")
 
-        # Create volumes for model cache
+        # Create volumes for model cache + scratch. A generic container that pulls
+        # no weights runs raw (no platform volumes mounted -- see container create
+        # below), so skip provisioning them; every other case mounts them.
         volume_name = self.get_volume_name(deployment.workspace, deployment.name)
-        try:
-            await asyncio.to_thread(self.create_volume, volume_name)
-            logger.info(f"Created volume: {volume_name}")
-        except Exception as e:
-            logger.warning(f"Failed to create volume {volume_name} (may already exist): {e}")
-
         scratch_volume_name = volume_name + "-scratch"
-        try:
-            await asyncio.to_thread(self.create_volume, scratch_volume_name)
-            logger.info(f"Created volume: {scratch_volume_name}")
-        except Exception as e:
-            logger.warning(f"Failed to create volume {scratch_volume_name} (may already exist): {e}")
+        provision_volumes = engine != ENGINE_GENERIC or weights_from_files
+        if provision_volumes:
+            try:
+                await asyncio.to_thread(self.create_volume, volume_name)
+                logger.info(f"Created volume: {volume_name}")
+            except Exception as e:
+                logger.warning(f"Failed to create volume {volume_name} (may already exist): {e}")
+
+            try:
+                await asyncio.to_thread(self.create_volume, scratch_volume_name)
+                logger.info(f"Created volume: {scratch_volume_name}")
+            except Exception as e:
+                logger.warning(f"Failed to create volume {scratch_volume_name} (may already exist): {e}")
 
         # Multi-LLM detection only applies to NIM images; vLLM/generic are never multi-LLM.
         is_multi_llm = False
@@ -1015,16 +1019,26 @@ class DockerDeploymentCreationReconciler:
         try:
             logger.info("Creating container %s with image %s...", container_name, full_image)
 
+            # Platform volumes (/model-store, /scratch) hold pulled weights + scratch
+            # space. NIM/vLLM always mount them. A generic container runs the user's
+            # image as-is, so only mount them when the platform actually pulls weights
+            # for it (a fileset-backed model deployment); otherwise the mounts would
+            # shadow the image's own contents at those paths.
+            mount_platform_volumes = engine != ENGINE_GENERIC or self._needs_puller(state)
+            volumes: dict[str, Any] = {}
+            if mount_platform_volumes:
+                volumes = {
+                    state.volume_name: {"bind": "/model-store", "mode": "rw"},
+                    state.scratch_volume_name: {"bind": "/scratch", "mode": "rw"},
+                }
+
             create_args: dict[str, Any] = {
                 "image": full_image,
                 "name": container_name,
                 "environment": env_vars,
                 "detach": True,
                 "device_requests": device_requests,
-                "volumes": {
-                    state.volume_name: {"bind": "/model-store", "mode": "rw"},
-                    state.scratch_volume_name: {"bind": "/scratch", "mode": "rw"},
-                },
+                "volumes": volumes,
                 "labels": {
                     "nmp.nvidia.com/deployment-workspace": deployment.workspace,
                     "nmp.nvidia.com/deployment-name": deployment.name,
--- a/services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
@@ -35,7 +35,7 @@ from nmp.common.sdk_factory import get_sdk_on_behalf_of
 from nmp.core.models.app import ModelWeightsType, get_model_weights_type, is_multi_llm_image, parse_model_name_revision
 from nmp.core.models.app.constants import MODEL_MANAGED_BY_LABEL, MODEL_MANAGED_BY_MODELS_CONTROLLER
 from nmp.core.models.app.utils import _get_k8s_safe_name
-from nmp.core.models.controllers.backends import vllm_compiler
+from nmp.core.models.controllers.backends import generic_compiler, vllm_compiler
 from nmp.core.models.controllers.backends.backends import DeploymentStatusUpdate
 from nmp.core.models.controllers.backends.common import deployment_config_view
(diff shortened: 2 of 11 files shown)


services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
```diff
@@ -24,6 +26,30 @@
...
+    missing: list[str] = []
+    if not (executor_config.image_name and executor_config.image_name.strip()):
+        missing.append("image_name")
+    if not (executor_config.health_check_path and executor_config.health_check_path.strip()):
+        missing.append("health_check_path")
+    if missing:
+        raise ValueError(
+            "The 'generic' engine requires executor_config."
+            + " and executor_config.".join(missing)
+            + " to be set (no platform default exists for a generic container)."
```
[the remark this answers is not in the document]
benmccown: Fixed in f35c40d. `_validate_engine_config` now rejects generic configs whose `image_name`/`health_check_path` have surrounding whitespace (raising rather than silently storing a value that would break downstream), in addition to the existing presence check — on both the create and update paths.

services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
```diff
@@ -492,6 +493,10 @@ async def register_deployment(
                 self._backend_config.default_vllm_image,
                 self._backend_config.default_vllm_image_tag,
             )
+        elif engine == ENGINE_GENERIC:
+            # Generic containers have no platform-default image; image_name is
+            # required (enforced at the API layer and again in the compiler).
+            image_name, image_tag = generic_compiler.resolve_generic_image(view)
```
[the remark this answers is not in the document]
benmccown: Addressed in f35c40d, though by a different design than gating the puller off. We decided to make the generic engine *weight-aware*: a generic deployment that resolves to a fileset-backed model now intentionally runs the puller and gets `/model-store`, while a self-contained generic image (no fileset) runs raw with no puller. So `_needs_puller()` staying engine-agnostic is now correct — the discriminator is weight presence, not engine.

services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
```diff
@@ -893,11 +898,20 @@ async def _advance_creating_container(
...
+        # via env, so it leaves the container command unset.
+        serve_args: list[str] | None = None
+        if engine == ENGINE_GENERIC:
+            # Generic: run the image with the user's raw env + args verbatim. The
+            # platform synthesizes nothing (no served-model-name, no LoRA, etc.).
+            env_vars = generic_compiler.compile_generic_env_vars(view)
+            generic_args = generic_compiler.compile_generic_args(view)
+            # Only override the image's command when the user supplied args.
+            serve_args = generic_args or None
+        elif engine == ENGINE_VLLM:
```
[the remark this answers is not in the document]
benmccown: Fixed in f35c40d. `create_args["volumes"]` is now conditional: for a generic deployment the platform `/model-store` + `/scratch` mounts are only attached when the deployment actually pulls weights (a fileset-backed model). A self-contained generic image runs with no platform volumes, so they no longer shadow the image contents. NIM/vLLM mounting is unchanged. (The volumes are also no longer provisioned at all for a weightless generic deployment.)

services/core/models/src/nmp/core/models/controllers/backends/generic_compiler.py
```diff
@@ -0,0 +1,57 @@
...
+    """Resolve the generic container image name and tag.
+
+    There is no platform default for a generic image (it is an arbitrary
+    user-supplied container), so ``image_name`` is required. ``image_tag``
+    defaults to ``latest`` when unset, mirroring Docker/Kubernetes conventions.
+    """
+    if not (view.image_name and view.image_name.strip()):
+        raise ValueError("The 'generic' engine requires executor_config.image_name to be set (no platform default).")
+    image_tag = view.image_tag or "latest"
+    return view.image_name, image_tag
```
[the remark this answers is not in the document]
benmccown: Fixed in f35c40d. `resolve_generic_image` now strips both fields: it returns the trimmed `image_name` and computes `image_tag` from the trimmed value, falling back to `"latest"` when the trimmed tag is empty (so `"   "` no longer yields an invalid reference). The API layer also rejects whitespace-padded values up front, but the compiler stays robust as a backstop for any other call path.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/backend.py
```diff
@@ -235,22 +235,21 @@ def _resolve(self, ctx: ModelContext) -> ResolvedDeployment:
...
+        defaults to the NIM-operator reconciler. ``None`` is reserved for a
+        genuinely unknown engine, which the callers treat as the "unsupported
+        engine" rejection (see :meth:`_unsupported_engine`).
         """
-        if engine == ENGINE_VLLM:
+        if engine in (ENGINE_VLLM, ENGINE_GENERIC):
             return self._k8s_reconciler
-        if engine == ENGINE_GENERIC:
-            return None
         return self._nim_reconciler
```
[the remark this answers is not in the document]
benmccown: Fixed in f35c40d. `_select_reconciler` is now explicit: `vllm`/`generic` -> direct-emission reconciler, `nim` -> NIM-operator reconciler, anything else -> `None`, which callers turn into `_unsupported_engine()`. Unknown engine strings are no longer silently reconciled as NIM, matching the docstring.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -321,7 +350,7 @@ def _vllm_objects_exist(self, resource_name: str) -> bool:
 
         def _has_vllm_engine_label(obj) -> bool:
             labels = getattr(getattr(obj, "metadata", None), "labels", None)
-            return isinstance(labels, dict) and labels.get("nmp.nvidia.com/engine") == ENGINE_VLLM
+            return isinstance(labels, dict) and labels.get("nmp.nvidia.com/engine") in (ENGINE_VLLM, ENGINE_GENERIC)
```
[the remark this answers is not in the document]
benmccown: Resolved in f35c40d. The `_vllm_objects_exist()` helper (and the generic `ENGINE_GENERIC` match it had grown) is gone. `K8sReconciler` was refactored so the staged-vs-immediate decision is driven by weight presence (`_has_weights`), not by sniffing engine labels on existing objects, so the cross-engine false-positive this flagged can no longer happen.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -464,6 +493,149 @@ def _create_vllm_serving_objects(
...
+        Deployment exists it is patched in place, otherwise it is (re)created.
+        """
+        deployment = resolved.deployment
+        logger.info(
+            f"Updating generic deployment: {deployment.workspace}/{deployment.name} "
+            f"(version: {deployment.entity_version})"
+        )
+        try:
+            self._create_generic_serving_objects(deployment, resolved.resource_name, resolved.view)
+            return DeploymentStatusUpdate(
```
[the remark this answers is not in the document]
benmccown: Fixed in f35c40d. The update path now calls `_apply_serving_objects`, which patches the existing Deployment and Service in place (`patch_namespaced_deployment` / `patch_namespaced_service`) so changed image/args/env/health/gpu/labels actually take effect, and creates them only when absent. The stale read-on-409 behavior (and the inaccurate docstring) are gone.

services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
```diff
@@ -24,6 +26,41 @@
...
+    padded: list[str] = []
+    for field in ("image_name", "health_check_path"):
+        value = getattr(executor_config, field)
+        if not (value and value.strip()):
+            missing.append(field)
+        elif value != value.strip():
+            padded.append(field)
+    if missing:
+        raise ValueError(
+            "The 'generic' engine requires executor_config."
```
mckornfield, line 63: lol this reads very interestingly, but I think I see how it becomes

`executor_config.image_name and executor_config.health_check_path`

fun lol

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -75,29 +97,52 @@ def __init__(
...
+        # vLLM (the only other engine routed here) is always weighted.
+        return True
+
     async def create(self, resolved: ResolvedDeployment) -> DeploymentStatusUpdate:
-        """Create phase P0: emit the PVC + weight-puller Job.
+        """Create the deployment's backend resources.
 
-        The Deployment + Service are created later by the status path once the Job
-        completes (controller-side weight-readiness gating).
+        Weighted deployments do phase P0: emit the PVC + weight-puller Job (the
```
mckornfield: "do Phase P0" sounds kinda weird

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -371,47 +417,54 @@ def _project_deployment_readiness(self, resource_name: str) -> DeploymentStatusU
...
-        Deployment cascades the rest.
+    # ------------------------------------------------------------------
+    # Engine-parameterized serving objects (shared by vLLM + generic)
+    # ------------------------------------------------------------------
+
+    def _pod_user(self, engine: str) -> tuple[Optional[int], Optional[int]]:
+        """Pod securityContext uid/gid for an engine.
+
+        vLLM pins its image's user (2000/0); a generic container runs as its own
+        image's user (unset), since we can't assume an arbitrary image tolerates a
```
mckornfield, line 456: do we allow passing one through though?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -371,47 +417,54 @@ def _project_deployment_readiness(self, resource_name: str) -> DeploymentStatusU
...
         startup_grace = self._backend_config.default_startup_probe_grace_period_seconds or 600
-
-        init_containers, sidecar_containers = self._build_lora_containers(deployment, view, model_entity)
+        user_id, group_id = self._pod_user(engine)
+
+        if engine == ENGINE_GENERIC:
+            image_name, image_tag = generic_compiler.resolve_generic_image(view)
+            args = generic_compiler.compile_generic_args(view)
+            env = generic_compiler.compile_generic_env_vars(view)
+            init_containers: Optional[list] = None
```
mckornfield, line 487: kinda similar question, will the generic engine support arbitrarily setting init or sidecar containers or no?

mckornfield (approved):
out of curiosity, do we have a deployment example with the generic compiler, like I'd guess sglang?

services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
```diff
@@ -24,6 +26,41 @@
 logger = logging.getLogger(__name__)
 
 
+def _validate_engine_config(engine: Engine, executor_config: ContainerExecutorConfig) -> None:
```
tylersbray: 1. Generic validation only in the service layer, not in Pydantic

engine=generic requirements (image_name, health_check_path) are documented on the schema fields but not enforced by a @model_validator:

services/core/models/src/nmp/core/models/schemas.py — ~L1400–1420 (ContainerExecutorConfig.image_name, health_check_path field docs)
Enforcement is only in:

services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py — _validate_engine_config() at ~L29–52, called from create_deployment_config() ~L144 and update_deployment_config() ~L297
Any code path that bypasses ModelDeploymentConfigService could still persist an invalid generic config. Should this be enforced at the schema/OpenAPI boundary too?

services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
```diff
@@ -1055,7 +1085,10 @@ async def cleanup_and_error(status_message: str, error_details: dict) -> tuple[D
             container_id = container.id[:12]
             logger.info("Container %s started successfully (ID: %s)", container_name, container_id)
 
-            if view.lora_enabled:
+            # The generic engine has no LoRA semantics (no engine compiler to
```
tylersbray, line 1088: 2. lora_enabled=True is silently ignored for generic

Docker skips the LoRA sidecar when engine=generic:

services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py — ~L1088–L1091 (if view.lora_enabled and engine != ENGINE_GENERIC)
K8s skips LoRA init/sidecar containers in _serving_spec():

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py — ~L455–L460 (init_containers / sidecar_containers set to None for ENGINE_GENERIC)
A user can set lora_enabled=True on a generic config and get no error. Should the API reject that combination?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
```diff
@@ -307,6 +307,7 @@ def compile_deployment(
     init_containers: Optional[list[k8s_client.V1Container]] = None,
     sidecar_containers: Optional[list[k8s_client.V1Container]] = None,
     extra_labels: Optional[dict[str, str]] = None,
+    mount_model_store: bool = True,
```
tylersbray, line 310: 3. Weightless generic k8s pods still mount /scratch and /dev/shm

mount_model_store=False correctly drops the model-store PVC, but compile_deployment() always mounts scratch + dshm emptyDirs:

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py — mount_model_store param ~L310; volume mounts ~L339–L345; volumes ~L364–L377
For an arbitrary generic container, those paths could shadow content the image expects. The test at test_vllm_k8s_compiler.py ~L190 treats them as harmless — is that always true for classification NIMs and other self-contained images?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -310,32 +373,15 @@ def _model_source(resolved: ResolvedDeployment) -> tuple[str, str]:
...
-        Deployment is created, so the Job alone is not a reliable marker), then the
-        puller Job for the pre-Deployment phase. Any lookup failure (including 404)
-        means "not (yet) a vLLM deployment".
-        """
-
-        def _has_vllm_engine_label(obj) -> bool:
-            labels = getattr(getattr(obj, "metadata", None), "labels", None)
-            return isinstance(labels, dict) and labels.get("nmp.nvidia.com/engine") == ENGINE_VLLM
-
+    def _serving_deployment_exists(self, resource_name: str) -> bool:
```
tylersbray, line 387: 4. Pull-phase update behavior change for weighted (vLLM / fileset-backed generic)

_serving_deployment_exists() (~L376–L385 in k8s.py) only checks for the serving Deployment, not the puller Job.

In update() (~L228–L236), when the model source is unchanged but the Deployment does not exist yet (still in the pull phase), the code now falls through to return await self.create(resolved) instead of returning a no-op “Update accepted”.

Previously, _vllm_objects_exist() also returned True when only the puller Job existed (with the vllm engine label), so mid-download updates were effectively no-ops.

Is re-calling create() during pull phase intentional? Are PVC/Job creates idempotent on conflict?

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -310,32 +373,15 @@ def _model_source(resolved: ResolvedDeployment) -> tuple[str, str]:
...
+        """True if the serving Deployment for this resource already exists."""
```
tylersbray, line 388: 5. _serving_deployment_exists() no longer checks the engine label

Replaces _vllm_objects_exist(), which verified nmp.nvidia.com/engine == vllm on the Deployment or Job:

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py — _serving_deployment_exists() ~L376–L385
The new helper returns True for any Deployment with that resource name. Safer on non-404 errors (re-raises instead of swallowing), but no longer confirms the object belongs to this reconciler. Probably fine if resource names are unique per deployment — worth confirming.

Line numbers are approximate (~) and refer to the PR branch; GitHub’s “Files changed” view may shift them slightly after further commits.

tylersbray (approved):
Giving a +1 approval.  Added comments for the medium-level things that got flagged by machine.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Reply submitted by Claude: Agreed — reworded the docstring to "Weighted deployments start at phase P0: emit the PVC + weight-puller Job …" (change staged locally, not yet pushed).

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Reply submitted by Claude: Not currently — there is no executor_config field to pass a uid/gid through, so generic runs as the image's own user (uid/gid unset) while vLLM pins its image's 2000/0. We are planning to address configurable uid/gid as follow-up work on a separate branch.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Reply submitted by Claude: Not in this PR — generic emits a single container from the image, with no API to inject init or sidecar containers (those are wired only for vLLM's LoRA path today). It is a reasonable future enhancement if a use case needs it.

services/core/models/src/nmp/core/models/controllers/backends/docker/creation_reconciler.py
benmccown: Reply submitted by Claude: Good call — added API validation that rejects engine=generic with model_spec.lora_enabled=True (in _validate_engine_config, alongside the image_name/health_check_path checks), so the combination now fails fast at create/update instead of being silently ignored. Change staged locally, not yet pushed.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/vllm_k8s_compiler.py
benmccown: Reply submitted by Claude: Leaving these as-is. /dev/shm is a standard, expected mount (a default-sized tmpfs that containers commonly rely on), and /scratch is an empty emptyDir a generic container simply won't use unless it writes there — neither shadows image content the way the read-only /model-store PVC could. mount_model_store=False already drops the one mount that mattered.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Reply submitted by Claude: Good catch — restored the mid-pull no-op. update() now checks for the puller Job when the serving Deployment is absent: if the Job is still present (mid-download) the update returns "Update accepted" without re-running create(), and only falls back to create() when the pull objects are genuinely gone. (For the record, create()'s PVC/Job creates are 409-tolerant via _create_or_skip, so the prior fall-through was idempotent, but the explicit Job check restores the intended no-op semantics.) Change staged locally, not yet pushed.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Reply submitted by Claude: Correct and intentional — resource_name is derived from workspace/name (get_deployment_resource_name), which is unique per deployment, so a name-only check cannot collide across engines. The new helper is also stricter than the old one: it re-raises non-404 errors instead of swallowing them.

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -142,42 +184,57 @@ async def create(self, resolved: ResolvedDeployment) -> DeploymentStatusUpdate:
...
 
+            _, source_tag = self._model_source(resolved)
             existing_source = self._existing_model_source(resource_name)
             if existing_source is not None and existing_source != source_tag:
                 logger.info(
                     f"Model source changed ({existing_source} -> {source_tag}); re-pulling weights for {resource_name}"
                 )
-                self._delete_vllm_resources(resource_name)
+                self._delete_serving_resources(resource_name)
                 return await self.create(resolved)
```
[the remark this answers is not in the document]
benmccown: Reply submitted by Claude: Confirmed this is a real (pre-existing, vLLM weighted re-pull) bug — after P3 deletes the puller Job, _existing_model_source() returns None, so a later revision change skips the re-pull and patches against stale PVC weights. The PVC does carry the model-source annotation, so the Job-then-PVC fallback is the right fix. We are deferring this to a follow-up PR (bundled with the other follow-up items for this feature) rather than expanding the scope of this one.

services/core/models/src/nmp/core/models/api/service/model_deployment_config_service.py
benmccown: Reply submitted by Claude: The Models API (ModelDeploymentConfigService) is the only write path for deployment configs, and _validate_engine_config runs on both its create and update paths, so invalid generic configs cannot be persisted in practice. We are comfortable keeping enforcement in the service layer rather than duplicating it across the request/config Pydantic models. If a second write path is ever introduced, a shared @model_validator would be the place to add boundary-level enforcement.

PR #479, opened 2026-06-25
feat(customizer): add RL backend in customizer plugin

anubhutivyas:
Linear ticket - https://linear.app/nvidia/issue/AALGO-292/add-dpo-support-for-customizer-plugin

## Summary by CodeRabbit

* **New Features**
  * Added NeMo-RL (DPO) workflow with a new submission/CLI surface, Kubernetes-only remote job compilation, and dedicated containerized steps for file I/O, training, and model-entity registration.
  * Introduced new RL Docker images/build targets (workspace, base, training, tasks) to support the end-to-end pipeline.
  * Added structured job inputs/outputs, RL dataset schema validation, and multi-family chat-template resolution.

* **Bug Fixes**
  * Hardened RL containers for non-root execution and applied vLLM CVE remediation.

* **Tests**
  * Added unit tests for compilation/config validation and GPU-oriented smoke testing for the new images.

### DIFF
--- a/docker/Dockerfile.nmp-rl-base
+++ b/docker/Dockerfile.nmp-rl-base
@@ -1,0 +1,56 @@
+# syntax=docker/dockerfile:1
+# nmp-rl base - NVIDIA NeMo-RL v0.6.0 NGC container + CVE hardening.
+#
+# Bases on the PUBLISHED NGC NeMo-RL container (amd64 + arm64 at v0.6.0), which
+# bundles the full RL stack on Python 3.13: PyTorch 2.10, Ray 2.54.0, vLLM
+# 0.17.1, Megatron-Core 0.18, Transformers 5.3.
+#
+# Publish target: nmp-rl-base.
+
+# Referenced by tag rather than a @sha256 digest pin: NGC publishes separate
+# per-architecture tags/digests for nemo-rl, so a single digest doesn't cleanly
+# cover the multi-arch (amd64 + arm64) base we build on. The :v0.6.0 tag is the
+# stable, arch-agnostic reference.
+ARG NEMO_RL_IMAGE=nvcr.io/nvidia/nemo-rl:v0.6.0
+
+FROM ${NEMO_RL_IMAGE} AS nmp-rl-base
+
+# NeMo-RL v0.6.0 ships its venv at /opt/nemo_rl_venv (Python 3.13).
+# VERIFY this path against the actual image if NeMo-RL relocates it in a future
+# tag — the training image's /opt/venv symlink and entrypoint depend on it.
+ENV VIRTUAL_ENV=/opt/nemo_rl_venv \
+    HF_HUB_ENABLE_HF_TRANSFER=1 \
+    OTEL_PYTHON_EXCLUDED_URLS="health"
+
+# The NGC base venv is built as root with locked-down cache dirs; relax them so
+# editable installs in the training image can write.
+RUN chmod 755 /root /root/.cache /root/.cache/uv /root/.local /root/.local/share /root/.local/share/uv 2>/dev/null || true
+
+# NeMo-RL is installed EDITABLE in this base: its source tree lives at
+# /opt/nemo-rl (root-owned, non-traversable) and the editable import finder
+# os.stat()s files under it at import time. The training image drops to
+# USER 1000, so `from nemo_rl... import ...` hits the finder and fails with
+# PermissionError on /opt/nemo-rl/nemo_rl/__init__.py. Make the source tree
+# world-readable + dir-traversable (a+rX) so the non-root runtime can import it.
+# Read-only is enough — nothing writes back into the editable tree at runtime.
+RUN chmod -R a+rX /opt/nemo-rl
+
+# CVE remediation: remove vLLM from the main venv and any pre-built Ray worker
+# venvs (NeMo-RL rebuilds those at runtime). vLLM is unused by DPO/SFT/LoRA.
+# Add it back when adding GRPO.
+RUN set -e; \
+    rm -rf ${VIRTUAL_ENV}/lib/python*/site-packages/vllm \
+           ${VIRTUAL_ENV}/lib/python*/site-packages/vllm-*.dist-info; \
+    for venv_dir in /opt/ray_venvs/*/; do \
+        [ -d "$venv_dir/lib" ] || continue; \
+        rm -rf "$venv_dir"/lib/python*/site-packages/vllm \
+               "$venv_dir"/lib/python*/site-packages/vllm-*.dist-info; \
+    done
+
+# Drop to a non-root user by default so DIRECT consumers of this published base do
+# not run with container root. The build steps above intentionally run as root (the
+# NGC base default — a USER directive only changes the final/default user, not the
+# RUNs preceding it). Derived images that need root for their own build steps
+# (e.g. Dockerfile.nmp-rl-training) re-assert `USER root` and then drop back to a
+# non-root UID at the end. UID 1000 is the NGC base's `ubuntu` user.
+USER 1000:1000
--- a/docker/Dockerfile.nmp-rl-tasks
+++ b/docker/Dockerfile.nmp-rl-tasks
@@ -1,0 +1,73 @@
+# syntax=docker/dockerfile:1
+# nmp-rl tasks - CPU file_io / model_entity steps for an nmp-rl DPO job.
+#
+# Deliberately does NOT build on nmp-rl-base: the download/upload/model-entity
+# steps only need the platform glue (SDK + customization-common), not NeMo-RL /
+# Ray / vLLM. Basing on the NGC image keeps it consistent with the platform's
+# CUDA userspace while staying far lighter than the training image.
+
+ARG SMOKE_MARKER=smoke_nmp_rl_tasks
+ARG PYTORCH_BASE=nvcr.io/nvidia/pytorch:26.02-py3
+
+FROM ${PYTORCH_BASE} AS base
+
+WORKDIR /app
+
+COPY --from=ghcr.io/astral-sh/uv:0.9.14 /uv /bin/uv
+ENV PATH="/bin:${PATH}"
+
+ENV VIRTUAL_ENV=/opt/venv \
+    UV_PROJECT_ENVIRONMENT=/opt/venv \
+    UV_LINK_MODE=copy \
+    UV_COMPILE_BYTECODE=1 \
+    HF_HUB_ENABLE_HF_TRANSFER=1 \
+    OTEL_PYTHON_EXCLUDED_URLS="health"
+ENV PATH="/opt/venv/bin:/root/.local/bin:${PATH}"
+
+RUN uv venv ${UV_PROJECT_ENVIRONMENT} --system-site-packages
+
+FROM base AS runtime
+
+ARG USERNAME=ubuntu
+ARG USER_UID=1000
+ARG USER_GID=1000
+
+COPY --from=platform-workspace / /app
+WORKDIR /app
+
+RUN mkdir -p /home/${USERNAME}/.cache && \
+    chown -R ${USER_UID}:${USER_GID} /home/${USERNAME} /app/services/rl
+
+# Only the glue + nmp-rl (compile/tasks side). No NeMo-RL extra → no Ray/vLLM.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --python ${VIRTUAL_ENV}/bin/python --no-cache \
+    -e /app/sdk/python/nemo-platform \
+    -e /app/packages/nemo_platform_plugin \
+    -e /app/packages/nmp_common \
+    -e /app/packages/nmp_customization_common \
+    -e /app/services/rl
+
+ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
+# Default CMD is a harmless help invocation; the platform overrides `command`
+# per step (e.g. -m nmp.rl.tasks.file_io / nmp.rl.tasks.model_entity).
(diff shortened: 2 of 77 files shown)


docker/Dockerfile.nmp-rl-base
```diff
@@ -0,0 +1,44 @@
+# syntax=docker/dockerfile:1
+# nmp-rl base - NVIDIA NeMo-RL v0.6.0 NGC container + CVE hardening.
+#
+# Bases on the PUBLISHED NGC NeMo-RL container (amd64 + arm64 at v0.6.0), which
+# bundles the full RL stack on Python 3.13: PyTorch 2.10, Ray 2.54.0, vLLM
+# 0.17.1, Megatron-Core 0.18, Transformers 5.3.
+#
+# Publish target: nmp-rl-base.
+
+ARG NEMO_RL_IMAGE=nvcr.io/nvidia/nemo-rl:v0.6.0
```
albcui: nit a more robust and secure way is to pin the SHA of the image digest with something like

```
ARG NEMO_RL_IMAGE=nvcr.io/nvidia/nemo-rl:v0.6.0@sha256:<64-char-hex-digest>
```

docker/Dockerfile.nmp-rl-tasks
```diff
@@ -0,0 +1,65 @@
...
+
+ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
+# Default CMD is a harmless help invocation; the platform overrides `command`
+# per step (e.g. -m nmp.rl.tasks.file_io / nmp.rl.tasks.model_entity).
+ENTRYPOINT ["/opt/venv/bin/python"]
+CMD ["-m", "nmp.rl.tasks.file_io", "--help"]
+
+USER ${USER_UID}:${USER_GID}
+
+FROM runtime AS smoke-test
```
albcui, line 66: There are no bake targets for this in `docker-bake.hcl`. Was this a miss or the intent is to wire them up in a follow-up?

docker/Dockerfile.nmp-rl-training
```diff
@@ -0,0 +1,92 @@
...
+    "distro==1.9.0" \
+    "pydantic-settings==2.14.2"
+
+ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
+ENTRYPOINT ["/opt/venv/bin/python"]
+CMD ["-m", "nmp.rl.tasks.training", "--help"]
+
+USER ${USER_UID}:${USER_GID}
+
+FROM runtime AS smoke-test
```
albcui, line 98: Similarly, this isn't wired up to `docker-back.hcl`

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/dpo_config.py
```diff
@@ -0,0 +1,559 @@
...
+
+def _build_data_config(customizer_config: TrainingStepConfig, prepared: PreparedDataset) -> dict[str, Any]:
+    """Build the NeMo-RL ``data`` config.
+
+    NeMo-RL's ``setup_preference_data`` reads nested ``train`` / ``validation``
+    dataset specs (``dataset_name`` + ``data_path``) and builds the datasets via
+    its own registered loaders — no custom preprocessor. ``detect_dpo_schema_name``
+    already returns the NeMo-RL loader class name (e.g. ``BinaryPreferenceDataset``).
+    """
+    dataset_name = detect_dpo_schema_name(prepared.train_file)
```
albcui: `detect_dpo_schema_name` can return either `HelpSteer3` or `Tulu3Preference`, which we pass directly as `dataset_name` to the RL library. In the library, these map to `HelpSeer3Dataset()`/`Tulu3PreferenceDataset()`, whose init methods are hardcoded to load [`nvidia/HelpSteer3`](https://github.com/NVIDIA-NeMo/RL/blob/v0.6.0/nemo_rl/data/datasets/preference_datasets/helpsteer3.py#L33) and [`allenai/llama-3.1-tulu-3-8b-preference-mixture`](https://github.com/NVIDIA-NeMo/RL/blob/v0.6.0/nemo_rl/data/datasets/preference_datasets/tulu3.py#L30) (from HF!), and **silently ignore** `data_path` (it's swallowed by `**kwargs`). So a user who uploads a custom HelpSteer3/Tulu3-shaped dataset either trains on the *public* HF dataset instead, or crashes if it doesn't have access to HF. 

Both cases are bad.

I think we should either only support Binary/Preference datasets for now, or figure out a way to support custom HelpSteer3/Tulu3 datasets.

albcui (changes requested):

docker/Dockerfile.nmp-rl-tasks
anubhutivyas: It's unwired because there are no RL smoke tests yet. `tests/smoke_gpu/` only has automodel and  generic tasks tests.
If I wire it now, `pytest -m smoke_nmp_rl_tasks` would collect 0 tests and exit which would fail the bake. So wiring it requires first adding RL import smoke tests, I can add a comment here.

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/dpo_config.py
anubhutivyas: Thanks for catching this. It was a miss when I upgraded from v0.4.0 to v0.6.0, have updated the logic to load the dataset from the given path if it's available, else use the HF dataset.

`detect_dpo_schema_name` calls `get_preference_dataset_discriminator`  which returns one of four strings, "PreferenceDataset", "BinaryPreferenceDataset", "HelpSteer3", "Tulu3Preference" which are nemo-rl `DATASET_REGISTRY` keys - https://github.com/NVIDIA-NeMo/RL/blob/r0.6.0/nemo_rl/data/datasets/preference_datasets/__init__.py#L24

During training, when `dpo_driver.py::main` loads the YAML, then before loading data, it calls `register_preference_datasets` that overrides the `DATASET_REGISTRY` with our own implementation that also supports loading the data from the path.

services/rl/src/nmp/rl/entities/values.py
```diff
@@ -0,0 +1,40 @@
...
+class TrainingType(str, Enum):
+    """RL training algorithm. DPO is wired; GRPO is headroom.
+
+    PPO is intentionally absent: the backend only compiles DPO/GRPO, so
+    accepting a PPO value here would defer the failure to a runtime crash inside
+    the training container instead of failing fast at request validation.
+    """
+
+    DPO = "dpo"
+    GRPO = "grpo"
```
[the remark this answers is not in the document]
anubhutivyas: GRPO is intentional headroom and is already gated at every reachable layer. The public job schema is DPO-only so GRPO can't be submitted, validation schemas only has only DPO, and NemoRLBackend raises a NotImplementedError.

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/dpo_config.py
```diff
@@ -0,0 +1,559 @@
...
+    cfg["dpo"] = {
+        "max_num_epochs": epochs,
+        "max_num_steps": max_steps,
+        "steps_per_epoch": steps_per_epoch,
+        "val_period": val_period,
+        "val_batches": 0,  # Run the entire validation dataset
+        "val_global_batch_size": batch_size,
+        "val_micro_batch_size": micro_batch_size,
+        "val_at_start": True,
+        "val_at_end": customizer_config.schedule.val_at_end,
```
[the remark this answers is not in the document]
anubhutivyas: The schema explicitly allows False to skip the final eval; forcing it would override that. I have updated the docstring

docker/Dockerfile.nmp-rl-tasks
albcui: Yea please add a comment for future devs.

services/rl/src/nmp/rl/app/jobs/compiler.py
```diff
@@ -0,0 +1,316 @@
...
+        fileset=FileSetRef(workspace=None, name=job_spec.output.fileset),
+        model_entity=job_spec.model,
+        base_model=job_spec.model,
+        peft=None,
+        trust_remote_code=trust_remote_code,
+        deployment_config=None,
+    )
+
+
+def _build_training_step_config(job_spec: RlJobOutput, *, trust_remote_code: bool) -> TrainingStepConfig:
```
albcui, line 180: Missing `integrations`, so no support for wandb nor mlflow. Was this intentionally?

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/backend.py
```diff
@@ -0,0 +1,296 @@
...
+        checkpointing_config = library_config.config_dict["checkpointing"]
+        if checkpointing_config is None or not isinstance(checkpointing_config, dict):
+            raise ValueError("Checkpointing config is required to find the best checkpoint")
+
+        checkpointing_config = cast(CheckpointingConfig, checkpointing_config)
+        checkpointer = CheckpointManager(checkpointing_config)
+
+        try:
+            best_checkpoint = checkpointer.get_best_checkpoint_path()
+        except KeyError:
```
albcui: dead code? The library already filters missing-metric checkpoints, warns, and internally returns `get_latest_checkpoint_path()`[checkpoint.py](https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/utils/checkpoint.py#L338-L343). So the library already does the fallback under the hood, and this exception handler is a bit misleading.

services/rl/src/nmp/rl/tasks/training/datasets/validation.py
```diff
@@ -0,0 +1,382 @@
...
+    - PreferenceDataset: Native format with context + ranked completions
+    - BinaryPreferenceDataset: Simple prompt/chosen_response/rejected_response
+    - HelpSteer3Dataset: NVIDIA HelpSteer3 format with preference scores
+    - Tulu3PreferenceDataset: AllenAI Tulu3 format with message lists
+
+    Args:
+        file_path: Path to the dataset file (JSONL format)
+
+    Returns:
+        The NeMo RL dataset class name (e.g., "BinaryPreferenceDataset", "HelpSteer3Dataset")
```
albcui:
```suggestion
        The NeMo RL dataset class name (e.g., "BinaryPreferenceDataset", "HelpSteer3")
```

services/rl/src/nmp/rl/tasks/training/runner.py
```diff
@@ -0,0 +1,321 @@
...
+            config_dict = yaml.safe_load(f)
+
+        logger.info(f"Loaded library config from: {config_path}")
+        return LibraryConfig(config_dict=config_dict, config_path=config_path)
+
+    def _training_phase(self, library_config: LibraryConfig) -> TrainingMetrics:
+        """
+        Execute training on all ranks.
+
+        Training itself is distributed via torchrun, which handles inter-process coordination internally.
```
albcui:
```suggestion
        Training itself is distributed via Ray, which handles inter-process coordination internally.
```

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/dpo_config.py
```diff
@@ -0,0 +1,585 @@
...
+        spec: dict[str, Any] = {"dataset_name": dataset_name, "data_path": str(path)}
+        # BinaryPreferenceDataset reads explicit prompt/chosen/rejected keys; our
+        # datasets use those exact field names.
+        if dataset_name == "BinaryPreferenceDataset":
+            spec.update({"prompt_key": "prompt", "chosen_key": "chosen", "rejected_key": "rejected"})
+        return spec
+
+    return {
+        "max_input_seq_length": customizer_config.model.max_seq_length,
+        "shuffle": False,
```
albcui, line 396: Here, it'd be good to explain why we are disabling shuffle. I'm assuming it's for determinism, but future readers shouldn't have to assume.

services/rl/src/nmp/rl/app/jobs/compiler.py
anubhutivyas: Fixed

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/backend.py
anubhutivyas: you're right, remove the try/except block

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/dpo_config.py
anubhutivyas: Added a comment with explanation

docker/Dockerfile.nmp-rl-base
anubhutivyas: Since there are separate tags for different archs, I feel it would be better if we do not add the sha here, see here - https://catalog.ngc.nvidia.com/orgs/nvidia/-/containers/nemo-rl/-/tags

services/rl/src/nmp/rl/tasks/training/datasets/preparation.py
```diff
@@ -0,0 +1,548 @@
...
+
+    # Second pass: stream each non-empty row to train/val by index — re-serializing
+    # to normalize JSON — without ever holding the whole dataset in memory.
+    train_count = 0
+    val_count = 0
+    with (
+        open(train_file, "r", encoding="utf-8") as src,
+        open(output_train, "w", encoding="utf-8") as train_out,
+        open(output_val, "w", encoding="utf-8") as val_out,
+    ):
```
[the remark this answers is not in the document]
anubhutivyas: this is fixed

albcui (approved):

PR #480, opened 2026-06-25
fix(customizer): fix dockerfile organization for customizer, add AV opencv build with ffmpeg 8.1.2+

soluwalana:


mckornfield (approved):
just a question about the added complexity for building these wheels: do we use the VLM/opencv/av.lib bits during customization?

soluwalana:
@mckornfield we are likely to enable it soon, unsloth and automodel both allow it, and with the most recent rework it is much easier to allow them.

ironcommit (approved):
Awesome.

PR #485, opened 2026-06-26
fix(studio): Fix input size on Studio Code Agent and hide artifact panel on landing

htolentino-nvidia:
## Summary by CodeRabbit

* **New Features**
  * Added support for controlling the minimum height of the chat input, improving composer sizing in assistant chat views.
  * Introduced an option to hide artifacts in the Claude Code layout and history panel, with updated panel behavior when artifacts are hidden.

* **Bug Fixes**
  * Updated the Claude Code chat composer to start with a taller default input area for a better typing experience.
  * Added a close control for the history panel when artifacts are hidden.


dmariali (approved):
Nice! The input size isn't clashing with the stuff Octavian changed right?

PR #494, opened 2026-06-26
fix(jobs): Fix pause/resume races in K8s and subprocess backends

matthewgrossman:
## Summary

Fixes [AIRCORE-853](https://linear.app/nvidia/issue/AIRCORE-853/job-pauseresume-races-with-errored-pod-detection-in-k8s-reconciler): Job pause/resume races with errored-pod detection in both the K8s and subprocess backends.

### K8s backend

When a K8s Job is suspended (paused), Kubernetes kills running pods via SIGTERM. The reconciler's `sync_step` was checking for errored pods **before** checking if the job was suspended. The terminated pod (exit code 137 from SIGTERM) was misclassified as an error, causing the job to transition to `error` instead of `paused`. The same race existed for cancellation.

**Fix:** Moved the `is_suspended` and `is_cancelling` checks before the `errored_pods` check in `kubernetes_job.py`. Pods killed during suspension or cancellation are expected — they must not cause the reconciler to report ERROR.

### Subprocess backend

The subprocess backend had no pause/resume support at all. When a pause request came in, the process was never terminated, and if it was killed externally, the non-zero exit code was treated as an error (same race as K8s).

**Fix:** Added PAUSING handling to `subprocess.py`:
- Terminate the subprocess when the step is PAUSING (mirrors existing CANCELLING behavior)
- Return PAUSED (not ERROR) when the process exits with a non-zero code during pause
- Handle missing metadata for PAUSING steps (return PAUSED, not ERROR)

### E2E tests

Removed the `@pytest.mark.skip(reason="AIRCORE-853")` from `test_job_pause_resume` and `test_job_pause_and_cancel`. Reduced `sleep 300` → `sleep 30` so tests complete within the default 120s timeout after resume.

## Changes

- `services/core/jobs/.../kubernetes_job.py` — Reordered `sync_step` to check suspended/cancelling state before errored pods
- `services/core/jobs/.../subprocess.py` — Added pause support: process termination on PAUSING, PAUSED status on non-zero exit during pause, metadata-missing handler
- `services/core/jobs/tests/.../test_kubernetes_backend.py` — Added 3 regression tests:
  - `test_sync_job_paused_with_errored_pods_from_sigterm` — suspended + errored pod → PAUSED
  - `test_sync_job_pausing_with_errored_pods_from_sigterm` — suspended + mix of running/errored → PAUSING
  - `test_sync_job_cancelling_with_errored_pods` — cancelling + errored pod → CANCELLED
- `e2e/test_jobs.py` — Removed skip markers, reduced sleep duration

## Test plan

- [x] All 63 unit tests in `test_kubernetes_backend.py` pass
- [x] All 34 subprocess-related unit tests pass
- [x] Full `e2e/test_jobs.py` suite passes against subprocess (`nemo services run`): 10 passed, 3 skipped
- [x] Full `e2e/test_jobs.py` suite passes against fresh minikube: 12 passed, 1 skipped
- [x] `test_job_pause_resume` and `test_job_pause_and_cancel` pass on both backends


e2e/test_jobs.py
```diff
@@ -390,7 +385,9 @@ def test_job_pause_resume(sdk: NeMoPlatform, workspace: str):
                     "executor": {
                         "provider": "cpu",
                         "container": {
-                            "command": ["sh", "-c", "sleep 300"],
+                            # Short sleep so the job completes quickly after resume.
```
mckornfield, line 388: could it be even shorter? or the problem is it might finish before cancel/resume

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
```diff
@@ -282,6 +282,11 @@ def sync(self, step: PlatformJobStepWithContext) -> JobUpdate:
                     status=PlatformJobStatus.CANCELLED.value,
                     status_details={"message": "Subprocess not found, job cancelled"},
                 )
+            if step.status == PlatformJobStatus.PAUSING:
+                return JobUpdate(
+                    status=PlatformJobStatus.PAUSED.value,
+                    status_details={"message": "Subprocess not found, job paused"},
```
mckornfield, line 288: the metadata none case is weird

mckornfield (approved):

e2e/test_jobs.py
ironcommit: I always hate seeing sleeps in code.  Don't need to do this here, but an alternative is to spin up a service that the job can call to and that the test can call too, and use that to determine when the job can exit.  Just an idea for the future.

ironcommit (approved):
excellent, thanks for implementing that.

e2e/test_jobs.py
matthewgrossman: Yeah, the problem is exactly that — if the sleep is too short, the job completes before the pause request gets processed by the reconciler. The full cycle is: wait for active, pause, wait for paused, resume, wait for active, wait for completed. The reconciler polls on an interval, and on K8s there is pod scheduling time after resume too. 30s gives enough headroom for the round-trip while still finishing well within the 120s default timeout. We could probably go to ~15s but it risks flakiness on slower clusters.

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess.py
matthewgrossman: This mirrors the existing CANCELLING handler at line 280 — if the subprocess registry has lost track of the process (e.g., controller restart while a job is pausing), we can't do anything with it, so the safest thing is to report it as paused. Without this, a controller restart during a pause would leave the job stuck in 'pausing' forever, or error it out because metadata is missing. The pattern is: if you can't find the thing you were supposed to stop, consider it stopped.

e2e/test_jobs.py
matthewgrossman: Agreed, good idea going forward.

PR #496, opened 2026-06-29
feat(evaluator): agent-evaluation plugin job (run_local + submit)

SandyChapman:
## What

Adds **`AgentEvalJob`** — a plugin-native job that runs the agent-eval SDK (`AgentEvaluator`) over a taskset against a **model, agent, or agent runner**, either locally (`run_local`) or submitted to a backend (subprocess / docker / k8s). This is the [AALGO-297](https://linear.app/) "agent evaluation as a plugin job" work.

## Design notes

- **Target model is kind-discriminated:** `ModelTarget | AgentTarget | AgentRunnerTarget` (`CodexRunnerTarget` today), unioned as `Target`. The prompt lives where it belongs — on the model target — rather than as a job-level `prompt_template`. A spec must provide **exactly one** trial source: an online `target`, or precomputed `trials` for an offline eval.
- **Online inference forwards the job's platform identity** (the task SDK's `X-NMP-*` headers — including `X-NMP-Principal-Id: service:evaluator`) to the evaluator, so a platform-routed target acts as the job's principal. No bearer is forwarded, so the platform token never leaks to a third-party endpoint. **Verified against an auth-enabled platform: these service-principal headers are sufficient** (the PDP grants `service:*` full permissions) — no bearer is needed. Identity propagation happens at run time (the on-behalf-of principal only exists when the task runs), and rides a generic `default_headers` seam, so the evaluator SDK stays platform-agnostic.

## Shared plumbing extracted

- `tasks/runner.py` — one container entrypoint (SIGTERM handling, task SDK build, dispatch) shared by both evaluator jobs; per-job modules are thin `python -m` targets.
- `secret_env.py` — `build_task_environment()` for metric/endpoint secrets.
- `metric_resolution.py` — stored-ref + model-ref resolution shared by both jobs. Stored refs need a real `AsyncNeMoPlatform`; model resolution duck-types sync/async clients via a `runtime_checkable` `ModelResolverSdk` protocol.

## Testing

Unit: full evaluator suite green (352 passed). Integration matrix (real platform fixtures):

| Test | Result |
|---|---|
| `run_local` model / agent (IGW mock) / codex (real) | ✅ pass |
| submit → subprocess backend | ✅ pass |
| submit with a stored `MetricRef` | ✅ pass |
| **submit Model target under `auth.enabled` → IGW** (forwarded identity authenticates online inference end-to-end) | ✅ pass |
| submit → docker backend | ⚠️ xfail (until the `cpu-tasks` image ships the task module) |

Auth-sufficiency probed directly too: a call to the IGW model route with the service-principal headers returns **200**, the same call with no auth returns **401** — so the forwarding is both necessary and sufficient under auth.

`nmp-testing` is now an explicit evaluator dev dependency.

## Follow-ups (Monday)

- SDK / OpenAPI regen for the new `/agent-evaluate/jobs` routes.
- AALGO-301 (container/runner image), AALGO-302 (`intent` leak into the agent prompt).
- 3 pre-existing `ty` diagnostics in `test_evaluate_job.py` (`format="nim"` literal, `unbundle_metric(MetricInline)`) are unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added an **agent-evaluate** job with model, agent, and Codex runner targets, including plugin-native compilation and local/platform execution.
  * Exposed new **job API routes** and a **nemo.jobs** entry point, plus a dedicated **OpenAPI** contract for agent-evaluation jobs.

* **Bug Fixes**
  * Strengthened job-completion polling with safety assertions.
  * Improved secret-backed task environment construction and identity header forwarding, and tightened evaluator/spec validation.

* **Tests**
  * Expanded unit and integration coverage (inline/stored metrics, auth, Codex, and docker backends).

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
@@ -44,7 +44,7 @@ class AgentOutput(BaseModel):
         default=None,
         description="User-visible final text produced by the agent, if any.",
     )
-    response: Any | None = Field(
+    response: dict[str, Any] | None = Field(
         default=None,
         description="Structured final response payload produced by the agent, if any.",
     )
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
@@ -15,7 +15,7 @@ from typing import Any, Protocol, runtime_checkable
 from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig, AgentEvalTask
 from nemo_evaluator_sdk.values import Agent, Model
 from nemo_evaluator_sdk.values.evidence import CandidateEvidence, EvidenceDescriptor
-from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, Field, JsonValue, field_validator, model_validator
 
 # Well-known evidence keys produced by ``standard_evidence_descriptors``. Harness
 # code may import these to tag evidence consistently; callers may still add
@@ -44,9 +44,10 @@ class AgentOutput(BaseModel):
         default=None,
         description="User-visible final text produced by the agent, if any.",
     )
-    response: dict[str, Any] | None = Field(
+    response: JsonValue | None = Field(
         default=None,
-        description="Structured final response payload produced by the agent, if any.",
+        description="Final response payload produced by the agent, if any. Any JSON value — a "
+        "structured object, or a raw JSON string/array for agents that don't return an object.",
     )
     metadata: dict[str, Any] = Field(
         default_factory=dict,
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
@@ -29,7 +29,21 @@ class Agent(BaseModel):
 
     # TODO: Much of this is duplicated between agent and model. Once we have aligned on model defination.
     # the duplication can be removed by defining EndPoint class and reusing it across both model and agent.
-    model_config = ConfigDict(extra="forbid")
+    #
+    # ``allOf``/``if``/``then`` mirrors the ``_validate_generic_fields`` validator into the OpenAPI
+    # schema: a generic-format agent must carry ``body`` + ``response_path`` (the generic HTTP path
+    # needs them), so the contract rejects a ``url``-only generic agent rather than only failing later.
+    model_config = ConfigDict(
+        extra="forbid",
+        json_schema_extra={
+            "allOf": [
+                {
+                    "if": {"properties": {"format": {"const": "generic"}}},
+                    "then": {"required": ["body", "response_path"]},
+                }
+            ]
+        },
+    )
 
     url: str = Field(description="Base URL of the agent endpoint.")
     name: str = Field(description="Agent name / identifier.")
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -45,6 +45,380 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs:
+    post:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Create Job
+      operationId: create_job_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs_post
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AgentEvaluateJobRequest'
+      responses:
+        '201':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AgentEvaluateJob'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: List Jobs
+      operationId: list_jobs_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: page
+        in: query
+        required: false
+        schema:
+          type: integer
+          exclusiveMinimum: 0
+          description: Page number.
+          default: 1
+          title: Page
+        description: Page number.
+      - name: page_size
+        in: query
+        required: false
+        schema:
+          type: integer
+          exclusiveMinimum: 0
+          description: Page size.
+          default: 10
+          title: Page Size
+        description: Page size.
+      - name: sort
+        in: query
+        required: false
+        schema:
+          allOf:
+          - $ref: '#/components/schemas/AgentEvaluateJobsSortField'
+          description: The field to sort by. To sort in decreasing order, use `-`
+            in front of the field name.
+          default: -created_at
+        description: The field to sort by. To sort in decreasing order, use `-` in
+          front of the field name.
+      - in: query
+        name: filter
+        style: deepObject
+        required: false
+        explode: true
+        schema:
+          $ref: '#/components/schemas/AgentEvaluateJobsListFilter'
+        description: Filter jobs on various criteria.
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AgentEvaluateJobsPage'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{job}/results/{name}:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Get Job Result
+      operationId: get_job_result_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__job__results__name__get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: job
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Job
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PlatformJobResultResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{job}/results/{name}/download:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Download Job Result
+      operationId: download_job_result_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__job__results__name__download_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: job
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Job
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+        '404':
+          description: Not Found
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{name}:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Get Job
+      operationId: get_job_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AgentEvaluateJob'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+    delete:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Delete Job
+      operationId: delete_job_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__delete
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '204':
+          description: Successful Response
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{name}/cancel:
+    post:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Cancel Job
+      operationId: cancel_job_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__cancel_post
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AgentEvaluateJob'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{name}/logs:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Get Job Logs
+      operationId: get_job_logs_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__logs_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      - name: limit
+        in: query
+        required: false
+        schema:
+          title: Limit
+          type: integer
+      - name: page_cursor
+        in: query
+        required: false
+        schema:
+          title: Page Cursor
+          type: string
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PlatformJobLogPage'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{name}/results:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: List Job Results
+      operationId: list_job_results_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__results_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PlatformJobListResultResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /apis/evaluator/v2/workspaces/{workspace}/agent-evaluate/jobs/{name}/status:
+    get:
+      tags:
+      - Evaluator Plugin Agent Eval Jobs Routes
+      summary: Get Job Status
+      operationId: get_job_status_apis_evaluator_v2_workspaces__workspace__agent_evaluate_jobs__name__status_get
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PlatformJobStatusResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
   /apis/evaluator/v2/workspaces/{workspace}/evaluate/jobs:
     post:
       tags:
@@ -605,6 +979,15 @@ paths:
 components:
   schemas:
     Agent:
+      allOf:
+      - if:
+          properties:
+            format:
+              const: generic
+        then:
+          required:
+          - body
+          - response_path
       properties:
         url:
           type: string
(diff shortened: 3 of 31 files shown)


plugins/nemo-evaluator/tests/integration/test_agent_evaluate_job.py
```diff
@@ -0,0 +1,356 @@
...
+CODEX_MODEL = "gpt-5.5"
+
+WORKSPACE = "default"
+
+
+def _codex_available() -> bool:
+    return shutil.which("codex") is not None
+
+
+requires_codex = pytest.mark.skipif(not _codex_available(), reason="codex CLI not on PATH")
```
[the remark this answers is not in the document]
SandyChapman: Intentionally not changing this. The `skipif` condition is evaluated at **import** (collection), so a `codex login status` subprocess would run during collection in *every* CI job — including unit/lint that never touch these tests — for a gate that, after the `RUN_AGENT_EVAL_INTEGRATION` opt-in, only matters on local runs where codex is both installed and logged in. The cheap `shutil.which` is the better tradeoff: if codex is on PATH but unauthenticated, the test fails loudly (correct signal locally) rather than paying a subprocess at collection time everywhere.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
```diff
@@ -29,7 +29,21 @@ class Agent(BaseModel):
...
+    # ``allOf``/``if``/``then`` mirrors the ``_validate_generic_fields`` validator into the OpenAPI
+    # schema: a generic-format agent must carry ``body`` + ``response_path`` (the generic HTTP path
+    # needs them), so the contract rejects a ``url``-only generic agent rather than only failing later.
+    model_config = ConfigDict(
+        extra="forbid",
+        json_schema_extra={
+            "allOf": [
+                {
+                    "if": {"properties": {"format": {"const": "generic"}}},
+                    "then": {"required": ["body", "response_path"]},
```
arpitsardhana, line 42: Today I learned :) that if/else can be done in pydantic serde

plugins/nemo-evaluator/src/nemo_evaluator/resolvers.py
```diff
@@ -54,8 +54,14 @@ class _InferenceResource(Protocol):
     providers: _ProvidersResource
 
 
-class _PlatformSDK(Protocol):
-    """Minimal platform SDK surface used by this resolver."""
+@runtime_checkable
+class ModelResolverSdk(Protocol):
```
arpitsardhana: nit: I guess convention is use ModelResolverSDK since SDK is itself acronym

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,259 @@
...
+
+Runs :class:`AgentEvaluator` over a set of tasks against a Model/Agent endpoint
+or an agent runner (e.g. Codex CLI), producing an ``AgentEvalResult`` (trials +
+per-trial scores + summary). The row-based counterpart is
+:class:`~nemo_evaluator.jobs.evaluate.EvaluateJob`.
+
+Per-task metrics may be given inline or as references to stored metrics;
+references are resolved into inline metrics during ``to_spec`` via the shared
+:mod:`nemo_evaluator.jobs.metric_resolution` helper. The result bundle is
+persisted as job artifacts; persisting it to the platform (entity + fileset) is
```
arpitsardhana: nit: file description need not contain follow-ups?
That can be TODO somewhere

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_compiler.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def _agent_eval_step(spec: AgentEvalSpec, profile: str | None) -> PlatformJobStep:
+    return PlatformJobStep(
+        name=AGENT_EVAL_STEP_NAME,
+        executor=CPUExecutionProviderSpec(
+            profile=profile or "default",
+            provider="cpu",
+            container=ContainerSpec(
+                image=get_qualified_image("nmp-cpu-tasks"),
```
arpitsardhana: nit: image name, entry point and command should be constant somewhere preferably.

In long run, do we see image name, entry point and command all to be configurable via job spec?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,259 @@
...
+    job_collection_path: ClassVar[str | None] = "/agent-evaluate/jobs"
+
+    @classmethod
+    async def to_spec(
+        cls,
+        input_spec: BaseModel,
+        *,
+        workspace: str,
+        entity_client: object,
+        async_sdk: AsyncNeMoPlatform | NeMoPlatform | None,
```
arpitsardhana: nit: shouldn't that async_sdk: by design of type AsyncNemoPlatform? and not overloaded

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -0,0 +1,184 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    kind: Literal["agent"] = "agent"
+    agent: Agent = Field(description="The agent endpoint to generate trials against.")
+    params: RunConfigOnline | None = Field(
+        default=None, description="Optional online-inference parameters for trial generation."
+    )
+
+
+class CodexRunnerTarget(BaseModel):
```
arpitsardhana, line 67: Should there be a common class representing AgentEvalTaskRunner and Codex being derived from it. So that Cursor, Claude or harbor runner can be added easily?

Put in another way, can this be extended easily to introduce more runner target?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,259 @@
...
+        NOTE: bearer-token auth for platform routes in an auth-enabled deployment is not yet
+        forwarded (the local/internal path relies on the ``X-NMP-*`` identity headers); see
+        AALGO-297 follow-ups.
+        """
+        identity_headers: dict[str, str] = {}
+        if platform is not None:
+            identity_headers = {
+                key: value
+                for key, value in platform.default_headers.items()
+                if key in _FORWARDED_IDENTITY_HEADERS and isinstance(value, str)
```
arpitsardhana, line 220: It may have potential to transfer PII information like id and email to third party URL(like chat.openai.com)  which user has provided in configs. Third party does not care about headers != best strategy to transfer header. I remember vividly we have added a check that checks if provider is IGW or internal based on if "host_url: shows up in url and then only add these headers in those cases
I am wondering if we can those check and add the these headers only for internally routed IGW based urls

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -0,0 +1,184 @@
...
+        description="What generates trials online: a Model or Agent endpoint, or an agent runner (e.g. Codex "
+        "CLI). Endpoint targets carry their own request config (prompt template / inference params). "
+        "Mutually exclusive with `trials`.",
+    )
+    trials: list[AgentEvalTrial] | None = Field(
+        default=None,
+        description="Precomputed trials to score directly (offline eval), instead of generating them from a "
+        "`target`. Mutually exclusive with `target`.",
+    )
+    parallelism: int = Field(default=4, ge=1, description="Maximum number of tasks scored concurrently.")
```
arpitsardhana: Is there a target level parallelism as well?
I am not very sure but I could sense that there are different place in full spec where parallelism is defined like spec.parallelism versus params.parallelism in same spec --> this may confuse end user/agent

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
```diff
@@ -44,7 +44,7 @@ class AgentOutput(BaseModel):
         default=None,
         description="User-visible final text produced by the agent, if any.",
     )
-    response: Any | None = Field(
+    response: dict[str, Any] | None = Field(
```
arpitsardhana: nit: there can be cases where response is json string and not a structured dictionary? This might limit spectrum of agents addressed

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -0,0 +1,184 @@
...
+
+class _AgentEvalSpecCommon(BaseModel):
+    """Fields shared by the submitter input and canonical agent-eval specs (everything but ``tasks``)."""
+
+    # ``oneOf`` mirrors the ``_require_exactly_one_trial_source`` validator into the OpenAPI schema, so
+    # the generated contract (and clients) reject a target-less or both-supplied request instead of
+    # only discovering it via a 422 at runtime.
+    model_config = ConfigDict(
+        extra="forbid",
+        json_schema_extra={"oneOf": [{"required": ["target"]}, {"required": ["trials"]}]},
```
arpitsardhana: if someone sends "target": null, trials: [...]

will oneOf  reject above as well right despite being correct?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,259 @@
...
+                AgentEvalTaskSpec(
+                    id=task.id,
+                    intent=task.intent,
+                    inputs=task.inputs,
+                    metrics=metrics,
+                    views=task.views,
+                    metadata=task.metadata,
+                )
+            )
+        return AgentEvalSpec(
```
arpitsardhana, line 145: Question: We create spec here and then spec is passed to container, where SDK executes them.
Do we need to have some sort of prechecks  like verifying that trials and task_id match before we pass on spec rather than fail at late stage. This something which can as followup post MVP

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
```diff
@@ -29,7 +29,21 @@ class Agent(BaseModel):
 
     # TODO: Much of this is duplicated between agent and model. Once we have aligned on model defination.
     # the duplication can be removed by defining EndPoint class and reusing it across both model and agent.
-    model_config = ConfigDict(extra="forbid")
+    #
+    # ``allOf``/``if``/``then`` mirrors the ``_validate_generic_fields`` validator into the OpenAPI
+    # schema: a generic-format agent must carry ``body`` + ``response_path`` (the generic HTTP path
```
ngoncharenko, line 34: This is probably out of scope, but wouldn't it make sense to decouple the `generic` and `nemo_agent_toolkit` into two different classes and make an Agent a union?

```py
class AgentBase(BaseModel):
    url: str
    name: str
    api_key_secret: SecretRef | None = None
    # shared cached_property helpers...

class GenericAgent(AgentBase):
    format: Literal[AgentFormat.GENERIC] = AgentFormat.GENERIC
    body: dict
    response_path: str
    trajectory_path: str | None = None

class NemoAgentToolkitAgent(AgentBase):
    format: Literal[AgentFormat.NEMO_AGENT_TOOLKIT] = AgentFormat.NEMO_AGENT_TOOLKIT

Agent = Annotated[
    GenericAgent | NemoAgentToolkitAgent,
    Field(discriminator="format"),
]
```

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Good catch. I had thought I caught all of the Claude docstring nonsense.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
SandyChapman: 🙂 Yep — keeps the wire contract explicit in the schema. No change needed.

plugins/nemo-evaluator/src/nemo_evaluator/resolvers.py
SandyChapman: Done — renamed to `ModelResolverSDK` in fca4215.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Agreed — dropped the follow-up note from the module docstring in fca4215.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_compiler.py
SandyChapman: Pulled image/entrypoint/command into module constants (fca4215). Making them spec-configurable is plausible later, but they're fixed for the MVP — noted in the constants' comment.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Good catch — the sync arm was unused (and unawaitable in an async resolver), so I narrowed it to `AsyncNeMoPlatform | None` in fca4215.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
SandyChapman: `AgentRunnerTarget` is already the kind-discriminated slot for exactly this. Holding off on a shared base class until a second runner (Cursor/Claude/Harbor) actually lands so we don't guess the abstraction — tracked as a follow-up.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Fixed in fca4215: identity headers now forward **only** when the target is platform-routed (its host matches the platform/IGW base URL); a third-party endpoint (or a runner) gets none, so the on-behalf-of id/email/groups never leaves the platform. This is the IGW/internal check you remembered.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
SandyChapman: Agreed — renamed the spec field to `max_concurrent_tasks` (fca4215) to disambiguate from a target's `params.parallelism`. They're genuinely different stages (task scoring vs. concurrent inference requests), now spelled out in the field doc; there's no separate target-level knob beyond `params.parallelism`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
SandyChapman: Good point — widened `AgentOutput.response` to `JsonValue` in fca4215, so a raw JSON string/array is accepted, not just a structured object.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
SandyChapman: Verified — yes, the presence-based `oneOf` would've rejected an explicit `target: null` + `trials`. Tightened each branch to exclude `null` (fca4215), so the schema now matches the runtime validator (non-null XOR).

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Agree it's worth doing; deferring to post-MVP as you suggested — tracked as a follow-up. (Today a mismatch fails inside the run rather than at submit.)

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
SandyChapman: Like this a lot — it'd also let the generic-format `body`/`response_path` requirement fall out of the schema instead of a validator. Out of scope for this PR (an SDK-values refactor with some blast radius), capturing it as a follow-up.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,242 @@
...
+        forwarded (the local/internal path relies on the ``X-NMP-*`` identity headers); see
+        AALGO-297 follow-ups.
+        """
+        identity_headers: dict[str, str] = {}
+        if platform is not None:
+            identity_headers = {
+                key: value
+                for key, value in platform.default_headers.items()
+                if key.startswith("X-NMP-") and isinstance(value, str)
+            }
```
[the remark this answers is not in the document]
SandyChapman: Already an explicit allowlist (`_FORWARDED_IDENTITY_HEADERS`), not an `X-NMP-*` prefix match — and as of fca4215 forwarding is additionally gated on the target being platform-routed, so identity never reaches a third-party endpoint.

plugins/nemo-evaluator/tests/integration/conftest.py
```diff
@@ -0,0 +1,220 @@
...
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            with urllib.request.urlopen(f"{base_url}/health/ready", timeout=2) as response:  # noqa: S310
+                if response.status == 200:
+                    return
+        except OSError:
+            pass
+        time.sleep(2)
+    raise RuntimeError(f"platform at {base_url} not ready within {timeout}s")
```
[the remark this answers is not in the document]
SandyChapman: Already handled — `_wait_for_ready` checks `process.poll()` each iteration and raises immediately if the subprocess exited before becoming ready.

plugins/nemo-evaluator/tests/integration/conftest.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    try:
+        _wait_for_ready(base_url, timeout=ready_timeout)
+        yield base_url
+    finally:
+        process.terminate()
+        try:
+            process.wait(timeout=20)
+        except subprocess.TimeoutExpired:
+            process.kill()
```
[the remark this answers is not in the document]
SandyChapman: Already handled — teardown calls `process.wait()` after `process.kill()` to reap the child.

plugins/nemo-evaluator/tests/integration/test_agent_evaluate_job.py
```diff
@@ -0,0 +1,356 @@
...
+        _codex_eval_input_spec(),
+        base_url=docker_platform,
+        workspace=WORKSPACE,
+        profile="default",
+    )
+    job_name = response.get("name") or response.get("id")
+    assert job_name, f"submit response carried no job name/id: {response}"
+
+    job = wait_for_platform_job(client, job_name, WORKSPACE, timeout=480)
+    assert job.status == "completed"
```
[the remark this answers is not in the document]
SandyChapman: Deferred — the docker path is `xfail(strict=False)` pending the cpu-tasks image rebuild (AALGO-301). We'll revisit the target choice when it un-xfails so it isolates the backend wiring.

plugins/nemo-evaluator/tests/test_agent_evaluate.py
```diff
@@ -0,0 +1,454 @@
...
+    forwarded so the platform token never leaks to a third-party endpoint. A submitted job is
+    injected the sync ``NeMoPlatform``; ``run_local`` may inject the async counterpart."""
+    sdk = sdk_cls(
+        base_url="http://platform",
+        default_headers={"X-NMP-Internal": "true", "X-NMP-Principal-On-Behalf-Of": "user-1"},
+    )
+
+    evaluator = AgentEvalJob._build_evaluator(sdk)
+
+    assert evaluator.default_headers == {"X-NMP-Internal": "true", "X-NMP-Principal-On-Behalf-Of": "user-1"}
```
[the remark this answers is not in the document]
SandyChapman: Already covered — the test injects `Authorization: Bearer` plus a non-identity `X-NMP-*` and asserts both are dropped. fca4215 adds a third-party-target case (no identity forwarded at all).

plugins/nemo-evaluator/tests/test_agent_evaluate.py
```diff
@@ -0,0 +1,454 @@
...
+    assert result["status"] == "completed"
+    assert fake.received_target is None
+    assert [trial.id for trial in fake.received_trials or []] == ["t-1"]
+
+
+def test_spec_requires_exactly_one_of_target_or_trials() -> None:
+    with pytest.raises(ValueError, match="exactly one"):
+        AgentEvalSpec(tasks=[_task_spec()])  # neither target nor trials
+    with pytest.raises(ValueError, match="exactly one"):
+        AgentEvalSpec(tasks=[_task_spec()], target=CodexRunnerTarget(model="gpt-5.5"), trials=[])  # both
```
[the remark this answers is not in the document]
SandyChapman: The XOR test does exercise the rule — `trials` has no min-length, so `[]` is a valid non-null value and the mutual-exclusion validator fires when a target is also set. Agreed it reads clearer with a seeded trial though; minor, can tidy in a follow-up.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -606,55 +980,539 @@ components:
...
+          description: 'What generates trials online: a Model or Agent endpoint, or
+            an agent runner (e.g. Codex CLI). Endpoint targets carry their own request
+            config (prompt template / inference params). Mutually exclusive with `trials`.'
+        trials:
+          title: Trials
+          description: Precomputed trials to score directly (offline eval), instead
+            of generating them from a `target`. Mutually exclusive with `target`.
+          items:
+            $ref: '#/components/schemas/AgentEvalTrialInput'
+          type: array
```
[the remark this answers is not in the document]
SandyChapman: Encoded — the schema already carries the `oneOf`; fca4215 tightens each branch to also reject an explicit `null`, matching the runtime validator.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -985,6 +1900,39 @@ components:
...
+          additionalProperties: true
+          type: object
+          title: Metadata
+          description: Free-form metadata associated with the evidence descriptor.
+      additionalProperties: false
+      type: object
+      required:
+      - kind
+      title: EvidenceDescriptor
+      description: Descriptor for a candidate trace, source, or artifact.
```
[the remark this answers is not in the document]
SandyChapman: Already encoded — `EvidenceDescriptor` carries an `anyOf` requiring `ref` or `data` in the generated spec.

plugins/nemo-evaluator/tests/integration/conftest.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Runs in ``pytest_configure`` (before collection) rather than at module import so it can be gated:
+    skipped under ``-m 'not integration'``, so it never leaks the env var into a unit run that merely
+    collects this conftest (which would break the inference-gateway ``is_mock_provider`` tests). A
+    plugin must not import ``nmp-common`` to set the config directly, so the documented env var is the
+    seam. The platform *subprocess* gets the same prefix via each fixture's ``env_overrides``.
+    """
+    if "not integration" in (config.option.markexpr or ""):
+        return
+    os.environ.setdefault(MOCK_PROVIDER_PREFIX_ENVVAR, MOCK_PROVIDER_PREFIX)
```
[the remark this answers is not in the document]
SandyChapman: Refactored since this comment — the fixture applies a single `MOCK_PROVIDER_PREFIX` via `igw_mock_provider_mode` (test process) and `env_overrides` (subprocess); it no longer reads a pre-set env var, so there's one source of truth.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -606,55 +980,539 @@ components:
...
+        response_path:
+          title: Response Path
+          description: JSONPath expression to extract the response text from the agent's
+            response body. Required for generic agents.
+          type: string
+        trajectory_path:
+          title: Trajectory Path
+          description: JSONPath expression to extract the trajectory from the agent's
+            response body. Optional.
+          type: string
```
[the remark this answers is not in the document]
SandyChapman: Already encoded — `Agent` carries an `allOf`/`if`/`then` requiring `body` + `response_path` when `format=generic` in the generated spec.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
SandyChapman: Filed as [AALGO-304](https://linear.app/nvidia/issue/AALGO-304) (Cycle 15) to revisit when a second runner lands.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py
SandyChapman: Filed as [AALGO-305](https://linear.app/nvidia/issue/AALGO-305) (Cycle 15) — the refactor would also let the generic-agent `body`/`response_path` requirement fall out of the schema instead of the validator/`if-then`.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Filed as [AALGO-306](https://linear.app/nvidia/issue/AALGO-306) (Cycle 15) for the post-MVP submit-time validation.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -0,0 +1,283 @@
...
+        cls,
+        input_spec: BaseModel,
+        *,
+        workspace: str,
+        entity_client: object,
+        async_sdk: AsyncNeMoPlatform | None,
+        is_local: bool,
+    ) -> BaseModel:
+        """Resolve each task's metric references into inline metrics for the canonical spec."""
+        del is_local
```
arpitsardhana, line 121: nit: Why is input parameter deleted at start of function?

arpitsardhana (approved):

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: I've noticed Claude does this frequently to suppress unused param errors from the typechecker/linter. It's a bit weird, but fine IMO.

PR #499, opened 2026-06-29
feat(studio): GroupExperiments sort based on metics

nakolean:
https://github.com/user-attachments/assets/440db4f1-de86-4d41-992c-cd8de1ab5439

## Summary by CodeRabbit

* **New Features**
  * Enhanced the experiment table’s sorting with additional sortable metric columns and evaluator score columns.
  * Improved recovery for recoverable sort failures, automatically restoring the last working sort and surfacing a clearer error message.

* **Bug Fixes**
  * Prevented full-page error states when a recoverable sorting issue occurs (sorting UI reverts instead).
  * Tooltip content is now only shown when a value is available, avoiding empty/incorrect tooltip displays.

### DIFF
--- a/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/useExperimentGroupExperiments.ts
+++ b/web/packages/studio/src/components/dataViews/ExperimentGroupDataView/useExperimentGroupExperiments.ts
@@ -121,6 +121,7 @@ export function useExperimentGroupExperiments({
     isLoading: isUnpinnedLoading,
     isFetching: isUnpinnedFetching,
     isSuccess: isUnpinnedSuccess,
+    isPlaceholderData: isUnpinnedPlaceholder,
     error: unpinnedError,
   } = useListExperiments(
     workspace,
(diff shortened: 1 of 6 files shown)


walston (approved):

web/packages/studio/src/components/dataViews/ExperimentGroupDataView/useExperimentGroupExperiments.ts
```diff
@@ -196,6 +203,9 @@ export function useExperimentGroupExperiments({
   // both lists have loaded once, rather than clearing as soon as the faster query returns.
   const isLoading = isPinnedLoading || isUnpinnedLoading;
   const isFetching = isPinnedFetching || isUnpinnedFetching;
+  // Only the unpinned query carries the caller's (potentially metric-backed) sort; the pinned query
+  // always sorts by `-pinned_at`. So sort-error recovery keys off the unpinned query's success.
+  const isSuccess = isUnpinnedSuccess;
```
shanaiabuggy: With `placeholderData: keepPreviousData` on this query, `isUnpinnedSuccess` stays `true` while a *new* sort is still in flight — it's showing the previous page as placeholder (`isPlaceholderData: true`), and keepPreviousData reports `status: 'success'` during that window.

That breaks `useSortErrorRecovery`: its `if (isSuccess) lastGoodSort = sortingState` effect records the new, about-to-fail sort as the "last good sort" *before* it errors. So when the request 413/503s, `isSameSort(current, lastGood)` is `true`, `isRecoverableSortError` is `false`, and we fall through to the full-page "Failed to load experiments" — the case this PR is trying to avoid.

Suggest gating out placeholder data so `isSuccess` means "the current sort actually loaded":

```ts
const {
  /* ... */
  isSuccess: isUnpinnedSuccess,
  isPlaceholderData: isUnpinnedPlaceholder,
} = useListExperiments(/* unpinned */);

// ...

const isSuccess = isUnpinnedSuccess && !isUnpinnedPlaceholder;
```

Heads up that the current unit tests pass only because the in-flight mocks hard-code `isSuccess: false`, which doesn't match keepPreviousData's real behavior — worth a test that exercises the placeholder window (in-flight render with `isSuccess: true` + `isPlaceholderData: true`).

shanaiabuggy (approved):
left 1 comment!

PR #508, opened 2026-06-29
fix(models): lora model revision re-pull logic

benmccown:
## Follow-ups to generic engine support (#476)

Addresses two items deferred from the generic engine PR review.

### 1. Fix stale-weights bug in the k8s re-pull check (`fix`)
`_existing_model_source()` only read the model-source annotation off the weight-puller Job, but that Job is deleted at P3 once weights are pulled. For an already-serving deployment the lookup returned `None`, so a later `model_revision` change skipped the re-pull and served stale weights.

Now reads the Job first, then falls back to the (long-lived) PVC annotation, and propagates non-404 API errors instead of swallowing them. Pre-existing bug in the vLLM weighted path; surfaced more broadly now that weight-aware generic deployments ride the same path.

### 2. Configurable `run_as_user` / `run_as_group` (`feat`)
Adds `executor_config.run_as_user` / `run_as_group` to override the serving pod's securityContext uid/gid on the k8s backend. An explicit value wins over the engine default (vLLM pins its image's 2000/0; generic runs as the image's own user). Applied to both the serving container and the puller Job so pulled weights are owned by the same user the server reads them as. Ignored by the docker backend (it sets no container user).

Includes OpenAPI + Python SDK regeneration.

## Summary by CodeRabbit

* **New Features**
  * Added optional `run_as_user` and `run_as_group` settings to container executor configuration.
  * When using the Kubernetes backend, these fields map to the serving container’s pod `securityContext` (k8s only; docker ignores them).

* **Bug Fixes**
  * Improved rollout/update behavior when model puller jobs or annotations are missing, including a safer fallback to PVC-stored metadata.
  * Non-recoverable Kubernetes API errors are no longer silently swallowed, improving reliability during updates.

### DIFF
--- a/services/core/models/src/nmp/core/models/controllers/backends/common.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/common.py
@@ -55,6 +55,8 @@ class DeploymentConfigView:
     image_name: Optional[str] = None
     image_tag: Optional[str] = None
     health_check_path: Optional[str] = None
+    run_as_user: Optional[int] = None
+    run_as_group: Optional[int] = None
     additional_envs: Optional[Dict[str, str]] = None
     additional_args: Optional[List[str]] = None
     k8s_nim_operator_config: Optional[K8sNIMOperatorConfig] = None
@@ -78,6 +80,8 @@ def deployment_config_view(config: Optional[_DeploymentConfigLike]) -> Deploymen
         image_name=getattr(executor, "image_name", None),
         image_tag=getattr(executor, "image_tag", None),
         health_check_path=getattr(executor, "health_check_path", None),
+        run_as_user=getattr(executor, "run_as_user", None),
+        run_as_group=getattr(executor, "run_as_group", None),
         additional_envs=getattr(executor, "additional_envs", None),
         additional_args=getattr(executor, "additional_args", None),
         k8s_nim_operator_config=getattr(executor, "k8s_nim_operator_config", None),
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
@@ -37,6 +37,7 @@ from nmp.core.models.app import ModelWeightsType, get_deployment_resource_name
 from nmp.core.models.app.constants import MODEL_MANAGED_BY_LABEL, MODEL_MANAGED_BY_MODELS_CONTROLLER
 from nmp.core.models.controllers.backends import generic_compiler, vllm_compiler
 from nmp.core.models.controllers.backends.backends import DeploymentStatusUpdate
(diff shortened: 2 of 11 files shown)


services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
```diff
@@ -718,14 +728,42 @@ def _find_job_pod_name(self, job_name: str) -> str | None:
...
             )
-        except k8s_client.exceptions.ApiException:
-            return None
-        annotations = (job.metadata.annotations or {}) if job.metadata else {}
+            annotations = (job.metadata.annotations or {}) if job.metadata else {}
+            source = annotations.get(vllm_k8s_compiler.MODEL_SOURCE_ANNOTATION)
+            if source:
+                return source
+        except k8s_client.exceptions.ApiException as e:
+            if e.status != 404:
```
mckornfield, line 754: nit- sorta weird to see this but below it's the opposite... though maybe we just want to return None only in the 404 case for the read, so it makes sense that they're flipped?

services/core/models/src/nmp/core/models/controllers/backends/common.py
```diff
@@ -55,6 +55,8 @@ class DeploymentConfigView:
     image_name: Optional[str] = None
     image_tag: Optional[str] = None
     health_check_path: Optional[str] = None
+    run_as_user: Optional[int] = None
```
mckornfield, line 58: ALL the params

mckornfield (approved):

services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/k8s.py
benmccown: Response from claude session: Your read is right — the inversion is intentional. The Job is the primary source but is deleted at P3, so a Job 404 should fall through to the PVC fallback (hence `if status != 404: raise`). The PVC is the last resort, so a PVC 404 is the terminal "nothing recorded" case (`if status == 404: return None`). Non-404s propagate in both. Left as-is.

benmccown:
Response from claude session: Addressed both CodeRabbit nitpicks in c348d0a:

- Added regression tests asserting `_existing_model_source` re-raises a non-404 `ApiException` from both the Job read and the PVC read (previously only the 404 fallback path was covered).
- Added concrete `int | None` type hints to the `run_as_user` / `run_as_group` params on the `_vllm_config` / `_generic_config` test helpers, per the repo typing guideline.

services/core/models/src/nmp/core/models/controllers/backends/common.py
benmccown: yeahhhh, this "view" thing is sort of a shim so that the "old" model deployment config, with everything flat on the top level object, would still work and not require a wall of code changes. But maybe worth a refactor PR dedicated to just the removal of this view object? Wdyt?

PR #512, opened 2026-06-29
feat(experiments): Experiment group 'ranked fields' (default sort + basis for rank)

shanaiabuggy:
**Problem.** No way to set a group's ranking, so users can't see the "winner" of a group at a glance.

**Approach.** Add a ranking: [{ field, direction }]| null config (ordered list, optional) on ExperimentGroup, settable at create/update. field is restricted to sortable (numeric) metrics — v1: cost_usd and latency_ms; evaluator-score fields follow once [ASE-320](https://linear.app/nvidia/issue/ASE-320/sort-experiments-list-by-a-rollup-metric) lands. The list is ordered by the fields in priority order (first = primary, rest = tiebreakers). This one config does double duty:

The experiments-list endpoint applies it as the default sort when no explicit sort is passed (explicit sort
still overrides). Pinned experiments float to top, then ranked-fields order. If no ranked field has resolved data, ordering falls back to -created_at.

It's the basis for the computed rank (see the rank ticket).

**Acceptance.** Setting a group's ranked fields changes the default order of its experiments list; an explicit sort
query overrides it; pinned-first is preserved; only numeric metrics / evaluator names are accepted; an expected evaluator with no data yet leaves those experiments unranked on that field.

## Summary by CodeRabbit

* **New Features**
  * Experiment groups can now configure an ordered `default_sort` for experiments.
  * When experiments are listed without an explicit sort, results use the experiment group’s default sort (pinned experiments first, then metric criteria, with `created_at` fallback).
  * Metric-based sorting documentation now includes the `p99` aggregate-stat variant.

* **Bug Fixes**
  * Improved behavior around missing rollup data: only requests that truly depend on rollups now fail, while default metric-based sorting degrades gracefully to `created_at`.

* **Tests**
  * Added coverage for default-sort ordering, validation, pinned-first behavior, and fallback/rollup failure semantics.

* **Documentation**
  * Updated OpenAPI contracts for the new `default_sort` and `SortCriterion` schema, and clarified `sort` parameter behavior.

### DIFF
--- a/services/intake/src/nmp/intake/entities/experiments.py
+++ b/services/intake/src/nmp/intake/entities/experiments.py
@@ -11,10 +11,19 @@ views. Rollups are derived from ClickHouse at read time.
 from __future__ import annotations
 
 from datetime import datetime
-from typing import Any, ClassVar
+from typing import Any, ClassVar, Literal
 
 from nmp.common.entities.client import EntityBase
-from pydantic import AnyUrl, Field
+from pydantic import AnyUrl, BaseModel, Field
+
+
+class RankingField(BaseModel):
+    """One field in a group's ranking: a sortable rollup-metric path and its direction."""
+
+    field: str = Field(
+        description="Rollup-metric sort path, e.g. cost_usd.mean, latency_ms.p95, or evaluators.<name>.mean."
+    )
+    direction: Literal["asc", "desc"] = Field(description="Sort direction for this field.")
 
 
 class ExperimentGroup(EntityBase):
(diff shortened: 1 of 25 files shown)


services/intake/src/nmp/intake/entities/experiments.py
```diff
@@ -11,10 +11,19 @@
...
 from datetime import datetime
-from typing import Any, ClassVar
+from typing import Any, ClassVar, Literal
 
 from nmp.common.entities.client import EntityBase
-from pydantic import AnyUrl, Field
+from pydantic import AnyUrl, BaseModel, Field
+
+
+class RankingField(BaseModel):
```
shanaiabuggy: rename to SortKey, default_sort_order as the entity field or smth

BrianNewsom (approved):
Discussed live, approved after updates!

PR #513, opened 2026-06-29
fix(kube): Update kube scripts to pull from ghcr

matthewgrossman:
## Summary

Add GHCR image pull support to the K8s e2e scripts, and consolidate duplicated code across setup scripts into a shared `lib.sh`.

## Changes

**New: `e2e/k8s/scripts/lib.sh`**
- Shared log helpers (`log_info`, `log_warn`, `log_error`) — previously duplicated in every setup script
- `create_platform_secrets()` — creates NGC, GHCR, and HuggingFace secrets when their corresponding env vars are set

**`e2e/k8s/scripts/install_helm_e2e.sh`**
- Sources `lib.sh` instead of defining its own log functions
- Automatically adds `imagePullSecrets` to the helm chart when `NGC_API_KEY` (for nvcr.io) and/or `GITHUB_TOKEN` (for ghcr.io) are set

**`e2e/k8s/scripts/setup_local_minikube_gpu.sh`**
- Sources `lib.sh`, removes duplicated log functions and secret creation
- Calls `create_platform_secrets` which now handles GHCR pull secret creation

**`e2e/k8s/scripts/setup_local_minikube_cpu.sh`**
- Same consolidation as GPU script

**`e2e/k8s/scripts/setup_local_kind_cpu.sh`**
- Same consolidation as GPU script

## Motivation

When deploying NeMo Platform on dev-blue (or any environment using GHCR pre-built images), the scripts needed to:
1. Create a `ghcr-pull` K8s secret for kubelet to authenticate to ghcr.io
2. Pass `imagePullSecrets` to the helm chart so pods reference the secret
3. Also pass `nvcrimagepullsecret` for NIM container images from nvcr.io

This logic was missing from the OSS scripts (it existed in Platform-Deploy for GitLab but was stripped during the port). The secret creation was also duplicated across 3 setup scripts, requiring identical changes in all 3 files.

## Test plan

- [x] Deployed NeMo Platform on dev-blue with GHCR images — all 19 services healthy
- [x] Deployed a NIM model (Llama 3.2 1B) — GPU inference working on A100
- [x] Verified `create_platform_secrets` creates correct secrets when env vars are set
- [x] Verified secrets are skipped with a warning when env vars are unset
- [x] Verified CI kind flow still works (passes `NGC_API_KEY` placeholder)


e2e/k8s/scripts/lib.sh
```diff
@@ -0,0 +1,62 @@
...
+#   - ghcr-pull: when GITHUB_TOKEN is set
+#   - huggingface-token: when HF_TOKEN is set
+create_platform_secrets() {
+    local namespace="${1:?namespace is required}"
+    local kubectl_ns=(kubectl -n "${namespace}")
+
+    if [ -n "${NGC_API_KEY:-}" ]; then
+        log_info "Creating NGC API secret..."
+        "${kubectl_ns[@]}" create secret generic ngc-api \
+          --from-literal=NGC_API_KEY="${NGC_API_KEY}" \
```
svvarom: I wonder if we can load these from a file instead, so they don’t show up in the process bits.

e2e/k8s/scripts/setup_local_minikube_cpu.sh
```diff
@@ -90,23 +76,8 @@ fi
 
 KUBECTL_NS=(kubectl -n "${KUBE_NAMESPACE}")
 
-log_info "Creating placeholder platform secrets in namespace ${KUBE_NAMESPACE}..."
-${KUBECTL_NS[@]} create secret generic ngc-api \
-  --from-literal=NGC_API_KEY="${NGC_API_KEY:-local-dev-placeholder}" \
```
svvarom, line 95: Does the chart handle `NGC_API_KEY` not being set gracefully? Or will we hit `CreateContainerConfigError`?

matthewgrossman: we will hit `CreateContainerConfigError`; fixing to use a placeholder like the script did before

e2e/k8s/scripts/lib.sh
matthewgrossman: done, no more secrets in argv

svvarom (approved):
Thanks, LGTM

PR #515, opened 2026-06-30
feat(evaluator): persist eval results as queryable entities

SandyChapman:
## Summary

Re-adds queryable persistence of eval-job results (removed with the legacy service in #231), now on the plugin-job API. Results become first-class, filterable entities again — without resurrecting the old service.

> **Stacked on #496.** Base is `aalgo-297-agent-eval-job/schapman`; the diff is just the persistence work. GitHub will retarget to `main` once #496 merges.

## What's included

- **Result entities** — `AgentEvalResultEntity` / `EvaluateResultEntity` store the aggregated scores plus the filterable traits (target, dataset); the full per-trial/per-row bundle stays in the run's fileset, referenced by `bundle_ref`. Jobs persist **best-effort** in `run()` — a transient entity-store error never fails an eval that already succeeded.
- **API DTOs** — `AgentEvalResult` / `EvaluateResult` (mapped from the entities, like `Metric` ↔ `MetricBundleEntity`) back the read routes, so `id`/`created_at`/`updated_at` round-trip cleanly on the wire and in the SDK (an `EntityBase` does not).
- **Read routes** — `/agent-eval-results` and `/eval-results` list/get/delete, with trait filtering. A small `DataFilter` base translates custom fields to `data.*` for the entity store; `MetricFilter` adopts it too, which **also fixes `metric_type` filtering on the existing `/metrics` route** (same latent bug).
- **SDK** — `client.evaluator.{agent_eval_results,eval_results}` resources (retrieve/list/delete, typed DTOs, trait filters), plus `metric_type` filtering on `client.evaluator.metrics.list`.
- **`get_async_task_sdk`** — async counterpart of `get_task_sdk` (protocol + both providers) so a synchronous job `run()` can drive the async entity-store write with the full on-behalf-of identity (id + email + groups).

## Verification

- **Live, against a real platform** (`RUN_AGENT_EVAL_INTEGRATION=1`): agent-eval submit **under auth** (proves the delegated identity authorizes the entity write), row-eval submit, and metric-type filtering. These caught three real bugs unit tests couldn't (response `exclude_none` round-trip, target wiring, custom-field filter 500).
- **Unit**: result entity round-trip, persistence helpers, service, routes, SDK resources, filter translation, async-task-SDK header parity. Full evaluator + plugin + nmp-common suites green.
- ruff + ty clean; OpenAPI spec regenerated.

## Notes

- Datetime-range filtering is intentionally not exposed in the SDK yet (needs an operator shape; equality-only for now).
- Intake remains the optional denormalized downstream copy; these entities are the evaluator's source of truth.

## Summary by CodeRabbit

* **New Features**
  * Added async task SDK initialization for job execution (so jobs can use async clients).
  * Added evaluator v2 persisted evaluation results APIs (agent and row) with matching SDK support for list/retrieve/delete.
  * Metrics listing now supports sorting and filtering by metric type.
* **Bug Fixes**
  * Improved metric filtering so custom fields are translated into the correct query namespace.
  * Result persistence is more robust: jobs continue even if persistence fails, with safe target URL handling.
* **Tests**
  * Expanded unit and integration coverage for async SDK wiring, results persistence, v2 routes, and metric/result filtering.

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/api/v2/results.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/api/v2/results.py
@@ -122,6 +122,9 @@ async def list_agent_eval_results(
 @agent_eval_results_router.get(
     "/agent-eval-results/{name}",
     summary="Get Agent Eval Result",
+    status_code=status.HTTP_200_OK,
+    response_model=AgentEvalResult,
+    response_model_exclude_none=True,
     responses={status.HTTP_404_NOT_FOUND: {"description": "Result not found"}},
 )
 @scope.read
@@ -132,10 +135,16 @@ async def get_agent_eval_result(
     service: ResultService = Depends(get_result_service),
 ) -> AgentEvalResult:
     """Get an agent-evaluation result record by workspace and name."""
-    result = await service.get_agent_eval_result(workspace, name)
-    if result is None:
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Result not found: {workspace}/{name}")
-    return result
+    try:
+        result = await service.get_agent_eval_result(workspace, name)
+        if result is None:
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Result not found: {workspace}/{name}")
+        return result
+    except HTTPException:
+        raise
+    except Exception:
+        logger.exception(f"Failed to get agent-eval result {_sanitize_for_log(workspace)}/{_sanitize_for_log(name)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Internal server error")
 
 
 @agent_eval_results_router.delete(
@@ -152,9 +161,15 @@ async def delete_agent_eval_result(
     service: ResultService = Depends(get_result_service),
 ) -> None:
     """Delete an agent-evaluation result record by workspace and name."""
-    if not await service.delete_agent_eval_result(workspace, name):
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Result not found: {workspace}/{name}")
-    return None
+    try:
+        if not await service.delete_agent_eval_result(workspace, name):
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Result not found: {workspace}/{name}")
+        return None
+    except HTTPException:
+        raise
+    except Exception:
+        logger.exception(f"Failed to delete agent-eval result {_sanitize_for_log(workspace)}/{_sanitize_for_log(name)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Internal server error")
 
 
 # --- (row) eval results ------------------------------------------------------
@@ -195,6 +210,9 @@ async def list_eval_results(
 @evaluate_results_router.get(
     "/eval-results/{name}",
     summary="Get Eval Result",
+    status_code=status.HTTP_200_OK,
+    response_model=EvaluateResult,
+    response_model_exclude_none=True,
     responses={status.HTTP_404_NOT_FOUND: {"description": "Result not found"}},
 )
 @scope.read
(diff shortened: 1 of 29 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/entities.py
```diff
@@ -70,3 +71,82 @@ class MetricBundleEntity(EntityBase):
...
+    """Persisted, queryable record of an ``AgentEvalJob`` run.
+
+    Carries only the shared record — its tasks are inline, so (unlike row-eval) it has no input ref
+    to record yet. Trials, per-metric coverage, and run counts live in the bundle's summary.
+    """
+
+    __entity_type__: ClassVar[str] = "agent_eval_result"
+
+
+class EvaluateResultEntity(_EvalResultCommon, EntityBase):
```
arpitsardhana, line 137: Should it be called ModelEvalResultEntity?
Or is it a common entity can be used for both kinds

plugins/nemo-evaluator/src/nemo_evaluator/sdk/result_resources.py
```diff
@@ -0,0 +1,273 @@
...
+
+    async def delete(self, name: str, *, workspace: str | None = None) -> None:
+        """Delete a result record by name."""
+        response = await self._http_client.delete(
+            self._item_url(name, workspace), headers=self._headers(), timeout=self._platform.timeout
+        )
+        response.raise_for_status()
+
+
+class EvaluatorAgentEvalResultsResource(_SyncResultsResource[AgentEvalResult]):
```
arpitsardhana, line 156: nit: Weird naming? EvaluatorAgentEval?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -280,4 +282,10 @@ def run(
         artifact = ctx.results.save(DEFAULT_RESULT_NAME, files.bundle_dir)
         ctx.results.save(SUMMARY_RESULT_NAME, files.summary)
 
+        # Persist the queryable result record (aggregates + coverage); the full bundle (trials) lives
+        # in the fileset referenced by `artifact`.
+        persist_agent_eval_result(
```
arpitsardhana: Does it make sense to have this function inn try: except block to ensure that failure in persistence not lead to job failure?

arpitsardhana (approved):
LGTM with minor nits

plugins/nemo-evaluator/src/nemo_evaluator/api/v2/results.py
```diff
@@ -0,0 +1,208 @@
...
+        return await service.list_agent_eval_results(
+            workspace=workspace, page=page, page_size=page_size, sort=sort, filter_operation=parsed_filter.operation
+        )
+    except Exception:
+        logger.exception(f"Failed to list agent-eval results for workspace {_sanitize_for_log(workspace)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Internal server error")
+
+
+@agent_eval_results_router.get(
+    "/agent-eval-results/{name}",
```
arpitsardhana, line 123: should this also have " response_model_exclude_none=True," so that get/list has same behaviour

arpitsardhana: Same for wrapping get under try: except

plugins/nemo-evaluator/src/nemo_evaluator/entities.py
SandyChapman: Good question. These are two intentionally-distinct typed entities: `AgentEvalResultEntity` ↔ `AgentEvalJob` and `EvaluateResultEntity` ↔ `EvaluateJob` — kept separate so the generated SDK sees the concrete result shape, and because they carry different traits (agent-eval has coverage/aggregates; evaluate has `dataset_ref`). Not a shared base.

On the name: `Model…` wouldn't fit since the target can be a model, agent, or runner. I kept `AgentEval*` for consistency with the SDK's `AgentEvalResult` and the `AgentEvalJob`/`/agent-evaluate` surface. You raise a fair broader point though — since targets are pluggable, "task eval" may be more accurate plugin-wide. Filed AALGO-315 to decide that holistically rather than renaming one entity piecemeal.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/result_resources.py
SandyChapman: Agreed it's redundant — the `Evaluator` prefix adds nothing since these mount under `client.evaluator.*` and users never see the class name. I kept it here only for consistency with the already-merged `EvaluatorMetricsResource`; renaming just the results resources would leave a mixed convention in one SDK. Filed AALGO-316 to drop the prefix across all resource families (metrics + results + tasks, sync + async) in a single sweep.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Good catch — done in a4e9e8dca9. Wrapped `persist_agent_eval_result` in try/except with a logged warning; the authoritative output (bundle + summary artifacts) is already saved before this call, so a persistence failure no longer fails an otherwise-successful eval. Applied the same to `EvaluateJob`, and added regression tests for both jobs.

plugins/nemo-evaluator/src/nemo_evaluator/api/v2/results.py
SandyChapman: Done in a4e9e8dca9 — added `response_model_exclude_none=True` to both single-GET routes so they match the list routes' serialization.

PR #523, opened 2026-06-30
fix(deployments): hash k8s-safe names from workspace/name identity (AIRCORE-860)

tylersbray:
## Summary

Fixes resource name collisions when hyphen-joining workspace and deployment/volume names produces the same display prefix (e.g. `foo` + `bar-baz` and `foo-bar` + `baz` both → `dep-foo-bar-baz`).

Both the **deployments plugin** and the **models service** now always append an 8-character SHA-256 hash suffix to DNS-safe resource names. Names remain ≤63 characters (RFC 1035) with regression tests for ambiguous pairs.

### Hash source: `workspace/name`, not the joined prefix

**Problem:** Hashing the hyphen-joined string (`dep-foo-bar-baz`) does **not** fix join ambiguity — different `(workspace, name)` pairs can produce identical joined strings.

**Decision:** Hash the unambiguous identity `"{workspace}/{name}"` (same format as `deployment_key()` / `get_deployment_key()`), while keeping the readable hyphenated prefix for operators (`dep-{ws}-{name}`, `md-{ws}-{name}`, etc.).

| Call site | Display prefix (`base_name`) | Hash input (`hash_input`) |
|-----------|------------------------------|---------------------------|
| `container_name(ws, name)` | `dep-{ws}-{name}` | `{ws}/{name}` |
| `docker_volume_name(ws, vol)` | `dep-vol-{ws}-{vol}` | `{ws}/{vol}` |
| `get_deployment_resource_name` | `md-{ws}-{name}` | `{ws}/{name}` |
| `get_nimcache_resource_name` | `md-{ws}-{name}` | `{ws}/{name}` |
| `get_deployment_secret_name` | `{prefix}-{ws}-{name}` | `{ws}/{name}` |
| Docker/K8s plugin puller | `md-plugin-{ws}-{name}` | `{ws}/{name}` |
| Generic `_get_k8s_safe_name` (init containers, etc.) | caller-provided | defaults to `base_name` |

### Always append `-{hash8}`

**Problem:** The old logic only appended a hash when truncation was needed, so short names with different identities but identical normalized prefixes could still collide.

**Decision:** Every primary resource name is `{normalized}-{hash8}{optional_suffix}`. Deterministic: same `(workspace, name)` → same name across restarts.

### Suffix chaining: `include_hash=False`

Init containers (`-prepare`, `-pull`, `-finalize`) and K8s LoRA sidecars (`-lora-sidecar`) derive names from an **already-hashed** base. A second `_get_k8s_safe_name` call with `include_hash=False` appends the suffix without adding another hash (avoids `{base}-{h1}-{h2}-lora-sidecar`).

### Length headroom for derived names

| Constraint | Limit | Reason |
|------------|-------|--------|
| NIMCache resources | 59 chars | k8s-nim-operator appends `-job` (4 chars) |
| Docker primary container | 55 chars | runtime appends `-sidecar` (8 chars) for LoRA |
| Secrets | 253 chars | RFC 1123 subdomain; suffix after hash: `{base}-{hash8}-hf-token` |

### Identity vs naming

Orphan cleanup and idempotency use **identity labels** (`deployment-workspace`, `deployment-name`, etc.), not container names alone. Container/volume names are for Docker/K8s API addressing; labels remain authoritative.

### Migration / breaking change

- **Existing running containers** keep old names after upgrade.
- **New deployments** get hashed names — acceptable for greenfield deployments plugin per ticket scope.
- **Models service** new Docker/K8s resources also rename; no dual-lookup migration in this MR. Operators should expect one-time orphan cleanup or manual teardown of pre-change resources.

### Out of scope (follow-up)

Extracting shared `k8s_safe_name` logic to `nmp_common` to deduplicate deployments `labels.py` and models `utils.py`.

## Test plan

- [x] `uv run pytest plugins/nemo-deployments/tests/unit/backends/docker/test_labels.py`
- [x] `uv run pytest services/core/models/tests/unit/common/test_utils.py -k "k8s_safe or deployment_resource or nimcache_resource or deployment_secret"`
- [x] `uv run pytest services/core/models/tests/unit/controllers/test_docker_backend.py`
- [x] `uv run pytest services/core/models/tests/unit/controllers/test_nimservice_compiler.py services/core/models/tests/unit/controllers/test_backend_config_fields.py`
- [x] Verified: `container_name("foo","bar-baz") != container_name("foo-bar","baz")`
- [x] Verified: same for `get_deployment_resource_name`

## Summary by CodeRabbit

* **New Features**
  * Docker and Kubernetes resource names now use shared DNS-safe deterministic hashing based on workspace/name identity.
  * Plugin puller containers (and optional LoRA sidecar) follow the same naming rules, including consistent truncation behavior.

* **Bug Fixes**
  * Improved disambiguation for ambiguous workspace/name combinations.
  * Orphan cleanup now relies on identity labels rather than name matching.
  * Existing containers keep their prior names after upgrade.

* **Documentation**
  * Added guidance on Docker resource naming and orphan cleanup behavior.

* **Tests**
  * Updated and expanded unit/integration coverage for the new naming rules.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
@@ -9,17 +9,15 @@ Resource names are derived from a human-readable prefix plus a deterministic
 pairs like ``("foo", "bar-baz")`` and ``("foo-bar", "baz")`` produce distinct
 names even when their joined prefixes collide.
 
-Orphan cleanup and idempotency rely on identity labels, not container names
-alone. Existing containers keep their pre-change names after upgrade; only new
-deployments receive hashed names.
+Naming logic is shared via ``nemo_platform_plugin.k8s_naming`` (plugins cannot
+import ``nmp_common``). Orphan cleanup and idempotency rely on identity labels,
+not container names alone.
 """
 
 from __future__ import annotations
 
-import hashlib
-import re
-
 from nemo_deployments_plugin.constants import MANAGED_BY_LABEL
+from nemo_platform_plugin.k8s_naming import k8s_safe_name, workspace_name_identity
 
 MANAGED_BY_KEY = "managed-by"
 DEPLOYMENT_WORKSPACE_LABEL = "nemo.nvidia.com/deployment-workspace"
@@ -29,43 +27,10 @@ CONFIG_NAME_LABEL = "nemo.nvidia.com/deployment-config"
 VOLUME_WORKSPACE_LABEL = "nemo.nvidia.com/volume-workspace"
 VOLUME_NAME_LABEL = "nemo.nvidia.com/volume-name"
 
-_HASH_SUFFIX_LENGTH = 8
-
-
-def k8s_safe_name(
-    base_name: str,
-    *,
-    max_length: int = 63,
-    suffix: str = "",
-    hash_input: str | None = None,
-) -> str:
-    """Generate a DNS-label-safe name (RFC 1035) with a mandatory hash suffix.
-
-    Normalizes ``base_name`` for display, truncates when needed, then returns
-    ``{normalized}-{hash8}{suffix}``. The hash is SHA-256 of ``hash_input`` when
-    provided, otherwise ``base_name``. Callers that represent a workspace/name
-    pair should pass ``hash_input=deployment_key(workspace, name)`` so the hash
-    reflects the unambiguous identity rather than the join-ambiguous prefix.
-    """
-    hash_source = hash_input if hash_input is not None else base_name
-    hash_suffix = hashlib.sha256(hash_source.encode()).hexdigest()[:_HASH_SUFFIX_LENGTH]
-    normalized = re.sub(r"[^a-z0-9-]", "-", base_name.lower())
-    normalized = re.sub(r"[-]+", "-", normalized)
-    if normalized and not normalized[0].isalpha():
-        normalized = f"x{normalized}"
-    normalized = normalized.rstrip("-")
-
-    reserved = len(suffix) + len(hash_suffix) + 1
-    if len(normalized) + reserved > max_length:
-        normalized = normalized[: max_length - reserved].rstrip("-")
-    if not normalized:
-        normalized = "x"
-    return f"{normalized}-{hash_suffix}{suffix}"
-
 
 def deployment_key(workspace: str, name: str) -> str:
     """Return the canonical identity string used for hashing and label keys."""
-    return f"{workspace}/{name}"
+    return workspace_name_identity(workspace, name)
 
 
 def container_name(workspace: str, deployment_name: str) -> str:
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
@@ -14,8 +14,6 @@ import ``nmp_common``). Orphan cleanup and idempotency rely on identity labels,
 not container names alone.
 """
 
-from __future__ import annotations
-
 from nemo_deployments_plugin.constants import MANAGED_BY_LABEL
 from nemo_platform_plugin.k8s_naming import k8s_safe_name, workspace_name_identity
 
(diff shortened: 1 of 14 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
```diff
@@ -18,10 +29,26 @@
...
+_HASH_SUFFIX_LENGTH = 8
+
 
-def k8s_safe_name(base_name: str, *, max_length: int = 63, suffix: str = "") -> str:
-    """Generate a DNS-label-safe name (RFC 1035) from arbitrary input."""
-    hash_suffix = hashlib.sha256(base_name.encode()).hexdigest()[:8]
+def k8s_safe_name(
+    base_name: str,
+    *,
+    max_length: int = 63,
```
mckornfield: would this ever be over 63? should we just hard code/constant it?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
```diff
@@ -30,24 +57,31 @@ def k8s_safe_name(base_name: str, *, max_length: int = 63, suffix: str = "") ->
...
     reserved = len(suffix) + len(hash_suffix) + 1
     if len(normalized) + reserved > max_length:
-        trim = max_length - reserved
-        normalized = normalized[:trim].rstrip("-")
-        normalized = f"{normalized}-{hash_suffix}{suffix}"
-    elif suffix:
-        normalized = f"{normalized}{suffix}"
-    return normalized
+        normalized = normalized[: max_length - reserved].rstrip("-")
+    if not normalized:
```
mckornfield: when would this happen?

services/core/models/src/nmp/core/models/app/utils.py
```diff
@@ -296,115 +310,140 @@ def _get_k8s_safe_name(
...
+        normalized = truncated
+
+    result = f"{normalized}-{hash_suffix}{suffix}"
+
+    if len(result) > max_length:
+        excess = len(result) - max_length
+        normalized = normalized[: len(normalized) - excess].rstrip("-.")
+        if not normalized:
+            normalized = "x"
+        result = f"{normalized}-{hash_suffix}{suffix}"
```
mckornfield: I guess this can't be easily shared with the plugin logic? or should this be moved to a sorta "core" place?

services/core/models/src/nmp/core/models/app/utils.py
```diff
@@ -296,115 +310,140 @@ def _get_k8s_safe_name(
...
-            # If nothing left, use first char if valid, else 'x'
-            truncated = "x"
 
-    result = f"{truncated}-{hash_suffix}{suffix}"
+def get_docker_volume_name(workspace: str, name: str) -> str:
+    """Docker model-cache volume name for a deployment."""
+    label_name = f"nim-cache-{workspace}-{name}"
+    return _get_k8s_safe_name(
+        label_name,
+        max_length=63,
```
mckornfield: nit- this is the default yeah?

mckornfield (approved):
some nits, wonder if there's a way to have the plugin share code with the core thing, maybe with some k8s_naming_util that lives in a place they can both see?

services/core/models/src/nmp/core/models/app/utils.py
tylersbray: Extracted the canonical implementation to nemo_platform_plugin.k8s_naming — the repo's intended shared location for plugin-accessible code (plugins can't import nmp_common). Both call sites now use it:

Deployments plugin — labels.py imports k8s_safe_name and workspace_name_identity
Models service — utils.py re-exports as _get_k8s_safe_name / _workspace_name_identity for existing call sites
Added unit tests in packages/nemo_platform_plugin/tests/test_k8s_naming.py.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
tylersbray: For deployments, yes — it was always 63 (Docker DNS labels). Removed the parameter from the plugin's local copy; the shared module uses DNS_LABEL_MAX_LENGTH = 63 as the default.

services/core/models/src/nmp/core/models/app/utils.py
tylersbray: Removed the redundant explicit arg; 63 is the default.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/labels.py
tylersbray: Documented in k8s_naming.py: the "x" fallback covers inputs where all characters are invalid/stripped, or truncation leaves no usable base (when max_length only fits hash + suffix).

benmccown (approved):
Thanks for tackling this!

PR #531, opened 2026-06-30
feat(auth): govern workspace creation through RBAC

ironcommit:
## Summary

Moves workspace creation from a hardcoded policy exception (any authenticated user can create) to a proper RBAC-controlled permission, while preserving the current default behavior through seeding.

### What changed

- **New `WorkspaceCreator` role** with `workspaces.create` permission added to the static authz config
- **OPA policy generalized**: the IAM-specific system-workspace fallback rule now applies to any workspace-less endpoint with required permissions, removing the `/apis/auth/v2/iam/` path prefix check
- **Workspace creation endpoint** (`POST /apis/entities/v2/workspaces`) now requires `workspaces.create` permission and `entities:write`/`platform:write` scopes instead of empty permissions/scopes
- **Platform seed** creates a wildcard `WorkspaceCreator` binding in the `system` workspace, preserving open workspace creation by default
- **Operators can restrict** workspace creation by rebinding the `system` workspace: remove the wildcard `WorkspaceCreator` and grant it only to specific users/groups

### Why

Previously, workspace creation was a special-cased policy exception — the endpoint had empty permissions, so any authenticated user could create workspaces with no way for operators to restrict it. This change brings workspace creation under the same RBAC model as every other operation, giving operators a knob to control who can create workspaces while keeping the default behavior unchanged.

### Files changed

- `services/core/auth/src/nmp/core/auth/assets/static-authz.yaml` — new role, permission, and endpoint config
- `services/core/auth/src/nmp/core/auth/app/policies/authz.rego` — generalized system-workspace fallback rule
- `services/core/auth/src/nmp/core/auth/app/seeding.py` — new `seed_workspace_creator()` function
- `services/core/auth/scripts/auth-tools.py` — updated permission/scope inference
- `services/core/auth/src/nmp/core/auth/app/policy_tests/` — updated Rego tests
- `services/core/auth/tests/test_embedded_pdp.py` — new system-scoped permission test
- `services/core/auth/tests/test_seeding.py` — tests for WorkspaceCreator seeding
- `services/core/entities/tests/integration/test_workspaces_crud_with_auth.py` — integration tests for default-open and restricted creation
- `docs/` — updated documentation across auth concepts, roles/permissions, OIDC setup, configuration, workspaces, and setup pages

## Test plan

- [ ] `make test-unit` passes (Rego policy tests, seeding tests, embedded PDP tests)
- [ ] Integration test `test_workspaces_crud_with_auth.py` covers both default-open and restricted creation flows
- [ ] Verify existing workspace creation still works without any configuration changes
- [ ] Verify operators can restrict creation by rebinding the `system` workspace roles

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated `workspaces.create` permission via a `WorkspaceCreator` role; the creator is granted Admin access in the newly created workspace.
  * Workspace creation now requires system-scoped authorization plus `entities:write` and `platform:write` scopes.
  * Default platform seeding enables open workspace creation for authenticated users via a wildcard `WorkspaceCreator` binding in the `system` workspace; operators can restrict it by removing/rebinding that wildcard.
* **Documentation**
  * Updated setup and permission/workspace creation docs to reflect the new `WorkspaceCreator` behavior and restriction steps.
* **Bug Fixes**
  * Authorization and permission-suggestion behavior were tightened to fail closed for unconfigured endpoints and correctly model workspace-creation requirements.

### DIFF
--- a/docs/auth/concepts.mdx
+++ b/docs/auth/concepts.mdx
@@ -145,9 +145,10 @@ flowchart TB
 
 ### Workspace Creation
 
-Creating a workspace is a special operation:
+Creating a workspace is a system-scoped RBAC operation:
 
-- All authenticated users can create workspaces (no special permission needed)
+- By default, all authenticated users can create workspaces because platform seed grants `WorkspaceCreator` to the wildcard principal `*` in the `system` workspace
+- Operators can restrict workspace creation by removing that wildcard binding and granting `WorkspaceCreator` only to specific users or groups
 - The creator automatically receives the **Admin** role
 - New workspaces are private by default (only the creator has access)
 - You can immediately start creating resources
@@ -158,6 +159,7 @@ NeMo Platform automatically provisions role bindings for the wildcard principal 
 
 - **`default`**: All authenticated users have **Editor** role, allowing everyone to create and manage resources immediately
 - **`system`**: All authenticated users have **Viewer** role, providing read-only access to system-level resources
+- **`system`**: All authenticated users also receive the seeded **WorkspaceCreator** role, preserving open workspace creation until operators rebind it
 
 ## Workspace Visibility
 
--- a/services/core/auth/scripts/auth-tools.py
+++ b/services/core/auth/scripts/auth-tools.py
@@ -322,9 +322,9 @@ def infer_permissions(path: str, method: str) -> List[str]:
         elif "embeddings" in path:
             return ["inference.embeddings"]
 
-    # Special case for workspace creation (empty permissions — open to authenticated users)
+    # Special case for workspace creation (system-scoped RBAC, not workspace-scoped RBAC)
     if path.endswith("/workspaces") and method == "post":
-        return []
+        return ["workspaces.create"]
 
     # Special case for workspace members (permission lives under workspaces.* regardless
     # of which API area exposes the route, e.g. /apis/entities/v2/workspaces/{ws}/members)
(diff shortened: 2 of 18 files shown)


docs/auth/concepts.mdx
```diff
@@ -158,6 +159,7 @@ NeMo Platform automatically provisions role bindings for the wildcard principal
 
 - **`default`**: All authenticated users have **Editor** role, allowing everyone to create and manage resources immediately
 - **`system`**: All authenticated users have **Viewer** role, providing read-only access to system-level resources
+- **`system`**: All authenticated users also receive the seeded **WorkspaceCreator** role, preserving open workspace creation until operators rebind it
```
mckornfield, line 162: thinking ahead, we might want this to be off by default for "non admins"? I am remembering in the Gretel days they wanted project creation to be off for new people who joined the org

this might require a users table/admin distinction? (which I'm not sure if this exists)

services/core/auth/scripts/auth-tools.py
```diff
@@ -322,9 +322,9 @@ def infer_permissions(path: str, method: str) -> List[str]:
         elif "embeddings" in path:
             return ["inference.embeddings"]
 
-    # Special case for workspace creation (empty permissions — open to authenticated users)
+    # Special case for workspace creation (system-scoped RBAC, not workspace-scoped RBAC)
     if path.endswith("/workspaces") and method == "post":
-        return []
+        return ["workspaces.create"]
```
mckornfield, line 327: does seem weird these are here, wonder if they should be a constant of sorts

mckornfield (approved):
some minor things, max might have more, I wonder about the default create though (I remember that being a no no for certain customers/larger orgs when anyone with an email could onboard)

docs/auth/concepts.mdx
drazvan: Yes. We should add an option in the config for the Auth service, to configure the workspace creation principals; something along the lines of:

```yaml
auth:
  workspace_creators:
  - "*"             # This is the default
```

If that option is set to an empty list, then only platform admins would be able to create workspaces. If it's set to group principals, then anyone in those groups would be able.

drazvan (approved):
LTGM

docs/auth/concepts.mdx
ironcommit: Definitely something to consider.  For now I'm just keeping the existing behavior.

PR #534, opened 2026-07-01
feat(studio): Adding range type filtering to dataview filters

nakolean:
https://github.com/user-attachments/assets/f8ad1a86-9150-40a7-bafc-fdc8ab421903

Screenshot 2026-07-01 at 9 57 31 AM

## Summary by CodeRabbit

* **New Features**
  * Added support for numeric range filtering in data views, including a slider and min/max inputs.
  * Filter chips and applied filter text now display numeric ranges in a clearer format.
  * Added Storybook examples for basic, custom-bound, and in-data-view range filtering.

* **Bug Fixes**
  * Improved range handling so values are clamped, normalized, and cleared correctly when they match default bounds.
  * Fixed formatting for range filters so active filters show the expected bounds instead of generic text.

### DIFF
--- a/web/packages/common/src/components/DataView/ColumnFilterPanel.tsx
+++ b/web/packages/common/src/components/DataView/ColumnFilterPanel.tsx
@@ -8,8 +8,10 @@ import {
   SingleSelectFilterControl,
   MultiSelectFilterControl,
   DateTimeFilterControl,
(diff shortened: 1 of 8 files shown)


shanaiabuggy (approved):
yahoo!

PR #536, opened 2026-07-01
feat(deployments): K8s PVC volumes (AIRCORE-757 phase 2)

tylersbray:
## Summary

- Implement `create_volume` / `read_volume_status` / `delete_volume` for `K8sDeploymentBackend` via `backends/k8s/volumes.py`
- Map PVC phases: `Pending`→PENDING, `Bound`→BOUND, `Lost`→FAILED
- Honor `K8sVolumeConfig.storageClass` and `namespace` on create; extend ABC + volume reconciler to pass `backend_config` on read/delete so custom namespaces work
- Idempotent create on 409 conflict; delete treats 404 as RELEASED

## Test plan

- [x] `uv run pytest plugins/nemo-deployments/tests/unit -v` (186 passed)
- New: `tests/unit/backends/k8s/test_volumes.py` — emission, status mapping, 409 idempotency, namespace on read

## Linear

AIRCORE-757 — Phase 2 of 7

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes-backed volume lifecycle (create, status, delete) using PVC helpers, including namespace and storage class resolution.
  * Volume operations now accept optional backend configuration for reconciler-driven behavior.
* **Bug Fixes**
  * Improved accuracy for existing/conflict PVCs and missing/not-found PVCs, including “not managed” detection.
  * More robust status handling, including reporting PVC deletion progress and safe handling of malformed configuration.
* **Tests**
  * Added/expanded unit tests for PVC phase-to-status mapping and create/read/delete flows.
  * Updated unit test fixtures for request-timeout expectations and reconciler forwarding.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
@@ -437,7 +437,8 @@ class DockerDeploymentBackend(DeploymentBackend):
         name: str,
         backend_config: dict[str, Any] | None = None,
     ) -> VolumeStatusUpdate:
-        del backend_config
+        # backend_config is part of the DeploymentBackend ABC so the reconciler can pass
+        # K8s namespace overrides; Docker volume names are global to the daemon.
         return await volume_ops.read_volume_status(self._client, workspace=workspace, name=name)
 
     async def delete_volume(
@@ -447,7 +448,6 @@ class DockerDeploymentBackend(DeploymentBackend):
         *,
         backend_config: dict[str, Any] | None = None,
     ) -> VolumeStatusUpdate:
-        del backend_config
         return await volume_ops.delete_volume(self._client, workspace=workspace, name=name)
 
     def _container_matches_deployment(
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
@@ -430,10 +430,24 @@ class DockerDeploymentBackend(DeploymentBackend):
             driver=driver,
         )
 
-    async def read_volume_status(self, *, workspace: str, name: str) -> VolumeStatusUpdate:
+    async def read_volume_status(
+        self,
+        *,
+        workspace: str,
+        name: str,
+        backend_config: dict[str, Any] | None = None,
+    ) -> VolumeStatusUpdate:
+        del backend_config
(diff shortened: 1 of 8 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -430,10 +430,24 @@ async def create_volume(
...
 
-    async def read_volume_status(self, *, workspace: str, name: str) -> VolumeStatusUpdate:
+    async def read_volume_status(
+        self,
+        *,
+        workspace: str,
+        name: str,
+        backend_config: dict[str, Any] | None = None,
+    ) -> VolumeStatusUpdate:
+        del backend_config
```
mckornfield: why do we delete the dict? or why do we even pass it and delete it lol, just because we're partway through?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/volumes.py
```diff
@@ -0,0 +1,211 @@
...
+    """Parse and validate the k8s section of entity backend_config."""
+    k8s_section = backend_config.get("k8s")
+    if not k8s_section:
+        return None
+    return K8sVolumeConfig.model_validate(k8s_section)
+
+
+def resolve_volume_namespace(*, default_namespace: str, volume_config: K8sVolumeConfig | None) -> str:
+    """Resolve target namespace from parsed volume config with executor default fallback."""
+    if volume_config is None or not volume_config.namespace:
```
mckornfield: should this be `and`?

mckornfield (approved):

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
tylersbray: Good catch — the `del` was just silencing an unused arg. The param is intentional: the reconciler always passes `backend_config` so K8s can resolve custom namespaces on read/delete; Docker ignores it because volume names are global to the daemon.

Removed the `del` stubs and added a brief comment on `read_volume_status` explaining that contract. Pushed in 7a265453b1.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/volumes.py
tylersbray: Yep — flipped to the positive guard:

```python
if volume_config is not None and volume_config.namespace:
    return volume_config.namespace
return default_namespace
```

Same behavior, reads clearer. Pushed in 7a265453b1.

PR #539, opened 2026-07-01
feat(client): Add nemoclient error hierarchy and fix streaming

matthewgrossman:
## Summary

Foundational improvements to the `NemoClient` typed HTTP client, extracted from the Files migration PR (#429) to keep that PR focused on Files-specific changes.

### Error hierarchy (`errors.py` — new)
- Moved `NemoHTTPError` out of `response.py` into its own module
- Added status-code-specific subclasses: `BadRequestError` (400), `AuthenticationError` (401), `PermissionDeniedError` (403), `NotFoundError` (404), `ConflictError` (409), `UnprocessableEntityError` (422), `RateLimitError` (429), `InternalServerError` (500+)
- Added `raise_for_status()` function that maps status codes to the correct subclass
- All error paths (JSON, binary, stream, paginated) now raise `NemoHTTPError` subclasses consistently — previously binary/stream paths raised `httpx.HTTPStatusError`

### `send()` raises on non-2xx (`client.py`)
- `send()` now calls `raise_for_status(raw)` before parsing the body, so callers don't need `.data()` to trigger errors
- `.data()` simplified to a plain body accessor
- Removed `_apply_client_options()` and `_raise_for_status()` — `exist_ok` infrastructure remains in `PreparedRequest.client_options` but is not acted on until the server returns entities on 409 (AIRCORE-866)

### Binary/stream response API (`response.py`)
- Replaced `__enter__`/`__exit__` context manager protocol with simpler `read()` + `stream()` API
- `read()` — opens the stream, reads all bytes, closes (one-shot); reads body before `raise_for_status()` so error detail is available
- `stream()` — `@contextmanager` yielding `Iterator[bytes]` (binary) or `Iterator[ModelT]` (stream)
- No more `_response` state or `http_response` property
- SSE support: auto-detects `Content-Type: text/event-stream` and strips `data:` prefixes before JSON parsing; skips non-data SSE fields (`event:`, `id:`, comments). Shared `_parse_stream_line()` helper used by both sync and async paths.

### Other improvements
- **`default_headers`** on `BaseNemoClient` — merged into every request; `client_from_platform()` propagates `_custom_headers`
- **JSON query param serialization** — dict/list values are `json.dumps()`-ed instead of Python `repr()`
- **`exclude_unset=True`** on body serialization — partial update bodies only send explicitly set fields

## Test plan

- [x] 54 client tests pass (`packages/nemo_platform_plugin/tests/client/`)
- [x] 25 example-plugin tests pass (`plugins/example-plugin/tests/test_sdk.py`), including new coverage for:
  - Binary upload/download via `read()` and `stream()`
  - NDJSON streaming via `stream()`
  - SSE framing (data: prefix stripping, non-data field skipping, sync + async)
  - Error detail extraction from binary error responses

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/client/client.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/client/client.py
@@ -29,7 +29,7 @@ from nemo_platform_plugin.client.auth import (
     StaticToken,
     TokenProvider,
 )
-from nemo_platform_plugin.client.errors import ConflictError, raise_for_status
+from nemo_platform_plugin.client.errors import raise_for_status
 from nemo_platform_plugin.client.response import (
     AsyncNemoBinaryResponse,
     AsyncNemoPaginatedResponse,
@@ -193,18 +193,6 @@ class BaseNemoClient:
                 filtered[k] = v
         return filtered or None
 
-    def _raise_for_status(self, raw: httpx.Response, request: PreparedRequest) -> None:
-        """Raise on non-2xx, unless a client option suppresses the error.
-
-        For example, ``exist_ok=True`` swallows 409 Conflict.
-        """
-        try:
-            raise_for_status(raw)
-        except ConflictError:
-            if request.client_options and request.client_options.get("exist_ok"):
-                return
-            raise
-
 
 class NemoClient(BaseNemoClient):
     """Sync HTTP client for NeMo Platform APIs."""
@@ -345,7 +333,10 @@ class NemoClient(BaseNemoClient):
             )
 
         raw = self._request_with_retry(request, url, req_headers, params, resolved_retry)
-        self._raise_for_status(raw, request)
+        # NOTE: client_options (e.g. exist_ok) from PreparedRequest are not
+        # acted on here yet — see AIRCORE-866 for the planned server-side
+        # fix that would let the client handle them properly.
+        raise_for_status(raw)
         body = None
         if request.response_type is not None:
             body = request.response_type.model_validate(raw.json())
@@ -531,7 +522,10 @@ class AsyncNemoClient(BaseNemoClient):
             )
 
         raw = await self._request_with_retry(request, url, req_headers, params, resolved_retry)
-        self._raise_for_status(raw, request)
+        # NOTE: client_options (e.g. exist_ok) from PreparedRequest are not
+        # acted on here yet — see AIRCORE-866 for the planned server-side
+        # fix that would let the client handle them properly.
+        raise_for_status(raw)
         body = None
         if request.response_type is not None:
             body = request.response_type.model_validate(raw.json())
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/client/client.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/client/client.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+import json
 import time
 from collections.abc import Mapping
 from pathlib import Path
@@ -28,6 +29,7 @@ from nemo_platform_plugin.client.auth import (
     StaticToken,
     TokenProvider,
 )
+from nemo_platform_plugin.client.errors import ConflictError, raise_for_status
 from nemo_platform_plugin.client.response import (
     AsyncNemoBinaryResponse,
     AsyncNemoPaginatedResponse,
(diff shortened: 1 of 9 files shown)


packages/nemo_platform_plugin/src/nemo_platform_plugin/client/client.py
```diff
@@ -174,32 +180,30 @@ def _is_paginated(self, request: PreparedRequest) -> bool:
...
-                    except (ValueError, TypeError):
-                        pass
-                return NemoResponse(http_response=response.http_response, body=body, request=request)
-
-        return response
+        try:
+            raise_for_status(raw)
+        except ConflictError:
+            if request.client_options and request.client_options.get("exist_ok"):
+                return
```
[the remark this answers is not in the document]
matthewgrossman: Fixed — `_raise_for_status` has been removed entirely. `send()` now calls `raise_for_status(raw)` directly with no `exist_ok` handling. The `exist_ok` blessed param infrastructure remains in `PreparedRequest.client_options` but is not acted on until the server-side fix lands (AIRCORE-866).

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/response.py
```diff
@@ -51,48 +53,43 @@ def data(self) -> ResponseT:
...
-
     def read(self) -> bytes:
         """Read and return the entire response body as bytes."""
-        return self.http_response.read()
-
-    def __iter__(self) -> Iterator[bytes]:
-        return self.http_response.iter_bytes()
+        with self._stream_ctx as raw:
+            raise_for_status(raw)
+            return raw.read()
```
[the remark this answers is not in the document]
matthewgrossman: Fixed — binary `read()` now calls `raw.read()` / `await raw.aread()` before `raise_for_status(raw)`, so the response body is available for `NemoHTTPError._extract_detail()` to parse `{"detail": "..."}`. Added a test (`test_binary_read_error_has_detail`) that verifies the error detail is extracted correctly on 404.

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/response.py
```diff
@@ -104,29 +101,21 @@ def __init__(
...
-
-    def __exit__(
-        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
-    ) -> None:
-        self._stream_ctx.__exit__(exc_type, exc_val, exc_tb)
+            def _iter() -> Iterator[ModelT]:
+                for line in raw.iter_lines():
+                    line = line.strip()
+                    if line:
+                        yield self._model_type.model_validate_json(line)
```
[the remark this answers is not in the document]
matthewgrossman: Fixed — added SSE support. The stream response classes now sniff `Content-Type: text/event-stream` and automatically strip `data:` prefixes before JSON parsing. Non-data SSE fields (`event:`, `id:`, `retry:`, comments) are skipped. NDJSON still works as before for other content types. Added tests for SSE framing, non-data field skipping, and async SSE.

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/errors.py
```diff
@@ -0,0 +1,112 @@
...
+            return resp.reason_phrase or f"HTTP {resp.status_code}"
+
+
+# ---------------------------------------------------------------------------
+# Status-code-specific errors
+# ---------------------------------------------------------------------------
+
+
+class BadRequestError(NemoHTTPError):
+    """HTTP 400"""
```
mckornfield, line 62: Lol didn't realize a comment would work instead of pass

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/response.py
```diff
@@ -191,29 +194,28 @@ def __init__(
...
-        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
-    ) -> None:
-        await self._stream_ctx.__aexit__(exc_type, exc_val, exc_tb)
+            async def _iter() -> AsyncIterator[ModelT]:
+                async for line in raw.aiter_lines():
+                    line = line.strip()
+                    if not line:
+                        continue
+                    if is_sse:
+                        if line.startswith("data:"):
```
mckornfield: Lol as much as i think magic numbers are w/e, it is weird to see it twice bc I assume async (the strip and the data prefix)

mckornfield (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/errors.py
matthewgrossman: it does!

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/response.py
matthewgrossman: ya I feel like we should probably have a separate helper for this that both sync/async utilize

matthewgrossman: done!

PR #544, opened 2026-07-02
feat(evaluator): Harbor-backed agent-eval runtime 

arpitsardhana:
## Summary

This PR delivers two things: the dependency-light Harbor **runtime**, and a runnable **example + native end-to-end test** that exercises it against a real Harbor job.

### 1. Harbor runtime (`nemo_evaluator_sdk.agent_eval.runtimes.harbor_runtime`)

- **`HarborAgentTaskRunner`** — runs a Harbor job and adapts it into `AgentEvalTrial`s. Execution is injected via a `run_job` callback, so the SDK does **not** import `harbor` or `nemo_oo_agents` (Harbor's JobConfig/wrapper/DinD/import-lock/`docker network prune` stay with the caller). Mirrors the existing dependency-light `CallableAgentTaskRunner`.
- **`build_trials_from_job_dir`** — pure adapter over Harbor's per-trial `result.json` (reward, exception type, token/cost → `TrialMeasurements` keys, standard evidence descriptors).
- **`HarborRewardMetric`** — scores the verifier reward stamped on trial metadata (same pattern as the example `VerifierRewardMetric`).
- **`reward_payload_from_result`** — Phase-1 adapter reproducing the optimizer's legacy `{reward, reward_details, exceptions}` payload.

No changes to existing SDK modules; this is purely additive.

### 2. Example + e2e test (`examples/harbor/`)

- **Hello-world Harbor local dataset** (`hello_world_dataset/`) — a task-folder directory (minimal Alpine Dockerfile, pure-bash verifier, deterministic oracle solution) that mirrors how real user Harbor task collections are discovered. `discover_tasks()` scans it into one `AgentEvalTask` per folder, so adding a task folder needs no code changes.
- **`run_harbor_example.py`** — one `build_hello_world_job_runner` helper driving two modes: `--mode native` (score through `AgentEvaluator`) and `--mode optimizer` (rebuild NeMo Optimizer's legacy reward payload via `reward_payload_from_result`). Scores with Harbor's oracle agent, so it runs offline with no model/API key.
- **`test_harbor_runtime_e2e.py`** — runs the hello-world job natively through the SDK and asserts `reward == 1.0`. Marked `e2e`/`slow`; auto-skips when `harbor` or Docker is unavailable.
- **`README.md`** — documents the dataset-directory pattern, both run modes, and the macOS/colima bind-mount caveat for `/logs` reward collection.

## Why experimental

This is the SDK half of a cross-repo integration (optimizer half lives in NeMo-Optimizer). Full runtime use also needs packaging follow-ups (lighten `nemo_evaluator_sdk/__init__.py`, declare the agent-eval dependency closure, vendor the new module). Tracked in the integration plan.

## Test plan

- [x] `uv run --frozen pytest packages/nemo_evaluator_sdk/tests/agent_eval/test_harbor_runtime.py` — no-Docker adapter + scoring coverage via `AgentEvaluator` + legacy adapter.
- [x] `uv run --frozen pytest packages/nemo_evaluator_sdk/tests/agent_eval/` — full suite green.
- [x] Native e2e (`test_harbor_runtime_e2e.py`) — real single-task Harbor Docker job scored end-to-end as `reward == 1.0`.
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run --frozen ty check` on the new files.
- [ ] Follow-up: packaging changes so the runtime is importable without the current example-driven wiring.

## Summary by CodeRabbit

* **New Features**
  * Added a Harbor example with dataset/task layout guidance, task verification, and native/optimizer-style run instructions.
  * Introduced Harbor-backed evaluation support, including a reward metric and richer per-trial metadata.
* **Bug Fixes**
  * Improved handling of rewards, exceptions, missing verifier outputs, and token/cost reporting.
  * Enhanced legacy reward payload compatibility, with clearer behavior when the expected reward key is absent.
* **Tests**
  * Added unit tests for Harbor trial adaptation and legacy payloads, plus an end-to-end test running the Harbor hello-world example.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+    ``<job_dir>/result.json`` is skipped because it is not nested). Each Harbor
+    trial whose ``task_name`` matches a supplied task id becomes one trial, with
+    the verifier reward, exception type, and token/cost measurements stamped on
+    ``metadata`` and standard evidence descriptors pointing at the trial's
+    on-disk artifacts.
+    """
+    job_path = Path(job_dir)
+    known_task_ids = {task.id for task in tasks}
+    trials: list[AgentEvalTrial] = []
+    for result_path in sorted(job_path.glob("*/result.json")):
```
SandyChapman, line 124: What happens if a task didn't produce a result? We might want to assert that each `known_task_ids` has a result. Additionally, if none of the inputted task_ids match results, then we end up with no trials and nothing scored.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def output_spec(self) -> list[MetricOutputSpec]:
+        return [MetricOutputSpec.continuous_score(self._output_name)]
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        metadata = input.candidate.metadata
+        reward = metadata.get("reward")
+        if reward is None:
+            reward = 1.0 if metadata.get("passed") else 0.0
+        return MetricResult(outputs=[MetricOutput(name=self._output_name, value=float(reward))])
```
SandyChapman: This is a great way to get the easy to find signal. Any thoughts on if it's worth seeing if there are additional metrics we could pull from the reward.json file or outputted artifacts? I suppose this would require users to supplement the harbor evals with evaluator metrics (either internally in the test.sh script) or externally, post-run.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+            out[str(key)] = float(value)
+        except (TypeError, ValueError):
+            continue
+    return out
+
+
+def _primary_reward(rewards: Mapping[str, float], reward_key: str) -> float | None:
+    if reward_key in rewards:
+        return rewards[reward_key]
+    return next(iter(rewards.values()), None)
```
SandyChapman: Can you expand on how this works? If a verifier emits multiple rewards, we just pick the first? is that valid?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+        return self._metric_type
+
+    def output_spec(self) -> list[MetricOutputSpec]:
+        return [MetricOutputSpec.continuous_score(self._output_name)]
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        metadata = input.candidate.metadata
+        reward = metadata.get("reward")
+        if reward is None:
+            reward = 1.0 if metadata.get("passed") else 0.0
```
SandyChapman: It looks like metadata is built like:
```
    metadata: dict[str, Any] = {
        "reward": reward,
        "reward_details": dict(rewards),
        "harbor_trial_dir": str(trial_dir),
    }
```
It seems like `passed` is never present?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+    read it off an :class:`AgentEvalResult`:
+
+    * ``reward`` — mean of each metric output, keyed ``"<metric_type>.<output>"``.
+    * ``reward_details`` — ``{output: {value_str: [task_id, ...]}}`` grouped from
+      per-trial scores (Harbor's ``reward_stats`` analogue).
+    * ``exceptions`` — ``{exception_type: [task_id, ...]}`` from trial metadata
+      (Harbor's ``exception_stats`` analogue).
+    """
+    reward = {
+        score.name: score.mean for score in result.summary.scores.scores if getattr(score, "mean", None) is not None
```
SandyChapman: The `getattr(score, "mean", None)` is a bit of a smell. Do we not have the right typing here?

packages/nemo_evaluator_sdk/tests/agent_eval/test_harbor_runtime.py
```diff
@@ -0,0 +1,74 @@
...
+    assert trials["pass-task"].metadata["reward"] == 1.0
+    assert trials["pass-task"].metadata["prompt_tokens"] == 100
+    assert trials["pass-task"].evidence is not None
+    assert trials["fail-task"].status == AgentEvalTrialStatus.PARTIAL
+    assert trials["fail-task"].metadata["exception_type"] == "NonZeroAgentExitCodeError"
+
+    # run_job is awaited exactly once, then the job dir is adapted and scored end-to-end.
+    calls = []
+    runner = HarborAgentTaskRunner(job_dir=job_dir, run_job=lambda: _record(calls))
+    result = await AgentEvaluator().run(tasks=tasks, target=runner, config=AgentEvalRunConfig(write_dashboard=False))
```
SandyChapman, line 69: I don't understand how job_dir relates to the `tasks` that are passed in here. It looks like `job_dir` is the _results_ of running Harbor? How do we enforce that run_job actually runs harbor? While this nicely sidesteps needing the harbor dependency in the lib, it doesn't present as nice of a work flow for dev UX.

SandyChapman: Oh and one more thing, how will this map to service side evals? The lambda makes it hard to think of a method for shipping execution over the wire.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: Added warning for trial with no matching signal.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: reward_key selects the names rewards.
the first value was sort of fallback if reward_key is not present

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: This is fair, passed was sort of fallback. 
this code path is removed

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: dropped the getattr

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: the full verifier reward mapping is present at metadata["reward_details"]

The followup MR #620 natively integrates the harbor and makes it easy to add extra metrics

packages/nemo_evaluator_sdk/tests/agent_eval/test_harbor_runtime.py
arpitsardhana: This is fair point, #620 addresses it by defining harbor runtime native to eval and user just need to supply harbor files artifacts

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
SandyChapman: Any thoughts on if we should either be loud about the fallback with a warn log or not do this?

SandyChapman (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
arpitsardhana: Done by failing loudly with warning

PR #553, opened 2026-07-02
feat(deployments): k8s PodSpec compiler for full workload config (AIRCORE-757 phase 5)

tylersbray:
## Summary

- Add `compiler.py` to compile `DeploymentConfig` + `K8sDeploymentConfig` into Kubernetes PodSpec objects: init containers, native sidecars (`restartPolicy: Always` on init, k8s ≥1.29), multi-container workloads, ConfigMap file mounts, tolerations, affinity, security context, and service account.
- Refactor `jobs.py` and `deployments.py` to use the shared compiler; wire ConfigMap create/delete lifecycle with ownership checks on 409 conflicts and rollback on partial creates.
- Add `Container.restart_policy` for native sidecar init containers; extend validation for duplicate port names and duplicate listen ports across containers.

## Test plan

- [x] `uv run --frozen pytest plugins/nemo-deployments/tests/unit` — 251 passed
- [x] `uv run ruff check plugins/nemo-deployments/`
- [x] `uv run --frozen ty check plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/`
- [x] Pre-commit hooks pass on commit
- [x] Golden/snapshot-style compiler tests in `test_compiler.py`
- [x] Deployment rollback tests (ConfigMap cleanup, adopted-deployment path preserves ConfigMap)

## Summary by CodeRabbit

* **New Features**
  * Kubernetes deployments and jobs now support multi-container workloads with sidecar-aware pod construction and improved status reporting.
  * Config files are packaged into generated ConfigMaps and mounted automatically.
  * Added `restartPolicy` support on container definitions for init containers, exposed in the API schema.
* **Bug Fixes**
  * Deployment/job create/delete cleanup is more robust, with better rollback on partial failures.
  * Service ports and in-cluster endpoints are derived from all containers with safer duplicate port handling.
  * Status checks now better reflect workloads that include init containers and `Always` restart behavior.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 import logging
 from dataclasses import dataclass
+from types import SimpleNamespace
 from typing import Any
 
 from kubernetes.client.rest import ApiException
@@ -122,6 +123,11 @@ class CompiledWorkload:
     service_containers: tuple[Container, ...]
 
 
+def _reraise_api_unless(exc: ApiException, *allowed_statuses: int) -> None:
+    if exc.status not in allowed_statuses:
+        raise exc
+
+
 def _validate_port_names(config: DeploymentConfig) -> None:
     seen_names: set[str] = set()
     seen_ports: set[tuple[int, str]] = set()
@@ -160,7 +166,7 @@ def validate_workload_config(config: DeploymentConfig) -> None:
 def validate_config_for_job(config: DeploymentConfig) -> None:
     validate_workload_config(config)
     if config.restart_policy == "Always":
-        raise DeploymentConfigError("restart_policy Always uses Deployment, not Job")
+        raise DeploymentConfigError("restart_policy Always requires a Deployment workload, not a Job")
 
 
 def validate_config_for_deployment(config: DeploymentConfig) -> None:
@@ -177,13 +183,8 @@ def configmap_data_key(path: str) -> str:
 
 def _deserialize_k8s(data: dict[str, Any], klass: str) -> Any:
     k8s = k8s_client_module()
-
-    class _Resp:
-        data: str | None = None
-
-    resp = _Resp()
-    resp.data = json.dumps(data)
-    return k8s.client.ApiClient().deserialize(response=resp, response_type=klass)
+    response = SimpleNamespace(data=json.dumps(data))
+    return k8s.client.ApiClient().deserialize(response=response, response_type=klass)
 
 
 def _build_probe(probe: Probe | None) -> Any | None:
@@ -419,8 +420,7 @@ def create_configmap(
     try:
         core_v1.create_namespaced_config_map(namespace=namespace, body=body, _request_timeout=timeout)
     except ApiException as exc:
-        if exc.status != 409:
-            raise
+        _reraise_api_unless(exc, 409)
         existing = core_v1.read_namespaced_config_map(
             name=body.metadata.name,
             namespace=namespace,
@@ -463,13 +463,11 @@ def delete_configmap(
     try:
         configmap = core_v1.read_namespaced_config_map(name=name, namespace=namespace, _request_timeout=timeout)
     except ApiException as exc:
-        if exc.status == 404:
-            return
-        raise
+        _reraise_api_unless(exc, 404)
+        return
     if not resource_labels_match(configmap, expected_labels):
         return
     try:
         core_v1.delete_namespaced_config_map(name=name, namespace=namespace, _request_timeout=timeout)
     except ApiException as exc:
-        if exc.status != 404:
-            raise
+        _reraise_api_unless(exc, 404)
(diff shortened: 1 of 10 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
```diff
@@ -0,0 +1,475 @@
...
+            raise DeploymentConfigError(
+                f"init container {init_container.name} has unsupported restart_policy "
+                f"{init_container.restart_policy!r}; only Always (native sidecar) is supported"
+            )
+
+
+def validate_config_for_job(config: DeploymentConfig) -> None:
+    validate_workload_config(config)
+    if config.restart_policy == "Always":
+        raise DeploymentConfigError("restart_policy Always uses Deployment, not Job")
```
mckornfield: lol this wording is weird

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
```diff
@@ -0,0 +1,475 @@
...
+def configmap_data_key(path: str) -> str:
+    normalized = path if path.startswith("/") else f"/{path}"
+    key = normalized.lstrip("/").replace("/", "__")
+    return key or "config"
+
+
+def _deserialize_k8s(data: dict[str, Any], klass: str) -> Any:
+    k8s = k8s_client_module()
+
+    class _Resp:
```
mckornfield: hmm inner class seems strange but 🤷

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
```diff
@@ -0,0 +1,475 @@
...
+        configmap = core_v1.read_namespaced_config_map(name=name, namespace=namespace, _request_timeout=timeout)
+    except ApiException as exc:
+        if exc.status == 404:
+            return
+        raise
+    if not resource_labels_match(configmap, expected_labels):
+        return
+    try:
+        core_v1.delete_namespaced_config_map(name=name, namespace=namespace, _request_timeout=timeout)
+    except ApiException as exc:
```
mckornfield, line 472: I've seen this pattern enough I almost want a `with raise_on_api_exception_except` but maybe that'll be messier?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/jobs.py
```diff
@@ -85,112 +89,31 @@ def trim_log_text(text: str) -> tuple[list[str], bool]:
...
 def build_job_body(
     *,
     job_name: str,
     labels: dict[str, str],
     config: DeploymentConfig,
-    container: Container,
     workspace: str,
+    deployment_name: str,
+    k8s_config: K8sDeploymentConfig | None,
 ) -> Any:
```
mckornfield: instead of this `Any` return type, could we return something a bit more concrete? I notice we do a comma separated return type as well, but I'd prefer just one data structure personally (I am not a fan of the following)

`_, something, _ = my_random_function`

plugins/nemo-deployments/tests/unit/backends/k8s/test_compiler.py
```diff
@@ -0,0 +1,231 @@
...
+        config=config,
+        workspace="default",
+        deployment_name="task",
+        labels={"managed-by": "nemo-deployments"},
+        k8s_config=None,
+        pod_restart_policy="Always",
+    )
+    pod_spec = _serialized(compiled.pod_spec_kwargs)
+    assert pod_spec["restart_policy"] == "Always"
+    assert [item["name"] for item in pod_spec["init_containers"]] == ["bootstrap", "sidecar"]
```
mckornfield: could we just do an assert `==` structurally on what we expect, or is that too messy?, for here and beyond

mckornfield (approved):
have some style things, but I think you can ship it assuming you address em

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
tylersbray: Updated to: `restart_policy Always requires a Deployment workload, not a Job` in 0e34d63.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
tylersbray: Replaced the inner `_Resp` class with `SimpleNamespace` in 0e34d63.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
tylersbray: Added a small `_reraise_api_unless(exc, *statuses)` helper for the ConfigMap create/delete paths in 0e34d63 — kept it minimal rather than a context manager for now.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/jobs.py
tylersbray: Added `BuiltJob` / `BuiltDeployment` frozen dataclasses (in `jobs.py` / `deployments.py`) so callers get a single struct instead of tuple unpacking. Pushed in 0e34d63.

plugins/nemo-deployments/tests/unit/backends/k8s/test_compiler.py
tylersbray: Refactored the init/sidecar golden test to a full structural `assert pod_spec == {...}` on the serialized dict in 0e34d63.

PR #557, opened 2026-07-02
test(guardrails): Add e2e tests for service endpoints

JashG:
## Summary by CodeRabbit

* **New Features**
  * Added end-to-end coverage for creating, retrieving, listing, updating, and deleting guardrail configurations.
  * Added validation for duplicate configuration names and unknown configuration references.
  * Added checks for safe and unsafe inputs and model outputs, including inline configurations.
  * Added verification of rail status and optional metadata for activated guardrails.

### DIFF
--- a/e2e/guardrails/test_checks.py
+++ b/e2e/guardrails/test_checks.py
@@ -1,0 +1,155 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""E2E tests for the Guardrails checks endpoint.
+
+These tests verify that ``/apis/guardrails/v2/workspaces/{workspace}/checks``
+runs content-safety rails through the real platform subprocess while using mock
+providers for deterministic rail-model responses.
+"""
+
+from collections.abc import Callable
+from typing import Any
+
+import httpx
+import pytest
+
+from e2e.guardrails.utils import (
+    BACKEND_RESPONSE,
+    CONTENT_SAFETY_INPUT_FLOW,
+    CONTENT_SAFETY_OUTPUT_FLOW,
+    GuardrailsChatTestCase,
+)
+
+
+def _post_check(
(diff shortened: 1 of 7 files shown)


e2e/guardrails/test_checks.py
```diff
@@ -0,0 +1,155 @@
...
+    test_case, _config_data = guardrails_check_test_case(
+        config_mode="referenced", outcome="unsafe_output", rail_types=("input", "output")
+    )
+
+    response = _post_check(
+        test_case,
+        extra_guardrails={"options": {"log": {"activated_rails": True}}},
+    )
+
+    assert response["guardrails_data"]["config_ids"] == [test_case.config_ref]
```
gabwow: I'm a bit confused on this as /private/tmp/nemo-pr557.thaytk/services/guardrails/src/nmp/guardrails/app/schemas/utils/response_transformers.py:285 sets config_ids=None and 
@router.post(
    "/v2/workspaces/{workspace}/checks",
    response_model=GuardrailCheckResponse,
    response_model_exclude_none=True,
)
excludes None. I'm guessing this passes but don't see where config_ids got set

e2e/guardrails/test_checks.py
```diff
@@ -0,0 +1,155 @@
...
+        guardrails = {"config_id": test_case.config_ref}
+    else:
+        if config_data is None:
+            raise ValueError("config_data is required for inline checks")
+        guardrails = {"config": config_data}
+
+    if extra_guardrails:
+        guardrails.update(extra_guardrails)
+
+    response = test_case.sdk._client.post(
```
gabwow: nit: In guardrails.py:74 we use sdk.guardrail.check(...). It seems like we could do this here to be more idiomatic.

gabwow (approved):
conditional approval provided my misunderstanding of that assert is a genuine misunderstanding.

e2e/guardrails/test_checks.py
JashG: You're right that `config_ids` is not included in the response - removed this assertion.

PR #563, opened 2026-07-03
feat(evaluator-sdk): typed workspace seed sources (inline/path/fileset)

SandyChapman:
## What

Follow-up to the general Codex runtime (#562). A task's `inputs["files"]` seed map now accepts three **JSON-serializable** seed shapes, chosen by a `kind` discriminator (a bare string stays sugar for inline text):

| `kind` | Source | Resolvable where |
|---|---|---|
| `inline` | contents in the task (`encoding="base64"` for binary) | anywhere — fully portable |
| `path` | a file on the authoring host | only there (local-authoring convenience) |
| `fileset` | a `workspace/name#glob` reference | only where the Files service is reachable (service-side) |

The three differ in **where they resolve**, which is what governs a task's portability — the reason a bare string alone (which can't tell content from a path from a ref) isn't enough.

## Layering

- `inline` + `path` resolve in the pure-SDK runtime (no platform dependency).
- `fileset` raises a clear *"cannot be resolved in local execution"* error — the same pattern as metric-reference resolution. It's **reserved** for the service-side path; a later change will resolve filesets there and **normalize** `inline`/`path` seeds into a fileset at the local→remote submit boundary (mirroring inline-metric → derived-metric normalization).

## Shape

- New `agent_eval/workspace_seeds.py`: the `SeedFile` discriminated union (`InlineSeed`/`PathSeed`/`FilesetSeed`), `parse_seed`, and the `seed_workspace` resolver — with workspace-escape protection and binary (base64) support.
- The Codex runtime delegates to `seed_workspace`; its private `_seed_workspace` is gone.
- `AgentEvalTask.inputs` stays a generic `dict[str, Any]` — seeding remains a documented convention over `inputs["files"]`, not a core task field.

## Testing

- New `test_workspace_seeds.py` (inline/base64/path/fileset/escape/unknown-kind); updated the Codex runtime's escape test (now `WorkspaceSeedError`).
- Full `agent_eval` suite (91) + plugin `test_agent_evaluate.py` pass; `ruff`/`format` clean; CI `lint-python-types.sh` exit 0.

**Stacked on #562** — targets `codex-runtime-general-coding/schapman`; review/merge after it.

## Summary by CodeRabbit

* **New Features**
  * Added a workspace seeding system supporting inline content, local file copies, and extensible custom seed handlers.
  * Added `fileset` seed support to load a single file from a referenced dataset into the workspace.
* **Bug Fixes**
  * Improved workspace safety by rejecting seed paths that could escape the workspace.
  * Standardized invalid/unsupported seed failures under `WorkspaceSeedError`, and ensured seeding runs off the main event loop.
* **Tests**
  * Added coverage for seed parsing, workspace writing, `fileset` handling, and Codex runtime thread/error behavior.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/workspace_seeds.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/workspace_seeds.py
@@ -1,0 +1,171 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Workspace seed files for agent-eval tasks.
+
+A task can stage starter files into the agent's workspace before it runs, under
+``inputs[SEED_FILES_INPUT_KEY]`` as a ``{relative_path: source}`` map. Each *source* is a
+JSON-serializable value whose ``kind`` selects a registered :class:`SeedHandler` (a bare string is
+sugar for inline text).
+
+The SDK ships handlers only for the kinds it can resolve with **no external dependency** — ``inline``
+and ``path``. Other kinds are contributed by consumers via :func:`register_seed_handler`; the SDK has
+no knowledge of them (e.g. a platform ``fileset`` handler lives in the plugin and resolves against the
+Files service at run time). An unregistered kind raises :class:`WorkspaceSeedError`.
+"""
+
+from __future__ import annotations
+
+import base64
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Any, Literal, Protocol, runtime_checkable
+
+from pydantic import BaseModel, ConfigDict, Field
+
+#: ``inputs`` key holding the ``{relative_path: seed}`` map of files to stage into the workspace.
+SEED_FILES_INPUT_KEY = "files"
+
+
+class WorkspaceSeedError(ValueError):
+    """A workspace seed could not be parsed, resolved, or written (bad value, unknown kind, ...).
+
+    Subclasses ``ValueError`` so a runner's per-task error handling still catches it.
+    """
+
+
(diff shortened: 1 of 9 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/workspace_seeds.py
```diff
@@ -0,0 +1,124 @@
...
+    path: str = Field(description="Filesystem path on the authoring host (absolute or relative to the cwd).")
+
+
+class FilesetSeed(BaseModel):
+    """A reference to a stored fileset. Resolvable only where the Files service is reachable."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    kind: Literal["fileset"] = "fileset"
+    ref: str = Field(description="Fileset reference, e.g. 'workspace/name' or 'workspace/name#glob'.")
```
arpitsardhana: nit: may be ref validation might be useful to ensure that ref confirms to files format

arpitsardhana (approved):
Just ensuring that my understanding is correct: that SDK may have files awareness so that we don't have too many duplicates files between SDK and plugin but SDK won't have any dependency or do resolution of files.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/workspace_seeds.py
SandyChapman: Ah, yeah, this is a good catch. We may also want to not have this `kind` in the SDK, but as a plugin extension here. Let me revisit to see if I can handle this.

SandyChapman: I fixed this. Only the plugin knows about filesets now and we properly validate the ref field.

PR #570, opened 2026-07-06
feat(cli): add anonymous usage telemetry

yamini:
## What this does

Adds anonymous usage telemetry to the NeMo Platform CLI and SDK, implementing Phase 1 of the usage-telemetry PRD:
https://docs.google.com/document/d/10zr4yno15Nx_O67inlvamuhVL5t1u3oNcWOE2CQ1_QY/edit

Three anonymous events, emitted from the choke points every command already flows through:

| Event | Fires when | Key fields (all anonymous) |
|---|---|---|
| `command_invoked` | any CLI command finishes | command, status, duration, agent-mode, CI flag |
| `onboarding_step` | each `nemo setup` step | step name, status, provider type, model-count bucket |
| `job_run` | a platform job reaches a terminal state | job type, status, duration, plugins, tokens/model when available |

## Privacy

No prompts, user data, file contents, or personal information leave the machine. Event models use pydantic `extra="forbid"`, so an event physically cannot carry a free-text or arbitrary field. A random per-run session id groups the events from one invocation and is discarded when the process exits; it is not a user or device identifier. stdout stays byte-identical whether telemetry is on or off, so agent-mode consumers that parse stdout are unaffected.

## Opt out (any one disables it)

- `NEMO_TELEMETRY_ENABLED=false` (environment)
- `telemetry_enabled: false` in the config file
- `--no-telemetry` on a single command

The three layers are independent: telemetry runs only if you have not disabled it in any of them, and no layer can turn it back on. If the config cannot be read, telemetry stays off. A first-run notice prints once to stderr explaining what is collected and how to opt out.

## Performance

Emission is best effort and can never break or change the outcome of a command. The send is bounded by a 2 second timeout with no retries on the exit path, so an unreachable telemetry endpoint adds at most about 2 seconds and normally well under a second. Moving the send fully off the command's thread (fire and forget) is a sensible follow-up if maintainers want zero added latency.

## Provenance

The queue/flush/retry handler is vendored from Safe-Synthesizer `src/nemo_safe_synthesizer/telemetry.py` (commit `8a86f8b`), adapted to the CLI's event models and logging. Source lives in `packages/nemo_platform_ext/`; `sdk/python/**` is the `make vendor` mirror.

## Tests

- Telemetry suite: 92 tests (ported handler suite plus per-hook suites), all passing.
- Full `nemo_platform_ext` suite green except pre-existing environment failures unrelated to this change (a local port 8080 conflict).
- A network-free wire-contract smoke test asserts each event serializes to the schema `1.9` envelope. ruff, `ty`, and the vendored-sync lint all pass.

## Open dependencies

- `nemoSource` value is set to `"platform"` as a placeholder. It must be registered with the telemetry schema before events validate end to end. Owner: Matt Kornfield (GXT). Live UAT against the telemetry endpoint is pending that registration; this PR validates against the wire contract locally only.
- `job_run` token and model fields ship now but stay `-1`/`"undefined"` until platform services populate `status_details`. That population is a services follow-up.
- The vendored handler drops non-retryable-looking HTTP responses (for example 429) without a dead-letter retry. Carried over from the reference implementation; flagging for Matt Kornfield to decide whether 429 should retry.

This is a proposal from PM, not assigned work. Happy to adjust scope or approach.

### DIFF
--- a/docs/cli/configuration.mdx
+++ b/docs/cli/configuration.mdx
@@ -151,6 +151,8 @@ Collected data includes:
 
 The events are batched and sent to NVIDIA's telemetry endpoint with no user or device identifiers attached.
 
+A random per-run session identifier groups the events from a single invocation. It is generated fresh each time you run a `nemo` command, carries no user or device identity, and is discarded when the process exits.
+
 ### What Data Is Not Collected
 
 - User data, prompts, or model outputs
--- a/docs/cli/configuration.mdx
+++ b/docs/cli/configuration.mdx
@@ -131,3 +131,75 @@ nemo models <Tab>
 ```
 
 If completion isn't working, make sure you've restarted your shell and that your shell supports programmable completion. For Bash, ensure `bash-completion` is installed on your system.
+
+## Telemetry
+
+NeMo Platform collects anonymous usage data to improve the product. On your first run, the CLI prints a notice to standard error explaining what data is collected and how to opt out.
+
+### What Data Is Collected
+
+The CLI and bundled plugins emit anonymous usage events. No prompts, user data, personal information, or file contents leave your machine.
+
+Collected data includes:
+
+- Command and job names you invoke
+- Completion status and duration of commands and jobs
+- Plugin and tool names you use
+- Provider type and model counts (bucketed, not exact)
+- Input and output token counts per model (where available)
+- Package version and CPU architecture
+
+The events are batched and sent to NVIDIA's telemetry endpoint with no user or device identifiers attached.
+
+### What Data Is Not Collected
+
+- User data, prompts, or model outputs
+- File contents or references
+- Usernames, email addresses, IP addresses, or hostnames
+- Machine fingerprints or unique identifiers
+
+### Opting Out
+
+Disable telemetry using any of these three methods (they all work):
+
+#### 1. Environment Variable
+
+```bash
+export NEMO_TELEMETRY_ENABLED=false
+```
+
+Then run any `nemo` command. This disables telemetry for the CLI and all bundled plugins for that session and all future sessions (until you unset it).
+
+#### 2. Configuration File
+
+Add this to your configuration file at `~/.config/nmp/config.yaml`:
+
+```yaml
+telemetry_enabled: false
+```
+
+#### 3. Per-Command Flag
+
+```bash
+nemo --no-telemetry workspaces list
+```
+
+The `--no-telemetry` flag disables telemetry for a single command only.
+
+<Note>
+
+These three layers are independent. NeMo Platform sends telemetry only when you have not disabled it in any of them. Setting any one layer to off disables telemetry, and no layer turns it back on. For example, if you leave the environment variable unset but set `telemetry_enabled: false` in your config file, telemetry stays off.
+
+</Note>
+
+### Bundled Plugins
+
+NeMo Platform ships with plugins that emit their own anonymous usage telemetry. The `NEMO_TELEMETRY_ENABLED` environment variable is the recommended single off switch: set it to `false` to disable telemetry for the CLI and for NeMo tools that share this telemetry system. The config file setting and the `--no-telemetry` flag apply to the CLI.
+
+### First-Run Notice
+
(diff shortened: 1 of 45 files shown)


docs/cli/configuration.mdx
```diff
@@ -131,3 +131,77 @@ nemo models <Tab>
...
+- Command and job names you invoke
+- Completion status and duration of commands and jobs
+- Plugin and tool names you use
+- Provider type and model counts (bucketed, not exact)
+- Input and output token counts per model (where available)
+- Package version and CPU architecture
+
+The events are batched and sent to NVIDIA's telemetry endpoint with no user or device identifiers attached.
+
+A random per-run session identifier groups the events from a single invocation. It is generated fresh each time you run a `nemo` command, carries no user or device identity, and is discarded when the process exits.
```
mckornfield, line 154: wonder if this will be a point of contention, if this is more than what currently the telemetry endpoint does

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -450,13 +461,57 @@ def _update_provider(
 _NON_COMPLIANT_MARKER = "Non-OpenAI compliant"
 
 
+def _bucket_model_count(count: int) -> str:
+    """Bucket a discovered-model count into a coarse range for telemetry.
```
mckornfield, line 465: I remember it's pretty high for nvidia-build (like ~100), do we care then to differentiate that?

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/help_formatter.py
```diff
@@ -47,36 +47,86 @@ def main(
...
+        if is_telemetry_root:
+            runtime.reset()
+            # Reading help (e.g. ``nemo docs --help``) is not usage, so it must not emit
+            # a command_invoked event. Detect a help request from the resolved args; the
+            # help option names are the same ones every command registers via
+            # ``_context_settings_with_help``.
+            import sys
+
+            resolved_args = args if args is not None else sys.argv[1:]
+            if any(arg in HELP_OPTION_NAMES for arg in resolved_args):
```
mckornfield, line 79: this is an interesting way to do it, couldn't it just be a variable that is conditional below, instead of being stateful into the other method deciding whether or not to fire?

packages/nemo_platform_ext/src/nemo_platform_ext/cli/telemetry/emit.py
```diff
@@ -0,0 +1,104 @@
...
+
+# One random session id per process. It groups every event emitted during a single
+# invocation without carrying any user or device identity, and is discarded when the
+# process exits. This is the same pattern the Agno framework uses for session_id/run_id.
+_SESSION_ID = uuid.uuid4().hex
+
+_NOTICE_TEXT = (
+    "NeMo Platform collects anonymous usage data to improve the product. "
+    "No prompts, data, or personal information leave your machine. "
+    "Turn it off any time with NEMO_TELEMETRY_ENABLED=false. "
```
mckornfield, line 27:
```suggestion
    "Turn it off at any time with NEMO_TELEMETRY_ENABLED=false. "
```

packages/nemo_platform_ext/src/nemo_platform_ext/cli/telemetry/events.py
```diff
@@ -0,0 +1,87 @@
...
+
+from __future__ import annotations
+
+import os
+from enum import Enum
+from typing import ClassVar
+
+from pydantic import BaseModel, ConfigDict, Field
+
+_CI_ENV_VARS = ("CI", "GITLAB_CI", "GITHUB_ACTIONS")
```
mckornfield, line 17: just thinking out loud, could other people be running the platform in CI, or we assume it will only be us?

mckornfield (commented):
submitting as maybe a review to myself if I take this over lol. Nothing crazy sticks out about it though

PR #561, opened 2026-07-03
feat(evaluator): add Taskset entity for grouping tasks (AALGO-307)

SandyChapman:
## What

Adds the **Taskset** half of the AALGO-307 entity work: a flexible grouping of stored tasks with metadata, mirroring the merged Task entity stack (#527) end-to-end.

- `TasksetEntity` (`entity_type="taskset"`) + `Taskset` / `TasksetInput` DTOs
- `TasksetService` (create/get/list/delete)
- `/tasksets` CRUD routes with decorator authz via `TasksetPerms`
- `client.evaluator.tasksets` SDK resource (sync + async)
- mounted in `service.py` and `sdk/resources.py`

## Design

- **Members by reference.** A taskset references its tasks by `workspace/name` (no inline tasks). Metadata reuses `MetadataItem` / `TaskMetadataList`; the grouping carries an optional `description`.
- **Set semantics.** Order is not significant and duplicate members are rejected — both byte-identically (field validator) and by resolved `(workspace, name)` in the service, so `task-a` and `default/task-a` can't both slip in.
- **Existence validated at create.** Each referenced task must resolve to a stored task; missing/duplicate refs surface as typed errors (`TaskRefNotFoundError` / `DuplicateTaskRefError`) mapped to 422, name collisions to 409.
- Factors the shared `workspace/name` ref parser into `schemas.parse_entity_ref`, now reused by `metric_refs.parse_metric_ref`.

## Not covered (follow-ups)

- **Validate-then-create is not atomic** (a referenced task could be deleted between the existence check and insert). Inherent to the "validate at create" design and consistent with metric-ref handling; left as-is.
- `task_refs` on the agent-eval submit path (run an eval over a stored taskset) — separate follow-up.

## Testing

- New: entity round-trip, service, routes, SDK resource tests (incl. resolved-dedup + missing-ref 422).
- Full evaluator unit suite: **469 pass**.
- `ruff check` / `format`, CI `lint-python-types.sh` (exit 0), `lint-all.sh` all pass.
- `make refresh-openapi`: only `plugins/nemo-evaluator/openapi/openapi.yaml` changed.

## Summary by CodeRabbit

* **New Features**
  * Added evaluator **tasksets** API (create, list, retrieve, delete) under workspace-scoped routes, including paginated filtering/sorting.
  * Added **tasksets** support to the evaluator SDK for both sync and async clients.
  * Added taskset DTO validation (task reference handling and duplicate-member protection).
* **Bug Fixes**
[...]

PR #573, opened 2026-07-06
feat(evaluator-sdk): per-task workspace + held-out overlay verifier for Fabric agent-eval

SandyChapman:
Two related changes that let Fabric run **workspace-based coding-agent evaluations** and grade them against held-out ground truth.

## 1. Per-task workspace in `FabricAgentRuntime`

`FabricAgentRuntime` was trajectory-only — it never seeded `inputs['files']` and exposed no `workspace` filesystem evidence, so the workspace-scoring metrics that work against the Codex runtime couldn't score a Fabric trial. Since Fabric drives the harness on the **host filesystem** in a dir we choose (no container), the agent's edited files are recoverable directly.

Now **every task runs in its own fresh per-task workspace**: `inputs['files']` are seeded into it (a no-op when there are none), the harness runs in it via an `environment.workspace` profile overlay, and the final tree is exposed as a `workspace` filesystem `EvidenceDescriptor` — identical key/kind to the Codex runtime, so existing workspace-reading metrics work unchanged. Staging is unconditional (not a flag): a uniform per-task workspace matches the Codex runtime and maps cleanly onto a per-task container volume later. A config-supplied `environment.workspace` is overridden per task.

## 2. `overlay_files` on `LocalFilesystemEvidence.run_verifier`

A `{relative_path: contents}` map of trusted files written *over* the throwaway verifier copy before the command runs — the reusable "load workspace + overlay held-out files + run verifier" primitive. This is how a grader supplies held-out artifacts (a canonical test suite, a reference implementation) that must not live in, and cannot be edited through, the agent's own workspace. Paths escaping the copy are rejected; stored evidence is never mutated.

Together: seed only agent-facing files into the workspace, keep graders' ground truth in the task's `reference`, overlay it at scoring time.

## Testing

- `pytest packages/nemo_evaluator_sdk/tests/agent_eval/` → 109 passed, 1 skipped (gated live test).
- Fabric runtime: workspace seeded + agent edit captured + evidence exposed; a no-files task still gets a workspace; a bad seed fails only that task.
- `run_verifier` overlay: held-out file appears in the copy, overlay wins over the agent's version, nested paths work, evidence untouched, path-escape rejected.
- `ruff` / `ruff format` / CI `ty` clean; `make vendor` synced.

## Stack

`main` ← #561 (Taskset) ← **#573** ← #565 (notebook). Base retargets to `main` as #561 merges.

## Summary by CodeRabbit

* **New Features**
  * Fabric task execution now stages each task in an isolated workspace, seeds provided input files, and includes the resulting workspace filesystem as evidence.
  * Evidence verification now supports optional overlay files applied to the verifier copy.
* **Bug Fixes**
  * Trial metadata now consistently reports `agent_ok` for both successful and failed trials.
  * Added stronger path containment protections for both workspace seeding and verifier overlays.
* **Tests**
  * Expanded coverage for overlay behavior, security validation, per-task workspace evidence/seeding (including empty inputs), traversal failures, and metric correctness.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
@@ -9,6 +9,13 @@ NeMo Fabric Python SDK and adapts each normalized Fabric ``RunResult`` into an
 ``harness.adapter_id`` (never inferred from a model); an optional ``model`` slug
 is applied as a final profile overlay, mirroring Fabric's own Harbor integration.
 
+Every task runs in its own fresh workspace: the runtime seeds it from
+``inputs['files']`` (a no-op when there are none), runs the harness in it (via
+``environment.workspace``), and exposes its final file tree as ``workspace``
+filesystem evidence, so workspace-reading metrics score a Fabric trial alongside
+the ATIF trajectory. Any ``environment.workspace`` set in the supplied config is
+overridden per task.
+
 ``nemo_fabric`` is an optional native dependency: its types are imported for
 annotations under ``TYPE_CHECKING`` and the package is loaded lazily at runtime,
 so this module stays importable without it.
@@ -24,6 +31,7 @@ from typing import TYPE_CHECKING, Any
 
 from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig, AgentEvalTask
 from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial, AgentEvalTrialStatus, AgentOutput
+from nemo_evaluator_sdk.agent_eval.workspace_seeds import SEED_FILES_INPUT_KEY, seed_workspace
 from nemo_evaluator_sdk.values.evidence import (
     EVIDENCE_FORMAT_ATIF,
     EVIDENCE_TRACE,
@@ -55,6 +63,14 @@ _MISSING_RELAY_MSG = (
 # them and hand them to Fabric/Relay, so they are not derived from either library.
 _RELAY_SUBDIR = "relay"
 _ARTIFACTS_SUBDIR = "artifacts"
+# Per-task workspace: where seed files are staged and where the harness reads/writes. We
+# create it, point Fabric's ``environment.workspace`` at it, and expose it as ``workspace`` evidence.
+_WORKSPACE_SUBDIR = "workspace"
+_WORKSPACE_PROFILE_NAME = "eval_workspace"
+# Evidence key + descriptor kind for the staged workspace, consumed by the
+# workspace-reading metrics.
+_WORKSPACE_EVIDENCE_KEY = "workspace"
+_WORKSPACE_EVIDENCE_KIND = "filesystem"
 # Trajectory profile identity + the file-exporter output names we choose (Relay accepts these as inputs).
 _TRAJECTORY_PROFILE_NAME = "eval_trajectory"
 _ATIF_FILENAME_TEMPLATE = "trajectory-{session_id}.atif.json"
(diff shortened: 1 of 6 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -148,12 +165,24 @@ async def _run_task(
             artifacts_dir.mkdir(parents=True, exist_ok=True)
             profiles.append(self._trajectory_profile(profile_cls, relay_dir=relay_dir, artifacts_dir=artifacts_dir))
 
+        # Every task runs in its own fresh workspace: seed any ``inputs['files']`` into it (a no-op when
```
arpitsardhana, line 167: by fresh workspace, you mean a fresh directory where original contents are copied?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -193,13 +224,19 @@ def _to_trial(self, task: AgentEvalTask, result: RunResult, evidence_dir: Path)
                 response=result.output,
                 metadata={**base_metadata, "evidence_dir": str(evidence_dir)},
             ),
-            evidence=self._evidence(result, result_path),
-            metadata={**base_metadata, "generated": True},
+            evidence=self._evidence(result, result_path, workspace_dir),
+            # agent_ok mirrors the Codex runtime: AgentPhaseSuccessMetric reads it to score whether the
```
arpitsardhana: I think claude adding these extraneous comment like We are doing x since Codex did x

arpitsardhana (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: Yes, exactly.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: Ah, thanks for catching this. I'll scrub it.

SandyChapman:
@coderabbitai Re: the "seed handling is outside the task timeout" note in `_run_task` — intentionally leaving this as-is (won't-fix for this PR):

- **Parity, not a regression.** The Codex runtime (already shipped) does the identical thing: `seed_workspace` via `asyncio.to_thread` *before* `asyncio.wait_for(client.run(...), timeout=self._timeout_s)`. This runtime matches that established pattern rather than diverging from it.
- **The suggested fix doesn't actually bound anything.** Wrapping the `to_thread` seed call in `wait_for` can't cancel a thread running in the default executor — on timeout the task is marked failed while the seed thread keeps running in the background. That's a misleading guarantee, not real cancellation.
- **Seeding is off the event loop.** Because it's offloaded via `to_thread`, a slow fileset handler doesn't stall other concurrent tasks; only that one task's wall-clock is unbounded.

If we ever need to bound seed/download time, the right place is a timeout inside the fileset handler itself (your secondary suggestion), not a `wait_for` here.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: Scrubbed the "mirrors / because the Codex runtime does X" editorializing across the file in the amended commit — kept only the factual harness/adapter references (the harness list, the config-example pointer, and the codex-cli adapter cwd/output-shape mechanics). Force-pushed.

PR #575, opened 2026-07-06
feat(deployments): grant k8s backend RBAC in deploy chart (AIRCORE-757 phase 6)

tylersbray:
## Summary

AIRCORE-757 Phase 6 — grants the k8s deployments backend (Phase 5, #553) the RBAC it needs, wired into the deploy chart.

- Extends the existing `nmp-core` controller Role (`k8s/helm/templates/core/controller-role.yaml`) with the two verbs the k8s `DeploymentBackend` needs but didn't have: `create`/`delete` on `apps/v1.Deployment`, and a new rule for `services` (`get`/`list`/`create`/`delete`).
- Documents the k8s executor config and RBAC scope in `plugins/nemo-deployments/README.md`.
- Adds `plugins/nemo-deployments/tests/unit/backends/k8s/test_rbac_manifest.py` asserting the chart's RBAC rules, plus a regression guard over the pre-existing base rules (pods, pods/log, jobs, PVCs, configmaps).

## Design decisions

**Extended the existing `nmp-core` controller Role instead of a dedicated ServiceAccount/Role for the deployments plugin.** `DeploymentsController` is registered via the `nemo.controllers` entry point and runs inside the existing `nmp-core` controller pod (`--controller-group=all`), under `controllerServiceAccountName`. A parallel Role bound to that same pod's ServiceAccount would be redundant, so this PR extends the Role already granting that pod's RBAC (pods, pods/log, jobs, PVCs, configmaps) rather than introducing a second one.

**RBAC audit was exhaustive against every k8s API call in the Phase 1-5 backend code**, not just the ticket's suggested verb list: `volumes.py`, `jobs.py`, `deployments.py`, `compiler.py`. Result: exactly two gaps — `apps/deployments` was `get`/`list`/`watch` only (backend also creates/deletes), and `services` had no rule at all (backend gets/creates/deletes). Both additions are minimal: no `update`/`patch` on Deployments since the backend recreates rather than patches.

**RBAC additions are ungated** (no new `rbac.*` values flag), consistent with the existing jobs/PVC/configmap base rules — keeps `values.yaml` and `helm-docs`-generated README untouched, since this isn't an optional third-party integration like Volcano or the NIM Operator.

**RBAC test uses raw-template regex parsing, not a `helm template` render** — this diverged from the original plan. `k8s/helm/Chart.yaml` declares an *unconditional* `k8s-nim-operator` chart dependency (only the *installed resources* are gated by `.Values.rbac.k8sNimOperatorEnabled`, not the dependency-presence check), so `helm template` refuses to run at all unless that subchart `.tgz` is fetched from an NGC repo into `charts/` first — verified locally that this fails even with `k8s-nim-operator.enabled=false` and `--show-only`. There's no existing helm lint/template CI job in this repo and no committed `Chart.lock`/`charts/` dir, so a skip-gated render would silently skip in CI 100% of the time. The test instead parses the static YAML rule entries directly from the Go-template source, which is safe because the base RBAC rules (unlike the Volcano/NIM-Operator blocks) are plain YAML with no Helm expressions inside them. Runs deterministically with no `helm` binary or network dependency.

**No functional carryover from Phase 5 review** — confirmed against #553's merge history; mckornfield's review comments were style-only and already addressed pre-merge.

### Forward-looking note: AIRCORE-759 interaction

[AIRCORE-759](https://linear.app/nvidia/issue/AIRCORE-759/deployments-plugin-models-adopts-the-plugin-drop-the-k8s-nim-operator) (models adopts this plugin, drops `k8s-nim-operator`) is backlogged but directly bears on the decisions above:

- It validates the Role-reuse decision: once models stops driving k8s directly and routes NIM/vLLM through this plugin's SDK, that traffic executes through the exact same `K8sDeploymentBackend`/`DeploymentsController`/`nmp-core` controller pod this PR grants RBAC to. A dedicated per-plugin ServiceAccount would have needed reconciling with models' path later; extending the shared Role avoids that.
- It explains and time-bounds the `helm template` blocker above: AIRCORE-759 requires removing the `k8s-nim-operator` subchart dependency, its CRDs, and its RBAC (`nimservices`/`nimcaches` rules) from this same chart. Once that lands, `helm template k8s/helm` should render standalone and `test_rbac_manifest.py` could be upgraded from raw-text parsing to a real chart render.
- The RBAC surface this PR grants (Deployment/Service/Job/PVC/ConfigMap/Pods/pods-log) should already be sufficient for AIRCORE-759's NIM/vLLM traffic, since both compile to the same generic `(Volume, puller, server)` `DeploymentConfig` primitives that already flow through the backend code audited here — no new verbs anticipated purely from that migration.
- One real gap for whoever picks up AIRCORE-759: its scope note mentions image-pull secrets moving to `backend_config.k8s.image_pull_secrets`, but that field doesn't exist yet on `K8sDeploymentConfig`/`compiler.py` today (verified) — will need a small compiler addition, and possibly broader `secrets` verbs (currently only `create`/`delete`) if the plugin needs to read/compare an existing pull secret.
- The dev-blue `tbray-dev` pod's missing `nemo-models-vllm` Role/RoleBinding (noted separately during manual-smoke-test setup) is models' current, pre-759 standalone k8s-emission identity from AIRCORE-694. It becomes obsolete once AIRCORE-759 lands and models delegates to this plugin's `nmp-core` Role — a natural line item for that ticket's own RBAC-cleanup requirement, not addressed here.

## Test plan

- [x] `uv run --frozen pytest plugins/nemo-deployments/tests/unit -q` — 258 passed
- [x] `uv run ruff check` / `uv run ruff format --check` / `uv run --frozen ty check` — clean
- [x] Manually verified with `helm template` (locally cached `k8s-nim-operator` subchart) that the rendered Role contains the expected verbs
- [x] `uv run pre-commit run -a` — all hooks pass except `studio-lint-staged`, a pre-existing Node-version environment gap scoped to `web/` (untouched by this PR) that self-skips in CI
- [x] Bugbot subagent review (clean context): no bugs found
- [ ] Phase 7 (follow-up PR) adds kind/kubeconfig integration tests and cluster-validation docs; Tier C manual smoke on dev-blue runs after Phase 6 + 7 merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded controller RBAC permissions to allow creating and deleting Kubernetes deployments and to grant required permissions for core services.
* **Documentation**
  * Updated Kubernetes executors backend documentation, including created resources, namespace override behavior, RBAC requirements, and native sidecar compatibility notes.
* **Tests**
  * Added unit tests that statically validate Helm chart RBAC rules and enforce required verb sets for key resources.


plugins/nemo-deployments/README.md
```diff
@@ -56,3 +56,52 @@ from the hyphen-joined string, so pairs like ``foo``/``bar-baz`` and
...
+### RBAC
+
+The `DeploymentsController` runs inside the `nmp-core` controller pod
+(registered via the `nemo.controllers` entry point), so it reuses that pod's
+existing ServiceAccount and Role rather than a dedicated one. The deploy
+chart's `k8s/helm/templates/core/controller-role.yaml` grants that Role the
+verbs the k8s backend needs in `.Release.Namespace`: `get`/`list`/`watch` on
+pods and pods/log, `create`/`get`/`list`/`watch`/`update`/`patch`/`delete` on
+`batch/v1.Job`, `get`/`list`/`create`/`delete` on PVCs, ConfigMaps, and
+Services, and `get`/`list`/`watch`/`create`/`delete` on `apps/v1.Deployment`.
```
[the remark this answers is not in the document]
tylersbray: Good catch, fixed in c282eceb2f — split `pods` (`get`/`list`/`watch`) from `pods/log` (`get`/`list` only) in the RBAC summary.

plugins/nemo-deployments/tests/unit/backends/k8s/test_rbac_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+
+import re
+from pathlib import Path
+
+import pytest
+
+
+def _repo_root() -> Path:
+    for parent in Path(__file__).resolve().parents:
+        if (parent / "k8s" / "helm" / "templates" / "core" / "controller-role.yaml").is_file():
```
mckornfield: nit- could this path construction be a fn or constant?

plugins/nemo-deployments/README.md
```diff
@@ -56,3 +56,52 @@ from the hyphen-joined string, so pairs like ``foo``/``bar-baz`` and
...
+existing ServiceAccount and Role rather than a dedicated one. The deploy
+chart's `k8s/helm/templates/core/controller-role.yaml` grants that Role the
+verbs the k8s backend needs in `.Release.Namespace`: `get`/`list`/`watch` on
+pods, `get`/`list` on pods/log, `create`/`get`/`list`/`watch`/`update`/`patch`/`delete` on
+`batch/v1.Job`, `get`/`list`/`create`/`delete` on PVCs, ConfigMaps, and
+Services, and `get`/`list`/`watch`/`create`/`delete` on `apps/v1.Deployment`.
+
+This Role is namespace-scoped to the release namespace. Pointing
+`backend_config.k8s.namespace` at a namespace outside the release namespace
+requires additional RBAC (a RoleBinding in that namespace, or a
```
mckornfield: is the parenthetical part necessary or "self evident"

mckornfield (approved):

plugins/nemo-deployments/tests/unit/backends/k8s/test_rbac_manifest.py
tylersbray: Good call — extracted into a `_CONTROLLER_ROLE_RELATIVE_PATH` constant so both `_repo_root()` and `_controller_role_source()` reference the same path. Fixed in 5265cea.

plugins/nemo-deployments/README.md
tylersbray: Agreed, not necessary — dropped the parenthetical in 5265cea. Anyone hitting the "additional RBAC" gap already knows they need a RoleBinding or ClusterRole; the sentence reads cleaner without spelling out the K8s primitives.

PR #577, opened 2026-07-06
feat(studio): Additional numerical range filters

nakolean:
## Summary

* **New Features**
  * Added dynamic filter-field mapping support (including function-based mapping and pattern-based remaps).
  * Enabled numeric range filtering for evaluator average scores, average cost, and run count.

* **Bug Fixes**
  * Improved backend filter key resolution so UI filter values map to the correct fields, with sensible fallback when a mapping is unavailable.
  * Kept evaluator score columns available and stably ordered even when the active filters produce zero matching rows.

* **Tests**
  * Expanded coverage for the new dynamic mapping and remapping behaviors.


shanaiabuggy (approved):
Beautiful! Can we include **Run Count** in the set of filters as well?

PR #578, opened 2026-07-06
feat(experiments): add group default metric sort, applied client-side

shanaiabuggy:
- Add a default metric sort to experiment groups — a single sort field (e.g. -cost_usd.mean) stored on the group as a sort-param string.
- Add UI to set it in the experiment group create and edit modals, via a new shared DefaultSortControl (pick a metric — Avg Cost, Avg Latency, Run count, or a discovered evaluator — plus an Asc/Desc toggle).
- Add an Edit action + edit modal on the experiment group detail page.
- The client now reads default_metric_sort from the group and applies it as the experiments list's sort param on load, so the sorted column's header reflects the active sort; users can still re-sort by clicking any column.
- Make the intake list endpoint a plain sort executor — it applies only the sort it's given and defaults to -created_at (pinned first) when omitted, instead of computing a group's default server-side.
- Change how default metric sort is stored from the old SortCriterion[] model to a single default_metric_sort string (and remove the now-unused SortCriterion model). This feature has no users today, so the change is safe to make with no migration.
- Regenerate the OpenAPI spec and the platform SDKs (web + python/Stainless) for the new shape.

https://github.com/user-attachments/assets/4d7814b4-19a3-49fe-bcc4-c03135c1b397

## Summary by CodeRabbit

* **New Features**
  * Added experiment group editing for **default sort** (description + default ordering).
  * Added a metric/evaluator-based default sort selector to create and edit flows.
  * Experiment group tables now initialize from the group’s stored **default_sort**.
* **Bug Fixes**
  * When experiments `sort` is omitted, ordering is now consistently **pinned first, then newest** (`-created_at`).
  * Validation now uses the single **sort-parameter string** format; legacy/unsupported values are coerced to `-created_at`.
* **Documentation**
  * Updated OpenAPI specs to reflect `default_sort` as a string (and removed the old structured sort criterion schema).


nakolean (approved):
FE code LGTM

services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -120,15 +120,15 @@ async def create_experiment_group(
     body: ExperimentGroupRequest,
     entity_client: EntityClientDep,
 ) -> ExperimentGroupResponse:
-    _validate_default_sort(body.default_sort)
+    _validate_default_metric_sort(body.default_metric_sort)
```
shanaiabuggy: remove concept of metric sort and just have default sort and any sortable experiment field can be set here

services/intake/src/nmp/intake/entities/experiments.py
```diff
@@ -41,14 +32,24 @@ class ExperimentGroup(EntityBase):
...
         ),
     )
+
+    @field_validator("default_sort", mode="before")
+    @classmethod
+    def _default_sort_fallback(cls, value: Any) -> Any:
+        """Upgrade legacy rows to the default. Groups persisted before this field was a non-null string
+        stored ``default_sort`` as ``null`` (or, earlier, a ``SortCriterion`` list); the entity store is
+        schema-on-read, so coerce anything that isn't a usable string to the default on read."""
+        return value if isinstance(value, str) else "-created_at"
```
shanaiabuggy, line 51: @BrianNewsom check this out!

BrianNewsom (approved):

PR #579, opened 2026-07-06
feat(client): Add exist_ok ConflictResolver to auto-retry on 409s

matthewgrossman:
## Summary

Restores `exist_ok=True` on create endpoints (starting with `create_fileset`) so a caller gets the existing entity back on conflict, without a manual follow-up GET:

```python
fs = client.create_fileset(body=CreateFilesetRequest(name="myname"), exist_ok=True)
# -> FilesetOutput, whether it was just created OR already existed
```

Fixes [AIRCORE-866](https://linear.app/nvidia/issue/AIRCORE-866/fix-exist-ok-semantics-for-create-endpoints).

## Background

`exist_ok` was previously removed (commit `1a78606e48`) because it was **inert**. The old implementation swallowed the `ConflictError` and parsed the 409 response body as the entity — i.e. it assumed the server returns the entity on 409 (Option A). The real server returns `{"detail": "... already exists"}`, so `model_validate(raw.json())` would have failed on a genuine conflict; it only ever "worked" in tests that mocked entity-shaped 409 bodies.

Rather than change every server endpoint to return the entity on 409 (non-standard — RFC 9110 treats the 409 body as diagnostic), this handles `exist_ok` **entirely client-side**: on a 409, the client replays a linked GET and returns that entity. This is exactly what `FilesetsSubResource.create` did by hand via try/except + GET — lifted into the client so every consumer gets it declaratively.

## How it works

A create endpoint declares a **`get_on_conflict` resolver** — a named, typed function that builds the retrieve request:

```python
def _get_fileset_on_conflict(body: CreateFilesetRequest, workspace: str | None) -> PreparedRequest[FilesetOutput]:
    return get_fileset(name=body.name, workspace=workspace)

@post("/apis/files/v2/workspaces/{workspace}/filesets", get_on_conflict=_get_fileset_on_conflict)
@abstractmethod
def create_fileset(
    *, workspace: str | None = None, body: CreateFilesetRequest, exist_ok: bool = False
) -> FilesetOutput: ...
```

- The decorator calls the resolver at **request-build time** (while the live `body` model still exists, before it's serialized to bytes) and stashes the resulting GET on `PreparedRequest.on_conflict_get`.
- `send()` (sync + async): on `409 + exist_ok`, it replays `on_conflict_get` and returns that entity instead of raising `ConflictError`. Without `exist_ok`, a 409 still raises as before.
- `exist_ok` set on an endpoint with no resolver raises a clear `ValueError` (not a parse crash).
- `NemoResponse.body` stays non-nullable (`T`) — no `T | None` tax on other callers.

Correctness is enforced by the type checker: the resolver returns `get_fileset(name=..., workspace=...)`, a real typed endpoint call, so a wrong kwarg/type is caught statically at the call site.

## Changes

| File | Change |
| --- | --- |
| `client/types.py` | `ConflictResolver` Protocol + `PreparedRequest.on_conflict_get` field |
| `client/endpoint.py` | `@post(..., get_on_conflict=...)`; resolver invoked at request-build time |
| `client/client.py` | `_should_resolve_conflict` helper; sync + async `send()` replay the linked GET on 409 + `exist_ok` |
| `files/endpoints.py` | `create_fileset` gains `exist_ok` + `get_on_conflict=_get_fileset_on_conflict` |
| `filesets/resources.py` | Drop the try/except + GET workaround in sync/async `create`; pass `exist_ok` through |
| `tests/client/test_client_options.py` | Auto-GET tests using a **real** `{"detail": ...}` 409 body + follow-up GET |

## Testing

- New tests cover: resolver wiring, 409+`exist_ok` returns the entity via auto-GET, 409 without `exist_ok` raises `ConflictError`, non-409 passthrough, missing-resolver `ValueError`, GET-404-after-409 surfaced, the flat `client.create_item(exist_ok=True)` method path, and async twins.
- `packages/nemo_platform_plugin/tests/client/` — 64 passed; full plugin suite — 871 passed.
- `ty` and `ruff` clean on changed code.

## Out of scope

- Server-side idempotency (200 + entity keyed on `Idempotency-Key`, or `PUT` create-or-replace) — the standards-aligned path if we ever change the server; deferred.
- Returning the entity in the 409 body (Option A) — rejected as non-standard.

## Summary by CodeRabbit

* **New Features**
  * Added conflict-aware create behavior for HTTP 409 by replaying a prebuilt read request when `exist_ok` is enabled (sync and async).
  * Extended fileset and example plugin item creation to support `exist_ok` conflict replay.

* **Bug Fixes**
  * `exist_ok=True` now returns the existing resource on 409 instead of requiring manual conflict handling.
  * If `exist_ok` is enabled without configuring conflict replay, a clear error is raised.

* **Tests**
  * Added coverage for resolver wiring, replay behavior, header preservation, and error cases across client and async client modes.

### DIFF
--- a/packages/filesets/src/filesets/resources.py
+++ b/packages/filesets/src/filesets/resources.py
@@ -255,15 +255,9 @@ class FilesetsSubResource:
             custom_fields=custom_fields or {},
             cache=cache,
         )
-        # The server returns an error body on 409, not the entity, so
-        # exist_ok is handled here with a follow-up GET rather than at
-        # the endpoint/client level.
-        try:
-            return self._client.create_fileset(workspace=workspace, body=body).data()
-        except nemo_platform.APIStatusError as e:
-            if e.status_code == 409 and exist_ok:
-                return self.retrieve(name=name, workspace=workspace)
-            raise
+        # exist_ok is handled by the client: on a 409 it replays the linked GET
+        # (get_fileset) and returns the existing entity. See AIRCORE-866.
+        return self._client.create_fileset(workspace=workspace, body=body, exist_ok=exist_ok).data()
 
     def retrieve(self, name: str, *, workspace: str | None = None) -> FilesetOutput:
         return self._client.get_fileset(workspace=workspace, name=name).data()
@@ -360,15 +354,9 @@ class AsyncFilesetsSubResource:
             custom_fields=custom_fields or {},
             cache=cache,
         )
-        # The server returns an error body on 409, not the entity, so
-        # exist_ok is handled here with a follow-up GET rather than at
-        # the endpoint/client level.
-        try:
-            return (await self._client.create_fileset(workspace=workspace, body=body)).data()
-        except nemo_platform.APIStatusError as e:
-            if e.status_code == 409 and exist_ok:
-                return await self.retrieve(name=name, workspace=workspace)
-            raise
+        # exist_ok is handled by the client: on a 409 it replays the linked GET
+        # (get_fileset) and returns the existing entity. See AIRCORE-866.
+        return (await self._client.create_fileset(workspace=workspace, body=body, exist_ok=exist_ok)).data()
 
     async def retrieve(self, name: str, *, workspace: str | None = None) -> FilesetOutput:
         return (await self._client.get_fileset(workspace=workspace, name=name)).data()
--- a/packages/filesets/src/filesets/resources.py
+++ b/packages/filesets/src/filesets/resources.py
@@ -255,8 +255,6 @@ class FilesetsSubResource:
             custom_fields=custom_fields or {},
             cache=cache,
         )
-        # exist_ok is handled by the client: on a 409 it replays the linked GET
-        # (get_fileset) and returns the existing entity. See AIRCORE-866.
         return self._client.create_fileset(workspace=workspace, body=body, exist_ok=exist_ok).data()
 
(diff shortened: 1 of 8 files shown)


matthewgrossman:
Addressed the review feedback in 045ee87 (all four items):

**Actionable — dropped headers on conflict replay** ✅
The recursive `send()` for the conflict-GET now passes `headers=headers` through (both sync and async paths), so caller-supplied per-request headers survive the replay. Added `test_caller_headers_are_preserved_on_conflict_replay`.

**Fail fast: `exist_ok` requires a resolver** ✅
Moved the check to decoration time in `_make_endpoint` — an endpoint declaring `exist_ok` without `get_on_conflict` now raises `TypeError` at import, not on the first production 409. Used `TypeError` to match the existing decoration-time validation style (`_validate_params`). The runtime `ValueError` in `_should_resolve_conflict` stays as a defensive backstop. Added `test_exist_ok_without_resolver_raises_at_decoration_time`; the former runtime-error test was replaced accordingly.

**`method` param shadowing in `_resp`** ✅
Renamed to `http_method`.

**Async flat-method replay coverage** ✅
Added `TestAsyncExistOkViaMethod` mirroring the sync `method()`-wrapper test on `AsyncNemoClient`.

All 66 client tests pass; `ty`/`ruff` clean on changed code.

maxdubrinsky (approved):

PR #581, opened 2026-07-06
test(deployments): add kind/kubeconfig integration tests for k8s backend (AIRCORE-757 phase 7)

tylersbray:
## Summary

- Adds kind/kubeconfig-backed integration tests for the k8s deployments backend (AIRCORE-757 Phase 7), mirroring the existing Docker integration-test layout: PVC lifecycle, Job success, ConfigMap mount round-trip, Deployment+Service readiness, and label-gated delete rejection.
- Adds a reconciler-level integration test (`test_reconcile_k8s.py`) exercising a puller→server prerequisite chain against a real cluster. PVC-mount gating is intentionally not covered here — kind's default `local-path` StorageClass uses `WaitForFirstConsumer` binding, which deadlocks against `DeploymentReconciler`'s requirement that mounted volumes already be `BOUND` before pod creation (documented in the test's docstring and in the README).
- Adds a `kubeconfig_availability.py` skip gate (mirrors `docker_availability.py`) so these tests skip cleanly in CI when no cluster is reachable — no new CI job required for this phase's closure.
- Extends `conftest.py`'s xdist grouping so k8s and docker integration tests run in separate serialized groups (they share cluster/daemon state within their own group but don't collide with each other).
- Documents local validation (against `kind` or dev-blue) in the plugin README's new "Cluster validation" section.

### Branched ahead of Phase 6

This branches from `origin/main` rather than waiting for Phase 6 (#575, RBAC + Helm chart integration) to merge. Phase 6's RBAC change only matters when the real `nmp-core` controller pod acts through its restricted ServiceAccount; these integration tests instantiate `K8sDeploymentBackend` directly in pytest, authenticated with the developer's own kubeconfig — not that constrained ServiceAccount — so there's no functional dependency. #575 was open/mergeable with green CI, blocked only on reviewer availability. Will rebase onto `main` once #575 lands (both touch the plugin README, but only via non-conflicting appends).

## Test plan

- [x] `uv run --frozen pytest plugins/nemo-deployments/tests/unit -q` — 251 passed
- [x] `uv run --frozen pytest plugins/nemo-deployments/tests/integration -v` against a local `kind` cluster — 10 passed (5 docker + 5 k8s, including the new tests)
- [x] `uv run pre-commit run -a` on changed files — ruff, ty, copyright headers, merge-conflict checks all pass
- [x] Bugbot review pass — found and fixed a prerequisite-chain race (puller loop broke on `status == SUCCEEDED` without also waiting on `exit_code`, which the `succeeded` prerequisite condition requires)
- [ ] CI: confirm these tests skip cleanly (no kubeconfig in the runner) rather than failing

## Summary by CodeRabbit

* **New Features**
  * Added “Cluster validation” documentation for running Kubernetes-backed integration checks (local kind and remote clusters), including kubeconfig/reachability behavior and namespace configuration.
  * Expanded Kubernetes integration coverage for PVC lifecycle, deployment job outcomes, ConfigMap mount verification, readiness/endpoint checks, and protection against deleting foreign resources.
  * Added an end-to-end reconciliation test validating prerequisite gating between two Kubernetes deployments.
* **Bug Fixes**
  * Improved integration-test skipping and parallel grouping so Kubernetes tests run only when a real cluster is reachable.
* **Documentation**
  * Documented expected coverage gaps and assumptions for PVC/gating edge cases.

### DIFF
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -137,11 +137,11 @@ export NMP_K8S_ITEST_NAMESPACE=<your-namespace>  # defaults to "default"
 uv run pytest plugins/nemo-deployments/tests/integration/backends/k8s -v
 ```
 
-Tests authenticate with your ambient kubeconfig identity, not the restricted
-`nmp-core` controller ServiceAccount, so they validate backend behavior against
-a real API server but do **not** exercise the RBAC scope granted by the deploy
-chart's `controller-role.yaml`. That still needs a manual smoke test of the
-deployed platform pod.
+Tests authenticate with whatever kubeconfig identity is currently active in
+your shell, not the restricted `nmp-core` controller ServiceAccount, so they
+validate backend behavior against a real API server but do **not** exercise
+the RBAC scope granted by the deploy chart's `controller-role.yaml`. That
+still needs a manual smoke test of the deployed platform pod.
 
 `test_reconcile_k8s.py` covers prerequisite gating (one deployment blocking on
 another) but not PVC-mount gating: kind's default `local-path` StorageClass uses
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -112,6 +112,15 @@ exercise the k8s backend and reconciler against a real cluster. Like the Docker
 integration tests above, they skip automatically (CI stays green) when no
 cluster is reachable.
 
+**Prerequisites:**
+
+- `kind` and `kubectl` on your `PATH` (for the local run) or an existing
+  cluster context (for the dev-blue run).
+- `uv` for running the test suite.
+- A kubeconfig pointing at a reachable cluster — either the one `kind create
+  cluster` sets automatically, or your own context via `kubectl config
+  use-context`.
+
 **Local run against kind** (no dev-blue access needed):
 
 ```bash
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -104,3 +104,41 @@ omit the per-container `restart_policy` on init containers and run that
 container as a regular main container instead — the compiler does not
 fall back to legacy sidecar emulation (emptyDir readiness files, etc.)
 automatically.
+
+## Cluster validation
+
+`tests/integration/backends/k8s/` and `tests/integration/test_reconcile_k8s.py`
+exercise the k8s backend and reconciler against a real cluster. Like the Docker
+integration tests above, they skip automatically (CI stays green) when no
+cluster is reachable.
+
+**Local run against kind** (no dev-blue access needed):
+
+```bash
+kind create cluster  # or reuse an existing one; sets your kubeconfig context
+uv run pytest plugins/nemo-deployments/tests/integration/backends/k8s -v
+uv run pytest plugins/nemo-deployments/tests/integration/test_reconcile_k8s.py -v
+```
+
+**Run against dev-blue** (uses your namespace and your own RBAC there):
+
+```bash
+kubectl config use-context <dev-blue>
+export NMP_K8S_ITEST_NAMESPACE=<your-namespace>  # defaults to "default"
+uv run pytest plugins/nemo-deployments/tests/integration/backends/k8s -v
+```
+
+Tests authenticate with your ambient kubeconfig identity, not the restricted
+`nmp-core` controller ServiceAccount, so they validate backend behavior against
+a real API server but do **not** exercise the RBAC scope granted by the deploy
+chart's `controller-role.yaml`. That still needs a manual smoke test of the
+deployed platform pod.
+
+`test_reconcile_k8s.py` covers prerequisite gating (one deployment blocking on
+another) but not PVC-mount gating: kind's default `local-path` StorageClass uses
+`WaitForFirstConsumer` binding, so an unconsumed PVC never reaches `BOUND`, and
+`DeploymentReconciler` requires a mounted volume to already be `BOUND` before it
+creates the pod that would consume it — a chicken-and-egg specific to
+`WaitForFirstConsumer` storage classes that doesn't arise on `Immediate`-binding
+classes (the common case outside kind). The standalone PVC lifecycle scenario in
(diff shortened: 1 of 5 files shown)


tylersbray:
Addressed CodeRabbit's feedback in 72a53c5:

**Actionable comments (all fixed):**
- `_wait_for_status` now calls `pytest.fail(...)` on timeout instead of silently returning, so a stuck poll fails with a clear "did not reach {statuses} within {attempts} attempts" message.
- `test_pvc_lifecycle`'s terminal assertion moved outside the `finally` block so a cleanup failure can't mask the real assertion error.
- `test_reconcile_k8s.py`'s teardown now deletes `puller` and `server` concurrently via `asyncio.gather(..., return_exceptions=True)` so a failure deleting one doesn't leak the other.

**Nitpicks (fixed):**
- README: added a prerequisites block (kind/kubectl/uv/kubeconfig) ahead of the run commands in the new "Cluster validation" section.
- `test_reconcile_k8s.py`'s `get_side_effect` now raises `KeyError` for any Deployment name other than `"puller"`/`"server"` instead of silently falling back to `server_dep`.

**Nitpicks (left as-is, noted for reviewers):**
- `from __future__ import annotations` — this matches the established convention already used in every file in this test suite, including the pre-existing `docker_availability.py` and `test_reconcile_docker.py`. Removing it only from the new k8s files would create inconsistency within the same directory.
- Import-time cluster probe in `kubeconfig_availability.py` — this mirrors the identical, already-accepted pattern in `docker_availability.py` (which does the same real network call via `docker.from_env().ping()` at import time). A lazy-check refactor would need to touch both modules together and is out of scope for this test-parity PR.

Verified after the fix: 251 unit tests + all 10 integration tests (5 docker + 5 k8s) pass locally against a `kind` cluster, and `pre-commit run` is clean.

plugins/nemo-deployments/tests/integration/backends/k8s/test_k8s_backend.py
```diff
@@ -0,0 +1,293 @@
...
+    created = await k8s_backend.create_volume(
+        workspace="itest-pvc",
+        name="data",
+        size="1Gi",
+        access_modes=["ReadWriteOnce"],
+        backend_config={},
+    )
+    try:
+        # Unconsumed PVCs stay Pending under kind's default WaitForFirstConsumer
+        # storage class, so both Pending and Bound are valid terminal states here.
```
mckornfield, line 149: lol this is weird

plugins/nemo-deployments/tests/integration/kubeconfig_availability.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared Kubernetes cluster availability check for integration tests."""
+
+from __future__ import annotations
+
+import pytest
+
+try:
```
mckornfield, line 10: lol this does feel weird without being in a method, but 🤷 probably the way to do it. Or KUBECONFIG_AVAILABLE could be something returned by a fn?

plugins/nemo-deployments/tests/integration/test_reconcile_k8s.py
```diff
@@ -0,0 +1,147 @@
...
+        name="server-cfg",
+        workspace="itest",
+        restart_policy="Always",  # ty: ignore[unknown-argument]
+        containers=[
+            Container(name="server", image="nginx:alpine", ports=[ContainerPort(name="http", containerPort=80)])
+        ],
+    )
+
+    config_cache = {
+        ("itest", "puller-cfg"): puller_cfg,
```
mckornfield, line 88: lol `itest`?

plugins/nemo-deployments/README.md
```diff
@@ -104,3 +104,50 @@ omit the per-container `restart_policy` on init containers and run that
...
+
+**Run against dev-blue** (uses your namespace and your own RBAC there):
+
+```bash
+kubectl config use-context <dev-blue>
+export NMP_K8S_ITEST_NAMESPACE=<your-namespace>  # defaults to "default"
+uv run pytest plugins/nemo-deployments/tests/integration/backends/k8s -v
+```
+
+Tests authenticate with your ambient kubeconfig identity, not the restricted
```
mckornfield: lol `ambient`

mckornfield (approved):
some funny wording things, yay integration and not e2e tests.

plugins/nemo-deployments/README.md
tylersbray: Fair, "ambient" was jargon-y. Reworded to "whatever kubeconfig identity is currently active in your shell" in cd2a3413cc.

plugins/nemo-deployments/tests/integration/kubeconfig_availability.py
tylersbray: Ha, fair. Left as-is: it mirrors the exact top-level try/except shape `docker_availability.py` already uses for the same skip-gate pattern, so a bare module attribute + `from fn import x` beats introducing an inconsistent lazy-fn variant just for k8s. CodeRabbit flagged the same thing on this PR and we kept it consistent for that reason.

plugins/nemo-deployments/tests/integration/test_reconcile_k8s.py
tylersbray: Haha yeah, shorthand for "integration test" — matches the workspace name `test_docker_backend.py` already uses for the same kind of scratch workspace, so kept it for consistency across backends.

plugins/nemo-deployments/tests/integration/backends/k8s/test_k8s_backend.py
tylersbray: Yeah, it is a little weird — that's the WaitForFirstConsumer chicken-and-egg the comment right above tries to explain: kind's default local-path StorageClass never binds an unconsumed PVC, so accepting PENDING here (alongside BOUND, for non-kind clusters) is intentional rather than an oversight. Left as-is.

PR #582, opened 2026-07-06
fix(agents): Honor shared CLI context for base URL and auth

tylersbray:
## Summary

The `nemo agents` command group resolved its own base URL (only `--base-url`/`NEMO_BASE_URL`, else localhost) and attached **no auth token**. On any real remote this means:

- **P0 #1 — wrong target, silently.** It ignored `nemo config set --base-url` and `NMP_BASE_URL` (the vars the rest of the CLI honors), so `deploy`/`invoke`/`list` silently hit `localhost:8080` — and reported success against the wrong platform.
- **P0 #2 — 401/403 on secured clusters.** With no `Authorization` header, every agents call is rejected on an auth-enabled cluster, even after `nemo auth login`.

Both gate the "build local, ship remote" flow. Reported in the local→remote test run (AIRCORE-885, follow-up to the AIRCORE-867 runbook).

## What changed

- **New shared module** `plugins/nemo-agents/src/nemo_agents_plugin/cli_context.py` — `resolve_base_url` / `resolve_context_headers` read the same `CLIContext` (`typer.Context.obj`) as the rest of the CLI. It lives outside `cli.py` to avoid the `cli.py` ↔ `usage/cli.py` import cycle.
- **Base URL precedence:** `--base-url`/`NEMO_BASE_URL` → `nemo config`/`NMP_BASE_URL` → localhost. Resolved target echoed to **stderr** (`Targeting <url>`) so a mis-point is visible; **stdout stays clean JSON**.
- **Auth:** bearer token from `nemo auth login` attached to the platform `httpx` calls (`_api_request`, gateway invoke) and to the `usage show` SDK client. Falls back to unauthenticated when no token is configured (local dev unchanged).
- **Scope:** covers the platform commands *and* `nemo agents usage show` (builds its own SDK client; resolution runs only in the fileset branch that actually contacts the platform).
- The previously misleading `nemo config view` error hint is now accurate.

Not regenerated: the CLI/SDK generator only owns `packages/nemo_platform_ext/.../cli/commands/api/`; the agents plugin CLI is hand-written, and no API routes/models changed.

## Test plan

- [x] New `test_cli_context_resolution.py`: base-URL precedence (flag/env/config/default), stderr-only banner, auth-header attach, gateway invoke, no-auth-without-context.
- [x] New usage test: `usage show <fileset>` builds the SDK client with the context base URL + `Authorization: Bearer`.
- [x] Updated JSON-output tests to read `result.stdout` (banner is on stderr).
- [x] `ruff check`, `ruff format --check`, `ty`: clean.
- [x] No regressions: full agents unit suite failure set byte-identical to `origin/main` (same pre-existing failures).

## Summary by CodeRabbit

* **New Features**
  * CLI commands now use a shared, configurable platform base URL (with `--base-url` defaulting from config/env) instead of a fixed local default.
  * Platform-related requests can now include resolved context headers (including authentication when available).
  * `nemo agents usage show` now follows the same base-URL and context behavior when platform downloads are needed.
* **Bug Fixes**
  * Improved CLI output routing so status/targeting banners go to stderr, keeping JSON on stdout clean.
* **Tests**
  * Added/updated unit tests for base-URL precedence, header attachment, and stdout/stderr behavior.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
@@ -45,6 +45,15 @@ from typing import Any, ClassVar, Literal, Optional, cast
 import httpx
 import typer
 import yaml
+from nemo_agents_plugin.cli_context import (
+    DEFAULT_BASE_URL as _DEFAULT_BASE_URL,
+)
+from nemo_agents_plugin.cli_context import (
+    resolve_base_url as _resolve_base_url,
+)
+from nemo_agents_plugin.cli_context import (
+    resolve_context_headers as _resolve_context_headers,
+)
 from nemo_agents_plugin.leaderboard.cli import register_leaderboard_commands
 from nemo_agents_plugin.usage.cli import register_usage_commands
 from nemo_platform.cli.core.formatters import Column, format_output
@@ -54,7 +63,6 @@ from nemo_platform_plugin.cli_progress import request_progress
 
 logger = logging.getLogger(__name__)
 
-_DEFAULT_BASE_URL = "http://localhost:8080"
 _DEFAULT_WORKSPACE = "default"
 _LIST_OUTPUT_FORMAT = Literal["table", "json", "yaml", "csv", "markdown", "raw"]
 _AGENT_LIST_COLUMNS = [
@@ -145,7 +153,16 @@ def _register_local_commands(app: typer.Typer) -> None:
             help="Name of a specific deployment to invoke (platform required).",
         ),
         workspace: str = typer.Option(_DEFAULT_WORKSPACE, "--workspace", "-w"),
-        base_url: str = typer.Option(_DEFAULT_BASE_URL, "--base-url", envvar="NEMO_BASE_URL"),
+        base_url: Optional[str] = typer.Option(
+            None,
+            "--base-url",
+            envvar="NEMO_BASE_URL",
+            help=(
+                "Platform base URL. Defaults to the base URL from the shared CLI "
+                "config (`nemo config set --base-url`) / NMP_BASE_URL env var, then "
+                "http://localhost:8080."
+            ),
+        ),
         timeout: float = typer.Option(
             300,
             "--timeout",
@@ -160,6 +177,7 @@ def _register_local_commands(app: typer.Typer) -> None:
         ),
     ) -> None:
         """Invoke an agent — locally (with --agent-config) or via the platform (with --agent or --agent-deployment)."""
+        base_url = _resolve_base_url(base_url)
         if agent_config:
             _local_invoke(agent_config, input, input_file, workspace=workspace, base_url=base_url)
         elif agent or agent_deployment:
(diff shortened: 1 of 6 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -145,7 +153,16 @@ def invoke(
             help="Name of a specific deployment to invoke (platform required).",
         ),
         workspace: str = typer.Option(_DEFAULT_WORKSPACE, "--workspace", "-w"),
-        base_url: str = typer.Option(_DEFAULT_BASE_URL, "--base-url", envvar="NEMO_BASE_URL"),
+        base_url: Optional[str] = typer.Option(
+            None,
+            "--base-url",
+            envvar="NEMO_BASE_URL",
+            help=(
+                "Platform base URL. Defaults to the base URL from the shared CLI "
```
mckornfield: might show better as a numbered/bulleted list

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -628,9 +646,19 @@ def create(
         ),
         description: str = typer.Option("", "--description"),
         workspace: str = typer.Option(_DEFAULT_WORKSPACE, "--workspace", "-w"),
-        base_url: str = typer.Option(_DEFAULT_BASE_URL, "--base-url", envvar="NEMO_BASE_URL"),
+        base_url: Optional[str] = typer.Option(
+            None,
+            "--base-url",
+            envvar="NEMO_BASE_URL",
+            help=(
+                "Platform base URL. Defaults to the base URL from the shared CLI "
```
mckornfield: same as above, and might be good as a constant since it'll be repeated/a bit clunky looking

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -950,7 +1068,16 @@ def deployments_wait(
         timeout: int = typer.Option(300, "--timeout", "-t", help="Maximum seconds to wait."),
         interval: float = typer.Option(2.0, "--interval", help="Poll interval in seconds."),
         workspace: str = typer.Option(_DEFAULT_WORKSPACE, "--workspace", "-w"),
-        base_url: str = typer.Option(_DEFAULT_BASE_URL, "--base-url", envvar="NEMO_BASE_URL"),
+        base_url: Optional[str] = typer.Option(
```
mckornfield: also after seeing this enough times, is there a decent way to share a typer option? maybe as a constant as well? or gotta be inline?

mckornfield (approved):
yay centralized config

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
tylersbray: Done in 0156e8cc0b — reformatted the help as a numbered resolution-order list: `(1) --base-url/NEMO_BASE_URL; (2) nemo config/NMP_BASE_URL; (3) localhost`.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
tylersbray: Agreed — pulled the help text into a `BASE_URL_HELP` constant in `cli_context.py` so it's defined once.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
tylersbray: Yep — there's a clean way via `Annotated`. Added a reusable `BaseUrlOption = Annotated[Optional[str], typer.Option("--base-url", envvar="NEMO_BASE_URL", help=BASE_URL_HELP)]` in `cli_context.py`; every command is now just `base_url: BaseUrlOption = None`. Collapsed all 12 inline copies here plus the one in `usage show` (net -98 lines). No behavior change.

PR #585, opened 2026-07-06
feat(client): Add new Jobs NemoClient

matthewgrossman:
## Summary

Migrates the Jobs service and its production consumers from the Stainless-generated `sdk.jobs.*` surface to the typed `NemoClient` HTTP client.

This PR is **AIRCORE-874**, part of **AIRCORE-827**. It follows the typed-client migration pattern documented in [`MIGRATION.md`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/packages/nemo_platform_plugin/src/nemo_platform_plugin/client/MIGRATION.md).

### Typed Jobs client

- Adds synchronous `JobsClient` and asynchronous `AsyncJobsClient` implementations for all 21 shipped Jobs endpoints.
- Covers job CRUD and lifecycle actions, status and status details, logs, execution profiles, results and binary downloads, steps, and tasks.
- Intentionally omits `rerun_job`: the `/rerun` route is test-only and is not mounted by the released service.
- Uses typed request models and typed response wrappers, including unions for backend execution profiles.

### Jobs wire types

- Moves pure Pydantic Jobs wire models into `nemo_platform_plugin.jobs` as shared leaf types.
- Keeps server-only behavior in thin service subclasses and re-exports. For example, Kubernetes models add `to_k8s()` only in the service, avoiding `kubernetes` and `docker` dependencies in the plugin package.
- Preserves the existing OpenAPI contract. No Stainless SDK regeneration is required.

### Consumer migration

- Migrates Jobs application and plugin consumers through `client_from_platform(..., JobsClient)` or `AsyncJobsClient`.
- Migrates the full Jobs controller subsystem: scheduler, reconciler, diagnostics, subprocess, Docker, Kubernetes, and Volcano backends.
- Migrates `api_factory`, result management, customization progress reporting, Safe Synthesizer, Data Designer helpers, model APIs, entity workspace cleanup, quickstart, and `nmp_testing` helpers.
- Adds `NemoHTTPError` handling alongside the existing Stainless error handling so typed-client service-to-service failures retain the expected HTTP response behavior.
- Preserves deep-object step filtering using `filter[status]`, matching the Jobs route parser rather than sending an ignored flat `filter=` value.

## Pagination and shared-client changes

Jobs logs use cursor pagination while Jobs and steps use the platform's standard offset pagination. This PR extends the shared client so both are first-class instead of adding a Jobs-specific page method:

- Endpoint return types use `Paginated[Item, Strategy]`, with offset pagination as the default and `CursorPagination` for `list_job_logs`.
- `.page()` returns typed items plus strategy-specific metadata; iteration and `.items()` auto-fetch subsequent pages in both sync and async clients.
- Offset metadata is validated against the complete service contract (`page`, `page_size`, `current_page_size`, `total_pages`, and `total_results`).
- Cursor metadata exposes `total`, `next_page`, and `prev_page`, and sends subsequent cursors as `page_cursor`.
- Existing offset-pagination consumers remain covered by regression and typing tests.

The shared client also now:

- Validates arbitrary endpoint response annotations with cached Pydantic `TypeAdapter`s, including lists and unions.
- Normalizes response-contract failures as `NemoResponseValidationError`.
- Uses sync and async retry paths with `time.sleep` and `asyncio.sleep`, respectively. The default retryable responses are `429`, `502`, `503`, and `504`, plus transport errors.
- Supports Python 3.11 pagination metadata validation by using `typing_extensions.TypedDict`.

## Intentional service/client asymmetries

- Server execution-profile subclasses retain behavior and environment-driven defaults that do not belong in plugin wire models.
- Server list filters remain entity-store `Filter` subclasses; client endpoints accept encoded query parameters.
- Server result/task list wrappers can contain entity instances, while client wrappers contain response DTOs. These types are intentionally separate even though their names and wire shapes are similar.

## Out of scope

- Generated `nemo jobs` CLI commands: [AIRCORE-893](https://linear.app/nvidia/issue/AIRCORE-893)
- Jobs SDK DI/factory construction: [AIRCORE-883](https://linear.app/nvidia/issue/AIRCORE-883)
- Removing the remaining Stainless-generated `*Param` aliases from `api_factory`: [AIRCORE-922](https://linear.app/nvidia/issue/AIRCORE-922/remove-stainless-generated-jobs-param-aliases-from-api-factory)
- Live and service-contract tests that intentionally exercise the generated Stainless SDK directly.
- Raw HTTP callers and `sdk.customization.jobs.*`, which is a separate API namespace.

## Validation

GitHub CI is green on the final branch, including:

- Python unit, integration, e2e, and auth-idp tests
- Kind CPU smoke and e2e tests
- `nemo-platform` and `nemo-platform-plugin` wheel build/test matrices on Python 3.11 through 3.14
- Lint, type checking, OpenAPI/SDK consistency, DCO, CodeQL, and secrets scanning

Focused coverage includes prepared request shapes, sync/async response parsing, strict offset and cursor pagination, retry behavior, typed-page inference, and `AsyncJobsClient` against the in-memory Jobs ASGI application.


packages/nemo_platform_plugin/tests/test_jobs_filter.py
```diff
@@ -324,30 +336,28 @@ class TestForwardedFilterSurvivesSdkSerialization:
...
+        # The migrated factory forwards a single JSON-encoded ``filter`` string
+        # via ``query_params`` — no deep-object array serialization involved.
+        filter_value = (sdk.list_kwargs.get("query_params") or {}).get("filter")
+        assert filter_value is not None, "expected a forwarded filter param"
         # Decode just like core jobs make_filter_dep would: see leading ``{``,
         # route through parse_json_filter.
         from nemo_platform_plugin.api.filter import parse_json_filter
 
-        operation = parse_json_filter(filter_values[0])
+        operation = parse_json_filter(filter_value)
```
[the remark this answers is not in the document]
maxdubrinsky: Not actioning: `TestFilterRoundTrip` already drives the full stack through `TestClient(app)`. `_round_trip` asserts exactly one `filter` query param is present, then decodes and compares the JSON, so a serializer regression that dropped or mangled the filter would fail the existing test. A raw-string equality assert would be belt-and-suspenders only.

packages/nmp_customization_common/src/nmp/customization_common/training/progress.py
```diff
@@ -62,14 +65,17 @@ def update_task(
...
                 job=self._job_ctx.job_id,
                 step=self._job_ctx.step,
-                status=status,  # ty: ignore[invalid-argument-type]
-                status_details=status_details or {},
-                error_details=error_details or {},
+                body=PlatformJobTaskUpdate(
+                    status=status,
+                    status_details=status_details or {},
+                    error_details=error_details or {},
+                ),
```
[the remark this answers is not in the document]
maxdubrinsky: False positive: `status_details or {}` maps `None` to `{}`, and the server gate `if task_update.status_details:` treats `None` and `{}` identically. `PlatformJobStatus` is a `str` enum, so the literal status values coerce fine. No type or serialization issue, no change needed.

maxdubrinsky (approved):

plugins/nemo-data-designer/src/nemo_data_designer_plugin/testing/utils.py
```diff
@@ -431,16 +433,19 @@ def __init__(self) -> None:
                 workspace="default",
             ) as client_context,
         ):
-            job = client_context.sdk.jobs.create(
+            jobs_client = client_from_platform(client_context.sdk, JobsClient)
```
mikeknep, line 436: I suppose the passing tests in CI answer this question for me, but: does this change lead to any shenanigans regarding loss of the ASGI transport? I saw some issues in DD tests related to the same `client_from_platform` change for Files:
- First updated to use `client_from_platform` in [this PR](https://github.com/NVIDIA-NeMo/nemo-platform/pull/584)
- Ryan reverted that change in [this PR](https://github.com/NVIDIA-NeMo/nemo-platform/pull/629)
- I'm fixing a separate (?) failing test in [this PR](https://github.com/NVIDIA-NeMo/nemo-platform/pull/659)
😅

matthewgrossman: In Files I thought we made it continue to work with the ASGI transport, but maybe something got missed

matthewgrossman: FWIW I think making it work with the ASGI transport is essential for all our tests, so I don't know how everything wouldn't just straight up explode if that didn't work

matthewgrossman: checking w/ agent:
```
• create_test_client provides separate sync and async in-process HTTP clients targeting the same ASGI app. The sync NeMoPlatform uses SDKTestClientAdapter, which forwards requests through Starlette’s TestClient; the async AsyncNeMoPlatform uses httpx.AsyncClient with ASGITransport.

  client_from_platform preserves these underlying clients: JobsClient receives the sync adapter, while AsyncJobsClient receives the async ASGI client. It does not replace the transport or convert between sync and async. The Data Designer code uses the sync path, and its integration test passes through the real in-process Jobs service
```

PR #586, opened 2026-07-07
fix(docker,jobs): scope container cleanup to owner labels and preserve pause/cancel status

ironcommit:
## Docker: scope container cleanup and orphan reconciliation to owner labels

Add `models_docker_container_labels` config field to `DockerBackendConfig` so containers created by the models controller carry arbitrary labels. Integration test fixtures use this to stamp containers with test-run and test-worker identity labels, scoping teardown cleanup and orphan reconciliation to the owning pytest-xdist worker.

- New `_managed_container_labels()` helper on the creation reconciler centralises label construction for NIM, sidecar, and puller containers
- `cleanup_model_deployment_containers()` accepts optional owner labels and only removes matching containers
- Pass `ignore_removed=True` on Docker container list calls to prevent races when another worker removes a container during enumeration
- Diagnostic output now includes owner labels for easier debugging

## Jobs: preserve pause/cancel status when pods report task errors

When a Kubernetes job is pausing, paused, cancelling, or cancelled, pod-level task errors (e.g. SIGTERM exit code 137) no longer override the job status to ERROR. This prevents user-initiated pause/cancel from incorrectly surfacing as a failure.

## Tests: fix intake ATIF ingest timestamp

Replace the frozen `_BASE_TIME` (2026-01-15) with a dynamic timestamp 7 days in the past so ingested spans stay within ClickHouse's 90-day TTL and tests don't start failing as time passes.

## Summary by CodeRabbit

* **New Features**
  * Added configurable `model_labels` support for applying custom labels to model resources across Docker and Kubernetes backends.
  * Test runs now tag model resources with run-/worker-specific labels to enable scoped teardown and better traceability.
* **Bug Fixes**
  * Docker cleanup now stops/removes only containers that match the expected label set; orphan detection and port selection ignore containers already marked removed.
  * Kubernetes job reconciliation now preserves `PAUSED`/`CANCELLED` when task errors occur during terminal transitions.
* **Documentation**
  * Updated the configuration reference with the new Docker Models service label option.
* **Tests**
  * Added unit and regression test coverage for label scoping and reconciliation edge cases.

### DIFF
--- a/docs/set-up/config-reference.mdx
+++ b/docs/set-up/config-reference.mdx
@@ -493,6 +493,8 @@ models:
         model_puller_max_workers: 1
         # Number of retries when the puller fails with a transient error (IncompleteRead, ChunkedEncodingError, connection broken). Same volume is reused so partial downloads can be completed. | default: 3
         model_puller_retries: 3
+        # Additional labels copied onto model resources managed by the models controller.
+        model_labels: {}
         # BusyBox image repository used for helper containers (permissions/find/chown). | default: 'busybox'
         busybox_image: busybox
         # BusyBox image tag used for helper containers. | default: 'latest'
@@ -577,6 +579,8 @@ models:
         default_tolerations:
         # Default Kubernetes node selector for all NIM deployments. Can be overridden per-deployment via k8s_nim_operator_config. Example: {'node-type': 'gpu-node', 'zone': 'us-west1-a'}
         default_node_selector:
+        # Additional labels copied onto model resources managed by the models controller.
+        model_labels:
         # Default Kubernetes labels applied to NIMService and NIMCache resources and their child resources (e.g. pods). Merged with controller-managed labels; controller labels take precedence on conflict. Example: {'team': 'ml-platform', 'environment': 'prod'}
         default_labels:
         # Default Kubernetes annotations applied to NIMService and NIMCache resources and their child resources (e.g. pods, PVCs). Merged with controller-managed annotations; controller annotations take precedence on conflict. Example: {'prometheus.io/scrape': 'true'}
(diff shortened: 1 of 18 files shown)


maxdubrinsky (approved):

docs/set-up/config-reference.mdx
```diff
@@ -577,6 +579,8 @@ models:
         default_tolerations:
         # Default Kubernetes node selector for all NIM deployments. Can be overridden per-deployment via k8s_nim_operator_config. Example: {'node-type': 'gpu-node', 'zone': 'us-west1-a'}
         default_node_selector:
+        # Additional labels copied onto model resources managed by the models controller.
+        model_labels:
```
mckornfield, line 583: hmm why does this not have `{}` after? I guess it's null but looks weird

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/kubernetes_job.py
```diff
@@ -311,7 +311,26 @@ def create_step_update(self, step: PlatformJobStepWithContext, job: V1Job) -> Jo
             error_details["message"] = status_details.get("message", "Job encountered an error")
         status_details["events"] = self.get_kube_job_events(job)
         task_has_error = update_all_tasks(self._nmp_sdk, self._core_v1, self.namespace, step)
-        if task_has_error:
+        teardown_lifecycle_statuses = {
```
mckornfield, line 314: should this be a constant/could it be used anywhere else?

mckornfield (approved):

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/kubernetes_job.py
ironcommit: It's not used anywhere else at the moment.

docs/set-up/config-reference.mdx
ironcommit: yeah, does look wierd.  The first one does have it, but the second one follows the convention around it which is null.

PR #593, opened 2026-07-07
test(nemo-agents): add k8s specific tests

mmogallapalli:
## Scope

This MR builds on the earlier basic nemo-agents e2e MR, which added backend-agnostic coverage for agent CRUD, CRUD edge cases, pagination/filtering/sorting, cross-workspace isolation, and SDK resource methods.

This MR adds the currently supportable deployment-oriented e2e coverage for nemo-agents:

- creating a deployment for a missing agent returns 404
- creating a deployment for an existing agent succeeds
- deployment appears in list results
- deployment get returns expected agent/config/status fields
- deployment delete removes the deployment
- deleting an agent is blocked while an active deployment exists
- gateway invocation for a missing deployment returns 404
- a mock-backed agent deployment reaches `running` against minikube

Validated against minikube with locally built images via:

```bash
NMP_BASE_URL=http://localhost:30080 uv run --frozen pytest e2e/test_nemo_agents.py --run-e2e --no-cov -v
```

## Known Limitation

The remaining K8s-specific cases from AIRCORE-829 are not included in this MR:

- gateway route by deployment
- gateway route by agent
- deployment stdout logs
- `ExternalLog` behavior

Those require a nemo-agents deployment backed by an actual Kubernetes pod/service that the agents gateway can reach. The current nemo-agents runner is `in_memory`, which starts NAT as a subprocess and records a loopback endpoint. That supports basic deployment lifecycle/status coverage, but it does not provide a true pod-backed runtime for gateway/log validation.

## Follow-Up

The existing nemo-deployments plugin already provides the K8s substrate needed here: `DeploymentConfig`, K8s Deployment/Service creation, in-cluster endpoints, status reconciliation, and log access.

A follow-up implementation can add a nemo-agents runner backend that delegates runtime creation to the deployments plugin, maps deployment status/endpoints back onto `AgentDeployment`, and then enables the remaining gateway/log e2e cases.

Once that runner integration lands, we can close the loop on the remaining AIRCORE-829 tests.

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for agent deployments, including create, list, get, delete, and readiness checks.
  * Added validation for key error cases such as missing agents, missing deployments, and preventing agent deletion while deployments are active.
  * Added coverage for deployment startup and endpoint availability using a mocked model response.


matthewgrossman (approved):
lgtm!

PR #597, opened 2026-07-07
fix: alias redundant anonymizer run run, import logging fix

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Plugin commands that fail to load now appear as unavailable in the CLI, with a clear error message when invoked.
  * Job help text is now more consistent and generic across local run, submit, and schema-explanation commands.

* **Bug Fixes**
  * Improved handling for plugin import failures so one broken plugin no longer hides its command entirely.
  * Fixed help and command behavior for jobs named `run`, including support for both `run` and `run run` usage.

### DIFF
--- a/packages/nemo_platform_plugin/tests/test_commands.py
+++ b/packages/nemo_platform_plugin/tests/test_commands.py
@@ -74,6 +74,14 @@ class _FailJob(NemoJob):
         raise RuntimeError("job exploded")
 
 
+class _RunNamedJob(NemoJob):
+    name = "run"
+    description = "A job whose name collides with the local run verb."
+
+    def run(self, config: dict) -> dict:
+        return config
+
+
 runner = CliRunner()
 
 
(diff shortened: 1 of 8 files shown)


packages/nemo_platform_plugin/tests/test_commands.py
```diff
@@ -74,6 +74,14 @@ def run(self, config: dict) -> dict:
         raise RuntimeError("job exploded")
 
 
+class _RunNamedJob(NemoJob):
+    name = "run"
+    description = "A job whose name collides with the local run verb."
```
tylersbray, line 79: Well it's accurate.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/cli.py
```diff
@@ -33,6 +35,27 @@ def _root() -> None:
         app.command("validate")(validate_command)
         return app
 
+    def update_job_cli(self, job_cls: type[NemoJob], group: typer.Typer) -> None:
+        if job_cls.name != "run":
```
tylersbray, line 39: I'm not going to pretend I've never done this.  "if not broken then don't fix else <unique edge case logic here>"  Tag me on the PR for the next edge case that gets added here.

tylersbray (approved):
lgtm lgtm, thanks thanks.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/cli.py
mckornfield: lol seriously

packages/nemo_platform_plugin/tests/test_commands.py
mckornfield: lolol

PR #604, opened 2026-07-08
feat(evaluator-sdk): sandboxed containerized Fabric runner (AALGO-321)

SandyChapman:
## What

Adds `FabricContainerRuntime` (AALGO-321): runs NeMo Fabric + a coding agent inside a **per-task Docker sandbox** and maps the result into the existing `CandidateEvidence` contract, so the workspace-file, `run_verifier`, and trajectory metrics score container trials with no metric changes.

## Highlights

- **Provider-neutral sandbox seam** (`SandboxSpec`/`SandboxProvider`/`AsyncSandbox`) with a Docker provider that shells out to the `docker` CLI. Context/artifacts cross the boundary by file transfer (`docker cp`), so a Kubernetes/agent-sandbox provider can slot in behind the same seam later. `AsyncSandbox` tears down only its own sandbox; the shared provider's lifetime is the batch owner's (`run_tasks` disposes it once).
- **Opaque build-if-missing image** — content-addressed tag, no Dockerfile for callers to write. A remote registry is the intended cache behind the same `ensure_fabric_image` seam.
- **`nemo-relay` is now a hard dependency**: the trajectory profile is built from Relay's own typed config (via the shared `fabric/_common.py`), so Relay owns its schema instead of a hand-maintained dict that silently drifts. Verified `cp311-abi3` wheels exist for every platform the workspace locks.
- **Host/container dedup**: the shared helpers (failed-trial builder, output extraction, evidence-subdir naming, trajectory telemetry) live in `fabric/_common.py` so the two runtimes cannot drift apart.

## Scope

Exercises the runtime directly (hermes-sdk harness, validated live end-to-end); the plugin `_resolve_target` wiring is a follow-up. See `examples/fabric_container/` for a runnable end-to-end.

## Verification

Full `agent_eval` suite green (143 passed, 1 skipped — the gated live Docker test); ruff + pyright clean.

## Summary by CodeRabbit

* **New Features**
  * Added containerized evaluation support for NeMo Fabric tasks.
  * Added Docker-based sandbox execution with file transfer, environment variables, resource settings, networking, and lifecycle management.
  * Added automatic Fabric image provisioning and reusable evaluation evidence, including results, logs, workspace files, and execution traces.
  * Added an end-to-end example for running and inspecting a Fabric evaluation.

* **Bug Fixes**
  * Improved failure reporting for timeouts, invalid results, execution errors, and sandbox failures.

* **Documentation**
  * Documented the sandbox contract, lifecycle, file transfer behavior, and Docker usage.


SandyChapman:
Thanks @coderabbitai — addressed in the latest push (also rebased onto `main` and vendored into `sdk/python`):

**Fixed**
- **`container_runtime.py` — provider leak on early failure:** moved `ensure_fabric_image()` + `resolve_secrets()` inside the `try/finally`, so `provider.aclose()` runs even if provisioning or secret resolution raises.
- **`image.py` — unbounded docker subprocess calls:** `image_exists` now uses a 30s `timeout` and raises `FabricImageError` on `TimeoutExpired`; the build gets a generous 3600s ceiling. Added `test_image_exists_raises_on_daemon_timeout`.
- **`test_sandbox_docker_provider_live.py` — readiness check:** `docker info` now has a 15s timeout and skips (rather than hangs) an unresponsive daemon.

**Deferred to a hardening follow-up** (tracked; kept out of this spike to preserve the verified live e2e)
- **`sandbox.Dockerfile` runs as root** and the provider doesn't default to `--network none`: both are real defense-in-depth gaps for a sandbox that execs agent code. Getting non-root right requires reworking `/in`, `/out`, and the `-w` workdir ownership against the `docker cp` flow, which I want to land with network-egress scoping and re-verify against the live run — so it's a dedicated follow-up rather than an untested Dockerfile change here. The isolation note in `docker.py` already flags egress control as future work.
- **Image tag doesn't hash the Fabric source:** intentional — the tag is content-addressed on the *recipe* (Dockerfile + extras), and `ensure_fabric_image(force_build=True)` is the escape hatch for a dev iterating on a local NeMo-Fabric checkout. The intended evolution (a registry-backed cache behind the same seam) makes source-hashing moot.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/sandbox.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# (CARGO_MANIFEST_DIR/../../python/src/nemo_fabric/adapters); ship just that dir to the baked path so a
+# wheel-only image can resolve them. Depends on NeMo-Fabric's installed-adapter-discovery layout
+# (see image.py); swap to installing the top-level adapters/* packages once that lands on main.
+COPY --from=builder /src/python/src/nemo_fabric/adapters /src/python/src/nemo_fabric/adapters
+# The CLI's baked path is the literal `<fabric-core crate>/../../python/src/nemo_fabric/adapters`; the
+# kernel needs `/src/crates/fabric-core` to exist to walk the `..`, even though nothing lives there.
+RUN mkdir -p /src/crates/fabric-core
+ENV PATH=/opt/venv/bin:$PATH
+RUN python -c "from nemo_fabric import FabricClient" && fabric version
+WORKDIR /out
```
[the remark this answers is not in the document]
SandyChapman: Added a non-root `sandbox` user (uid 1000) with `/in` and `/out` pre-created and owned by it, then `USER sandbox`. Pre-owning those dirs is required because the runtime `mkdir`s the `/out/{workspace,relay,artifacts,logs}` subtree at exec time and a non-root process cannot create dirs under `/`. Smoke-tested the mechanics with an equivalent minimal image: `docker exec` runs as sandbox, `mkdir` of the `/out` subtree succeeds, and the `docker cp`ed (root-owned, mode 644) `/in/input.txt` is readable. Fixed in the latest push.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/image.py
```diff
@@ -0,0 +1,158 @@
...
+        try:
+            subprocess.run(
+                [docker_bin, "build", "--build-arg", f"EXTRAS={_extras_arg()}", "-t", tag, str(ctx)],
+                check=True,
+                env={**os.environ, "DOCKER_BUILDKIT": "1"},
+            )
+        except subprocess.CalledProcessError as exc:
+            raise FabricImageError(f"docker build failed for {tag}: {exc}") from exc
+    logger.info("Built Fabric image %s.", tag, extra=dict(tag=tag))
+    return tag
```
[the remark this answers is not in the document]
SandyChapman: Acknowledged, and real for local iteration on a NeMo-Fabric checkout. Leaving the tag recipe-only by design: it is content-addressed for a registry cache and the deterministic no-arg `fabric_image_tag()` is part of the contract (folding a source hash would break `ensure_fabric_image(...) == fabric_image_tag()` and make the tag vary per working-tree state). `force_build=True` is the escape hatch for editing Fabric source locally. If we want automatic source-drift busting later, the cheap path is folding the `fabric_repo` git HEAD sha + dirty flag into the tag rather than hashing the whole tree — filing as a follow-up rather than changing the caching model here.

packages/nemo_evaluator_sdk/examples/fabric_container/run_e2e.py
```diff
@@ -0,0 +1,81 @@
...
+# typed config. A plain mapping works too — that's what the plugin passes — if you'd rather not import it.
+from nemo_fabric import FabricConfig, HarnessConfig, MetadataConfig, RuntimeConfig  # ty: ignore[unresolved-import]
+
+
+async def main() -> int:
+    model = os.environ.get("FABRIC_MODEL", "nvidia/nemotron-3-nano-30b-a3b")
+
+    # Resolve the model credential into the process env so the default LocalSecretResolver can find it.
+    if not os.environ.get("NVIDIA_API_KEY"):
+        key_file = os.environ.get("NVIDIA_API_KEY_FILE")
```
ngoncharenko: allowing providing key in a file might be more prone to key leakage compared to env var if user add to the repo and forgets to delete. Do we want to have this option?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -0,0 +1,372 @@
...
+        if status != "succeeded":
+            return self._failed_trial(task, evidence_dir, _result_error(result_payload), extra_metadata=base_metadata)
+
+        return AgentEvalTrial(
+            id=f"{task.id}:fabric_container",
+            task_id=task.id,
+            status=AgentEvalTrialStatus.COMPLETED,
+            output=AgentOutput(
+                output_text=_common.extract_output_text(result_payload.get("output")),
+                response=result_payload if isinstance(result_payload, dict) else None,
```
ngoncharenko: The host Fabric runtime sets `AgentOutput.response` to `RunResult.output`, but this container path stores the entire normalized result envelope. Metrics expecting the shared response shape now see the payload under `sample.response.output` instead. Should we set this to `result_payload.get('output')`?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -0,0 +1,372 @@
...
+                metadata={**base_metadata, "evidence_dir": str(evidence_dir)},
+            ),
+            evidence=self._evidence(out_dir, result_path),
+            metadata={**base_metadata, "generated": True},
```
ngoncharenko: Successful container trials omit `agent_ok`. `AgentPhaseSuccessMetric` treats a missing value as `False`, so every successful container run receives a failed phase-success score. Should we match the host Fabric and Codex runtimes by setting `agent_ok=True` here and `False` on failed trials?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/sandbox/providers/docker.py
```diff
@@ -0,0 +1,262 @@
...
+            start_new_session=True,
+        )
+        try:
+            stdout_b, stderr_b = await asyncio.wait_for(proc.communicate(input=stdin), timeout=timeout_s)
+        except (asyncio.TimeoutError, TimeoutError) as exc:
+            with contextlib.suppress(ProcessLookupError):
+                os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+            with contextlib.suppress(Exception):
+                await proc.wait()
+            raise TimeoutError(f"docker command timed out after {timeout_s:g}s: {argv}") from exc
```
ngoncharenko: reported by codex: 

> The create argv contains resolved values such as `NVIDIA_API_KEY=...`, and this timeout message renders the raw argv. `FabricContainerRuntime` catches the resulting error and persists it in `error.json`, so a Docker startup timeout leaks credentials into evaluation artifacts. Redact argv before constructing exceptions, using the existing environment-command redaction pattern, and avoid secret values in process arguments where practical.

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/sandbox/providers/docker.py
SandyChapman: Fixed. `_run` now redacts argv via the existing `_redact_for_logging` before interpolating it into the TimeoutError, so no `-e KEY=<secret>` reaches `error.json`. The `create`/`exec` paths re-raise that message, so they're covered too. Regression test: `test_run_timeout_redacts_secrets_in_argv`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: Fixed. Set `response = result_payload.get("output")` (matching the host runtime's `RunResult.output`) instead of the whole normalized envelope, so metrics reading `sample.response` see one shape. Regression test: `test_success_response_is_output_payload_not_full_envelope`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: Fixed. Success trials now stamp `agent_ok=True`, and the shared `build_failed_trial` stamps `agent_ok=False` on failures — matching the host Fabric and Codex runtimes. Regression tests: `test_success_trial_stamps_agent_ok_true` / `test_failed_trial_stamps_agent_ok_false`.

packages/nemo_evaluator_sdk/examples/fabric_container/run_e2e.py
SandyChapman: Removed the `NVIDIA_API_KEY_FILE` option from the example; it now reads only from the `NVIDIA_API_KEY` env var, eliminating the committed-key-file risk.

PR #605, opened 2026-07-08
fix(agent-eval): stop leaking task.intent to the agent under evaluation

SandyChapman:
## Problem

The agent-eval runtimes framed the harness prompt with `task.intent` — the eval-side description of the **desired behavior** (what the grader checks for). Handing that to the agent under evaluation is a reward-hacking hole: the agent can read the grader's intent and target it directly. Same class of concern as the held-out-`reference` field work (AGENT-EVAL).

Three runtimes leaked it:
- `FabricAgentRuntime` — `f"Task id: {id}\nIntent: {intent}\nInputs: {inputs}\n"`
- `default_codex_prompt` — an explicit `Intent: {task.intent}` line
- `docker_sandbox._task_prompt` — `intent` as the last-resort fallback, written into the agent's `instruction.md`

## Change

Unify the framing so every runtime builds the prompt from `task.inputs` only:

- **New `runtimes/fabric/_common.py`** with a shared, intent-free `fabric_input`: task id → instruction (`inputs["instruction"|"prompt"]`) → remaining `inputs` content → seed files listed by name (not dumped inline). The host `FabricAgentRuntime` now calls it. Built as the shared home so the AALGO-321 container runtime (PR #604) reuses it instead of carrying its own copy.
- **`default_codex_prompt`** drops the `Intent:` line, lifts the instruction from inputs, keeps its workspace-edit trailer.
- **`docker_sandbox._task_prompt`** no longer falls back to `task.intent`; a task with no instruction in inputs yields an empty prompt rather than leaking intent.

## Tests

Updated `test_fabric_runtime.py`, `test_codex_runtime.py`, and `test_docker_sandbox_runtime.py` to lock in the intent-free contract across all three runtimes (assert instruction is surfaced, intent text and `Intent:` label are absent).

- `pytest` (all three suites) → 34 passed
- `ruff check` / `ruff format` → clean
- `ty` → no new diagnostics

Relates to AALGO-321 fabric-runtime dedup (PR #604).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated prompt framing across Codex, Docker sandbox, and Fabric to use only task `instruction`/`prompt` (with `instruction` taking priority) and to omit any evaluation intent details.
  * Simplified prompt payloads by generating the `Inputs:` JSON from remaining task inputs only, excluding instruction-related fields and the seeded-files key.
  * Adjusted seeded-workspace behavior so staged file metadata/content isn’t inlined into the agent prompt.
* **Tests**
  * Tightened Codex prompt redaction checks and updated seed-related expectations.
  * Updated Docker sandbox and Fabric tests to ensure intent is never present and seeded content isn’t inlined.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
@@ -68,7 +68,7 @@ class CodexCliAgentRuntime:
         self._work_root = Path(work_root).expanduser() if work_root is not None else None
         self._codex_bin = codex_bin
         self._timeout_s = timeout_s
-        self._prompt_builder = prompt_builder or default_codex_prompt
+        self._prompt_builder = prompt_builder or AgentEvalTask.agent_prompt
         self._process_factory = process_factory or asyncio.create_subprocess_exec
         self._runtime_name = runtime_name
 
@@ -95,15 +95,17 @@ class CodexCliAgentRuntime:
         evidence_dir.mkdir(parents=True, exist_ok=True)
         workspace_dir.mkdir(parents=True, exist_ok=True)
 
-        prompt = self._prompt_builder(task)
         prompt_path = evidence_dir / "prompt.txt"
         task_path = evidence_dir / "task.json"
         stdout_path = evidence_dir / "stdout.jsonl"
         stderr_path = evidence_dir / "stderr.txt"
         final_output_path = evidence_dir / "final_output.txt"
 
-        prompt_path.write_text(prompt, encoding="utf-8")
-        task_path.write_text(task.model_dump_json(indent=2), encoding="utf-8")
+        # Persist the task for debugging, but never the grader-only fields: the docker variant mounts
+        # this evidence dir into the sandbox (danger-full-access), so serializing `intent` (desired
+        # behavior) or `reference` (held-out ground truth) here would let the agent read them back out
+        # of `/evidence/task.json` — the same reward-hacking leak the intent-free prompt closes.
+        task_path.write_text(task.model_dump_json(indent=2, exclude={"intent", "reference"}), encoding="utf-8")
 
         command = self._command(workspace_dir=workspace_dir, final_output_path=final_output_path)
         process: Any | None = None
@@ -113,6 +115,10 @@ class CodexCliAgentRuntime:
             # synchronous (a handler may do blocking I/O, e.g. the plugin's fileset download), and this
             # runs on the event loop shared by every concurrent task, so a blocking seed would stall them all.
             seeded_files = await asyncio.to_thread(seed_workspace, workspace_dir, task.inputs.get(SEED_FILES_INPUT_KEY))
+            # Build the prompt after seeding and inside the guarded block: an instruction-less task
+            # raises here, failing just this task instead of aborting the run (and seeding wins if both).
+            prompt = self._prompt_builder(task)
+            prompt_path.write_text(prompt, encoding="utf-8")
             process = await self._process_factory(
                 *command,
                 stdin=subprocess.PIPE,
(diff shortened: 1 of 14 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/_common.py
```diff
@@ -0,0 +1,69 @@
...
+    if instruction:
+        lines += ["", instruction]
+
+    body_inputs = {
+        key: value for key, value in task.inputs.items() if key not in _INSTRUCTION_KEYS and key != SEED_FILES_INPUT_KEY
+    }
+    if body_inputs:
+        lines += ["", "Inputs:", json.dumps(body_inputs, indent=2, default=str)]
+
+    if seeded_files:
```
arpitsardhana: nit: This seems like Agent prompt.  especially "These files ..." and more importantantly "Complete the task by reading, creating etc.."

This should ideally be configurable. One naive idea is to have "named key" represent it? if not pesent in input items then simply add default

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
```diff
@@ -389,13 +389,19 @@ def print_codex_agent_models(*, codex_bin: str = "codex") -> None:
...
+    files, rather than constraining the answer to a single text reply. Seed files
+    (``inputs[SEED_FILES_INPUT_KEY]``) are listed by name instead of dumped inline — the agent finds
+    them already in its workspace. Pass a custom :data:`CodexPromptBuilder` to the runtime to override
+    this framing for a specific benchmark.
     """
-    body_inputs = {key: value for key, value in task.inputs.items() if key != SEED_FILES_INPUT_KEY}
-    lines = [f"Task id: {task.id}", f"Intent: {task.intent}"]
+    instruction = task.inputs.get("instruction") or task.inputs.get("prompt")
+    body_inputs = {
+        key: value for key, value in task.inputs.items() if key not in ("instruction", "prompt", SEED_FILES_INPUT_KEY)
```
arpitsardhana: Can this still inject intent?

SandyChapman: No, intent isn't a member of `inputs`. It has it's own property at `task.intent`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/_common.py
SandyChapman: Addressed by removing the templated framing entirely rather than making it configurable. The prompt handed to the agent is now exactly the task's `inputs["instruction"]` — no "These files…", no "Complete the task…" footer, no task-id/inputs scaffolding. `_common.py` is deleted; resolution lives in `AgentEvalTask.agent_prompt()`, and the fabric/codex/docker runtimes all route through it. Anything an author wants the agent to see now goes in the instruction text, so there's no runtime-injected boilerplate left to configure.

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
```diff
@@ -113,6 +111,10 @@ async def _run_task(self, index: int, task: AgentEvalTask, config: AgentEvalRunC
             # synchronous (a handler may do blocking I/O, e.g. the plugin's fileset download), and this
             # runs on the event loop shared by every concurrent task, so a blocking seed would stall them all.
             seeded_files = await asyncio.to_thread(seed_workspace, workspace_dir, task.inputs.get(SEED_FILES_INPUT_KEY))
+            # Build the prompt after seeding and inside the guarded block: an instruction-less task
+            # raises here, failing just this task instead of aborting the run (and seeding wins if both).
+            prompt = self._prompt_builder(task)
```
ngoncharenko, line 120: Sanitizing does not close the leak: `_run_task()` writes the full serialized task: `task_path.write_text(task.model_dump_json(indent=2), encoding="utf-8")`, including `intent` and `reference`, into `evidence_dir`, while `CodexDockerCliAgentRuntime._command` mounts that directory at `/evidence` and launches Codex with `danger-full-access`. The agent can therefore read `/evidence/task.json`

SandyChapman: Good catch — the prompt sanitization alone did leave `task.json` as a second vector. Fixed by redacting the grader-only fields from the persisted task: `task.model_dump_json(exclude={"intent", "reference"})`, so `/evidence/task.json` no longer carries `intent` or the held-out `reference` even though the docker variant mounts that dir with `danger-full-access`. Added a regression test (`test_codex_task_json_omits_grader_only_fields`) asserting both are absent while agent-safe fields still persist.

One thing left as-is: the docker variant still bind-mounts the whole evidence dir rw (it needs to write `final_output.txt` back). Redaction closes the reported leak, but tightening that mount to a dedicated output dir would be stronger defense-in-depth — happy to do it as a follow-up if you want the whole class closed rather than just the grader fields.

PR #612, opened 2026-07-08
chore(experiments): Rename all child resources of Experiment entity to "Evaluation"

shanaiabuggy:
Rename the child resource everywhere user-facing (API paths/schemas, SDK, CLI, RBAC, docs, Studio) to Evaluation. Keep the stored entity as Experiment (__entity_type__="experiment", parent_experiment_id); the API layer maps Evaluation ⇄ the Experiment entity so existing rows keep working.

- **No breaking changes to API consumers**-- For any legacy API endpoint, we register the same handler under a second path with `include_in_schema=False` so the old URL keeps working but doesn't clutter the docs
- Includes UI renames
- No data migration (deferred to the final migration step)

## Summary by CodeRabbit

* **New Features**
  * Added Evaluations as a first-class workspace resource (create, view, update, delete), including pin/unpin, advanced sorting/filtering, and pagination.
  * Added Evaluation session browsing with `summary` vs `detailed` mode, filtering, and pagination.
* **Improvements**
  * Experiment Group UI and API contracts now surface evaluation-focused fields, including `evaluation_count`, updated descriptions, and evaluation-aware sorting.
  * Studio navigation and detail pages now route and render evaluations (including trace and session views).
* **Bug Fixes**
  * Improved behavior when evaluation metric/rollup data is unavailable, returning deterministic errors and maintaining session browsing and sort recovery behavior.


sdk/python/nemo-platform/scripts/lint
```diff
@@ -13,7 +13,7 @@ else
 fi
 
 echo "==> Running pyright"
-uv run pyright
+uv run pyright -p .
```
shanaiabuggy, line 16: why?

BrianNewsom (approved):

PR #613, opened 2026-07-08
fix(pip): Set some dep lower-bound to fix `ResolutionTooDeep`

matthewgrossman:

## What

Add lower/upper version bounds on two transitive dependencies to stop pip's
backtracking resolver from failing with `ResolutionTooDeep: 200000` on
`pip install "nemo-platform[all]"`.

In `plugins/nemo-agents/pyproject.toml`:

```toml
"litellm>=1.83.14",
"pyopenssl>=26.0.0,<27.0.0",
```

These flow through `make vendor` into the generated `nemo-agents-plugin` extra
in `packages/nemo_platform/pyproject.toml`, which `[all]` pulls in via
`[services]` → `[plugins]`.

## Why

`nemo-platform` doesn't depend on `litellm` directly — it comes in transitively
via `nvidia-nat-langchain` → `langchain-litellm` → `litellm`, and `nvidia-nat-langchain`
also requires `pyopenssl` directly. `litellm` has ~1,200 released versions on
PyPI and enters the graph unbounded, so pip's `resolvelib` backtracker explores
a huge candidate space and blows past its 200,000-round `ResolutionTooDeep` cap.

The two bounds restate the effective floors `nvidia-nat-langchain` already
imposes (`langchain-litellm>=0.6.5` requires `litellm>=1.83.14`; nvidia-nat-langchain
requires `pyopenssl>=26.0.0,<27.0.0`). Surfacing them at the top level collapses
pip's candidate set and lets the resolver converge. This mirrors the existing
`ASTD-164` pins in the same file, which do the same thing for the
`aioboto3`/`langchain-aws` chain.

## Verification

Reproduced the reporter's exact environment (pip 24.0, Python 3.12.12,
`pip install "<wheel>[all]"` in a clean venv):

| | Before | After |
|---|---|---|
| Outcome | `ResolutionTooDeep: 200000` | ✅ success (exit 0) |
| Time | ~2 h → fail | ~4.5 min |

CI wheel build + test jobs (which install `[services]` via uv across py3.11–3.14)
continue to pass.

## Notes / follow-ups (out of scope)

- These bounds track today's PyPI state; a future `litellm`/`langchain-litellm`
  release could shift the backtracking again, same as the existing ASTD-164 pins.
  This is mitigation, not a permanent cure.
- `pip install "nemo-platform[all]"` remains broken on **Python 3.13/3.14**
  independent of this change, because `litellm` and `nemo-safe-synthesizer`
  declare `Requires-Python <3.14`. pip strictly enforces that cap; uv does not,
  which is why CI (uv) is green there. Worth a separate issue to either bump the
  upstream deps or narrow `nemo-platform`'s advertised `requires-python`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation reliability for the platform and agents plugin bundles by tightening optional dependency requirements.
  * Added explicit compatible version bounds for key dependencies to reduce resolution backtracking and lower the chance of install-time failures.


mckornfield (approved):
lol uv.lock 485 changes on this one pin

PR #615, opened 2026-07-08
fix(jobs): prevent Docker job scheduling races under queue pressure

ironcommit:
## Summary

Fixes job scheduling reliability under queue pressure by introducing admission control to the Docker backend, moving TTL enforcement out of individual backends into the scheduler/reconciler, and adding comprehensive debug logging across the scheduling pipeline.

### Admission control for Docker container starts
- Added a `BoundedSemaphore`-based admission gate (`_container_start_admission`) that limits concurrent container start operations to `DOCKER_CONTAINER_START_WORKERS` (10)
- When the admission gate is full, scheduling raises `SchedulingDeferred` instead of blocking — the step stays in PENDING and is retried on the next scheduler tick
- Container argument preparation (`_prepare_container_args_for_start`) is extracted and runs inside the admission gate, so storage setup and image pulls don't block unrelated steps
- The semaphore is released in a `finally` block in `run_container`, ensuring slots are always freed even on failure

### TTL enforcement moved to scheduler/reconciler
- Removed per-backend `ttl_seconds_before_active` checks from `DockerJobBackend.schedule()` and `SubprocessJobBackend.schedule()` — these were racy because the backend could time out a step that the scheduler would then re-queue
- PENDING steps that already have a running/exited/dead container now route through `sync_pending` in the reconciler, preventing the TTL path from killing steps that are actually making progress

### Scheduling order fix
- Changed step fetch sort from `-created_at` (newest first) to `created_at` (oldest first), so older queued steps are scheduled before newer ones

### Debug logging
- Added structured `logger.debug` timing spans across the entire scheduling and reconciliation pipeline: fetch, backend sync, status updates, container create, image pull, storage setup, launcher lookup, cancellation checks, and cleanup
- Added `docker_state_debug_fields()` helper that extracts Docker container state (exit code, OOM, timestamps, auto-remove config, cleanup TTL) into a dict for structured logging
- Added `parse_docker_timestamp()` to safely handle Docker's zero-time sentinel and parse errors
- Wrapped `update_status` calls in `_update_step_status_with_timing` helpers in both scheduler and reconciler for consistent timing and error logging

### Cleanup improvements
- Container cleanup now uses `docker_state_debug_fields()` for consistent state extraction and logs detailed context for each cleanup decision (immediate removal, TTL retention, skip due to non-terminal state)
- `parse_docker_timestamp` handles Docker's `0001-01-01T00:00:00Z` sentinel without crashing

### Tests
- Updated Docker backend tests for the new admission semaphore and `_prepare_container_args_for_start` extraction
- Updated scheduler tests for `SchedulingDeferred` handling and sort order change
- Updated subprocess backend tests to remove TTL enforcement expectations
- Added `SchedulingDeferred` to Kubernetes and Volcano backend test mocks

## Test plan
- [x] Unit tests updated for all changed backends and controllers
- [ ] Stress test with concurrent job submissions to verify admission backpressure
- [ ] Verify PENDING steps with existing containers transition correctly through reconciler

## Summary by CodeRabbit

* **New Features**
  * Added bounded concurrency control for Docker container starts, including queue-delay tracking.
  * Added deferred-scheduling behavior when backend capacity is full.
  * Improved end-to-end scheduling/reconciliation visibility with monotonic timing and richer structured logs.

* **Bug Fixes**
  * Prevented premature “before-active” TTL failures before backends accept steps (Docker/Subprocess/Kubernetes/Volcano).
  * Tightened pending sync and updated TTL/cleanup logic using parsed Docker timing; improved handling of Docker kill race (409).

* **Tests**
  * Updated and expanded unit/regression coverage for TTL, deferred scheduling, and reconciliation outcomes across backends.

### DIFF
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
@@ -7,11 +7,13 @@ import json
 import logging
 import os
 import tarfile
+import threading
 import time
 import uuid
 from abc import abstractmethod
 from concurrent.futures import ThreadPoolExecutor
-from typing import Generic, Literal, TypeVar
+from dataclasses import dataclass
+from typing import Any, Generic, Literal, TypeVar
 
 import docker.types
 from docker.errors import APIError, ImageNotFound, NotFound
@@ -79,6 +81,7 @@ from nmp.core.jobs.controllers.backends.exceptions import (
     FailedToScheduleError,
     JobStorageError,
     ResourceAllocationError,
+    SchedulingDeferred,
 )
 from opentelemetry import trace
 from pydantic import BaseModel, Field
@@ -87,6 +90,7 @@ import docker
 
 tracer = trace.get_tracer(__name__)
 logger = logging.getLogger(__name__)
+DOCKER_CONTAINER_START_WORKERS = 10
 
 
 def k8s_shm_quantity_to_docker(quantity: str) -> str:
@@ -114,6 +118,13 @@ DOCKER_STOP_TIMEOUT = int(os.getenv("NEMO_JOBS_DEFAULT_DOCKER_STOP_TIMEOUT", "30
 
 
 ProviderT = TypeVar("ProviderT", bound=ExecutionProviderT)
+
+
+@dataclass(frozen=True, slots=True)
+class DockerTimestampParseResult:
+    parsed: datetime.datetime | None
+    parse_error: str | None
+    is_zero: bool
 
 
 class DockerVolumeMount(BaseModel):
@@ -181,7 +192,8 @@ class DockerJobBackend(JobBackend[ProviderT, DockerJobExecutionProfileConfig], G
     BACKEND_NAME: str = "docker"
 
     def init(self) -> None:
-        self._container_run_threadpool = ThreadPoolExecutor(max_workers=10)
+        self._container_start_admission = threading.BoundedSemaphore(DOCKER_CONTAINER_START_WORKERS)
+        self._container_run_threadpool = ThreadPoolExecutor(max_workers=DOCKER_CONTAINER_START_WORKERS)
         self._client = docker.from_env(timeout=180)
         if NEMO_JOBS_IMAGE_REGISTRY:
             logger.info(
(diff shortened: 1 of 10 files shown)


services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -659,6 +703,15 @@ def schedule_single_container(
             additional_volumes_mounts=additional_volume_mounts,
             step_config_json=step_config_json,
         )
+        logger.debug(
+            "Docker job storage ensured",
+            extra={
+                "job": step.job,
+                "step": step.name,
+                "task": task_id,
+                "duration_seconds": time.monotonic() - ensure_storage_started_at,
```
mckornfield, line 720: my nitty comment, what place do all these duration_seconds go to?

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -770,10 +818,24 @@ def get_jobs_launcher_binary(self) -> io.BytesIO | None:
...
+    def run_container(
+        self,
+        step: PlatformJobStepWithContext,
+        container_args: dict,
+        submitted_to_threadpool_at: float | None = None,
+    ):
         with start_span_with_ctx(
             tracer, "jobs_controller/docker_backend/run_container", JobContext(id=step.job, step_name=step.name)
         ):
+            if submitted_to_threadpool_at is not None:
```
mckornfield, line 839: could it still log it, even without the time, just skip the queue_delay_seconds addition to extras?

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -1121,13 +1299,79 @@ def sync_stop_container(self, step: PlatformJobStepWithContext, container: Conta
                 else:
                     raise e
 
+    @staticmethod
+    def parse_docker_timestamp(timestamp: str | None) -> tuple[datetime.datetime | None, str | None, bool]:
```
mckornfield: nit- return a dataclass or something clearer on what the other fields are that are returned

services/core/jobs/src/nmp/core/jobs/controllers/scheduler.py
```diff
@@ -73,21 +75,34 @@ def step(self):
...
-                            step.name,
-                            workspace=step.workspace,
-                            job=step.job,
+                        self._update_step_status_with_timing(
+                            step=step,
+                            phase="schedule",
                             status=update.status,
-                            status_details=update.status_details,  # type: ignore
-                            error_details=update.error_details,  # type: ignore
+                            status_details=update.status_details,
```
mckornfield, line 118: why could you remove this one, but you had to have it in the other call site?

mckornfield (approved):

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
ironcommit: these are just for debugging, and it's super helpful for profiling, which was what I was doing when I fixed this bug.  They don't go anywhere other than into the logs where they can be analyzed.

PR #617, opened 2026-07-08
fix: update for auditor, nss and fine tuning cves

mckornfield:
## Summary by CodeRabbit

* **Security**
  * Updated runtime components with newer security-patched versions and refreshed system packages.
  * Added safeguards to prevent vulnerable package downgrades during installation.

* **Compatibility**
  * Updated Safe Synthesizer support to version 0.1.7.
  * Improved CUDA 12.9 runtime setup using the appropriate package index.

* **Performance**
  * Refreshed Hugging Face tooling versions for unsloth training environments.

### DIFF
--- a/docker/Dockerfile.safe-synthesizer-tasks
+++ b/docker/Dockerfile.safe-synthesizer-tasks
@@ -11,10 +11,10 @@
 ARG PYTHON_VERSION=3.13
 ARG PYTHON_IMAGE=python:${PYTHON_VERSION}-slim-trixie
 ARG UV_IMAGE=ghcr.io/astral-sh/uv:0.9.30
-ARG SAFE_SYNTHESIZER_RUNTIME_PACKAGE="nemo-safe-synthesizer[engine,cu129]==0.1.2"
+ARG SAFE_SYNTHESIZER_RUNTIME_PACKAGE="nemo-safe-synthesizer[engine,cu129]==0.1.7"
 ARG FLASHINFER_CU129_INDEX_URL="https://flashinfer.ai/whl/cu129"
 ARG PYTORCH_CU129_INDEX_URL="https://download.pytorch.org/whl/cu129"
-ARG VLLM_CU129_WHEEL="vllm @ https://github.com/vllm-project/vllm/releases/download/v0.20.0/vllm-0.20.0%2Bcu129-cp38-abi3-manylinux_2_31_x86_64.whl"
+ARG VLLM_CU129_INDEX_URL="https://wheels.vllm.ai/ee0da84ab9e04ac7610e28580af62c365e898389/cu129"
 
 # =============================================================================
 # uv binary
@@ -29,7 +29,7 @@ ARG PYTHON_IMAGE
 ARG SAFE_SYNTHESIZER_RUNTIME_PACKAGE
 ARG FLASHINFER_CU129_INDEX_URL
 ARG PYTORCH_CU129_INDEX_URL
-ARG VLLM_CU129_WHEEL
+ARG VLLM_CU129_INDEX_URL
 ARG CONTAINER_VARIANT=cu129
 ARG USERNAME=nemo
 ARG USER_UID=1000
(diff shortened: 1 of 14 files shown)


docker/automodel/Dockerfile.nmp-automodel-base
```diff
@@ -155,6 +155,9 @@ RUN rm -rf \
     /usr/local/lib/python3.12/dist-packages/tornado-*.dist-info \
     /usr/local/lib/python3.12/dist-packages/onnx \
     /usr/local/lib/python3.12/dist-packages/onnx-*.dist-info \
+    /usr/local/lib/python3.12/dist-packages/flash_attn \
```
mckornfield: hmmm feel like these are required, so this might be bad lol

docker/Dockerfile.safe-synthesizer-tasks
```diff
@@ -14,7 +14,7 @@ ARG UV_IMAGE=ghcr.io/astral-sh/uv:0.9.30
 ARG SAFE_SYNTHESIZER_RUNTIME_PACKAGE="nemo-safe-synthesizer[engine,cu129]==0.1.2"
 ARG FLASHINFER_CU129_INDEX_URL="https://flashinfer.ai/whl/cu129"
 ARG PYTORCH_CU129_INDEX_URL="https://download.pytorch.org/whl/cu129"
-ARG VLLM_CU129_WHEEL="vllm @ https://github.com/vllm-project/vllm/releases/download/v0.20.0/vllm-0.20.0%2Bcu129-cp38-abi3-manylinux_2_31_x86_64.whl"
+ARG VLLM_CU129_WHEEL="vllm @ https://github.com/vllm-project/vllm/releases/download/v0.22.0/vllm-0.22.0%2Bcu129-cp38-abi3-manylinux_2_28_x86_64.whl"
```
mckornfield: I have a PR to bump to 0.24.0 in nss, but it's in limbo. This is probably fine for now though, since I'll have to come back and update this

docker/Dockerfile.safe-synthesizer-tasks
```diff
@@ -176,6 +189,17 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     UV_CACHE_DIR=/root/.cache/uv uv sync --package nemo-safe-synthesizer-plugin --no-dev --no-editable --inexact \
         --no-install-package nemo-platform
 
+RUN --mount=type=cache,target=/root/.cache/uv \
```
ironcommit: can we just do this once?

mckornfield: think it might be right to move to the constraints anyhow

mckornfield: Done in 866c30daea. The Safe Synthesizer Dockerfile now installs the patched cryptography/pyarrow versions once in the runtime dependency layer, then skips those packages during the later plugin uv sync so the workspace lock cannot downgrade them. I also backed vLLM back to the Safe Synthesizer 0.1.2 default and removed the separate vLLM install split; we can pick up the vLLM bump with the next NSS release.

.github/workflows/ci.yaml
```diff
@@ -1087,6 +1087,12 @@ jobs:
             exit 1
           fi
           WHEEL="${wheels[0]}"
+          if [[ "${PYTHON_VERSION}" == "3.14" ]]; then
+            # litellm 1.92.0 builds its Rust bridge from source on Python 3.14.
+            # Its pinned PyO3 supports the stable ABI, but rejects 3.14 unless
+            # this forward-compatibility opt-in is present.
+            export PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1
+          fi
```
svvarom, line 1094: is this var missing for user installs? will they hit the same PyO3 failure?

mckornfield: It is... I can try and find a way to avoid it for them

svvarom (approved):

ironcommit (approved):
LGTM

PR #619, opened 2026-07-09
ci: manage NGC metadata programmatically

crookedstorm:
Using PEP 723 inline requirements to avoid polluting the actual product, this is a simple python script and test file to manage NGC metadata in a visible and reviewable fashion. The markdown files include optional frontmatter
which can be used to apply fields besides the overview with non-default values.

## Summary by CodeRabbit

* **New Features**
  * Added NGC metadata synchronization via a new script and the **NGC Metadata Sync** workflow (manual or reusable calls).
  * Extended CI path-change detection with a new **ngc-metadata** flag and exposed it in the changes action output for targeted runs.
* **Tests**
  * Added/expanded unit tests for metadata parsing, sync create-vs-update behavior, and dry-run output.
* **Documentation**
  * Added documentation pages for the NeMo Platform Helm chart and related container images used during metadata sync.
* **Chores**
  * Improved CI disk usage by enabling Go cleanup in a test-image job.

### DIFF
--- a/.github/assets/ngc/charts/nemo-platform.md
+++ b/.github/assets/ngc/charts/nemo-platform.md
@@ -1,0 +1,12 @@
+## NeMo Platform Helm Chart
+
+NVIDIA NeMo Platform Helm Chart for the deployment of NeMo Platform on Kubernetes.
+
+### Resources
+
+[Documentation](https://docs.nvidia.com/nemo-platform)
+
+### License
+
+This container is licensed under the [Apache License 2.0](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/LICENSE).
+
--- a/.github/assets/ngc/charts/nemo-platform.md
+++ b/.github/assets/ngc/charts/nemo-platform.md
@@ -8,5 +8,5 @@ NVIDIA NeMo Platform Helm Chart for the deployment of NeMo Platform on Kubernete
 
 ### License
 
-This container is licensed under the [Apache License 2.0](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/LICENSE).
+This chart is licensed under the [Apache License 2.0](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/LICENSE).
 
--- a/.github/assets/ngc/containers/nmp-api.md
+++ b/.github/assets/ngc/containers/nmp-api.md
@@ -1,0 +1,14 @@
+---
+description: API component of NeMo Platform
+---
+## NeMo Platform API Container
+
+This container image provides a deployable API interface to NeMo Platform.
+
+### Resources
+
+[Documentation](https://docs.nvidia.com/nemo-platform)
+
+### License
+
+This container is licensed under the [Apache License 2.0](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/LICENSE).
--- a/.github/assets/ngc/containers/nmp-api.md
+++ b/.github/assets/ngc/containers/nmp-api.md
@@ -3,7 +3,7 @@ description: API component of NeMo Platform
 ---
 ## NeMo Platform API Container
 
-This container image provides a deployable API interface to NeMo Platform.
+This container image provides a REST API for NeMo Platform.
 
 ### Resources
 
--- a/.github/scripts/ngc_metadata.py
+++ b/.github/scripts/ngc_metadata.py
@@ -1,0 +1,218 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "ngcsdk==4.20.1",
+#   "pyyaml>=6.0.2",
+#   "typer>=0.24.1,<0.25",
+# ]
+# ///
+
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Create or update NGC metadata from Markdown files."""
+
+from __future__ import annotations
+
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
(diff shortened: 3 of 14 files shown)


.github/assets/ngc/charts/nemo-platform.md
```diff
@@ -0,0 +1,12 @@
+## NeMo Platform Helm Chart
```
svvarom, line 1: Do we also want to include the SDKs?

crookedstorm: Not really? It wouldn't download very well from a resource compared to pypi.

.github/workflows/ngc-metadata.yaml
```diff
@@ -0,0 +1,38 @@
...
+  workflow_run:
+    workflows: ["Release"]
+    types: [completed]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  sync:
```
svvarom, line 11: Do we want to gate this job with [Environment secrets](https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments#environment-secrets)?

.github/workflows/ngc-metadata.yaml
```diff
@@ -0,0 +1,38 @@
+name: NGC Metadata Sync
+
+on:
+  workflow_run:
```
svvarom: Do want this to be opt-in?

svvarom: I was thinking of doing something like:

```yaml
update_ngc_metadata:
    description: Update NGC catalog metadata
    type: boolean
    default: false
```

svvarom: Then `ngc-metadata.yaml` does not need to worry about the release workflow.

.github/workflows/ngc-metadata.yaml
crookedstorm: Doesn't work with workflow call...

.github/workflows/ngc-metadata.yaml
crookedstorm: Yes...I may have chosen the wrong trigger. Still reviewing that...

.github/scripts/ngc_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+    metadata, overview = split_front_matter(path.read_text(encoding="utf-8"))
+    name = path.stem
+    display_name = str(metadata.get("display_name") or default_display_name(name))
+    default_description = (
+        f"Deploy {display_name} to Kubernetes"
+        if asset_type == "chart"
+        else f"{display_name} is part of the NeMo Platform"
+    )
+    description = str(metadata.get("description") or default_description)
+    labels = metadata.get("labels", DEFAULT_LABELS)
```
svvarom, line 91: Could we avoid overwriting fields that aren’t set in front matter? I checked `ngcsdk==4.20.1`: it preserves omitted fields, but these defaults are always sent. For example, `nmp-automodel-tasks` would lose its `Fine Tuning` label and change its display name

crookedstorm: That would be more complex, requiring two-way communication for little benefit. I'd rather add the label to our frontmatter. Nobody seemed to care much about the labels on the last run, but that one might be work adding.

.github/workflows/ngc-metadata.yaml
crookedstorm: Fixed. It should just be workflow_call.

.github/scripts/ngc_metadata.py
svvarom: yeah, the labels might matters more for NVAIE stuff. This is a NIT - feel free to ignore.

.github/scripts/ngc_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+"""Create or update NGC metadata from Markdown files."""
+
+from __future__ import annotations
+
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Annotated, Literal
+
+import typer
```
svvarom, line 22: Nice!! this is much much better than argparse 🤮

.github/scripts/ngc_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+    if not isinstance(metadata, dict):
+        raise ValueError("Markdown front matter must be a mapping")
+    return metadata, overview
+
+
+def load_asset(path: Path, asset_type: Literal["container", "chart"]) -> Asset:
+    """Load one asset and apply its metadata defaults."""
+    metadata, overview = split_front_matter(path.read_text(encoding="utf-8"))
+    name = path.stem
+    display_name = str(metadata.get("display_name") or default_display_name(name))
```
svvarom, line 84: I like how we infer the name from path.

crookedstorm: Yeah, the files are the config.

.github/scripts/ngc_metadata.py
crookedstorm: I put the Fine Tuning labels back and verified it works against my test project/team

svvarom (approved):
LGTM!

.github/assets/ngc/containers/nmp-api.md
```diff
@@ -0,0 +1,14 @@
+---
+description: API component of NeMo Platform
+---
+## NeMo Platform API Container
+
+This container image provides a deployable API interface to NeMo Platform.
```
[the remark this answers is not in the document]
svvarom: Makes sense lol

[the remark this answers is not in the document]
svvarom: Like Release: RC 😆

PR #625, opened 2026-07-09
feat(evaluator-sdk): agent-skill A/B evaluation for Fabric agent-eval

SandyChapman:
## What

Adds **agent-skill evaluation** to the Fabric agent-eval runtime: inject an
[agentskills.io](https://agentskills.io) skill into an eval and A/B the same
taskset with vs. without it, to measure the skill's effect.

- **`AgentSkill` + `install_skill`** (`runtimes/fabric/skills.py`) — a
  directory-backed agentskills bundle, staged per task and wired to the harness.
  Namespaced under `<name>/` so it can't collide with task-seeded workspace files.
  `SkillProvenance` (name, content hash, mode, adapter, location) is stamped into
  trial metadata for an auditable A/B diff.
- **How the skill reaches the harness is decided by querying Fabric's own
  capability planner at runtime**, not a hardcoded adapter list. The runtime plans
  a copy of the config with a sentinel skill path attached and reads
  `RunPlan.capability_plan` routes:
  - a `skills` route targeting `harness_native` (the adapter accepts the native
    Fabric `skills` config, e.g. Hermes/Claude) → native `skills.paths` overlay;
  - otherwise a `codex` harness → `.agents/skills/<name>/` placement, which the
    Codex CLI self-discovers;
  - otherwise → **fail fast** rather than run a skill-free trial mislabeled "with
    skill" (which would corrupt the A/B).

  Driving the decision from Fabric means it tracks whatever the installed adapters
  declare — including **end-user adapters we don't ship** (e.g. the `claude`
  adapter, which a static allow-list would miss) — instead of a list that silently
  goes stale.
- **`FabricAgentRuntime`** gains a `skill=` param and a **`with_skill()` clone** for
  confound-free A/B: baseline (`with_skill(None)`) vs. treated (`with_skill(skill)`)
  differ in exactly the skill and nothing else.
- **`SkillUsedMetric`** — `skill_present` / `skill_used` booleans, to flag a failure
  to engage an injected skill (best-effort trajectory detection; strongest for
  Codex filesystem discovery).
- **`run_id`** is folded into the Fabric evidence path (`<root>/<run_id>/<task>`) so
  an A/B baseline and skilled variant sharing a `work_root` don't collide.
- **`examples/skill_eval/`** — a runnable baseline-vs-treated A/B where the skill
  (the "Supercool Coding Guidelines": `supercool_` prefix + French `enieme` param)
  is *required* to pass `follows_guidelines`.

## Notes

- Rebased onto the merged `FabricClient` → `Fabric` SDK migration (#648), so this is
  built on the current config-first Fabric run model (`Fabric().run(...)` /
  `Fabric().plan(...)`). No longer blocked on that migration.
- The runtime-query decision (`resolve_skill_mode` / `native_skills_route`) is pure
  and unit-testable without the native SDK; the actual `client.plan(...)` probe was
  verified live against a NeMo-Fabric checkout (hermes→native, codex→unsupported,
  claude→native).

## Verification

- `test_fabric_skills.py`, `test_fabric_runtime.py`, `test_skill_used_metric.py`
  and the surrounding agent-eval suite — **160 passed, 2 skipped**.
- `ruff`, `ty` (only the expected `unused-ignore` warnings CI relies on),
  `lint-sdk-vendored`, and `lint-cli` all clean; vendored SDK mirror regenerated
  via `make vendor`.

## Summary by CodeRabbit

* **New Features**
  * Added agent-skill injection for Fabric evaluations with `with_skill(...)`, including per-run evidence/run isolation and skill provenance metadata.
  * Added `skill_present`/`skill_used` reporting via a new metric.
  * Added a complete A/B skill evaluation example, including a sample `supercool-guidelines` skill bundle.
* **Bug Fixes**
  * Unsupported skill-injection configurations now fail fast with a clear message.
  * Failed trials retain the relevant skill metadata for easier diagnosis.
* **Documentation**
  * Updated example documentation and skill instructions.

### DIFF
--- a/packages/nemo_evaluator_sdk/examples/skill_eval/README.md
+++ b/packages/nemo_evaluator_sdk/examples/skill_eval/README.md
@@ -1,0 +1,107 @@
+# skill_eval — A/B evaluation of an injected agent skill
+
+Runs one taskset twice through the Fabric agent-eval runtime — once **without** a
+skill (baseline) and once **with** an injected [agentskills.io](https://agentskills.io)
+skill (treated, via `runtime.with_skill(skill)`) — then compares the scores. The
+two arms differ in *exactly* the skill, so the delta is attributable to it.
+
+## The setup: the skill is *required* to pass
+
+Each task asks the agent to *"write a Python function that ..., following the
+Supercool Coding Guidelines."* Those guidelines live **only** in the injected
+`supercool-guidelines` skill and are **not inferable from the prompt**:
+
+- function names must start with `supercool_`;
+- a positional-index parameter must be named `enieme` (French for "nth").
+
+So the `follows_guidelines` metric can pass **only when the agent actually uses
+the skill** — giving a clean, measured baseline-vs-treated difference rather than
+a fuzzy "did it help".
+
+## What it demonstrates
+
+- Injecting an agentskills bundle into `FabricAgentRuntime` and running a
+  confound-free A/B with `with_skill(...)` (baseline vs. treated as distinct
+  `run_id`s).
+- Task-authored guidelines metrics whose `follows_guidelines` output is
+  skill-dependent by construction, scored off the **parsed function signature**
+  (AST) so a convention only *mentioned* in prose doesn't count. Each task gets the
+  metric matching its parameters: `GuidelinesMetric` (index tasks) checks for the
+  `enieme` positional-index name — the unguessable discriminator — while
+  `GcdGuidelinesMetric` (gcd, no index param) checks its parameters are French
+  words. `supercool_prefix` is weakly guessable and reported on its own.
+- `SkillUsedMetric` — `skill_present` / `skill_used` surface whether the agent
+  used the injected skill.
+
+The bundled skill is `skills/supercool-guidelines/` (a spec-compliant `SKILL.md`).
+
+## Run it
+
+**Prerequisites** — this example imports `nemo_evaluator_sdk` and drives the native
+Fabric stack, so run it from the project virtualenv (a bare repo-root `python`
+won't have the workspace on its import path):
+
+- `make bootstrap-python` — creates `.venv` and `uv sync --all-packages`, which
+  installs the workspace packages (including `nemo_evaluator_sdk`);
+- `script/dev-install-fabric.sh` — the native `nemo-fabric` + Hermes SDK adapter +
+  `nemo-relay` gateway (not in the lockfile, so installed separately);
+- `NVIDIA_API_KEY` for an account **provisioned for** `MODEL` in `run_skill_eval.py`.
+
+Then, from the repo root, run with the venv interpreter:
+
+```bash
+NVIDIA_API_KEY=... ADAPTER_PYTHON="$(pwd)/.venv/bin/python" \
+  .venv/bin/python -m packages.nemo_evaluator_sdk.examples.skill_eval.run_skill_eval
+```
+
+`ADAPTER_PYTHON` is required whenever the `python3` on your `PATH` is not this venv
+(common on macOS/Homebrew, pyenv, etc.): the Fabric Hermes adapter runs as a
+subprocess and otherwise falls back to a bare `python3` off `PATH`, which won't
+have `nemo_fabric_adapters` installed (`python_adapter_exit_nonzero` /
+`ModuleNotFoundError`).
+
+Each arm writes a run bundle under `skill-eval-output/<arm>/`, with per-task
+Fabric evidence under `evidence/fabric/<run_id>/`. If a trial fails (bad model
+id, missing credential, harness crash), the run prints a `⚠️ N trial(s) FAILED`
+block and exits non-zero rather than showing an empty-but-tidy table.
+
+Example output (with `nvidia/nemotron-3-super-120b-a12b`):
+
+```text
+Harness: nvidia.fabric.hermes.sdk   model: nvidia/nemotron-3-super-120b-a12b   tasks: 2
+runs: baseline (baseline) vs treated (treated)
+
+  metric.output                             baseline   with-skill
+  agent_phase_success.agent_phase_success   2/2        2/2
+  follows_guidelines.enieme_param           0/1        1/1
+  follows_guidelines.follows_guidelines     0/2        2/2
+  follows_guidelines.french_params          0/1        1/1
+  follows_guidelines.supercool_prefix       0/2        2/2
+  skill_used.skill_present                  0/2        2/2
+  skill_used.skill_used                     0/2        2/2
+```
+
+(`enieme_param` and `french_params` each total `/1` — they are the per-task checks,
+emitted only by the index task and the gcd task respectively.)
+
+## Notes
+
+- `follows_guidelines` is the causal metric — it is what the skill directly
+  controls. `skill_used` is the *mechanism* signal; it detects the skill's staged
+  `location` in the trajectory and can under-report for the **Hermes** harness
+  (in-context loading), so treat `follows_guidelines` as the source of truth for
+  "did the skill take effect". See `SkillUsedMetric` for the detail.
+- The model is the `MODEL` constant in `run_skill_eval.py`. It must be
+  **provisioned for your account** — some catalog-listed models return HTTP 404
+  (`Function ... not found for account`). It also has to be capable enough to
+  *obey* the injected skill: weaker models read the guidelines but ignore them,
+  producing a flat A/B (a valid, if undramatic, result).
+  `nvidia/nemotron-3-super-120b-a12b` produces the lift shown above.
+- The Hermes agent loop budget is `harness.settings.max_iterations` (set to 50
+  here). The Fabric Hermes adapter defaults it to **1**, which starves any
+  multi-step task — leave it set.
+- The guidelines metrics score the **parsed function signature** (`ast`, with a
+  `def`-regex fallback), scoping to the `supercool_`-named answer function so a
+  helper definition or a convention mentioned only in prose doesn't count. The gcd
+  `french_params` check uses a small illustrative French wordlist (`_FRENCH_PARAMS`)
+  — extend it if a run picks a French word not listed.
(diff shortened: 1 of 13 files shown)


packages/nemo_evaluator_sdk/examples/skill_eval/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+**If the `python3` on your `PATH` is not the interpreter running this example**
+(common with a venv on macOS/Homebrew, pyenv, etc.), also set `ADAPTER_PYTHON` to
+that interpreter. The Fabric Hermes adapter runs as a subprocess and otherwise
+falls back to a bare `python3` off `PATH`, which won't have `nemo_fabric_adapters`
+installed — you'd get a `python_adapter_exit_nonzero` / `ModuleNotFoundError`:
+
+```bash
+NVIDIA_API_KEY=... ADAPTER_PYTHON="$(pwd)/.venv/bin/python" \
+  python -m packages.nemo_evaluator_sdk.examples.skill_eval.run_skill_eval
```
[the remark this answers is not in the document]
SandyChapman: Addressed — the README now documents the workspace install prerequisite (`make bootstrap-python`, which installs `nemo_evaluator_sdk`, plus `script/dev-install-fabric.sh` for the native Fabric stack) and runs via the venv interpreter (`.venv/bin/python -m ...`) so the import path is established.

packages/nemo_evaluator_sdk/examples/skill_eval/run_skill_eval.py
```diff
@@ -0,0 +1,203 @@
...
+
+    async def compute_scores(self, input: MetricInput) -> MetricResult:
+        text = input.candidate.output_text or ""
+        prefix = bool(self._PREFIX.search(text))
+        french = bool(self._FRENCH_PARAM.search(text))
+        return MetricResult(
+            outputs=[
+                MetricOutput(name="supercool_prefix", value=prefix),
+                MetricOutput(name="french_param", value=french),
+                MetricOutput(name="follows_guidelines", value=prefix and french),
```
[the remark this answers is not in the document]
SandyChapman: Addressed — scoring now parses the function **signature** via `ast` (with a `def`-regex fallback), scoped to the `supercool_`-named answer function so helper defs and prose mentions no longer count. Also made task-aware: `GuidelinesMetric` (index tasks → `enieme`) and a new `GcdGuidelinesMetric` (gcd → French params), so `enieme` is only required where the skill actually mandates it.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
```diff
@@ -0,0 +1,246 @@
...
+    namespaced under ``<name>/`` so it never collides with task-seeded workspace-root files; the content
+    hash is computed over the staged bytes so provenance tracks the actual skill content.
+    """
+    if mode == SKILL_MODE_NATIVE:
+        skill_root = skill_stage_dir / skill.name
+        _stage_bundle(skill.directory, skill_root)
+        overlay: dict[str, object] = {
+            "name": SKILL_PROFILE_NAME,
+            "description": "Make the evaluation skill available via the native Fabric skills config.",
+            "skills": {"paths": [str(skill_root)]},
```
ngoncharenko: **[P1] Preserve the baseline skill set**

Fabric applies profile arrays last-wins: adding this overlay after existing `skills.paths` resolves to only `skill_root`. The treated arm therefore drops every preconfigured skill while the baseline retains them, invalidating the claimed one-variable A/B. Build the trailing profile from the effective existing paths plus the evaluated skill, with a regression for config/profile-provided skills.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def _stage_bundle(directory: Path, skill_root: Path) -> None:
+    """Copy the skill ``directory`` into ``skill_root`` (the ``<name>/`` bundle dir)."""
+    src = directory.expanduser()
+    if not (src / PRIMARY_SKILL_DOC).is_file():
+        raise SkillInjectionError(f"skill directory {str(directory)!r} has no {PRIMARY_SKILL_DOC}")
+    skill_root.parent.mkdir(parents=True, exist_ok=True)
+    # OS-level copy — file contents never pass through Python memory.
+    shutil.copytree(src, skill_root, dirs_exist_ok=True)
```
ngoncharenko: **[P1] Stage an exact bundle instead of merging**

`dirs_exist_ok=True` preserves files removed from the source bundle when an explicit run ID is reused; the bundled example reuses `treated`, and I reproduced a deleted `old.md` surviving the second install. It can also merge task-seeded files under the reserved Codex skill path. Recreate the evaluator-owned destination and reject reserved-path collisions so provenance and behavior reflect exactly the supplied bundle.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
```diff
@@ -0,0 +1,246 @@
...
+            "skills": {"paths": [str(skill_root)]},
+        }
+        return SkillInstallation(
+            profiles=[overlay],
+            provenance=_provenance(skill, _hash_directory(skill_root), mode, adapter_id, str(skill_root)),
+        )
+
+    if mode == SKILL_MODE_CODEX_SKILLS_DIR:
+        skill_root = workspace_dir / CODEX_SKILLS_DIR / skill.name
+        _stage_bundle(skill.directory, skill_root)
```
ngoncharenko: **[P2] Exclude the treatment bundle from workspace evidence**

For Codex this copies the injected bundle into `workspace_dir`, which `_evidence()` later exposes wholesale to filesystem metrics. A treated run with no agent-created files is therefore non-empty solely because `.agents/skills/.../SKILL.md` exists, producing a score delta from the treatment artifact rather than agent behavior. Remove or filter the injected subtree before publishing final workspace evidence.

ngoncharenko (approved):

packages/nemo_evaluator_sdk/examples/skill_eval/run_skill_eval.py
```diff
@@ -0,0 +1,319 @@
...
+            tasks=tasks,
+            target=baseline_runtime.with_skill(
+                skill
+            ),  # We include the skill in the treated arm, so the two runs differ in *exactly* the skill.
+            config=AgentEvalRunConfig(run_id="treated", output_dir=output_dir / "treated", write_dashboard=False),
+        )
+    except RuntimeError as exc:
+        print(f"skill eval failed: {exc}")
+        print("This example needs the native NeMo Fabric stack and NVIDIA_API_KEY.")
+        return 1
```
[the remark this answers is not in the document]
SandyChapman: Addressed — `AgentSkill.from_directory(...)` now loads inside the guarded block, with an explicit `except SkillInjectionError` that prints a friendly message and returns non-zero instead of a traceback.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/metrics.py
```diff
@@ -105,6 +114,73 @@ async def compute_scores(self, input: MetricInput) -> MetricResult:
...
+        if evidence is None or evidence.get(self._trace_evidence) is None:
+            return False
+        try:
+            trajectory = await (await evidence.trace(self._trace_evidence)).trace()
+        except (KeyError, ValueError) as exc:
+            logger.warning(
+                "SkillUsedMetric scored skill_used=False: could not read trace %r: %s", self._trace_evidence, exc
+            )
+            return False
+        return _trajectory_references(trajectory, location)
```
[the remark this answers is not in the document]
SandyChapman: Addressed — broadened the best-effort `except` in `_skill_used` to `(KeyError, ValueError, ValidationError, OSError)`, so a malformed trajectory (`Trajectory.model_validate` → `ValidationError`) or a file-read error scores `skill_used=False` rather than escaping the fallback.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
SandyChapman: Fixed — the native overlay now re-lists the pre-existing skill paths (gathered from the config/profiles via `_existing_skill_paths()`) ahead of the evaluated skill, so the treated arm no longer drops preconfigured skills. Regressions: `test_install_native_preserves_existing_skill_paths` and `test_fabric_runtime_native_skill_preserves_preconfigured_skills`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
SandyChapman: Fixed — `_stage_bundle` now stages an exact copy: the evaluator-owned native stage dir is recreated (no file removed from the source survives a reused run id), and the reserved Codex path raises on a pre-existing collision instead of merging task-seeded files. Regressions: `test_install_native_recreates_stale_stage`, `test_install_codex_rejects_reserved_path_collision`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
SandyChapman: Fixed — for Codex the runtime removes the injected `.agents/skills/<name>` subtree (and prunes emptied parents) after the run, before the workspace is published as evidence, so the treatment artifact no longer skews filesystem metrics. The trajectory already captured the skill during the run, so SkillUsedMetric (reads the trace) is unaffected. Covered by `test_fabric_runtime_codex_skill_staged_for_run_then_excluded_from_evidence`.

PR #627, opened 2026-07-09
fix(docker): scope Docker cleanup and labels by controller owner

ironcommit:
## Summary

Prevents multiple Docker-backend jobs controllers from interfering with each other by scoping container lifecycle operations (cleanup, stop, kill) to the owning controller instance.

## Changes

- **Controller instance ID label**: Each controller stamps containers with a `nmp.nvidia.com/jobs_controller_instance_id` label derived from `NMP_JOBS_DOCKER_OWNER_ID` env var (if set) or a SHA-256 hash of the resolved data directory path.
- **Owner-scoped operations**: `cleanup_container`, `sync_stop_container`, and `sync_kill_container` now check `_is_container_owned_by_this_controller()` instead of only checking the `managed-by` label. Containers belonging to a different controller instance or legacy containers without an owner label are skipped.
- **Scoped cleanup filters**: `cleanup_steps` passes owner-scoped Docker label filters (`managed-by`, `controller-instance-id`, `backend`, `profile`) so the Docker daemon only returns containers relevant to this controller.
- **Consolidated base labels**: A `_base_controller_labels()` helper provides the common label set (`managed-by`, `controller-instance-id`, `backend`, `profile`) used by both job containers and storage-cleanup containers, reducing duplication.
- **Tests**: Added dedicated tests for cross-owner and legacy-container skip behavior, owner label assertions on storage-cleanup containers, and updated existing tests to include the new label.

## Summary by CodeRabbit

* **Bug Fixes**
  * Docker-managed jobs are now tracked with a controller-instance label, so stop, kill, and cleanup actions only affect containers owned by the current controller.
  * Cleanup flows now skip legacy or foreign containers more reliably, reducing accidental interruption of unrelated workloads.
  * Persistent-storage cleanup jobs now carry the same ownership metadata, improving consistency across job lifecycle operations.

### DIFF
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
@@ -253,6 +253,7 @@ class DockerJobBackend(JobBackend[ProviderT, DockerJobExecutionProfileConfig], G
                 f"{JOB_MANAGED_BY_LABEL}={JOB_MANAGED_BY_JOBS_CONTROLLER}",
                 f"{JOB_CONTROLLER_INSTANCE_ID_LABEL}={self._jobs_controller_instance_id}",
                 f"{JOB_EXECUTION_BACKEND_LABEL}={self.BACKEND_NAME}",
+                f"{JOB_EXECUTION_PROFILE_LABEL}={self._profile_name}",
             ]
         }
 
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import datetime
+import hashlib
 import io
 import json
 import logging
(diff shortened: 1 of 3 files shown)


services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -115,11 +117,21 @@ def k8s_shm_quantity_to_docker(quantity: str) -> str:
...
 
 ProviderT = TypeVar("ProviderT", bound=ExecutionProviderT)
 
 
+def _resolve_jobs_controller_instance_id() -> str:
+    configured = os.getenv(NMP_JOBS_DOCKER_OWNER_ID_ENVVAR)
+    if configured:
+        return configured
+
+    owner_source = f"nmp-data-dir:{nmp_user_data_dir().expanduser().resolve()}"
```
mckornfield, line 131: interesting... so a hash of the path with that prefix? i.e., if multiple are running on the same machine?

mckornfield (approved):

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
ironcommit: yeah, just to handle the situation where you have multiple instances talking to the same docker daemon, which is likely going to be the case during testing at least.

PR #631, opened 2026-07-09
feat(auth): implement workload identity token exchange

ironcommit:
## Post-Merge Note

The Authentik Compose merge-group CI failure from https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/29758464168/job/88408033334 is addressed separately in #788. The deterministic `authentik-blueprint-init` compose change is not part of this merged PR.

## Summary

Replace legacy static workload bearer-token envs (`NEMO_WORKLOAD_TOKEN` / `NEMO_WORKLOAD_TOKEN_FILE`) with managed workload identity token exchange. Managed job backends now inject `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`, pointing at a backend-owned subject-token file, and the SDK exchanges that subject token for a NeMo Platform access token using OAuth 2.0 Token Exchange (RFC 8693).

`NMP_ACCESS_TOKEN` remains the explicit user override. The legacy `NEMO_WORKLOAD_TOKEN` / `NEMO_WORKLOAD_TOKEN_FILE` env vars are no longer client bootstrap sources; managed job specs also reserve/reject those names and `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` so users cannot override platform-owned workload auth material.

### Workload Token Exchange

- Add NeMo auth service workload exchange endpoints: `/apis/auth/token` for RFC 8693 token exchange and `/apis/auth/jwks` for the NeMo-signed workload access-token JWKS
- Validate JWT subject tokens through configured workload subject JWKS, and validate Kubernetes projected service account tokens through the TokenReview API when enabled
- Mint short-lived NeMo-signed workload access tokens with configurable issuer, audience, TTL, signing key, subject JWKS, accepted subject issuers, and JWKS cache TTL
- Extend auth discovery with workload exchange metadata (`workload_token_exchange_enabled`, `workload_client_id`, `workload_token_endpoint`, `workload_audience`, `workload_scope`)

### SDK / Client

- Add `WorkloadTokenExchangeProvider`, which reads `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`, sends an RFC 8693 exchange request, and caches the returned access token until near expiry
- Wire automatic workload auth into both `nemo_platform_ext` and `nemo_platform_plugin` client bootstrap paths
- Keep `NMP_ACCESS_TOKEN` as the highest-precedence explicit bearer-token override
- Remove config-layer auth bootstrap from `NEMO_WORKLOAD_TOKEN` and `NEMO_WORKLOAD_TOKEN_FILE`; those legacy env vars are now ignored by client config loading
- Add client TLS env support so demo workloads can trust the local Envoy certificate without disabling verification

### Job Backends

- **Docker backend**: create a dedicated workload identity volume, obtain and refresh an Authentik subject token from the configured Docker workload identity issuer, mount the token file read-only into the workload, and inject `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`
- **Kubernetes backend**: mount a kubelet-rotated projected service account token volume with configurable audience and expiration, and inject `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`
- Thread job runtime service URLs via shared platform env vars (`NMP_BASE_URL`, `NMP_AUTH_URL`, `NMP_JOBS_URL`, `NMP_FILES_URL`, `NMP_MODELS_URL`, `NMP_SECRETS_URL`)
- Reject user-provided managed auth env vars in execution profiles and job step environments
- Move jobs-launcher log auth to workload token exchange and remove the old `NEMO_WORKLOAD_TOKEN` OTEL-header injection path

### Authentik Reference Demo

- Use the checked-in shared Authentik blueprint at `contrib/auth/authentik/helm/files/blueprints/nemo.yaml`
- Seed separate local-only identities for CLI login, workload identity (`svc-nemo`), and E2E setup (`nemo-setup`)
- Update the Compose reference under `contrib/auth/authentik/compose/` to route through Envoy, share generated local TLS/signing material, and use the NeMo auth service for workload token exchange
- Add a Helm umbrella chart under `contrib/auth/authentik/helm/` that depends on the official Authentik chart and the repo `k8s/helm` NeMo Platform chart
- Configure the Kubernetes demo chart for shared Postgres, Authentik blueprint application, Envoy auth routing, workload token signing key material, projected-token TokenReview, and workload TLS trust
- Update `run.sh` with `compose`, `k8s`, `prepare-local`, `render-blueprint`, `run-local`, and `down` flows plus reuse/diagnostics options

### Platform Config, OpenAPI, And Docs

- Add workload identity/token-exchange config fields to `OIDCConfig`
- Add Kubernetes secret/config-map volume schema support needed by the Helm demo and generated SDK
- Regenerate OpenAPI, Stainless SDK, vendored SDK, and CLI/docs surfaces for the new auth/job schemas and endpoints
- Update authentication, IdP integration, credential propagation, deployment configuration, and config-reference docs to describe `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` and the managed backend ownership model

### Hello-World Workload Task

- Simplify `workload_workspace_get` to construct `NeMoPlatform()` directly; SDK bootstrap now handles managed workload token exchange automatically

## Testing

- Unit tests for workload token exchange, JWKS publishing, subject-token validation, client workload bootstrap, token refresh behavior, TLS env handling, config no-op behavior for legacy env vars, and Docker/Kubernetes backend injection
- Jobs-launcher Go tests for workload auth token exchange and log header behavior
- Structural Authentik tests for the checked-in blueprint, Compose layout, Helm chart rendering, Envoy config, generated local inputs, and absence of legacy workload-token env injection
- Auth-idp contract tests under `tests/auth_idp/contracts` for CLI refresh, discovery, gateway routing, tokens, workspace access, and workload jobs
- Compose and Kubernetes E2E harnesses via `contrib/auth/authentik/run.sh compose` and `contrib/auth/authentik/run.sh k8s`


contrib/auth/authentik/blueprints/nemo.yaml
```diff
@@ -53,6 +53,7 @@ entries:
       client_id: nemo-platform-cli
       sub_mode: user_username
       include_claims_in_id_token: true
+      access_token_validity: "hours=1"
```
mckornfield: out of curiosity, are the names always spelled out or does h=1 work too?

contrib/auth/authentik/config/platform-compose-authentik.yaml
```diff
@@ -23,6 +25,13 @@ auth:
     client_id: "nemo-platform-cli"
     token_endpoint: "http://127.0.0.1:18080/application/o/token/"
     device_authorization_endpoint: "http://127.0.0.1:18080/application/o/device/"
+    workload_token_exchange_enabled: true
+    workload_client_id: "nemo-platform-workload"
+    workload_token_endpoint: "http://workload-token-exchange:8080/application/o/token/"
+    workload_subject_token_type: "urn:ietf:params:oauth:token-type:jwt"
```
mckornfield: ietf?

contrib/auth/authentik/kubernetes/authentik.yaml
mckornfield, line 1: intresting, so this won't be in helm yeah?

contrib/auth/authentik/kubernetes/platform-authentik-kubernetes.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+    default_scopes: "openid email offline_access groups"
+
+jobs:
+  executors:
+    - provider: cpu
+      profile: workload
+      backend: kubernetes
+      config:
+        workload_identity_token_audience: "nemo-platform"
+        workload_identity_token_expiration_seconds: __NMP_WORKLOAD_IDENTITY_TOKEN_TTL_SECONDS__
```
mckornfield: is this the templated part?

contrib/auth/authentik/kubernetes/README.md
```diff
@@ -0,0 +1,100 @@
+# Authentik Kubernetes Reference Demo
+
+This directory contains a raw Kubernetes + Kustomize version of the Authentik
+reference example. It is intended for local development clusters and uses
+demo-only secrets.
```
mckornfield: guess that answers my above q's lol. Do think a helm chart would be more consistent with what's already in the repo, no?

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
```diff
@@ -0,0 +1,218 @@
...
+ALLOWED_SUBJECT_ISSUERS = {
+    issuer.strip()
+    for issuer in os.environ.get(
+        "WORKLOAD_EXCHANGE_ALLOWED_SUBJECT_ISSUERS",
+        "http://127.0.0.1:18080/application/o/nemo-workload/,http://authentik-server:9000/application/o/nemo-workload/",
+    ).split(",")
+    if issuer.strip()
+}
+
+_private_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
```
mckornfield: so this does thjis by loading the file, and not lazily?

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
```diff
@@ -0,0 +1,273 @@
...
+
+    token_review = response.json()
+    if not isinstance(token_review, dict):
+        raise jwt.InvalidTokenError("Kubernetes TokenReview response was not an object")
+
+    status = token_review.get("status", {})
+    if not isinstance(status, dict):
+        raise jwt.InvalidTokenError("Kubernetes TokenReview response did not include a status object")
+    if not status.get("authenticated"):
+        raise jwt.InvalidTokenError(status.get("error") or "Kubernetes TokenReview rejected subject token")
```
mckornfield, line 432: maybe a nit, should we gather all the issues into a list of errors and then raise one thing at the end?

contrib/auth/authentik/docker-compose.yml
```diff
@@ -35,10 +35,13 @@ services:
...
     networks:
-      - nemo-internal
+      nemo-internal: {}
+      workload:
+        aliases:
+          - nemo-direct
 
   gateway:
-    image: envoyproxy/envoy:v1.33-latest
+    image: envoyproxy/envoy@sha256:f91c972d5c99bc133233a079b5663b903e7d56b3b0b0216398924f7b80d09e47
```
mckornfield: better sec or?

contrib/auth/authentik/run.sh
```diff
@@ -186,14 +250,118 @@ run_tests() {
...
+      - nemo.yaml=blueprints/nemo.yaml
+  - name: authentik-envoy-config
+    files:
+      - envoy.yaml=gateway/envoy.yaml
+  - name: workload-token-exchange-broker
+    files:
+      - workload_exchange_broker.py=workload-exchange-broker/workload_exchange_broker.py
+
+images:
+  - name: my-registry/nmp-api
```
mckornfield: I guess this part can't be templated?

packages/nemo_platform_ext/src/nemo_platform_ext/client/enhanced.py
```diff
@@ -100,8 +101,13 @@ def __init__(
         """
         # Backward compatibility: an explicit base_url means direct mode (no config bootstrap),
         # unless config-specific overrides are provided.
+        has_workload_identity_token_file = bool(os.environ.get("NMP_WORKLOAD_IDENTITY_TOKEN_FILE"))
```
mckornfield: so if this is set, and we have no http client, we _should_ bootstrap?

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/oidc_factory.py
```diff
@@ -23,11 +23,13 @@
     DEFAULT_REFRESH_MARGIN_SECONDS,
     OIDCTokenProvider,
     TokenSet,
+    WorkloadTokenExchangeProvider,
     _discover_oidc_client_settings,
     build_effective_scope,
 )
 
 logger = logging.getLogger(__name__)
+WORKLOAD_IDENTITY_TOKEN_FILE_ENVVAR = "NMP_WORKLOAD_IDENTITY_TOKEN_FILE"
```
mckornfield: nit- do we want to centralize this env var?

mckornfield (commented):
fun pr lol

personally my beef is just the authentik bit using kustomize, I'd rather it just be in helm if it's all the same, since we already have a helm chart in the repo (understanding it's a demo object, but unless you have a strong reason it should be kustomize, idk why we wouldn't just use helm)

contrib/auth/authentik/blueprints/nemo.yaml
ironcommit: It's specific to authentik, and it needs to be spelled out. It's not well documented by authentik, but you can see the parser here: https://raw.githubusercontent.com/goauthentik/authentik/main/authentik/lib/utils/time.py

contrib/auth/authentik/kubernetes/README.md
```diff
@@ -0,0 +1,91 @@
+# Authentik Kubernetes Reference Demo
```
crookedstorm: They don't provide one for us to use?

contrib/auth/authentik/kubernetes/authentik.yaml
ironcommit: No, this will be a separate chart eventually.  It's kustomize at the moment, but I plan to switch it to helm shortly.  Just not part of the main helm chart.

contrib/auth/authentik/kubernetes/README.md
crookedstorm: This is a lot to maintain vs their documented helm install they maintain? https://docs.goauthentik.io/install-config/install/kubernetes/#install-authentik-helm-chart

crookedstorm: https://github.com/goauthentik/helm

.github/workflows/ci.yaml
```diff
@@ -993,10 +993,128 @@ jobs:
...
+        with:
+          name: policy-wasm
+          path: services/core/auth/src/nmp/core/auth/assets
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+        with:
+          python-version: "3.11"
+          enable-cache: true
+          cache-dependency-glob: uv.lock
+      - name: Install kind
```
crookedstorm: There is already a composite action to setup a kind cluster in `.github/actions`.  It also installs the nemo-platform, but it can be made more generic if needed. I'd rather not duplicate it all a lot.

contrib/auth/authentik/docker-compose.yml
ironcommit: Just wanting to pin it so it doesn't drift and get's cached.

crookedstorm:
If there's something wrong with the way the helm chart installs the api container, please change the helm chart. I don't see why we are going to all this trouble to install authentik in what seems like an unusual way in addition to an apparently unusual install of the api for a test like this. There must be more context I don't know here?

contrib/auth/authentik/config/platform-compose-authentik.yaml
ironcommit: Removed.  We only support the one type of exchange at the moment.

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/oidc_factory.py
ironcommit: done

.github/workflows/ci.yaml
ironcommit: fixed.

contrib/auth/authentik/run.sh
ironcommit: fixed

contrib/auth/authentik/kubernetes/README.md
ironcommit: updated to use authentik official helm chart.

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
ironcommit: updated

.github/actions/setup-kind-cluster/action.yaml
```diff
@@ -1,13 +1,17 @@
...
-  pre-pulls images, deploys NeMo Platform via Helm, and waits for the
-  API to become healthy.
+  optionally deploys NeMo Platform via Helm, and waits for the API to become
+  healthy.
 
 inputs:
   kind_cluster_name:
     description: Kind cluster name
     required: true
+  install_nemo_platform:
```
crookedstorm: awesome!

crookedstorm: By convention, all inputs should be kebab-case. The fact that the old inputs in this are not means I'd humbly ask if you can fix those as well.

.github/workflows/ci.yaml
```diff
@@ -993,10 +993,101 @@ jobs:
...
+      - name: Verify official Authentik Helm chart is reachable
+        shell: bash
+        run: |
+          helm show chart authentik --repo https://charts.goauthentik.io --version 2026.5.4
+      - name: Download policy WASM
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: policy-wasm
+          path: services/core/auth/src/nmp/core/auth/assets
+      - name: Pre-pull auth-idp image into Kind
```
crookedstorm: This is probably unnecessary. The prepulling stuff for containers is mostly because of bugs and quirks in NGC. I don't see the same problems with docker.io generally or other registries.

.github/workflows/ci.yaml
```diff
@@ -993,10 +993,101 @@ jobs:
...
+          image_registry: ${{ needs.build-cpu-smoke-images.outputs.image_registry }}
+          image_tag: ${{ needs.build-cpu-smoke-images.outputs.image_tag }}
+          kind_image_pull_token: ${{ github.token }}
+          kind_image_pull_user: ${{ github.actor }}
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+      - name: Lint Authentik umbrella Helm chart
+        shell: bash
+        run: |
+          helm repo add authentik https://charts.goauthentik.io --force-update
```
crookedstorm, line 1069: It would be a bit more predictable for CI if we set a version rather than force the absolute latest.

contrib/auth/authentik/kubernetes/README.md
ironcommit: now using authentik helm

.github/actions/setup-kind-cluster/action.yaml
ironcommit: updated

crookedstorm (approved):
LGTM

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
anastasia-nesterenko, line 304: should shutdown stop _workload_identity_refreshers?

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
```diff
@@ -0,0 +1,529 @@
...
+
+        try:
+            async with httpx.AsyncClient(timeout=30.0) as client:
+                response = await client.get(jwks_uri)
+            response.raise_for_status()
+        except httpx.HTTPError as exc:
+            raise jwt.InvalidTokenError(f"Subject token JWKS request failed: {exc}") from exc
+
+        jwks = response.json()
+        if not isinstance(jwks, dict):
```
anastasia-nesterenko, line 269: could it pass this but be invalid?

ironcommit: Yes, it could previously pass the object check but still be an invalid JWKS, and that invalid value could be cached. I updated the fetch path to validate the JWKS with `jwt.PyJWKSet.from_dict()` before caching it.

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
ironcommit: Good catch. I updated `DockerJobBackend.shutdown()` to stop all active workload identity refreshers before closing the Docker client. Shutdown now attempts every refresher, logs stop failures, clears the refresher registry, and still closes the client.

PR #632, opened 2026-07-10
feat(studio): compact latency duration format across experiments and intake

rrhyne:
## What

Latency was rendered as long raw-millisecond numbers and formatted three different ways across the app. This consolidates all latency/duration rendering onto one shared formatter that produces compact segments:

- `10m 12s 13ms`
- `12s 34ms`
- `34ms`

## Why

Before, three formatters coexisted:
- **Experiment surfaces** (group list "Avg Latency", detail header, session/task table) inlined `` `${Math.round(ms)} ms` `` — long unscaled milliseconds.
- **Intake surfaces** used `formatDurationMs` (ms/s/min, 2 decimals).

## Changes

- **New `web/packages/studio/src/util/duration.ts`** — `formatDuration(ms)`: renders highest non-zero unit down to lowest, drops leading/trailing zero units, keeps interior zeros (`1h 0m 5s`), supports hours, preserves sub-ms precision (`0.34ms`), returns `—` for null/undefined.
- **`intakeTelemetry.ts`** — `formatDurationMs` now aliases `formatDuration`, so all six intake trace/span surfaces adopt the new format with zero call-site churn.
- **`ExperimentGroupDataView`, `ExperimentDetailMetrics`, `ExperimentSessionsDataView`** — call `formatDuration` instead of inline millisecond strings.
- **`SpanTriggerMeta.tsx`** — removed a `.replace(/\s+/g,'')` that would collapse the segmented format into `10m12s13ms`.

## Testing

- New `duration.test.ts` — 10/10 pass (covers the examples above plus edges: hours, interior zeros, rounding, sub-ms, zero/negative).
- ESLint clean on all touched files.
- Typecheck: no new errors introduced (the `SpanEvaluationContext` errors in `intakeTelemetry.ts` are pre-existing on `main` from an SDK schema change, unrelated to this PR).

## Out of scope

The Agent Monitor inference-logs table has its own route-local `formatDuration` (ms/s/min, 1 decimal); left untouched intentionally.

## Summary by CodeRabbit

- **Enhancements**
  - Improved duration and latency formatting across experiment metrics, sessions, and intake details by using a shared formatter for consistent, compact time units.
  - Experiment latency and “Avg Latency” now display human-readable durations with standardized handling for missing values.
  - Intake duration display no longer strips whitespace, reflecting the formatted output directly.
- **Tests**
  - Added dedicated test coverage for the shared duration formatter and updated duration-related UI expectations.


web/packages/studio/src/util/duration.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Format a duration given in milliseconds as compact human-readable segments,
+ * e.g. `10m 12s 13ms`, `12s 34ms`, or `34ms`.
+ *
+ * Segments run from the highest non-zero unit down to the lowest non-zero unit
+ * (leading and trailing zero units are dropped, interior zeros are kept so
+ * `1h 0m 5s` stays unambiguous). Sub-millisecond values keep two decimals of
+ * precision (`0.34ms`) so span timings are not rounded away.
+ */
+export const formatDuration = (value: number | null | undefined): string => {
```
walston: one red flag:
[`formatTimeInSeconds`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/web/packages/common/src/utils/date.ts#L47-L56) already exists. it might be better to extend that with an "include ms" field argument if you need it. or just auto-include it if the time is sub second and drop it if it's more than 1 second.

walston (changes requested):
Open to pushback, just don't want parallel solutions to the same problem sprinkled throughout the codebase.

web/packages/studio/src/util/duration.ts
rrhyne: Ok, will exend!

rrhyne: Done in 1600293. Consolidated into `@nemo/common/src/utils/date`: added an ms-aware `formatDurationMs` primitive right next to `formatTimeInSeconds`, and refactored `formatTimeInSeconds` to delegate to it so there's a single segmentation implementation. `formatDurationMs` drops zero-valued units (including interior ones, e.g. `1h 5s`) to match `formatTimeInSeconds`, keeps sub-ms precision, and adds an `ms` component (`10m 12s 13ms`, `12s 34ms`, `34ms`). Deleted the studio-local util; `formatTimeInSeconds`'s existing tests still pass unchanged.

I kept it as a separate ms-input primitive (rather than an `includeMs` flag on `formatTimeInSeconds`) because latency fields are in milliseconds and the design calls for showing ms even above 1s (`12s 34ms`), so the "drop ms if > 1s" variant wouldn't fit.

walston (approved):

web/packages/studio/src/routes/ExperimentDetailRoute/ExperimentDetailMetrics.tsx
```diff
@@ -21,9 +22,7 @@ export const ExperimentDetailMetrics: FC<ExperimentDetailMetricsProps> = ({ expe
     experiment?.cost_usd?.mean != null ? `$${experiment.cost_usd.mean.toFixed(3)}` : undefined;
 
   const avgLatency =
-    experiment?.latency_ms?.mean != null
-      ? `${Math.round(experiment.latency_ms.mean)} ms`
-      : undefined;
+    experiment?.latency_ms?.mean != null ? formatDurationMs(experiment.latency_ms.mean) : undefined;
```
walston: why do a pre-check before passing to `formatDurationMs` since it also handles invalid values by returning `'-'`?

If return value isn't useful maybe it shouldn't return it; either that or we should use the built-in error handling (of `formatDurationMs`)

rrhyne: Good catch — removed the pre-check in e9ccd7c. `formatDurationMs` returns `—` for null/undefined, and that's exactly `KVPair`'s own `defaultValue`, so the guard produced identical output. Now just `formatDurationMs(experiment?.latency_ms?.mean)`.

Note I kept the explicit `: '-'` fallback in the two experiment *table* cells (ExperimentGroupDataView / ExperimentSessionsDataView) on purpose: every sibling column in those tables renders empty as `-` (hyphen), whereas `formatDurationMs` returns `—` (em dash). Dropping the guard there would make the latency cell inconsistent with its neighbors. Happy to switch the whole table to `—` if you'd prefer that direction instead.

web/packages/common/src/utils/date.ts
```diff
@@ -40,17 +40,42 @@ export const utcToLocalDate = (utcIsoString?: string): Date | undefined => {
...
-  const hours = Math.floor(seconds / 3600);
-  const minutes = Math.floor((seconds % 3600) / 60);
-  const remainingSeconds = Math.floor(seconds % 60);
-  return [hours, minutes, remainingSeconds]
-    .map((val, idx) => `${val}${idx === 0 ? 'h' : idx === 1 ? 'm' : 's'}`)
-    .filter((val) => parseInt(val) > 0)
-    .join(' ');
+  // Whole seconds only — this variant never renders a sub-second (ms) component.
+  return formatDurationMs(Math.floor(seconds) * 1000);
 };
```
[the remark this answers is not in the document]
rrhyne: Confirmed and fixed in e9ccd7c. `formatTimeInSeconds(0.9)` was returning `0ms` because `Math.floor(0.9) * 1000` collapses to `0`. Added a `seconds < 1` guard so fractional-second (and negative) inputs render `''` as before, with a regression test covering `0.9` and `-5`.

PR #638, opened 2026-07-10
feat(guardrails): expose chat completions inference endpoint

aray12:
Register the previously-unmounted chat completions router on the guardrails
service so guardrailed inference is reachable without a VirtualModel. Enables
ad-hoc testing of a guardrail config against a target model with OpenAI-
compatible streaming. Regenerate the OpenAPI spec to include the new route.

Signed-off-by: Alex Ray <alray@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Added an OpenAI-compatible chat completions endpoint for Guardrails.
  * Supports standard JSON and streaming responses.
  * Added support for messages, tool calls, function calls, token usage, and log probabilities.
  * Documented request formats, response schemas, examples, and standard error responses.

PR #639, opened 2026-07-10
feat(studio): add chat playground to guardrail config UI

aray12:
Split the guardrail detail page into Details and Chat tabs (Details is the
default sub-route). The Chat tab pairs a model picker and inference-params
popover with a streaming chat thread that routes through the guardrails chat
completions endpoint, applying the config's rails via a config_ids body field.

Thread an optional bodyExtra prop through the shared chat components so a chat
can inject extra request-body fields without forking the runtime.

Signed-off-by: Alex Ray <alray@nvidia.com>


aray12:

- #638
    - #639 ◀

PR #644, opened 2026-07-10
feat(agents): route gateway to container-mode agents via Deployment e…

benmccown:
### What & why

Teaches the agents-plugin gateway proxy to route to **container-mode** (docker/k8s) agents by resolving their address from the projected `AgentDeployment.endpoints`, instead of the hardcoded loopback `endpoint`. Subprocess mode is unchanged and stays the default.

This is the stop-gap that unblocks agents-in-k8s: without it, a containerized agent can be deployed but inbound traffic can't reach it, since the gateway is hardwired to a loopback `127.0.0.1:<port>` that's only valid for a same-host subprocess.

### Changes

- **`entities.py`** — promote `endpoint` → `endpoints: list[Endpoint]` and add a `deployment_mode` discriminator (`subprocess|docker|k8s`, default `subprocess`). `endpoint` stays as the subprocess loopback compat shim for one release.
- **`gateway.py`** — add `_get_deployment_endpoint` / `_is_deployment_routable`, the single place resolution branches on mode: container modes read the first http(s) URL from `endpoints`; subprocess reads `endpoint`. Routable once status is `running`; not-ready → `503`, missing → `404` (unchanged contract).
- The SSRF origin guard is origin-relative (not a host allow-list), so it works unchanged for k8s Service DNS and docker `host:port`.

### Out of scope

- Load-balancing across deployments; per-deployment authN.
- **Possible future direction (not committed to):** folding agent routing into the Inference Gateway would likely retire this proxy. This stop-gap stands on its own.

### Coordination

The container producer path (compile → `DeploymentConfig`, status/endpoint projection) lands separately and this PR consumes it. Preferably that change absorbs the self-contained `entities.py` hunk, after which this branch rebases and drops it. It's assumed the controller projects the deployments-plugin `READY` onto the agents-local `running`.

### Testing

- **Unit** (49 pass): subprocess unchanged (regression), docker/k8s resolves from `endpoints` (both routes), `503` not-ready, `404` missing, Service-DNS origin guard. `ruff` + `ty` clean.
- **Manual E2E (docker mode, dev pod):** seeded a container-mode `AgentDeployment` pointing at a docker-in-docker mock agent; reached it `200` via both agent-name and deployment-name routes; verified `503` (not-ready) and `404` (missing). k8s Service-DNS routing is covered by unit tests.

## Summary by CodeRabbit

* **New Features**
  * Gateway proxy now supports container-based deployments by resolving and routing to the deployment’s available HTTP/HTTPS endpoints.
  * Agent-name routing now selects only “routable” (running/ready) deployments with resolvable endpoints.
* **Bug Fixes**
  * Improved 503 responses for unroutable/pending deployments, including clearer details about routability and deployment state.
  * Added defensive handling when a resolved proxy target is unexpectedly unavailable.
* **Tests**
  * Added unit coverage for container-mode endpoint resolution for both deployment-name and agent-name proxying, plus SSRF/origin guard validation.
* **Chores**
  * Updated plugin installation to enable docker/k8s extras for deployment support.


tylersbray (approved):
👍

PR #648, opened 2026-07-13
refactor(evaluator-sdk): migrate Fabric agent-eval runtime to the updated Fabric SDK

SandyChapman:
## What & why

Aligns the **host** `FabricAgentRuntime` (`agent_eval/runtimes/fabric/`) with the current NeMo Fabric Python SDK. The SDK renamed its client, moved to a config-first composition model, and introduced the `RunOutput` response contract — this branch migrates our runtime to match and keeps everything importable/testable without the (still optional, unlocked) `nemo-fabric` native package.

## Key changes

- **`FabricClient` → `Fabric`.** `Fabric` is a plain reusable facade (not an async context manager); per-invocation `input` + `request_id` now go through `RunRequest`.
- **Config-first composition.** Per-task workspace / model / trajectory settings are applied directly onto a copied `FabricConfig` (`model_copy` + `environment` + `enable_relay`) instead of layered profile overlays. The old profile-builder helpers are removed; caller-supplied `profiles=` still pass through.
- **`RunOutput` response contract (#52).** `RunResult.output` is now a `RunOutput` (a `Mapping`), not a raw JSON value. `_normalize_output` copies it to a plain dict so it round-trips through the trial's `JsonValue` response field.
- **Lazy imports, not threaded types.** `nemo_fabric` stays an optional lazy import (guarded in `run_tasks`); helpers re-import the two runtime types locally where used rather than threading `type[...]` params through signatures. Annotations name the real SDK types via the `TYPE_CHECKING` import.
- **Dev script fix.** `script/dev-install-fabric.sh` smoke check updated `FabricClient` → `Fabric`.

## Verification

- **Hermetic tests** (fake `nemo_fabric`, run in CI): green — runner → evaluator → metric → evidence chain.
- **Live end-to-end**: `test_fabric_codex_live_eval_captures_atif_trajectory` **passed** — a real codex agent run through the migrated runtime → relay → ATIF trajectory → metric.
- `ruff` (check + format), copyright headers, merge-conflict, and `ty` (CI config) clean for all changed files.

## Notes

- Only the **codex** harness was exercised live this session; the Hermes-SDK path is covered by the hermetic tests but not a live run.
- `nemo-fabric` remains an optional lazy import because there is no published, cross-platform, indexed wheel to lock yet (its CI builds Linux-only artifacts, unpublished). CI stays on the `ty: ignore[unresolved-import]` path; local dev uses `script/dev-install-fabric.sh`.
- The **container** runtime (`container_runtime.py`, #604) is a separate, unmerged branch and still needs the same migration — out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Improved agent evaluation runs with a config-first Fabric setup and a reusable Fabric session.
  * Enhanced trajectory/relay capture using task-specific workspace and per-task evidence/artifact locations.
  * Normalized structured Fabric run outputs into standard trial response fields.
* **Bug Fixes**
  * Added fail-fast validation for trajectory capture dependencies when relay is enabled.
* **Tests**
  * Updated Fabric runtime/integration tests and hermetic Fabric fakes to match new config composition, workspace handling, relay settings, and output normalization.
* **Chores**
  * Updated the Fabric dev-install verification script to use the current public Fabric API.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
@@ -7,7 +7,11 @@
 NeMo Fabric Python SDK and adapts each normalized Fabric ``RunResult`` into an
 :class:`AgentEvalTrial`. The harness is chosen by the supplied Fabric config's
 ``harness.adapter_id`` (never inferred from a model); an optional ``model`` slug
-is applied as a final profile overlay, mirroring Fabric's own Harbor integration.
+is applied as the config's default model, mirroring Fabric's own Harbor integration.
+
+Per-task settings (workspace, model, trajectory capture) are composed directly onto
+a copy of the supplied config via the SDK's config helpers (``model_copy`` +
+``enable_relay`` + ``environment``), rather than layered as profile overlays.
 
 Every task runs in its own fresh workspace: the runtime seeds it from
 ``inputs['files']`` (a no-op when there are none), runs the harness in it (via
@@ -38,6 +42,7 @@ from nemo_evaluator_sdk.values.evidence import (
     CandidateEvidence,
     EvidenceDescriptor,
 )
+from pydantic import JsonValue
 
 if TYPE_CHECKING:
     # Annotations use nemo_fabric's real types (single source of truth). nemo_fabric is an optional
@@ -45,9 +50,10 @@ if TYPE_CHECKING:
     # loaded lazily at runtime (see ``run_tasks``). Drop the ty:ignore once nemo-fabric is a
     # resolvable dependency and the checker can see it.
     from nemo_fabric import (  # ty: ignore[unresolved-import]
-        FabricClient,
+        Fabric,
         FabricConfig,
         FabricProfileConfig,
+        RunOutput,
         RunResult,
     )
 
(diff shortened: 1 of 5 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -121,69 +122,72 @@ async def run_tasks(
...
+                # ``Fabric.run`` folds the per-invocation input + request id into a ``RunRequest``.
+                # Caller-supplied ``base_profiles`` still pass through; this runtime's per-task settings
+                # live on ``task_config`` (config-first) rather than in profile overlays.
                 client.run(
-                    agent_config,
-                    profiles=profiles,
-                    input=_fabric_input(task, seeded_files),
-                    request_id=task.id,
+                    task_config,
+                    profiles=base_profiles,
```
ngoncharenko: Caller profiles are applied by Fabric after _compose_config(), so they can overwrite the evaluator-owned workspace, model, and Relay settings set in the _compose_config()

More details by an agent:
---
They are passed here in runtime.py:186:

```python
client.run(
    task_config,
    profiles=base_profiles,  # caller profiles
    ...
)
```

Inside Fabric, `run()` calls `plan()` with those profiles. Fabric’s resolver then applies each profile over `task_config`:

```rust
let mut effective = config;
for profile in profiles {
    apply_profile_config(&mut effective, profile)?;
}
```

`apply_profile_config()` recursively merges the profile as the overlay, so its specified values win. See [[Fabric’s resolver](https://github.com/NVIDIA/NeMo-Fabric/blob/8d7ad4ab819dfc9cb69fc964e691565bfc3c7bbe/crates/fabric-core/src/config.rs#L854-L863)](https://github.com/NVIDIA/NeMo-Fabric/blob/8d7ad4ab819dfc9cb69fc964e691565bfc3c7bbe/crates/fabric-core/src/config.rs#L854-L863).

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: I adjusted it so that the required evaluator configs go last. Nice find.

PR #649, opened 2026-07-13
ci: pin wheel install-test deps to uv.lock (supply-chain)

SandyChapman:
## Problem

The wheel install-smoke-test jobs (`nemo-platform wheel build + test`, `nemo-platform-plugin wheel build + test`) install the freshly-built wheel with **no dependency pinning**:

```bash
uv tool install --force --python "${PYTHON_VERSION}" "${WHEEL}[services]"   # nemo-platform
uv pip install  --python .venv/bin/python "${WHEEL}"                        # plugin
```

Neither uses `--constraints`, `uv.lock`, `--frozen`, or hashes, so they resolve the **entire transitive closure fresh from PyPI, latest-in-range**. That's:

- **A supply-chain risk** — these jobs then *execute* the installed code (CLI smoke test / imports) in the runner, so a compromised newer in-range release would run with the runner's permissions/secrets.
- **Non-reproducible** — a run's dependency set drifts as PyPI publishes new releases.

This is what let `litellm==1.92.0` slip in over the locked `1.90.2` and turn the py3.14 job red (1.92.0 added a Rust/PyO3 core that fails to build on Python 3.14). ([see this CI link](https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/29261103381/job/86855558947))

> Note: the rest of CI (unit/integration tests, builds) already runs `--frozen` against `uv.lock`; this gap was specific to these two install-test steps.

## Fix

Export the locked dependency set from `uv.lock` and pass it as `--constraints` to both installs:

```bash
uv export --frozen --all-packages --no-emit-workspace --no-emit-local --no-hashes \
  -o "${RUNNER_TEMP}/wheel-constraints.txt"
uv tool install --force --python "${PYTHON_VERSION}" \
  --constraints "${RUNNER_TEMP}/wheel-constraints.txt" "${WHEEL}[services]"
```

- `--all-packages` pins the full external dependency universe of the workspace.
- `--no-emit-workspace --no-emit-local` drop the local/workspace path & editable entries (which aren't installable as constraints), leaving only PyPI-pinned versions.
- Now the smoke test installs **only versions vetted in `uv.lock`**; a newer (potentially malicious) in-range release can no longer be pulled.

**Side effect:** this also fixes the py3.14 wheel+test failure — the constraints force `litellm==1.90.2` (pure-Python wheel, installs cleanly on 3.14).

## Verification

- `uv pip compile` with the exported constraints resolves `litellm==1.90.2`; without it, `1.92.0`.
- The exported constraints file has **0** local/editable/path entries (clean, install-safe).
- `[services]` extra has no heavy/native deps (no torch/vllm/flashinfer), so pinning to the lock is safe across the Python matrix.
- Workflow YAML validated; `Run actionlint` in CI covers the shell.

## Follow-up (not in this PR)

Hash-pinning (`--require-hashes`) is stronger defense-in-depth but needs the local wheel hashed too; version-pinning here closes the practical "auto-pull a newer release" vector first.

## Summary by CodeRabbit

* **Chores**
  * Improved CI wheel smoke tests by installing dependencies from committed, hash-pinned lock/constraint files.
  * Updated the wheel install/validation flow to install the wheels without resolving new dependencies.
  * Added lockfile artifacts for the platform services and plugin wheels, plus a helper script to regenerate them for consistent, reproducible builds.


crookedstorm (approved):
Looks like an improvement to me.

PR #650, opened 2026-07-13
feat(datasets): add DatasetProfile contract

albcui:
## Summary by CodeRabbit

* **New Features**
  * Introduced a versioned dataset profiler storage contract to persist dataset metadata, sampling details, per-partition feature schemas, column statistics, and objective classifications, including optional verifiability and evidence.

* **Tests**
  * Added fixture-driven coverage validating schema version defaults, lossless JSON round-trips, dataset-specific contract shape constraints, deep semantic-role parsing, permissive handling of open vocabularies, forward-compatible parsing of unknown fields, and correct construction of quantiles and message statistics.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
@@ -118,7 +118,13 @@ class MessageStats(BaseModel):
     content_chars: Quantiles = Field(description="Per-row total content length -> chat sequence length.")
     roles_seen: list[str] = Field(
         default_factory=list,
-        description='e.g. ["system", "user", "assistant", "tool"].',
+        description=(
+            'The distinct role strings actually present in the sampled rows, verbatim — e.g. ["system", '
+            '"user", "assistant", "tool"], but equally ShareGPT\'s ["human", "gpt"] or a house convention. '
+            "A measurement of row content, not a vocabulary the profiler picks from, so it is deliberately "
+            "not an enum: an unexpected role is the finding worth reporting, and normalizing or dropping it "
+            "would hide exactly what a consumer needs to see before choosing a chat template."
+        ),
     )
     ends_with_assistant_rate: float = Field(
         ge=0.0,
@@ -283,10 +289,20 @@ class SplitProfile(BaseModel):
             "train, with the variant's intent kept in `name`."
         ),
     )
-    files: list[FileRecord]
+    files: list[FileRecord] = Field(
+        description=(
+            "Every file resolved into this split, measured. Partitioning is exhaustive and disjoint: each "
+            "file of the partition lands in exactly one split, so concatenating `files` across splits "
+            "reconstructs the partition's file list with no gaps or repeats."
+        ),
+    )
     num_examples: int | None = Field(
         default=None,
-        description="Exact when exhaustive / from parquet footers, else estimated.",
+        description=(
+            "Rows in this split, counting every file in `files` whether or not it was scanned. Exact when "
+            "read from parquet footers or an exhaustive scan, otherwise extrapolated from the rows sampled — "
+            "check `SamplingInfo.exhaustive` before treating it as a fact. None when nothing usable was found."
+        ),
     )
 
 
@@ -344,12 +360,19 @@ class SamplingInfo(BaseModel):
     rows_scanned: int = Field(description="Total rows actually parsed across all files.")
     rows_total: int | None = Field(
         default=None,
-        description="Exact when cheaply known (parquet footer, exhaustive scan).",
+        description=(
+            "How many rows the whole fileset holds, scanned or not — the denominator `rows_scanned` is a "
+            "fraction of, so a consumer can judge how representative the stats are. Populated only when the "
+            "count is exact and cheap (summed parquet footers, or an exhaustive scan); None means unknown, "
+            "never zero and never an estimate."
+        ),
     )
     files_scanned: int = Field(
         description=(
-            "Every file should be probed (head-sampling a subset of files hides late columns / schema "
-            "drift); < the fileset's file count only when scale forces file-level sampling."
+            "How many files were opened and read from (a count, not a list — the files themselves are "
+            "`SplitProfile.files`). Every file should be probed, since head-sampling a subset hides columns "
+            "that appear only in later shards; expect this to equal the fileset's file count, and be lower "
+            "only when scale forces file-level sampling."
         ),
     )
     per_file_row_cap: int | None = Field(default=None, description="Cap that bounded per-file reads, if any.")
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
@@ -185,6 +185,19 @@ class FeatureSchema(BaseModel):
     fields: list[FeatureSchema] | None = Field(default=None, description="dtype == struct: named child fields.")
     items: FeatureSchema | None = Field(default=None, description="dtype in {list, messages}: element schema.")
 
+    @model_validator(mode="after")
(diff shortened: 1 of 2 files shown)


packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,353 @@
...
+class Verifiability(BaseModel):
+    """A found verification target. Present only when one exists; absence *is* the claim (not verifiable)."""
+
+    method: str = Field(
+        description="extractable_final_answer | ground_truth_column | constraint | test_cases",
+    )
+    coverage: float | None = Field(
+        default=None,
+        description="Fraction of sampled rows with a usable verification target.",
+    )
```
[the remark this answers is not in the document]
albcui: can you re-review?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,353 @@
...
+    fixed_length: int | None = Field(
+        default=None,
+        description=(
+            "dtype == list: constant observed element count (e.g. an embedding vector's 768), None when "
+            "variable. Multi-dimensional shapes compose via nesting."
+        ),
+    )
+    fields: list[FeatureSchema] | None = Field(default=None, description="dtype == struct: named child fields.")
+    items: FeatureSchema | None = Field(default=None, description="dtype in {list, messages}: element schema.")
```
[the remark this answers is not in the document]
albcui: Can you re-review?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,353 @@
...
+    distinct_count: int = Field(
+        description=(
+            "Distinct values among scanned rows: ~=rows_scanned -> id-like; a small bounded set "
+            "corroborates score / category roles."
+        ),
+    )
+    values: list[str] | None = Field(
+        default=None,
+        description="The proven enumeration; only when the scan was exhaustive and distinct_count <= 32.",
+    )
```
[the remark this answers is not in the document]
albcui: I think you mis-understood the design here -- the producer already stringifies `(sorted(str(v) for v in distinct), stats.py:100)`, so list[str] is the intended uniform representation. Widening to str | int would break that invariant.

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,366 @@
...
+    strategy: str = Field(
+        description=(
+            "full | stratified_probes | random. Kept explicit alongside `exhaustive` because, with an open "
+            "strategy vocabulary, consumers can't derive exhaustiveness from the strategy name alone."
+        ),
+    )
+    rows_scanned: int = Field(description="Total rows actually parsed across all files.")
+    rows_total: int | None = Field(
+        default=None,
+        description="Exact when cheaply known (parquet footer, exhaustive scan).",
```
anubhutivyas: can we explain a bit more on what this field means?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,366 @@
...
+    )
+    files_scanned: int = Field(
```
anubhutivyas, line 370: Is it a list of files scanned, the desc is not that descriptive of what this field has.

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,366 @@
...
+    name: str = Field(description="The on-disk name: train | test | train_prefs | ...")
+    canonical: str | None = Field(
+        default=None,
+        description=(
+            "Normalized concept: train | validation | test; None when nothing matches. E.g. train_prefs -> "
+            "train, with the variant's intent kept in `name`."
+        ),
+    )
+    files: list[FileRecord]
+    num_examples: int | None = Field(
```
anubhutivyas, line 299: can the desc be a bit more descriptive?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+class FileRecord(BaseModel):
+    """One physical file, measured.
+
+    Stores the exact digest inputs (so a profile self-describes its ``content_digest`` and per-file
+    staleness is computable) plus what the reader learned cheaply.
+    """
+
+    path: str = Field(description="Relative path within the fileset.")
```
anubhutivyas, line 252: should this be of type `Path`?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -0,0 +1,366 @@
...
+
+    chars: Quantiles = Field(description="Per-row character-length distribution.")
+
+
+class MessageStats(BaseModel):
+    """Measurements for a ``messages`` column (a list of ``{role, content}``)."""
+
+    turns: Quantiles = Field(description="Per-row turn count (p99 -> packing long chats).")
+    content_chars: Quantiles = Field(description="Per-row total content length -> chat sequence length.")
+    roles_seen: list[str] = Field(
```
anubhutivyas, line 119: can be an enum?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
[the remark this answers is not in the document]
albcui: Yes wrt your comment on `FeatureSchema`, but no wrt `stats`. Can you re-review my changes?

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
albcui: This is the total number of rows that contribute to the data profile. However, the subtle thing is that sometimes we don't need to scan the entire dataset, or scan at all. With parquet files, we can cheat a bit because parquet has built-in schema

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
albcui: Good catch, updated with a better description. It's the number of files not the list of file names.

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
albcui: Oof, `files` had no description at all. Added a description to `files`, and updated the description in `num_examples` to have the distinction between when files are scanned or not (e.g. parquet).

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
albcui: At this layer of the code (the output of the dataset profiler), we don't really need `Path`. `Path` will be used internally by the dataset profiler engine/script itself for robust file path resolution. However, in the exported dataset profile here, we actually want to control exactly the value of the string we are exporting. `Path` could potentially overwrite the string representation.

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
albcui: `roles_seen` is open-ended. This is the role strings actually extracted from the dataset itself, which can be arbitrary.

anubhutivyas (approved):

PR #653, opened 2026-07-13
feat(experiments): Update metadata shape and add filter support

shanaiabuggy:
Updates Experiment.metadata and ExperimentGroup.metadata type to be `dict[str, str]` and adds filter support.

## Summary by CodeRabbit

- **New Features**
  - Added metadata key/value filtering for experiment and experiment-group listings.
  - Documented rollup metric range filter examples for experiment searches.

- **Improvements**
  - Standardized experiment metadata values as strings across requests, responses, and stored records.
  - Preserved compatibility with existing metadata by converting legacy non-string values during deserialization.

- **Tests**
  - Added coverage for metadata filtering and legacy metadata conversion.


services/intake/src/nmp/intake/entities/experiments.py
```diff
@@ -10,13 +10,23 @@
...
 from nmp.common.entities.client import EntityBase
 from pydantic import AnyUrl, Field, field_validator
 
 
+def _stringify_metadata(value: Any) -> Any:
+    """Schema-on-read coercion for ``metadata``, which tightened from ``dict[str, Any]`` to
+    ``dict[str, str]``. Legacy rows may hold non-string values; stringify them (JSON-encoding
+    structured values) so old rows still read. Non-dict values pass through for pydantic to handle."""
+    if not isinstance(value, dict):
+        return value
```
shanaiabuggy, line 26: why do we allow this shape?

shanaiabuggy: Non-dict values fall through to pydantic's dict[str, str] | None validation, which rejects anything that isn't a dict or null (422) — so a bad shape already errors here; the passthrough just delegates the raise. Only None (unset) legitimately passes. Leaving as-is.

BrianNewsom (approved):

PR #658, opened 2026-07-13
feat(profiler): profiler plugin for datasets

albcui:
## Dataset Profiler Engine

The goal of this dataset profiler is to answer the question: **what is this dataset, and what can I train on it?**

This PR adds the core profiler engine: given a fileset's files it derives the row schema, measures per-column stats, classifies what the data is, and emits a typed `DatasetProfile`. Nothing consumes this contract yet, the Files service integration follows in a stacked PR.

### Basic usage

Given a local directory containing two datasets (partitions):
```
demo
├── helpsteer2
│   ├── train-00000-of-00001.parquet
│   └── validation-00000-of-00001.parquet
└── hh-rlhf-helpful-base
    ├── test-00000-of-00001.parquet
    └── train-00000-of-00001.parquet
```

Run
```python
from nemo_datasets_plugin.profiler.file_source import LocalFileSource
from nemo_datasets_plugin.profiler.pipeline import profile

profile(LocalFileSource("demo")).model_dump_json(indent=2, exclude_none=True)
```

```jsonc
{
  "profile_schema_version": "1.0",
  "created_at": "2026-08-07T14:25:08.524649Z",
  "profiler_info": { "name": "nemo-dataset-profiler", "version": "0.1.0" },
  "sampling": {
    "rows_scanned": 13392,        // 10k budget per partition, split across its files
    "rows_present": 67551,        // exact, from parquet footers — read or not
    "files_read": 4,
    "files_present": 4,
    "bytes_present": 46514872,
    "row_budget": 10000
  },
  "partitions": [
    {
      "name": "helpsteer2",
      "file_formats": ["parquet"],
      "splits": [
        { "name": "train",      "canonical": "train",      "num_files": 1,
          "size_bytes": 18495985, "num_examples": 20324,
          "data_files": "helpsteer2/train*.parquet"},
        { "name": "validation", "canonical": "validation", "num_files": 1,
          "size_bytes":   963692, "num_examples":  1038,
          "data_files": "helpsteer2/validation*.parquet"}
      ],
      "features": [
        { "name": "prompt",      "dtype": "string", "semantic_role": "prompt",     "semantic_role_source": "detected" },
        { "name": "response",    "dtype": "string", "semantic_role": "completion", "semantic_role_source": "detected" },
        { "name": "helpfulness", "dtype": "int64",  "semantic_role": "score",      "semantic_role_source": "detected" },
        ...  // correctness, coherence, complexity, verbosity — all score
      ],
      "stats": {
        "prompt": {
          "null_rate": 0.0,
          "text": { "chars": { "p50": 265, "p95": 2692, "p99": 3089, "max": 3950 } },
          "categorical": { "distinct_count": 3019 },
          "quality": { "whitespace_ratio": 0.1699, "non_ascii_ratio": 0.0025,
                       "repetition_score": 0.0009 }
        },
        "helpfulness": {
          "numeric": { "min": 0.0, "max": 4.0, "mean": 2.8538 },
          "categorical": { "distinct_count": 5 }
        },
        ...
      },
      "stats_complete": false,
      "classification": {
        "modality": "text",
        "dataset_type": "scored_response",
        "candidates": ["scored_response", "prompt_completion"],
        "format": "standard",
        "prompt_form": "explicit",
        "evidence": [
          { "kind": "column_name",
            "detail": "columns matched roles: prompt -> prompt, response -> completion, helpfulness -> score, ..." },
          { "kind": "column_dtype", "detail": "standard format from role column dtypes" }
        ]
      }
    },
    {
      "name": "hh-rlhf-helpful-base",
      "splits": [ { "name": "test", ... }, { "name": "train", ... } ],
      "features": [
        { "name": "prompt",   "dtype": "messages", "semantic_role": "prompt",   ... },
        { "name": "chosen",   "dtype": "messages", "semantic_role": "chosen",   ... },
        { "name": "rejected", "dtype": "messages", "semantic_role": "rejected", ... }
      ],
      "stats": {
        "prompt": { "messages": {
          "turns":         { "p50": 3, "p95": 9, "p99": 13, "max": 61 },
          "content_chars": { "p50": 342, "p95": 1463, "p99": 2230, "max": 3923 },
          "roles_seen": ["user", "assistant"],
          "ends_with_assistant_rate": 0.0,      // ends on a *user* turn: it is a prompt
          "valid_alternation_rate": 0.9983
        } },
        "chosen": { "messages": { ..., "ends_with_assistant_rate": 1.0 } },
        ...
      },
      "classification": {
        "candidates": ["preference_pair"],
        "format": "conversational",
        "prompt_form": "explicit",
        ...
      }
    }
  ],
  "file_errors": []
}
```

### Some design decisions worth calling out

1\. The `DatasetProfile` schema lives at `packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py`, while the profiler itself lives in  a separate `plugins/nemo-datasets`.

If we put the contract in the dataset plugin, then a *core* service depends on an optional plugin, which goes against the architecture we are aiming for. It makes more sense to define the profile schema in the Files package because it is the first-class consumer of this schema, and will store it once we integrate with FileSets.

2\. Partitions vs splits

A partition is the top-level container of a set of files with shared schema, and it can be further broken down into splits (e.g. `train`, `val`, `test` etc). Partition names are based on top-level directory names, except when there are no top-level directories, or when the top-level directories happen to be split names like `train` or `val`, in which case a empty-string `""` is used for the partition name (the root partition), and split accordingly.

In the future, we will add the capability to leverage user provided split configurations (either via the `profile` entrypoint, or we can parse the optional `README.md` front-matter in HF dataset repos).

3\. Splits and glob patterns

Given files
```
test.parquet
train1.parquet
train2.parquet
```
We resolve two splits, and each split gets a glob pattern to represent the files
```
"train": "train*.parquet"
"test": "test*.parquet"
```
Checkout `infer_data_files` (`splits.py`) for how we implement this.

4\. Batching, folds, accumulators, and stats

In addition to understanding the schema of a dataset, we also extract stats from individual columns. For example, the length of the string sequences in the dataset is useful for coming up with a meaningful training budget (`max_seq_length`). However, this means we need some way of measuring sequence lengths across all rows, and do some bookkeeping to keep track of the p50, p95, p99s etc. We also need to do this in a scalable way for large datasets. The solution is to accumulate batches of data in a loop, "fold" them in (counting, min, max etc), then throw away the batch so we never accumulate more than one batch worth in memory.

At a high-level the loop looks like:
```
for batch in batches:
    fold.update(batch)   # counting, or unit of measurement per batch
fold.finalize()  # process the counts, summary stats, e.g. quantiles
```

The unit of measurement is type specific. For example, the `MessageAccumulator` contains histograms for sequence lengths, which gets bucketed (`_length_bucket()`), and counted. Finally, the quantiles are computed from the bucket counts.

5\. JSONL vs Parquet, and deferred dtype inference

In order to accumulate a specific column's values, we need to dispatch to the an accumulator that is aware of the dtype of the value, so that a number gets accumulated by `NumericAccumulator`, while a string gets accumulated by `StringAccumulator`.

However, with JSONL, we actually don't know the dtype of each column upfront, without actually loading the data. To solve this, we have `RoutedAccumulator`, which is a higher level abstraction on top of the type specific accumulators. `RoutedAccumulator` resolves the dtype using `isinstance(value, T)`, and dispatches to the type specific accumulator accordingly. 

The same isn't true for Parquet files, where the dtype is known in the footer. In this case, build the type-specific accumulator, and short-circuit. However, both go through `RoutedAccumulator`, so we have a single entrypoint.

6\. **Removed content-digest**. There was a `content_digest` field, which was originally meant as a field for change detection, where a subsequent job can try to compute a content_digest and see if it conflicts with a previous one's. However, it gets complicated when the content_digest might not be based on all the files in the dataset (we only profile the files we care about like jsonl and parquet). I don't want to deal with this complexity right now, and I don't think it should be the dataset profiler's responsibility to compute the content digest.

## Summary by CodeRabbit

* **New Features**
  * Added dataset profiling for JSONL and Parquet files, including schema inference, column statistics, semantic classification, split detection, and sampling metadata.
  * Added support for profiling partitioned datasets and mixed file formats.
  * Added configurable row budgets and optional column-role hints.
  * Profiles now include file formats, read status, completeness details, unreadable-file records, and classification candidates.
  * Added a platform task that publishes profiles as `profile.json` artifacts.

* **Bug Fixes**
  * Profiling now isolates read and measurement failures while preserving available dataset metadata.


plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/pipeline.py
```diff
@@ -0,0 +1,179 @@
...
+        split_examples = 0
+        split_counts_known = True  # every file's exact total row count is known (footer or full scan)
+        split_scanned = True  # every row of every file was actually parsed
+        for entry in split.entries:
+            try:
+                result = get_reader(file_format).read(source, entry)
+            except Exception:
+                # Failure isolation: an unreadable file (or missing reader) keeps its identity,
+                # skips its rows, and does not abort the profile.
+                result = None
```
anubhutivyas: nit: should we log the exception here?

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/pipeline.py
```diff
@@ -0,0 +1,179 @@
...
+            partitions.append(partition)
+            rows_scanned += partition_rows_scanned
+            all_scanned = all_scanned and partition_scanned
+
+    sampling = SamplingInfo(
+        exhaustive=all_scanned,
+        strategy="full",
+        rows_scanned=rows_scanned,
+        rows_total=rows_scanned if all_scanned else None,
+        files_scanned=len(data_entries),
```
anubhutivyas: should it be the number of successful reads only?

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/partition.py
```diff
@@ -0,0 +1,39 @@
...
+    A single top-level group — every file at the root, or all under one container like ``data/`` —
+    is one "default" partition. Multiple top-level directories (e.g. ``main/`` and ``socratic/``)
+    each become their own partition, named after the directory; root-level files fall under
+    "default".
+    """
+    by_dir: dict[str | None, list[FileEntry]] = {}
+    for entry in entries:
+        by_dir.setdefault(_top_dir(entry.path), []).append(entry)
+
+    if len(by_dir) == 1:
```
anubhutivyas, line 47: what happens when the dir structure is like `data/main/*.parquet` and `data/socratic/*.parquet`, it has a single top-level dir `data`, does it become one default partition mixing two schemas?

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/stats.py
```diff
@@ -0,0 +1,208 @@
...
+def _column_stats(feature: FeatureSchema, values: list[Any], total: int, exhaustive: bool) -> ColumnStats | None:
+    present = [value for value in values if value is not None]
+    null_rate = (total - len(present)) / total if total else 0.0
+
+    text = numeric = messages = categorical = quality = None
+    if feature.dtype == "string":
+        strings = [value for value in present if isinstance(value, str)]
+        if strings:
+            text = TextStats(chars=_quantiles([len(value) for value in strings]))
+            quality = _text_quality(strings)
```
soluwalana: For a 10GB dataset do we roughly know how much the _text_quality function adds to the runtime?

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/pipeline.py
```diff
@@ -0,0 +1,179 @@
...
+                # Failure isolation: an unreadable file (or missing reader) keeps its identity,
+                # skips its rows, and does not abort the profile.
+                result = None
+            if result is None:
+                num_rows = None
+                scanned_all = False
+            else:
+                num_rows = result.num_rows
+                rows_scanned += result.rows_scanned
+                partition_rows.extend(result.rows)
```
soluwalana: Are we worried about OOM on large datasets here?

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/readers/jsonl.py
```diff
@@ -0,0 +1,35 @@
...
+    file_format = "jsonl"
+
+    def read(self, source: FileSource, entry: FileEntry, *, row_cap: int | None = None) -> ReadResult:
+        rows: list[dict] = []
+        with source.open(entry.path) as stream:
+            for raw_line in stream:
+                stripped = raw_line.strip()
+                if not stripped:  # tolerate blank lines between records
+                    continue
+                record = json.loads(stripped)
```
soluwalana: first failed entry kills the read, is this intended?

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/pipeline.py
albcui: yup good catch, we now track how many files are actually read here: https://github.com/NVIDIA-NeMo/nemo-platform/pull/658/changes#diff-93f007b3a0eaca45c054317c6d41fb11540e65d29e83b3c54a70d1186d564009R104

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/partition.py
albcui: Currently, partitions are based on top-level directories only. So both `data/main/*.parquet` and `data/socratic/*.parquet` would be grouped under the `data` partition.

Note that a partition can be further broken down into splits. Typically, splits are `train`, `val`, `test` etc.  Currently, we hard-code this in `_CANONICAL_ALIASES`, and make a best-effort guess. However, in the future, I'm thinking of extending this to allow the user to specify their own split names in the `README.md`'s YAML front-matter like HF's format: https://huggingface.co/datasets/trl-lib/OpenMathReasoning/blob/main/README.md?code=true#L16-L22

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/readers/jsonl.py
albcui: Good catch, wrapped this with a `try/except/continue`

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/pipeline.py
```diff
@@ -0,0 +1,370 @@
...
+def _format_of(path: str) -> str:
+    """The registered format of a data file. Callers pass only pre-filtered ``data_entries``, so the
+    format is always known; a None here would mean that invariant was broken."""
+    file_format = detect_format(path)
+    if file_format is None:
+        raise ValueError(f"no registered format for {path!r}")
+    return file_format
+
+
+def profile(
```
albcui, line 73: start here: this is the main entrypoint

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/partition.py
albcui: Updated the PR description with an explanation of how partitions and splits work.

plugins/nemo-profiler/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+]
+version = "0.1.0"
+
+[tool.uv.sources]
+nemo-platform-plugin = { workspace = true }
+nemo-platform-sdk = { workspace = true }
+
+# Deliberately contributes no `nemo.cli` entry point. The profiler runs as a job task
+# (`python -m nemo_datasets_plugin.tasks.profile`), which is invoked by the platform rather than
+# typed by a user, so its inputs can keep moving while the feature is new.
```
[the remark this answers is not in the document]
albcui: no I decided not to contribute to the `nemo` CLI surface.

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/pipeline.py
```diff
@@ -0,0 +1,439 @@
...
+            num_rows: int | None = None
+            scanned_all = False
+            scanned = 0
+            try:
+                # Inside the guard: resolving the reader can fail too, and a format with no reader
+                # registered is a file the profiler could not use like any other.
+                reader = get_reader(_format_of(entry.path))
+                preview = previews[entry.path]
+                read_errors: list[str] = []
+                for batch in reader.batches(source, entry, row_cap=row_cap, errors=read_errors):
```
albcui, line 418: main loop where we iterate through batches, and fold them into the accumulators.

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/stats.py
```diff
@@ -0,0 +1,735 @@
...
+        self._measurements: dict[str, ColumnAccumulator] = {}
+
+    def _measurement(self, key: str) -> ColumnAccumulator:
+        """The named measurement, built if this is the first value to call for it."""
+        accumulator = self._measurements.get(key)
+        if accumulator is None:
+            accumulator = self._measurements[key] = _MEASUREMENTS[key]()
+        return accumulator
+
+    def _observe(self, present: list[Any]) -> None:
```
albcui, line 604: For both JSONL and Parquet, we go through `RoutedAccumulator`, this keeps the entrypoint the same, and the underlying implementation share most of the logic. The only difference is that with Parquet, we know the dtype beforehand, so we can pass `declared: FeatureSchema` in, and we short-circuit below and use the accumulator that matches `declared.dtype` directly.

For JSONL, we're not sure about the `dtype` until we observe all the rows. This is why we try to resolve all the different types -- currently just string, numeric, bool, messages.  Of course, this introduces the possibility of a column carrying multiple dtypes that may or may not agree with each other. This is handled by `SchemaFold` (`self._schema`), where a disagreement leads to resolving to `json` dtype, which is not "measurable" (i.e. not in `_MEASUREMENTS`), and is discarded through `folds.finalize()`. The column still appears in `features` but just not in `stats`.

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/stats.py
```diff
@@ -0,0 +1,735 @@
...
+
+def _is_numeric(dtype: str) -> bool:
+    return dtype.startswith(("int", "uint", "float"))
+
+
+# How finely a length distribution is recorded. Lengths below the slice count get a counter each and
+# are exact; above it each octave is cut into this many slices, a fixed *relative* width of 1/32.
+# Reporting a bucket's midpoint puts every estimate within ~1.6% whatever the magnitude.
+_HISTOGRAM_SLICE_BITS = 5
+_HISTOGRAM_SLICES = 1 << _HISTOGRAM_SLICE_BITS  # 32
```
albcui, line 719: The trick to computing the quantiles of some statistic (currently just seq length) without keeping all rows in memory, is to bin them into buckets, and keeping track of the count.

We divide the number line into "octaves" -- slices of 32 buckets each, where the spacing doubles each octave. The range of numbers covered in each octave can be defined as `[2^k, 2^k+1)`, so effectively, each octave doubles the coverage.

First 31 numbers/buckets can't evenly divide 32, so we just special case them.

| bucket spacing  | buckets | seq lengths covered
| ----------- | ----------- | ----------- |
| 1      | 0-63       | 0-63
| 2   | 64-95       | 64-127
| 4   | 96-127       | 128-255
| ...   | ...       | ...
| 64   | 224-255       |  2048–4095

The choice of `32` is a bit arbitrary, if we go higher, then we increase precision, if we go lower then we lose precision. For our use case, the distribution of text sequences is that of a long tail, where the avg is typically around 1000 tokens. This gives us about `2^10/32 = 32` of precision (in chars for seq length).

plugins/nemo-profiler/src/nemo_profiler_plugin/dataset/stats.py
```diff
@@ -0,0 +1,735 @@
...
+            if value is None:
+                self._nulls += 1
+            else:
+                present.append(value)
+                if value not in ("", [], {}):
+                    self._non_empty += 1
+            text = _probe_text(value)
+            if text is not None:
+                self._texts += 1
+                if _GSM8K_ANSWER.search(text) or _BOXED_ANSWER.search(text):
```
albcui, line 280: We use content probes to gain information from the actual content of the data -- information that the structure of the data can't tell us.

Consider two datasets with the same structure
```
{question: "Natalia sold clips to 48 .... How many clips did Natalia sell ... ?", answer: "Natalia sold ... #### 72"}  <- GSM8K
{question: "What is the capital of France", answer: "Paris"} <- normal prompt completion
```

The GSM8K dataset contains reasoning and a verifiable answer, which favours RL training tasks. The only way to tell them apart is with the `#### 72` at the end.

However, currently, we run this probe on **every column**! You may argue that it's more efficient to probe on columns that we know contain the "answer". However, at this layer of the code, we are actually agnostic to the column name. The part where we actually care about the column name is downstream in `classify.py`, where we map column names to their "roles".

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/pipeline.py
albcui: The implementation now batches rows and processes per batch in an iterative way. This does limit us to "streaming" computations, but I think that's pretty much all we need (in most cases we're just counting).

packages/nemo_platform_plugin/src/nemo_platform_plugin/files/dataset_profile.py
```diff
@@ -77,7 +96,25 @@ class PartitionClassification(BaseModel):
...
+        default_factory=list,
+        description=(
+            "Every dataset type the assigned roles satisfy, most specific first, so "
+            "`candidates[0] == dataset_type`. prompt + completion + score + label is genuinely both "
+            "scored_response and unpaired_preference; reporting only the first made rule order an "
+            "invisible tie-break and hid that the data supports more than one use. Deliberately not a "
+            'capability list ("supports DPO") — trainer requirements shift and differ per framework, '
+            "so that mapping belongs in the consumer, computed from the roles."
+        ),
+    )
```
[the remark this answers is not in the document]
albcui: I got rid of `dataset_type`. Now, `candidates` is all there is, and `candidates[0]` is the preferred `dataset_type`.

plugins/nemo-datasets/src/nemo_datasets_plugin/profiler/file_source.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, root: str | Path) -> None:
+        self._root = Path(root)
+        if not self._root.is_dir():
+            raise NotADirectoryError(f"{self._root} is not a directory")
+
+    def list_files(self) -> list[FileEntry]:
+        entries = [
+            FileEntry(path=path.relative_to(self._root).as_posix(), size_bytes=path.stat().st_size)
+            for path in sorted(self._root.rglob("*"))
+            if path.is_file()
```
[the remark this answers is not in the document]
albcui: Can you re-review?

PR #662, opened 2026-07-13
feat(deployments): OpenShell sandboxed deployment backend (AIRCORE-872)

maxdubrinsky:
## Summary

Adds an OpenShell deployment backend to the `nemo-deployments` plugin, so an agent deployed through the standard deployments API runs inside an OpenShell sandbox. The backend generates a Landlock filesystem/process policy and a default-deny egress rule that only permits traffic back to the platform Inference Gateway, so the agent's model calls go through the platform and other egress is blocked at the sandbox boundary.

The create/package/deploy/invoke flow works end to end. The demo routes model traffic through gateway-managed `inference.local`, which drops the docker-specific host and proxy workarounds the earlier version needed. Getting-started notebook: `plugins/nemo-deployments/examples/openshell/DEMO.ipynb`.

## What's here (AIRCORE-872)

Twenty commits, ordered so each builds on the last. The first six:

- consume the published openshell SDK (892). Depends on the `openshell` PyPI package as the plugin's optional `[openshell]` extra and dev group. No vendored gRPC stubs and no wheel pin. Pinned `>=0.0.92`, with the exact version in `uv.lock`. The floor matters: the gateway started enforcing a 19-char routable-name limit in v0.0.88, so an older client contract is not compatible with a current gateway.
- generate SandboxPolicy (891). Default-deny filesystem plus the always-injected platform-egress rule. A static YAML policy can override it.
- OpenShell deployment backend (887). `OpenShellDeploymentBackend` implements the `DeploymentBackend` ABC (create, read_status, delete, expose) over the OpenShell gRPC gateway and applies the policy and serve workdir. Covered by mocked unit tests, an executor-config test, and a live-gateway integration test.
- sandbox-runtime image profiles (888). A `--sandbox-runtime openshell` profile for `nemo agents package`, registered through the `nemo.sandbox_profiles` entry point.
- wire the local openshell executor (889). Adds the `openshell-local` executor to `local.yaml` for the local end-to-end path.
- deploy-sandbox skill and inference.local demo. The `deploy-sandbox` skill plus the getting-started notebook.

## Review round (28 Jul)

Addressing @matthewgrossman's findings:

- **sandbox names were rejected by any gateway from v0.0.88 on.** `_sandbox_name` emitted 20 chars against a 19-char limit, so every deployment failed with `INVALID_ARGUMENT`. Verified live: a 0.0.92 gateway rejects the old name and accepts the new 18-char one. Port names are clamped to the same limit, since `ExposeService` caps service names identically.
- **a dead workload reported READY.** The detached serve command is unsupervised, so a process that exited (unresolvable model, rejected config) left an exposed port answering 502 behind a READY deployment. The reconciler now probes the workload's pid before exposing ports and on every poll, and reports FAILED with the tail of the serve log.
- **`get_logs` returned supervisor logs, not the agent's.** `GetSandboxLogs` is fed by supervisor tracing events, never by the workload's redirected stdout, so the serve log is tailed over `ExecSandbox` instead.
- **static policy typos failed open.** A policy mapping is now parsed with `json_format.ParseDict` against `SandboxPolicy` rather than rebuilt field by field, so an unknown key raises instead of vanishing, and every field OpenShell supports is accepted rather than the subset this backend writes. The two free-form strings the proto cannot type (`landlock.compatibility`, `enforcement`) are checked separately, because the supervisor reads anything unrecognised as the weaker setting.
- **the live integration test could not pass.** It asserted `created.endpoints` after create went asynchronous; it now polls `read_status` to READY.

## Status

In review. `openshell` is a normal PyPI dependency (`>=0.0.92`, resolved to `0.0.92` in the lock), so CI can resolve it and `uv sync` picks it up like any other dependency. The example compose pins the gateway image to the matching tag instead of `:latest`, overridable through `IMAGE_TAG`, so client and gateway cannot drift apart. Local checks pass: openshell unit tests (69), ruff check and format, ty. The live integration test passes against a 0.0.92 gateway.

Linear: AIRCORE-872

## Summary by CodeRabbit

* **New Features**
  * Added OpenShell deployment backend with default-deny sandbox policies and configurable platform egress controls.
  * Added an `openshell-local` executor for sandboxed local deployments.
  * Added `--sandbox-runtime` to `nemo agents package`, enabling sandbox-runtime-aware Dockerfile generation.
* **Documentation**
  * Updated the OpenShell demo notebook and refreshed the `deploy-sandbox` runbook with governed sandbox guidance and troubleshooting notes.
* **Tests / CI**
  * Added path-based CI execution plus a dedicated OpenShell test job and `make test-deployments-openshell` target.


maxdubrinsky:
Thanks @coderabbitai. Triaged everything against the code and `openshell/proto/sandbox.proto` (the source of truth for the egress value sets). Disposition:

**Addressing now (pushed to this branch):**

- **#5 (marker-probe `RpcError` escapes `read_status`)**: correct. Wrapped the provisioning call so a transient exec RPC error maps to `UNKNOWN` and retries on the next poll instead of propagating into the reconciler. The deliberate launch/expose failure paths already return `FAILED` and clean up, so this only catches unhandled transients.
- **Nitpick: config `extra="forbid"` + `Literal`**: adopted `extra="forbid"` on all three models (the real win: a typo'd `platform_egres` no longer silently falls back to the default egress). Added `Literal` constraints, but using the proto value sets, which are broader than the field docstrings: `protocol` also allows `graphql` and `""` (L4-only), and `tls` allows `passthrough`. `Literal`s built from the docstrings would have rejected valid configs.

**Acknowledged, deferred to follow-up work (not this PR):**

- **#1** (`--sandbox-runtime` + `--dockerfile`/`--template` bypass): valid gap in the nemo-agents packager; no guard rejects the combo.
- **#6** (enforce glibc >= 2.39 in `sandbox_profile`): real limitation; modeling it plus a build-time preflight is a larger change.
- **#3** (compose gateway is root + holds the docker socket + `0.0.0.0`): fair hardening note on the local-dev example.
- **#4** (arch/version-pinned interpreter path in the example policy): will add a host-specific note.
- **#2 / stale step reference** (DEMO.ipynb step 3 vs `platform_egress: null`): correct, the narration still describes the old direct-egress model.
- **Nitpick: `touch` always succeeds** (instant serve death still marked launched): known and documented in `_launch_serve`; a real fix is a liveness probe in `read_status` before declaring READY.
- **Nitpick: pin `generate-certs` off `:latest`**: a sha digest is impractical to carry in a skill doc; will revisit pinning to a release tag when there is a canonical one.

No redesign required. Thanks for the review.

matthewgrossman:
A few more things from my agent:

---

## Findings

1. **[P1] Generated sandbox names are rejected by OpenShell 0.0.90.** [`[_sandbox_name](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L561-L569)`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L561-L569) always produces 20 characters (`nmp-` plus 16 hex digits), but [[OpenShell v0.0.90 limits routable names to 19](https://github.com/NVIDIA/OpenShell/blob/v0.0.90/crates/openshell-server/src/grpc/mod.rs#L108-L113)](https://github.com/NVIDIA/OpenShell/blob/v0.0.90/crates/openshell-server/src/grpc/mod.rs#L108-L113) and rejects longer names during creation. Every deployment against that release fails with `INVALID_ARGUMENT`. Reduce the generated name and update the unit test’s obsolete `<=28` assertion.

2. **[P1] A dead workload is reported as `READY`.** The launcher backgrounds the command and immediately touches the marker, so shell success does not prove the serve process started. The next poll exposes the port and returns `READY`; afterward, [[existing service metadata is treated as proof of health](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L435-L437)](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L435-L437), even after the process exits. The runbook confirms bad configs/models produce `READY` plus persistent 502s. This needs supervised execution or an actual readiness/liveness check before and after reaching `READY`.

3. **[P2] `[get_logs](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L366-L380)` cannot return the agent’s logs.** The workload redirects stdout/stderr to `/tmp/nemo-serve.log`, while [`get_logs`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L366-L380) calls `GetSandboxLogs`. OpenShell populates that API from supervisor tracing events, not the redirected workload file ([[implementation](https://github.com/NVIDIA/OpenShell/blob/v0.0.90/crates/openshell-supervisor-process/src/log_push.rs#L4-L20)](https://github.com/NVIDIA/OpenShell/blob/v0.0.90/crates/openshell-supervisor-process/src/log_push.rs#L4-L20)). Tail `_SERVE_LOG` through `ExecSandbox`, or launch the workload through a mechanism whose output OpenShell captures.

4. **[P2] Static policy typos can silently weaken confinement.** [`[normalize_loaded_policy](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/policy.py#L101-L124)`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/policy.py#L101-L124) and the proto builder accept arbitrary mappings and ignore unknown keys. For example, `compatibilty: hard_requirement` silently becomes `compatibility: best_effort`, allowing filesystem confinement to fail open on unsupported kernels. Validate override YAML with strict typed models before constructing the security policy.

5. **[P2] The live integration test cannot pass after the asynchronous-create refactor.** [`create_deployment` now returns `STARTING` without endpoints](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py#L287-L294), but the [live test immediately requires `created.endpoints`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/680a1c346d1ff08d2e4bc8a3052d3d1e4e3b139e/plugins/nemo-deployments/tests/integration/backends/openshell/test_openshell_backend.py#L75-L88). It should poll `read_status` until `READY`, then obtain the endpoint.

The 57 targeted OpenShell unit tests pass, and current CI is green; the live gateway path is skipped without a running gateway. I did not repeat the already-open CodeRabbit findings or post comments to GitHub.

.github/workflows/ci.yaml
```diff
@@ -1514,6 +1515,39 @@ jobs:
           retention-days: 7
           path: web/packages/studio/playwright-report/
 
+  deployments-openshell-tests:
+    # The openshell extra is a platform-restricted manylinux_2_39 wheel that the
```
mckornfield, line 1565: any reason we can't install that wheel in the unit tests? is it really slow or?

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
```diff
@@ -70,6 +71,8 @@ Verification: confirm the deployment reached ready state.
 
 If `DEPLOY_NOT_READY`: jump to the recovery table at the bottom.
 
+> **Governed sandbox deployment.** To deploy this agent as a policy-governed OpenShell sandbox instead of the default executor (Landlock filesystem isolation plus default-deny network egress, so its model traffic can only reach the platform), use the `deploy-sandbox` skill once the image is built. It swaps this step's deploy path for the `openshell-local` executor and an auto-generated SandboxPolicy.
```
mckornfield, line 74: will we let people configure it or no?

packages/nemo_platform_plugin/src/nemo_platform_plugin/sandbox.py
```diff
@@ -0,0 +1,65 @@
...
+
+    The packager renders this into ``groupadd``/``useradd`` invocations. The
+    runtime typically resolves the user *by name*, so the uid is not
+    load-bearing; ``system=True`` keeps it out of the uid range a packager may
+    reclaim for its own default runtime user.
+    """
+
+    name: str
+    group: str | None = None
+    """Primary group name. Defaults to :attr:`name` when ``None``."""
```
mckornfield, line 38: feel like some comments are redundant with the code that's in line of sight below

packages/nmp_platform/config/local.yaml
```diff
@@ -130,6 +129,34 @@ deployments:
         pull_images: false
         port_range_start: 9000
         port_range_end: 9100
+    # OpenShell-backed sandboxed agent deployments (opt-in per deployment; docker
+    # stays the default). The openshell executor talks to a local docker-driver
+    # gateway on :17670 (OpenShell's documented docker default; NOT :8080, the
+    # platform owns that) and generates a per-deployment default-deny SandboxPolicy
+    # (AIRCORE-891). The serve command runs in a sandbox-writable workdir because a
```
mckornfield: comment is a bit verbose, also do we want to use the ticket numbers?

packages/nmp_platform/config/local.yaml
```diff
@@ -130,6 +129,34 @@ deployments:
...
+        # https://inference.local/v1, so the sandbox needs NO direct egress -> platform_egress
+        # is null, yielding a pure default-deny SandboxPolicy (the gateway makes the upstream call).
+        #
+        # To keep the older direct-egress path instead (agent -> host.docker.internal:8080), set:
+        #   platform_egress:
+        #     host: host.docker.internal
+        #     port: 8080
+        #     binaries: [/workspace/.venv/bin/python3.13]
+        # To pin a hand-written policy, set default_policy_path, e.g.:
+        # default_policy_path: plugins/nemo-deployments/examples/openshell/local-sandbox-policy.yaml
```
mckornfield: ok so this is how you'd set a policy, at least locally, but not in kube?

plugins/nemo-agents/src/nemo_agents_plugin/container/sandbox.py
```diff
@@ -0,0 +1,78 @@
...
+    logical Dockerfile line (with escaped continuations for readability).
+    """
+    parts: list[str] = []
+    for user in profile.users:
+        name = _validate(user.name, _NAME_RE, "user name")
+        group = _validate(user.resolved_group(), _NAME_RE, "group name")
+        home = _validate(user.resolved_home(), _PATH_RE, "home dir")
+        shell = _validate(user.shell, _PATH_RE, "shell")
+        sysflag = "--system " if user.system else ""
+        groupadd = f"groupadd {sysflag}{group}"
```
mckornfield, line 72: maybe a nit, could this just be a multiline, heredoc string so it's a bit more readable?

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
```diff
@@ -161,6 +161,16 @@ def is_plugin_managed(path: Path) -> bool:
 ENV NAT_CONFIG_FILE={{ config_file_path }}
 
 ENV PATH="/workspace/.venv/bin:$PATH"
+{% if sandbox_user_setup %}
+# Sandbox-runtime compatibility ({{ sandbox_runtime }}). The runtime's supervisor
+# runs the workload as the profile's non-root user (resolved by name), so the
+# uid is not load-bearing; a ``--system`` user stays out of the 1000/1001 range
```
mckornfield: `load-bearing`? lol

plugins/nemo-deployments/examples/openshell/agent/config.yaml
```diff
@@ -0,0 +1,44 @@
...
+  llm:
+    _type: openai
+    # OpenShell gateway-managed inference: the gateway routes https://inference.local to a
+    # registered provider (the platform IGW) and injects the real credential. The sandbox
+    # needs no direct egress and no proxy handling. Operator wiring (once per gateway):
+    #   openshell provider create --name nemo-igw --type openai --credential OPENAI_API_KEY=empty \
+    #     --config OPENAI_BASE_URL=http://host.openshell.internal:8080/apis/inference-gateway/v2/workspaces/default/openai/-/v1
+    #   openshell inference set --provider nemo-igw --model <model-id>
+    base_url: https://inference.local/v1
+    api_key: not-used                  # MUST be non-empty; the gateway injects the real credential
```
mckornfield, line 35: does `not-used` == empty?

plugins/nemo-deployments/examples/openshell/DEMO.ipynb
mckornfield, line 1: Commenting on notebooks is tough. For the `how to run this`, could it be a little simpler on how to launch a workbook?

Should we have shell command examples that can be copied out for the prereqs? Or could they just be runnable as `%%bash`?

Will `nemo models list` be too honking long of a list?

Any reason we use httpx/ the CLI versus using the SDK, something that feels more natural in a notebook?

plugins/nemo-deployments/examples/openshell/docker-compose.yml
```diff
@@ -0,0 +1,91 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mckornfield, line 1: nooo not docker-compose lol

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/config.py
```diff
@@ -0,0 +1,148 @@
...
+
+from __future__ import annotations
+
+from typing import Literal
+from urllib.parse import urlparse
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+
+class OpenShellTLSConfig(BaseModel):
```
mckornfield, line 18: Their stuff is all in rust yeah? so they don't expose any models we can just import?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/policy.py
```diff
@@ -0,0 +1,261 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+import yaml
+from nemo_deployments_plugin.backends.base import MissingBackendDependencyError
+
+if TYPE_CHECKING:
+    from openshell._proto import sandbox_pb2 as sb  # ty: ignore[unresolved-import]
```
mckornfield, line 30: ohhh heck yeah.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/policy.py
```diff
@@ -0,0 +1,261 @@
...
+# the sandbox closed on such kernels. Kept as the default to preserve the local-dev
+# docker-driver happy path; harden via OpenShellExecutorConfig.landlock_compatibility.
+DEFAULT_LANDLOCK_COMPATIBILITY = "best_effort"
+
+# The binary that opens the egress socket for NAT's LLM calls is the venv/uv python,
+# NOT curl. The exact uv-managed interpreter path is image/patch-version specific
+# (e.g. /opt/uv/python/cpython-3.13.7-.../bin/python3.13); an executor can add it via
+# its platform_egress config. AIRCORE-890 tracks an upstream ask for binary path
+# prefixes/globs so egress rules need not pin an interpreter patch version.
+DEFAULT_EGRESS_BINARIES: tuple[str, ...] = ("/workspace/.venv/bin/python3.13", "/usr/bin/curl")
```
mckornfield, line 53: yeah this looks kinda weird

uv.lock
mckornfield, line 1: interesting diff, seems to take out a lot of weird wheel variants?

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
```diff
@@ -0,0 +1,376 @@
...
+not-for:
+  - nemo-build-agent (use to scaffold + build the agent image first)
+  - nemo-try-agent (use to query an already-deployed agent)
+  - nemo-setup (use to install and start the platform first)
+  - nemo-status (use for a read-only health dashboard)
+compatibility: >-
+  nemo-platform >= 0.1.0; requires the nemo-deployments plugin installed with the
+  openshell extra (`openshell>=0.0.83` from PyPI); a running OpenShell docker-driver
+  gateway on :17670 (not :8080, which the platform owns) whose JWT signing keys were
+  generated once with `generate-certs`; a NeMo Platform started bound to 0.0.0.0 with
```
mckornfield: `a NeMo Platform started bound to 0.0.0.0` the phrasing of this confused me for a bit lol

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
```diff
@@ -0,0 +1,376 @@
...
+   generate the signing keys once before the gateway starts (see `gateway.toml`'s
+   `[openshell.gateway.gateway_jwt]` and the `docker-compose.yml` header). Run the
+   one-time key generation, then bring the compose stack up and register it:
+
+   ```bash
+   # one-time: generate the JWT signing keys the docker driver needs to mint
+   # sandbox tokens (writes /var/lib/openshell/tls/jwt/*). Re-run only if the
+   # /var/lib/openshell state dir is wiped.
+   docker run --rm --user 0 -v /var/lib/openshell:/var/lib/openshell \
+     ghcr.io/nvidia/openshell/gateway:latest generate-certs \
```
mckornfield: do we want to use `latest` or the same version as the version of openshell we're using?

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
```diff
@@ -0,0 +1,376 @@
...
+
+   ```bash
+   export NMP_BASE_URL=http://localhost:8080
+   nemo services run --host 0.0.0.0 --port 8080 \
+     --config packages/nmp_platform/config/local.yaml
+   curl -sf http://localhost:8080/health/ready      # {"status":"ready"}
+   ```
+
+   If the platform is down, route to `nemo-setup`. Binding `0.0.0.0` matters: the
+   OpenShell gateway resolves `inference.local` by calling the platform IGW at
```
mckornfield: will an agent reading this know what IGW is?

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
```diff
@@ -0,0 +1,376 @@
...
+```
+
+The clean directory and the `config.yaml` name both matter (verified the hard way): `nemo agents package` uses the agent file's **parent directory as the Docker build context**, so a shared dir like bare `/tmp` can hold a file that breaks `docker buildx` (`failed to unmarshal ... invalid UTF-8`); and it bakes the file into the image at **`/workspace/<basename>`**, so naming it `config.yaml` makes it land at `/workspace/config.yaml`, the exact path Step 3's serve command loads. A name mismatch is invisible: the deployment still reaches `READY`, but `nat serve` never finds its config and every invoke returns `502`.
+
+Build the image with the OpenShell runtime profile (adds the `sandbox` user and required packages). Use the shipped demo config:
+
+```bash
+nemo agents package \
+  --agent  plugins/nemo-deployments/examples/openshell/agent/config.yaml \
+  --nat-version 1.8.0 \
```
mckornfield, line 218: also wonder if this should be inferred

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
```diff
@@ -0,0 +1,376 @@
...
+```bash
+SBX=$(curl -sf "$BASE/deployments/igw-agent" | jq -r '.endpoints[0].url' | sed -E 's#^https?://##; s#--.*##')
+openshell sandbox exec --name "$SBX" -- cat /tmp/nemo-serve.log
+```
+
+The most common root causes are: the `MODEL` is not served on your IGW (`nemo models list --all-pages`), the config did not bake to `/workspace/config.yaml`, or a leftover `general.telemetry` block made NAT reject the config. An empty or error `value` means the LLM call failed at the gateway or the model misbehaved. Either way, check the recovery table.
+
+## Step 7: Prove the sandbox has zero direct egress
+
+The security punchline. Show the sandbox can reach nothing on the network directly:
```
mckornfield, line 300: lol `punchline`

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -0,0 +1,612 @@
...
+        )
+
+    async def delete_volume(
+        self,
+        workspace: str,
+        name: str,
+        *,
+        backend_config: dict[str, Any] | None = None,
+    ) -> VolumeStatusUpdate:
+        del workspace, name, backend_config
```
mckornfield, line 461: lol what do these deletes accomplish?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -0,0 +1,612 @@
...
+
+
+def _sandbox_name(workspace: str, name: str) -> str:
+    """OpenShell sandbox name, <=28 chars for service routing.
+
+    OpenShell rejects longer names for service routing, so we cannot reuse the shared
+    ``dep-<ws>-<name>-<hash>`` scheme. The human-readable workspace/name mapping is carried
+    in identity labels (used by list/idempotency), not the sandbox name.
+    """
+    digest = hashlib.blake2s(f"{workspace}/{name}".encode(), digest_size=8).hexdigest()
```
mckornfield: why blake2s?

mckornfield (commented):
hehe

packages/nmp_platform/config/local.yaml
maxdubrinsky: Oh god I thought I caught these. What is it with agents and putting ticket numbers in code.

plugins/nemo-deployments/examples/openshell/agent/config.yaml
maxdubrinsky: Basically, needs to have some value but it's never checked. I personally want to start using `sirnotappearinginthisfilm`, but it's a bit of a mouthful.

plugins/nemo-deployments/examples/openshell/docker-compose.yml
maxdubrinsky: Oh it's so much better than the monster `docker` command... This you can read and add comments to!

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
maxdubrinsky: We _should_ pin this to whatever version we have in `uv.lock`, but then this is another place we need to update every time that happens.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
maxdubrinsky: yep, you can override our default (landlocked) policy with `default_policy_path`, and `platform_egress` sets up the egress rule back to the platform, which we inject on top of whatever policy you supply

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
maxdubrinsky: I thiiiiiink so? Probably better to make it say `Inference Gateway` just to be sure.

plugins/nemo-agents/src/nemo_agents_plugin/container/template.py
maxdubrinsky: no matter how many times you bully an agent to not write this type of stuff in comments...

plugins/nemo-deployments/examples/openshell/DEMO.ipynb
maxdubrinsky: I could definitely add more, key concern is making sure they're actually runnable. Key thing I wanted to avoid was duplicating any docs on how to set up the platform/openshell since doing so would add _another_ location we need to maintain.

`nemo models list` will be a monster, problem is we can't assume what models will be available on a particular platform.

Currently we don't have complete coverage in the SDK/CLI for raw deployments, my next project once this lands is to clean up/simplify the deployment flow in a slightly smaller diff.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/config.py
```diff
@@ -0,0 +1,148 @@
...
+        default=None,
+        description="Force plaintext (True) or TLS (False). When None, derived from the endpoint scheme.",
+    )
+    tls: OpenShellTLSConfig | None = Field(default=None, description="mTLS material for an https gateway.")
+    request_timeout_seconds: int = Field(
+        default=120,
+        ge=1,
+        description="Per-RPC deadline for control-plane calls (create/get/expose/delete/logs).",
+    )
+    ready_timeout_seconds: int = Field(
```
tylersbray: Is this used, or will be used eventually?  I see the key values below a couple places with hardcoded values.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -0,0 +1,692 @@
...
+        Returns None on success, or a FAILED update if the launcher itself failed (bad
+        workdir/shell or a non-zero launcher exit). The backgrounded serve process is
+        not supervised, so this catches launch-time failures, not later serve crashes.
+        """
+        serve = shlex.join(serve_command)
+        # The inner shell records its own pid and then execs, so the pidfile holds the
+        # workload's pid whether or not setsid forks. The marker is written synchronously
+        # so a poll racing the background start does not relaunch.
+        inner = f"echo $$ >{_SERVE_PIDFILE}; exec {serve} >{_SERVE_LOG} 2>&1"
+        launch = f"setsid /bin/sh -c {shlex.quote(inner)} </dev/null >/dev/null 2>&1 & touch {_LAUNCH_MARKER}"
```
tylersbray: The pidfile + `kill -0` probe closes the "serve died after starting" case, but this launch path still has a hole: `setsid … & touch marker` always succeeds, and `_LIVENESS_PROBE` treats a missing pidfile as "not dead" (`exit 0`). So a launch that writes the marker but never writes the pidfile can still advance to expose → READY.

Worth either requiring the pidfile before expose, or an HTTP readiness check against the exposed port before declaring READY (or ticket the known gap explicitly if deferring).

tylersbray (approved):
Nonblocking approval, modulo conflicts, happy CI, stripping ticket numbers out, etc etc.  Looks good.

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
maxdubrinsky: Yeah it's not the clearest, updating with a bit more context (and fixing a few other places)

packages/nmp_platform/config/local.yaml
maxdubrinsky: kube has not been in my crosshairs for this work, suspect there will be some changes but tbd.

.github/workflows/ci.yaml
maxdubrinsky: not per se, more what guardrails does for extras atm

plugins/nemo-deployments/src/nemo_deployments_plugin/skills/deploy-sandbox/SKILL.md
maxdubrinsky: potentially... but with the switch to fabric I might hold off?

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
maxdubrinsky: I am not entirely sure, think the robot just picked a library. We can use a normal sha256 though.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/config.py
maxdubrinsky: it's not used now so I'd rather just drop it.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/config.py
maxdubrinsky: yes but not everything is exposed. at some point the Python SDK will get better, but until then...

also this is an our-config thing, theirs maps a bit differently.

uv.lock
maxdubrinsky: not sure... but tests pass so ¯\_(ツ)_/¯

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
maxdubrinsky: the vars are unused, probably just to avoid that particular lint warning. not strictly necessary though.

svvarom (approved):

PR #668, opened 2026-07-14
ci: authenticate nightly GHCR polling

svvarom:
The nightly release run successfully pushed wheels, containers, and the Helm chart, but final artifact polling never got past the first GHCR container. The poll was anonymous because nightly GHCR artifacts were expected to be public; these packages are currently private, so `docker manifest inspect` received `denied` and retried as though the artifact had not arrived.

This PR gives only the final polling job `packages: read` and authenticates Docker and Helm with the repository's built-in `GITHUB_TOKEN` for nightly runs. Stable container and Helm polling remains anonymous, and no PAT or new secret is introduced.

## Human attention needed

- **Decision:** accept authenticated GHCR availability as the nightly readiness contract while public access is tracked separately in [AIREINF-269](https://linear.app/nvidia/issue/AIREINF-269/request-that-we-can-make-our-pre-release-containers-public)
- **Read carefully:** the job-level package permission and the two nightly-only registry logins
- **Safe to skim:** comment updates describing nightly versus stable access
- **Proof:** `actionlint`, `act workflow_dispatch --list`, and `git diff --check` pass locally; the [failed nightly poll](https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/29298618095/job/86977532472) reproduced the anonymous `denied` response after the container build had pushed successfully
- **Biggest risk:** each private container package must grant the `nemo-platform` repository Actions read access; a fresh nightly is still required to prove that package-side configuration end to end

## Rollout

Merge this change, then start a fresh nightly from `main`. Rerunning the existing workflow would reuse its original workflow revision and would not exercise this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nightly release readiness checks by authenticating access to container images and Helm charts.
  * Added the required package-read permissions for nightly artifact validation.
  * Preserved existing behavior for stable release checks.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -123,9 +123,9 @@ env:
   RELEASE_HELM_REGISTRY: nvcr.io/0921617854601259/nemo-platform
   RELEASE_NIGHTLY_WHEEL_INDEX: https://pypi.nvidia.com
   RELEASE_STABLE_WHEEL_INDEX: https://pypi.org/simple
+  # Nightly GHCR readiness checks authenticate with the repository GITHUB_TOKEN.
   RELEASE_NIGHTLY_CONTAINER_REGISTRY: ghcr.io/nvidia-nemo/nemo-platform
   RELEASE_STABLE_CONTAINER_REGISTRY: nvcr.io/nvidia/nemo-platform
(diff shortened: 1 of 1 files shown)


crookedstorm (approved):
I think this covers my PR faster than I could get it up. LGTM

PR #670, opened 2026-07-14
chore: update fine tuning and nss images

mckornfield:
## Summary by CodeRabbit

* **Updates**
  * Updated bundled ML and notebook dependencies to newer pinned versions, including Hugging Face Transformers and related components.
  * Refreshed Markdown/HTML and image-processing dependency constraints (e.g., mistune, soupsieve, pillow) to align with supported versions.
  * Updated the Safe Synthesizer runtime dependency on Weights & Biases to the newer pinned release.
  * Improved dependency consistency and cleanup during image builds to remove stale preinstalled artifacts so the intended versions are used.


ironcommit (approved):
LGTM

PR #673, opened 2026-07-14
docs: Consolidate DD library docs links

mikeknep:
## Summary by CodeRabbit

* **Documentation**
  * Updated Data Designer documentation links to use internal library documentation paths.
  * Added redirects for Data Designer library pages to the corresponding NVIDIA documentation.
  * Removed a duplicate external “Library Documentation” link from the Next Steps section.


docs/data-designer/index.mdx
```diff
@@ -99,10 +99,4 @@ Learn through examples: basics, seeding, and more.
 Move configurations between local CLI and NeMo Services execution.
 
 </Card>
-<Card title="Library Documentation" href="https://docs.nvidia.com/nemo/datadesigner/v0.7.0/getting-started/welcome">
```
mikeknep, line 102: This card component is the only link to upstream DD docs that **isn't** a standard Markdown link, and it isn't especially valuable as a card here anyways, so I'm just deleting it.

docs/fern/docs.yml
```diff
@@ -21,6 +21,10 @@ versions:
   slug: latest
 
 redirects:
+  # Upstream Data Designer library pages, targeting a specific library version
+  # TODO: add version path component once v0.7.0 is accessible explicitly
```
mikeknep, line 25: Currently, the DD library only publishes the latest version of the docs with no version component at all, so we can't actually explicitly target a specific version when that version is the latest. This is getting fixed in the upstream library

docs/fern/docs.yml
```diff
@@ -21,6 +21,10 @@ versions:
...
+  - source: "/data-designer-library/:slug*"
+    destination: "https://docs.nvidia.com/nemo/datadesigner/:slug*"
```
mikeknep, line 27: This redirect approach seems the best option for consolidating the upstream DD library links:
- it achieves the goal of specifying the library version just once, here
- it allows authors to write normal-looking Markdown links (don't have to use a custom `<DataDesignerDocsLink>` component
- it's short

There is one unfortunate side-effect: fern's `broken-links` check doesn't understand redirects, and so all the DD links are flagged as broken 😒 These are currently only warnings, not errors, but still: not great. My agent suggested wrapping the broken-links check in a script that ignores the data-designer-library references. I'm tempted to punt that until we start actually blocking in CI on broken links (there is an existing one on `main` related to auth).

I also considered a pattern in which authors provide full-but-**unversioned** links in the markdown, and we use redirects to target the specific version. So for example the author would write `[custom columns](https://docs.nvidia.com/nemo/datadesigner/custom-columns)` and the redirect config would add in the version. Unfortunately, this does not work: you can't redirect from external sources, only from internal references (e.g. `/foo`); this also produced a slew of TypeErrors in Fern, including for completely unrelated links in the docs.

mikeknep: The other weird thing about this approach is that if you right-click the link and select "Copy link address", you get the invalid, pre-redirected link. Example from the preview build:

> https://nvidia-preview-dd-docs-constant-mknepper.docs.buildwithfern.com/nemo-platform/data-designer-library/getting-started/welcome

If you paste that into a browser and try to hit it, it **does** correctly redirect, but it's still a little weird...

mckornfield: wow this is annoying lol. I guess it's valid technically as a link, just not to the final destination

mckornfield (approved):
I honestly don't know if this is an improvement with how weird it is. I will stamp it though if you feel it is hahaha

mikeknep:
Chatted in Slack and we don't like any of the options enough. In particular, the `/data-designer-library/*` approach looks too much like an absolute link and/or a link that should exist in the repo's docs, and therefore is more confusing than it is valuable. We'll just keep relying on grep/fastmod/agents to update all the docs links as we update lib dep versions.

PR #678, opened 2026-07-14
ci: wire the real release deployment signal

svvarom:
The [nightly run](https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/29345313121) showed **Signal deployment** as skipped. That looked like the real deployment handoff, but inspection revealed it was still an echo-only placeholder; the actual dispatch lived in a separate job with a different name.

Consolidating those jobs exposed the reason both had skipped: nightlies intentionally skip stable-only release registration, and GitHub propagates that skipped state through dependent jobs unless they explicitly evaluate their own status. Artifact polling already handled that case and succeeded, but deployment signaling did not.

This PR makes the real dispatch the single `signal-deployment` job. It now runs only when the workflow was not cancelled, artifact polling succeeded, Helm was selected, and the run is not a dry run. The final notification therefore observes the job that actually performs the handoff.

| Release state | Deployment signal |
|---|---|
| Nightly or stable; polling succeeded; Helm selected | Dispatch `create-release-deployment` with the source SHA, cadence, release label, and Helm version |
| Nightly skipped stable-only registration | Dispatch still runs after successful polling |
| Polling failed or the run was cancelled | Skip |
| Helm not selected or dry run | Skip |

The downstream workflow remains responsible for creating the pending GitHub Deployment and completing it independently.

## Evidence

- `actionlint .github/workflows/release.yaml` passes.
- No live release or nightly was triggered while validating this change.

Linear: [AIREINF-267 — Create the deployment workflow](https://linear.app/nvidia/issue/AIREINF-267/create-the-deployment-workflow)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -755,6 +755,8 @@ jobs:
     name: Signal deployment
     needs: [plan-release, stage-helm, poll-final-release]
     if: >-
+      !cancelled() &&
+      needs.poll-final-release.result == 'success' &&
       needs.plan-release.outputs.include_helm == 'true' &&
       needs.plan-release.outputs.dry_run != 'true'
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


crookedstorm (approved):
That should do it!

PR #687, opened 2026-07-14
feat(evaluator): Add optional diagnostics for a MetricResult

JashG:
This PR adds support for attaching structured diagnostic data to a `MetricResult`. This optional field should be used to attach context explaining why a metric produced an unexpected value. This would be useful for debugging false positives/negatives and may also be useful on successful results to preserve the reasoning trail.

## Summary by CodeRabbit

- **New Features**
  - Added structured metric diagnostics with messages and optional details.
  - Diagnostic information now flows through scoring, benchmark results, multi-metric evaluations, and row exports.
  - Added exact-match evaluation diagnostics showing expected output, actual output, and unified differences when results mismatch.
  - Made `MetricDiagnostic` publicly available through the SDK.

- **Documentation**
  - Updated task metric authoring guidance to recommend structured diagnostics.

- **Tests**
  - Added coverage for diagnostic preservation, namespacing, merging, exports, and mismatch details.


packages/nemo_evaluator_sdk/examples/examples.py
```diff
@@ -252,6 +254,64 @@ async def compute_scores(self, input: MetricInput) -> MetricResult:
...
+                prediction.splitlines(),
+                fromfile="expected",
+                tofile="actual",
+                lineterm="",
+            )
+        )
+        return MetricResult(
+            outputs=outputs,
+            diagnostics=[
+                MetricDiagnostic(
```
ngoncharenko, line 302: each diagnostics[i] corresponds to outputs[i]? I wonder if we put it into MetricResult.diagnostic, would it make it more straightforward?

SandyChapman (approved):
lgtm

ngoncharenko (approved):

packages/nemo_evaluator_sdk/examples/examples.py
JashG: Not intended to necessarily be a 1-to-1 map to `outputs[i]`. I made diagnostics a list just to allow flexibility for attaching more than 1 piece of metadata to the result. Let me know if you think that is more confusing than a single `diagnostic`.

PR #689, opened 2026-07-14
feat(studio): Virtual Models view/delete page with search, filtering, and autoprovisioned exclusion

marcusds:
## Summary

Adds a **Virtual Models** page to NeMo Studio (ASTD-300) — view and delete inference-routing VirtualModels, with a read-only middleware-pipeline details panel. Nav lives under the **Models** group.

Because nearly all VirtualModels are controller-managed passthroughs (`autoprovisioned`), the IGW list endpoint gains a way to exclude them, and the UI hides them by default. Search + column filtering round out the table to match the other Studio data views.

## What's included

**Studio UI** (`web/packages/studio`)
- `VirtualModelsListRoute` + `VirtualModelsDataView` + `VirtualModelDetailsSidePanel`
- List columns: name, default model, middleware call count, created; row actions **View** / **Delete**
- Read-only details panel: core fields + all three middleware pipelines (plugin, config_type, config)
- Name search (`$like`) + column filters (default model text, created-at date range), search/filter-aware empty state
- Route + sidebar nav wiring, no feature flag

**Inference Gateway** (`services/core/inference-gateway`)
- `exclude_autoprovisioned` query param on `list_virtual_models` (default `false`) — server ANDs a real-boolean `data.autoprovisioned == false` condition
- `VirtualModelFilter` for `name` / `default_model_entity` / `created_at` / `updated_at` filtering

**Regenerated:** OpenAPI spec, Python SDK (stainless), CLI (`--exclude-autoprovisioned` switch), web SDK (CI-rebuilt).

## Design notes

- **Why `exclude_autoprovisioned` and not a generic `filter[autoprovisioned]`?** A string-based deep-object filter delivers the value as `"false"`, which never matches the SQLite entity store's JSON boolean. Inverting to a dedicated boolean param (default `false`) both fixes the match and lets the generated CLI expose a working positive `--exclude-autoprovisioned` switch. (Both issues were caught in a Codex review and fixed in `afc16b7`.)

## Verification

- `pytest services/core/inference-gateway/tests/unit/test_virtual_models_router.py` — 34 pass (incl. real-store exclusion + `name $like`)
- `pnpm --filter nemo-studio-ui test` (VM suites) — 7 pass; `typecheck` + `typecheck:go` + `eslint` clean
- `ruff` + `ty` clean on the IGW route
- Manual: list + details rendered live against a local platform; `--exclude-autoprovisioned` CLI switch verified

## Summary by CodeRabbit

* **New Features**
  * Added a Virtual Models page in Studio with workspace navigation, server-side search/filtering (including time ranges), sorting, pagination, and virtual model deletion.
  * Added a Virtual Model details side panel with middleware/pipeline metadata.
* **Documentation**
  * Updated CLI and API docs to support excluding controller-managed (autoprovisioned) models via `--exclude-autoprovisioned` / `exclude_autoprovisioned`, plus expanded `filter` options (including `default_model_entity`, `name`, `project`, `workspace`, `created_at`, `updated_at`).
* **Tests**
  * Added unit and UI coverage for exclusion behavior, name matching, empty states, and details-side-panel rendering.


docs/cli/reference.mdx
```diff
@@ -3464,11 +3464,25 @@ nemo inference virtual-models list [OPTIONS]
...
 * `--workspace`
+* `--exclude-autoprovisioned`: When true, controller-managed (autoprovisioned) passthrough VirtualModels are excluded from the results.
 * `--page <INTEGER>`: Page number (1-indexed).
 * `--page-size <INTEGER>`: Number of results per page.
 * `--sort`: Sort field. Prefix with `-` for descending order.
 * `--all-pages`: Fetch all pages
 
+**Filter Options:**
+
+* `--filter FILTER_JSON`: Use --filter with JSON for complex/nested queries, or --filter. FIELD options for simple fields. Both can be combined, with field options taking precedence.
```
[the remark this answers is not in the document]
marcusds: Claude says: Skipping — this is a pre-existing docs-generator artifact, not something regeneration fixes. The CLI source help text is already correct (`--filter.FIELD`, no space); the space is introduced by `docs_generator.py` reflowing the long help string into `reference.mdx`. It appears 10× in the file, 9 of which already exist on `main` for the other filter commands (models/filesets/traces/…). Regenerating reproduces the space verbatim. A real fix belongs in the docs generator and should cover all commands — filing that separately rather than hand-editing one generated line here.

services/core/inference-gateway/src/nmp/core/inference_gateway/api/v2/virtual_models.py
```diff
@@ -336,18 +357,33 @@ async def list_virtual_models(
         default="-created_at",
         description="Sort field.  Prefix with ``-`` for descending order.",
     ),
+    exclude_autoprovisioned: bool = Query(
+        default=False,
+        description=(
+            "When true, controller-managed (autoprovisioned) passthrough VirtualModels are excluded from the results."
```
mckornfield, line 363: just so I understand, this would get rid of all the ones on an external provider from showing up? like build?

web/packages/studio/src/components/dataViews/VirtualModelsDataView/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+  const { data, isFetching, error } = useListVirtualModels(
+    workspace,
+    {
+      page: dataViewState.pagination.state.pageIndex + 1,
+      page_size: dataViewState.pagination.state.pageSize,
+      sort: sortParam,
+      filter,
+      // Autoprovisioned VMs are controller-managed passthroughs; hide them from the UI.
+      exclude_autoprovisioned: true,
```
mckornfield, line 95: ok so we never want to query for these on the UI?

mckornfield (approved):

web/packages/studio/src/components/dataViews/VirtualModelsDataView/index.tsx
marcusds: Yeah just a lot of noise and nothing we can do with them from a CRUD perspective.

services/core/inference-gateway/src/nmp/core/inference_gateway/api/v2/virtual_models.py
marcusds: I'm not totally sure when auto provisioned ones come from, but its something like that.

PR #690, opened 2026-07-14
feat: Check models before opening preview response stream

mikeknep:
Today on `main`, the model health check calls (a simple "say hello" ping) don't happen until we call `DataDesigner.preview`, which happens _after_ we open the response stream. This means that it's possible to send a `200` response and start streaming back log messages, only for the nearly-first-and-only message to be "model X failed a health check".

The upstream library just added `DataDesigner.check_models`, a method to run the model health checks for a given config without triggering the full workload execution. This allows us to do a health check **before** opening the response stream, meaning if something goes wrong we can return a more appropriate 4xx or 5xx response.

## Summary by CodeRabbit

* **New Features**
  * Preview streaming now includes dataset metadata, dataset records, processor outputs/artifacts, and analysis.
  * Preview performs model health checks up front and reports failures as configuration validation errors.

* **Bug Fixes**
  * Improved preview error handling for invalid configs, worker failures, and timeouts, with clearer client-facing validation errors.
  * Preview streaming now closes deterministically and only completes when no error was emitted.
  * Person reader retrieval is now optional in both local and remote contexts.

* **Tests**
  * Updated and extended preview tests to cover worker failure, model-health-check, and streaming behaviors.


plugins/nemo-data-designer/src/nemo_data_designer_plugin/functions/preview.py
```diff
@@ -66,38 +76,48 @@ def send_from_thread(frame: BaseModel) -> None:
...
-                    model_providers,
-                    model_configs,
-                    num_records,
-                )
-            except Exception as exc:
+                data_designer.check_models(config_builder)
+            except DataDesignerError as e:
+                raise NDDInvalidConfigError(str(e))
+            except TimeoutError as e:
+                raise NDDInternalError(str(e))
```
mikeknep: This and the integration test are the meat of this PR: call `check_models` and potentially raise _before_ we start the worker stream. The rest of the diff is refactoring to inline ("in-module"?) a previously separate function.

packages/data_designer_nemo/src/data_designer_nemo/context.py
```diff
@@ -93,8 +92,8 @@ def get_seed_readers(self) -> list[SeedReader]:
             FilesetFileSeedReader(self._sdk),
         ]
 
-    def get_person_reader(self) -> PersonReader:
-        return create_person_reader(str(MANAGED_ASSETS_PATH))
+    def get_person_reader(self) -> PersonReader | None:
+        return None
```
mikeknep, line 98: By returning None here, we delegate to the upstream library's default, which means no code duplication, no drift, and lets the library handle any necessary setup

mckornfield: might be worth a code comment with exactly this?

plugins/nemo-data-designer/src/nemo_data_designer_plugin/functions/_preview_worker.py
mckornfield, line 1: more of this code was moved into the lib yeah? or just shuffled into the next file?

packages/data_designer_nemo/src/data_designer_nemo/context.py
benmccown: Nit: Maybe worth logging as a code comment? No strong feelings from my side tho.

benmccown (approved):

plugins/nemo-data-designer/src/nemo_data_designer_plugin/functions/preview.py
```diff
@@ -117,3 +137,42 @@ def _validate_and_get_num_records(requested_num_records: int | None, is_local: b
...
+    Synchronous function that runs on a worker thread under
+    :func:`anyio.to_thread.run_sync`. SDK calls bridge back to the API
+    process's event loop via :func:`anyio.from_thread.run` inside the
+    helpers. Sends frames back to the async context via the
+    ``send_frame`` callback.
+    """
+    with forward_data_designer_logs(send_frame):
+        preview_results = data_designer.preview(config_builder, num_records=num_records)
+
+    if (dataset_metadata := preview_results.dataset_metadata) is not None:
```
mckornfield, line 158: look at you, walrus guy

mckornfield (approved):

packages/data_designer_nemo/src/data_designer_nemo/context.py
mikeknep: Yup, good calls, added a comment for this

plugins/nemo-data-designer/src/nemo_data_designer_plugin/functions/_preview_worker.py
mikeknep: just shuffled into the next file as a private function; didn't warrant a standalone python module

plugins/nemo-data-designer/src/nemo_data_designer_plugin/functions/preview.py
mikeknep: 🦭

PR #691, opened 2026-07-14
chore: update e2e safe-synthesizer tests

mckornfield:
## Summary by CodeRabbit

* **Tests**
  * Reworked Safe Synthesizer end-to-end coverage to run via a local subprocess smoke harness by default (replacing container-completion assertions).
  * Added Safe Synthesizer API health/status readiness checks.
  * Added Fileset upload/download round-trip validation with temporary test datasets and automatic cleanup.
  * Added Safe Synthesizer job lifecycle coverage: create, list, retrieve, cancel, and delete, with more robust status waiting.
  * Kept GPU/Kubernetes workflow tests opt-in, including cancel transitions, PII replacement completion, and full workflow artifact downloads.

### DIFF
--- a/e2e/test_safe_synthesizer.py
+++ b/e2e/test_safe_synthesizer.py
@@ -24,7 +24,7 @@ import subprocess
 import sys
 import time
 import uuid
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from contextlib import suppress
 from pathlib import Path
 from typing import Any
@@ -51,8 +51,12 @@ SMOKE_JOB_TIMEOUT_SECONDS = 180.0
 K8S_JOB_TIMEOUT_SECONDS = float(os.environ.get("NSS_E2E_JOB_TIMEOUT_SECONDS", "5400"))
 POLL_INTERVAL_SECONDS = float(os.environ.get("NSS_E2E_POLL_INTERVAL_SECONDS", "10"))
 RESULT_DOWNLOAD_TIMEOUT_SECONDS = float(os.environ.get("NSS_E2E_RESULT_DOWNLOAD_TIMEOUT_SECONDS", "600"))
+MODEL_FILESETS_TIMEOUT_SECONDS = float(os.environ.get("NSS_E2E_MODEL_FILESETS_TIMEOUT_SECONDS", "300"))
+DELETE_VERIFY_TIMEOUT_SECONDS = float(os.environ.get("NSS_E2E_DELETE_VERIFY_TIMEOUT_SECONDS", "60"))
 DEFAULT_INPUT_ROWS = int(os.environ.get("NSS_E2E_INPUT_ROWS", "250"))
 DEFAULT_NUM_RECORDS = int(os.environ.get("NSS_E2E_NUM_RECORDS", "250"))
+
+NssJobFactory = Callable[[str, str, dict[str, Any]], dict[str, Any]]
 
 
 def _unique_name(prefix: str) -> str:
@@ -181,6 +185,10 @@ def _list_nss_jobs(sdk: NeMoPlatform, workspace: str) -> dict[str, Any]:
     return response.json()
 
 
+def _job_names(jobs: dict[str, Any]) -> set[str]:
+    return {str(entry["name"]) for entry in jobs.get("data", [])}
+
+
 def _retrieve_nss_job(sdk: NeMoPlatform, workspace: str, name: str) -> dict[str, Any]:
     response = sdk._client.get(
         _nss_url(sdk, workspace, f"jobs/{name}"),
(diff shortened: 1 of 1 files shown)


crookedstorm (approved):
LGTM (not that you need my approval in this folder)

e2e/test_safe_synthesizer.py
```diff
@@ -79,65 +359,214 @@ def nss_model_filesets(sdk: NeMoPlatform, _services: str) -> None:
...
+
+        _cancel_nss_job(sdk, workspace, job_name)
+        status, _ = _wait_for_status(
+            sdk,
+            workspace,
+            job_name,
+            timeout_seconds=SMOKE_JOB_TIMEOUT_SECONDS,
+            poll_interval_seconds=2.0,
+        )
+        assert status in TERMINAL_STATUSES
```
tylersbray: `TERMINAL_STATUSES` includes `"error"`, so a job that fails immediately would still pass this test even though the intent is cancel lifecycle. Consider asserting `status == "cancelled"` after the explicit cancel (or at least `status != "error"` before cancel).

The k8s cancel test has the same looseness at line 484: `final_status in {"cancelled", "completed", "error"}` allows non-cancel outcomes, and `assert history + final_history` is tautological once either list is non-empty. Prefer `assert final_status == "cancelled"` and something like `assert "cancelled" in final_history`.

e2e/test_safe_synthesizer.py
```diff
@@ -79,65 +359,214 @@ def nss_model_filesets(sdk: NeMoPlatform, _services: str) -> None:
...
+        config={
+            "enable_synthesis": False,
+            "enable_replace_pii": True,
+        },
+    )
+    try:
+        _assert_job_completed(sdk, workspace, job_name)
+        results = _list_nss_results(sdk, workspace, job_name)
+        assert {"summary", "synthetic-data"}.issubset(_result_names(results))
+        _assert_csv_rows(
```
tylersbray: Row-count checks alone do not prove PII replacement worked. The fixture CSV includes identifiable emails and phone numbers (`customer{N}@example.com`, `415-555-...` in the review text). Consider asserting known PII strings are absent or changed in the downloaded `synthetic-data` CSV, e.g. `assert "customer1@example.com" not in content`.

e2e/test_safe_synthesizer.py
```diff
@@ -1,74 +1,354 @@
...
 @pytest.fixture(scope="module")
-def nss_model_filesets(sdk: NeMoPlatform, _services: str) -> None:
-    """Register HuggingFace-backed model filesets required by Safe Synthesizer tasks."""
+def nss_model_filesets(sdk: NeMoPlatform) -> None:
+    if os.environ.get("NSS_E2E_SKIP_MODEL_FILESETS") == "1":
+        return
+
+    platform_root = _platform_root()
+    script = platform_root / "plugins/nemo-safe-synthesizer/scripts/setup_model_filesets.py"
     result = subprocess.run(
```
tylersbray: This `subprocess.run` has no `timeout=`. If `setup_model_filesets.py` hangs (e.g. stuck HF/network call), the module-scoped fixture can block the suite indefinitely — pytest test-level timeout will not reliably kill a blocked subprocess.

Suggest adding something like `timeout=300` (or an env override) and handling `subprocess.TimeoutExpired` with the captured stdout/stderr in the failure message.

e2e/test_safe_synthesizer.py
```diff
@@ -79,65 +359,214 @@ def nss_model_filesets(sdk: NeMoPlatform, _services: str) -> None:
...
+            sdk,
+            workspace,
+            job_name,
+            timeout_seconds=SMOKE_JOB_TIMEOUT_SECONDS,
+            poll_interval_seconds=2.0,
+        )
+        assert status in TERMINAL_STATUSES
+    finally:
+        _cancel_nss_job(sdk, workspace, job_name)
+        _delete_nss_job(sdk, workspace, job_name)
```
tylersbray: Delete is called in `finally` but never verified. Consider asserting the job is gone after cleanup, e.g. expect 404 from `_retrieve_nss_job` (or confirm the name is absent from `_list_nss_jobs`) so a silent delete failure does not leave orphaned jobs in shared workspaces.

e2e/test_safe_synthesizer.py
```diff
@@ -79,65 +359,214 @@ def nss_model_filesets(sdk: NeMoPlatform, _services: str) -> None:
...
+
+def test_safe_synthesizer_job_create_list_retrieve_cancel_delete(
     sdk: NeMoPlatform,
     workspace: str,
+    nss_dataset: tuple[str, str],
+) -> None:
+    _, data_source = nss_dataset
+    job_name = _unique_name("nss-smoke")
+
+    job = _create_nss_job(
```
tylersbray: The create / try / finally / cancel / delete pattern is duplicated across four tests here. Consider a factory fixture, e.g. `nss_job`, that yields a `_create(prefix, data_source, config) -> name` helper and performs cancel+delete cleanup in `finally`. Would DRY the repeated boilerplate without hiding test-specific assertions.

tylersbray (approved):
Adding one more non-blocking approval modulo those hopefully easy feedback items, thanks.

PR #696, opened 2026-07-15
chore: make helm versioning follow repo version

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Added a manual `helm-version` override for stable, Helm-only releases, with validation.
  * Resolved the Helm chart version during release planning for both nightly and stable builds (nightly uses a timestamped version from the latest reachable tag; stable uses the release version unless overridden).
  * Updated Helm packaging to use the resolved chart version and rewrite Helm image repositories to the correct nightly vs stable registry.
* **Documentation**
  * Updated releasing guidance with the optional `helm-version` input, including an example and updated “What the workflow does” Helm versioning details.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -148,6 +148,7 @@ jobs:
       release_label: ${{ steps.plan.outputs.release_label }}
       nightly_timestamp: ${{ steps.plan.outputs.nightly_timestamp }}
       wheel_version: ${{ steps.wheel-version.outputs.wheel_version }}
+      helm_version: ${{ steps.helm-version.outputs.helm_version }}
       wheel_ids: ${{ steps.plan.outputs.wheel_ids }}
       container_ids: ${{ steps.plan.outputs.container_ids }}
       has_wheels: ${{ steps.plan.outputs.has_wheels }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -148,6 +153,7 @@ jobs:
       has_wheels: ${{ steps.plan.outputs.has_wheels }}
       has_containers: ${{ steps.plan.outputs.has_containers }}
       include_helm: ${{ steps.plan.outputs.include_helm }}
+      helm_version_override: ${{ steps.plan.outputs.helm_version_override }}
```
crookedstorm, line 157: Is that an output from plan?

crookedstorm: Ugh...expanded the view. Now I see it.

.github/workflows/release.yaml
```diff
@@ -680,17 +705,37 @@ jobs:
...
-            chart_version="${chart_version}-nightly-${NIGHTLY_TIMESTAMP}"
             release_registry="${NIGHTLY_IMAGE_REGISTRY}"
           else
-            # TODO: Decide whether stable chart versions should track the platform
-            # release version or remain independently managed in Chart.yaml.
-            chart_version="${RELEASE_VERSION}"
             release_registry="${STABLE_IMAGE_REGISTRY}"
           fi
 
+          chart_version="${HELM_VERSION_OVERRIDE}"
```
crookedstorm: What if we just move helm version logic into the plan job entirely. This just consumes a helm-version output. That would make this bash bit easier to read.

Also, we cannot rely on tags. Tag happens at the end.

.github/workflows/release.yaml
```diff
@@ -643,6 +665,8 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           ref: ${{ needs.plan-release.outputs.source_sha }}
+          fetch-depth: 0
```
crookedstorm: no need for this since the tag isn't going to be useful most likely at this point in the setup.

mckornfield: I guess my thought was for the nightlies, not for a full release (but maybe this is the wrong spot)

crookedstorm (approved):
Looks good. Follows the general patterns of the repo, and we can always try a manual nightly once all the containers are fixed.

PR #706, opened 2026-07-15
fix(platform-runner): seed NMP_BASE_URL from config platform.base_url

benmccown:
## What this touches

`NMP_BASE_URL` — the URL other components use to reach the platform's own API.
The platform injects it into deployed agents so they can call back through the
Inference Gateway; if it's wrong for a containerized agent, that agent can't
reach any models. It can be set three ways, resolved in this order:

```
NEMO_BASE_URL env  >  NMP_BASE_URL env  >  platform.base_url (config file)
```

- **`NMP_BASE_URL` env** exists so a deployer (Helm/k8s) can point the platform
  at its real service address (e.g. `http://nemo-platform-api:8080`); it takes
  priority because the deployer knows the true address.
- **`platform.base_url` in config** is the operator-facing equivalent for a
  self-hosted/standalone run — you'd set it to whatever address is reachable in
  your environment (e.g. the docker bridge `http://172.17.0.1:8080` so
  containers can reach the host).

## The bug

When you start the platform with `nemo services run`, the runner
(`apply_run_environment`) **auto-fills `NMP_BASE_URL`** for standalone users who
haven't set it — it derives a value from the server's own bind address:

```python
connect_host  = _connect_host_for_internal_clients(config.host)  # 0.0.0.0 -> 127.0.0.1
default       = f"http://{connect_host}:{port}"                  # http://127.0.0.1:8080
env.setdefault("NMP_BASE_URL", default)
```

`--host 0.0.0.0` means "bind all interfaces," which isn't itself a
connectable address, so `_connect_host_for_internal_clients` translates it to
the loopback `127.0.0.1` (fine for in-process clients on the same host). The
result — `http://127.0.0.1:8080` — is written into `NMP_BASE_URL` via
`setdefault`.

The problem: `NMP_BASE_URL` (env) ranks **above** `platform.base_url` (config).
So the runner's auto-derived loopback value **silently overrides** whatever you
set as `platform.base_url` in your config — the config setting becomes a no-op,
with no error or warning. The only way to make it stick today is to *also*
export `NMP_BASE_URL` yourself before launch, which defeats the purpose of the
config field.

The user-visible symptom: an agent deployed in a container gets
`llms.*.base_url = http://127.0.0.1:8080`, which inside the container points at
the container itself, so its model calls fail. (Helm/k8s deployments are
unaffected — there `NMP_BASE_URL` is set externally to the real, reachable
address, so nothing is being shadowed.)

## Fix

`apply_run_environment` now uses an explicit `platform.base_url` from the config
file as the default it seeds `NMP_BASE_URL` with (read from raw YAML so a
file-set value is distinguishable from the schema default), and only falls back
to the bind-derived loopback when the config doesn't set one. An externally
provided `NMP_BASE_URL` still wins via `setdefault`, so Helm/k8s is unchanged.

New precedence: **external env > config `platform.base_url` > bind-derived default.**

## Testing

- New unit tests in `test_config.py`: seeds from config, external env still
  wins, falls back when config omits `base_url`, falls back when the file is
  missing. Full runner suite green.
- Verified end-to-end in a dev pod: with only `platform.base_url` set (no
  `NMP_BASE_URL` env), a docker-mode agent now deploys and invokes through the
  gateway (previously failed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined how runtime `NMP_BASE_URL` is derived: existing environment values are preserved; otherwise the configured `platform.base_url` supplies the host while the bind port is applied.
  * Improved handling of missing/malformed config, including IPv6 and wildcard hosts, and ensured `NMP_AUTH_POLICY_DECISION_POINT_BASE_URL` matches the final value.

* **Tests**
  * Expanded coverage for these precedence and fallback behaviors, including config present/absent cases.

### DIFF
--- a/packages/nmp_platform_runner/src/nmp/platform_runner/config.py
+++ b/packages/nmp_platform_runner/src/nmp/platform_runner/config.py
@@ -156,20 +156,23 @@ def apply_run_environment(
     In standalone mode these variables are absent, so setdefault fills them in.
 
     Precedence for NMP_BASE_URL: an externally-provided value (Helm / k8s) wins,
-    then an explicit ``platform.base_url`` *host* from the config file combined
-    with the actual bind port, then a value derived entirely from the bind
-    host/port. Seeding the host from the config file lets operators set a base
+    then the scheme + host of an explicit ``platform.base_url`` from the config
+    file combined with the actual bind port, then a value derived entirely from
+    the bind host/port. Seeding from the config file lets operators set a base
     URL reachable from inside deployed agent containers (e.g. the Docker bridge
     address) via config alone; without this the bind-derived loopback default
     would silently shadow the configured value.
 
-    Only the host (and scheme) of the configured ``platform.base_url`` is
+    Only the scheme and host of the configured ``platform.base_url`` are
     honored — its port is replaced with the port the server actually binds. A
     config that hardcodes ``:8080`` must not point internal clients (and the
     embedded PDP) at 8080 when the platform is launched on another port (which
     the e2e harness always does, and any ``nemo services run --port`` differing
     from 8080 does); doing so leaves internal HTTP clients unable to reach the
-    server and the platform never becomes ready.
+    server and the platform never becomes ready. The configured host is run
+    through the same wildcard -> loopback normalization as the bind host, so a
+    config like ``http://0.0.0.0:8080`` (the bundled ``local.yaml`` default)
+    still yields a connectable internal base URL.
     """
     if env is None:
         env = os.environ
@@ -177,13 +180,14 @@ def apply_run_environment(
     connect_host = _connect_host_for_internal_clients(config.host)
     effective_host = env.setdefault("NMP_SERVICE_HOST", connect_host)
     effective_port = env.setdefault("NMP_SERVICE_PORT", str(config.port))
-    normalized = effective_host.strip("[]")
-    url_host = f"[{normalized}]" if ":" in normalized else normalized
-    config_base_url_host = _config_file_base_url_host(config.config_path)
-    if config_base_url_host is not None:
-        default_base_url = f"http://{config_base_url_host}:{effective_port}"
+    config_base_url_parts = _config_file_base_url_parts(config.config_path)
+    if config_base_url_parts is not None:
+        scheme, config_host = config_base_url_parts
+        host_for_url = _bracket_ipv6(_connect_host_for_internal_clients(config_host))
+        default_base_url = f"{scheme}://{host_for_url}:{effective_port}"
     else:
-        default_base_url = f"http://{url_host}:{effective_port}"
+        host_for_url = _bracket_ipv6(effective_host)
+        default_base_url = f"http://{host_for_url}:{effective_port}"
     base_url = env.setdefault("NMP_BASE_URL", default_base_url)
     # Embedded PDP is served from the same platform process; keep the auth client
     # origin aligned with NMP_BASE_URL when services run on a non-default port.
@@ -201,19 +205,21 @@ def _set_or_clear_env(env: MutableMapping[str, str], name: str, values: set[str]
         env.pop(name, None)
 
 
-def _config_file_base_url_host(config_path: str) -> str | None:
-    """Return the host of an explicit ``platform.base_url`` from the config file.
+def _config_file_base_url_parts(config_path: str) -> tuple[str, str] | None:
+    """Return the (scheme, host) of an explicit ``platform.base_url`` from config.
 
     Reads the raw YAML rather than the merged config object so a value present
     in the file can be told apart from the schema default (the merged config
-    always carries ``base_url``). Returns the host component (already bracketed
-    for IPv6 so it can be dropped into an ``http://<host>:<port>`` URL), or
-    ``None`` when the file is missing, unreadable, does not set
-    ``platform.base_url``, or the value has no parseable host.
+    always carries ``base_url``). Returns the URL scheme (defaulting to
+    ``"http"`` when the config omits one) and the host component, or ``None``
+    when the file is missing, unreadable, does not set ``platform.base_url``, or
+    the value has no parseable host.
 
-    Only the host is returned — callers pair it with the actual bind port, so a
-    config that hardcodes a port (e.g. ``:8080``) does not point internal
-    clients at the wrong port when the platform runs on a different one.
+    The host is returned unbracketed (as ``urlparse`` yields it) so callers can
+    normalize it (e.g. wildcard -> loopback) before composing the final URL.
+    Only the scheme and host are returned — callers pair them with the actual
+    bind port, so a config that hardcodes a port (e.g. ``:8080``) does not point
+    internal clients at the wrong port when the platform runs on a different one.
     """
     try:
         global_settings = Configuration.get_global_settings_from_file(config_path)
@@ -225,12 +231,16 @@ def _config_file_base_url_host(config_path: str) -> str | None:
     base_url = platform_settings.get("base_url")
     if not isinstance(base_url, str) or not base_url:
         return None
-    parsed = urlparse(base_url)
+    try:
+        parsed = urlparse(base_url)
+    except ValueError:
+        # Malformed value (e.g. an unterminated bracketed IPv6 like
+        # ``http://[::1``). Fall back to the bind-derived default rather than
+        # aborting startup — a bad config value should fail soft here.
+        return None
     if not parsed.hostname:
         return None
-    # urlparse lowercases and strips IPv6 brackets from hostname; re-bracket so
-    # the host can be safely composed back into an http://<host>:<port> URL.
-    return f"[{parsed.hostname}]" if ":" in parsed.hostname else parsed.hostname
+    return parsed.scheme or "http", parsed.hostname
 
 
 def _connect_host_for_internal_clients(host: str) -> str:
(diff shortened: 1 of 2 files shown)


packages/nmp_platform_runner/src/nmp/platform_runner/config.py
```diff
@@ -179,6 +187,25 @@ def _set_or_clear_env(env: MutableMapping[str, str], name: str, values: set[str]
         env.pop(name, None)
 
 
+def _config_file_base_url(config_path: str) -> str | None:
+    """Return an explicit ``platform.base_url`` from the config file, if set.
```
mckornfield: the joys of merging configs

mckornfield (approved):
e2e tests go brr?

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
```diff
@@ -153,6 +154,22 @@ def apply_run_environment(
...
+
+    Precedence for NMP_BASE_URL: an externally-provided value (Helm / k8s) wins,
+    then an explicit ``platform.base_url`` *host* from the config file combined
+    with the actual bind port, then a value derived entirely from the bind
+    host/port. Seeding the host from the config file lets operators set a base
+    URL reachable from inside deployed agent containers (e.g. the Docker bridge
+    address) via config alone; without this the bind-derived loopback default
+    would silently shadow the configured value.
+
+    Only the host (and scheme) of the configured ``platform.base_url`` is
```
tylersbray: **Non-blocking — docstring vs implementation:** This says the configured scheme is honored, but the URL construction below always prefixes `http://`. Consider either preserving `parsed.scheme` (when present) or narrowing the docstring to "host only" so it matches behavior.

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
```diff
@@ -162,7 +179,12 @@ def apply_run_environment(
     effective_port = env.setdefault("NMP_SERVICE_PORT", str(config.port))
     normalized = effective_host.strip("[]")
     url_host = f"[{normalized}]" if ":" in normalized else normalized
-    base_url = env.setdefault("NMP_BASE_URL", f"http://{url_host}:{effective_port}")
+    config_base_url_host = _config_file_base_url_host(config.config_path)
+    if config_base_url_host is not None:
+        default_base_url = f"http://{config_base_url_host}:{effective_port}"
```
tylersbray: **Non-blocking — HTTPS configs:** A configured `https://172.17.0.1` (or similar) would be downgraded to `http://` here. Unlikely for the docker-bridge standalone case this PR targets, but worth a follow-up test/fix if anyone runs TLS-terminated or HTTPS base URLs from config.

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
```diff
@@ -179,6 +201,38 @@ def _set_or_clear_env(env: MutableMapping[str, str], name: str, values: set[str]
...
+        global_settings = Configuration.get_global_settings_from_file(config_path)
+    except (OSError, ValueError):
+        return None
+    platform_settings = global_settings.get("platform")
+    if not isinstance(platform_settings, dict):
+        return None
+    base_url = platform_settings.get("base_url")
+    if not isinstance(base_url, str) or not base_url:
+        return None
+    parsed = urlparse(base_url)
```
tylersbray: **Non-blocking — malformed IPv6 hardening:** `urlparse()` can raise `ValueError` on malformed bracketed IPv6 (e.g. `http://[::1`). That would abort startup even when `NMP_BASE_URL` is already set externally. A small try/except here returning `None` (fall back to bind-derived default) would make bad config fail soft.

packages/nmp_platform_runner/tests/test_config.py
```diff
@@ -210,3 +210,60 @@ def test_derives_base_url_from_effective_host_port(self):
         }
         apply_run_environment(_make_config(host="0.0.0.0", port=8080), env=env)
         assert env["NMP_BASE_URL"] == "http://nemo-platform-api:9090"
+
+
+class TestApplyRunEnvConfigBaseUrl:
```
tylersbray, line 217: **Non-blocking — bundled default config:** The bundled `local.yaml` files set `platform.base_url: http://0.0.0.0:8080`. With this PR that value is now explicit in raw YAML, so default `nemo services run` will seed `NMP_BASE_URL` with `0.0.0.0` instead of the previous bind-derived `127.0.0.1`. Docker agent deploy may still work (loopback rewrite in `container_gateway_url`), but a test using `default_config_path()` would document the intended behavior and guard against regressions for internal PDP/readiness clients.

tylersbray (commented):
Non-blocking feedback from review — looks sufficient for the stated bug. Inline notes below are optional polish, not merge blockers.

packages/nmp_platform_runner/tests/test_config.py
```diff
@@ -210,3 +210,60 @@ def test_derives_base_url_from_effective_host_port(self):
...
+    port hardcoded in the config. An externally-provided NMP_BASE_URL still
+    wins.
+    """
+
+    def _write_config(self, tmp_path, body: str) -> str:
+        path = tmp_path / "config.yaml"
+        path.write_text(body, encoding="utf-8")
+        return str(path)
+
+    def test_seeds_base_url_host_from_config_file(self, tmp_path):
```
tylersbray: **Non-blocking — test gap:** Nice coverage for the core scenarios. Optional follow-ups: an HTTPS `platform.base_url` case (if scheme preservation is added), and an automated e2e for docker-mode agent invoke (manual dev-pod verification is great; an e2e would lock it in).

tylersbray:
**Non-blocking — documentation follow-up:** `SETUP.md` still tells users to `export NMP_BASE_URL=...` without mentioning that `platform.base_url` now seeds `NMP_BASE_URL` when the env var is unset. A one-line note there would help operators discover the config-only path this PR enables. (Not in this diff, so filing here rather than inline.)

tylersbray (approved):
+1 approval, left some optional non-blockers.

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
```diff
@@ -162,7 +179,12 @@ def apply_run_environment(
     effective_port = env.setdefault("NMP_SERVICE_PORT", str(config.port))
     normalized = effective_host.strip("[]")
     url_host = f"[{normalized}]" if ":" in normalized else normalized
-    base_url = env.setdefault("NMP_BASE_URL", f"http://{url_host}:{effective_port}")
+    config_base_url_host = _config_file_base_url_host(config.config_path)
+    if config_base_url_host is not None:
+        default_base_url = f"http://{config_base_url_host}:{effective_port}"
+    else:
+        default_base_url = f"http://{url_host}:{effective_port}"
+    base_url = env.setdefault("NMP_BASE_URL", default_base_url)
```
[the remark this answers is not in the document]
benmccown: Good catch — fixed in `ee811b11`. `urlparse` is now wrapped in `try/except ValueError`, so a malformed value (e.g. `http://[::1`) falls back to the bind-derived default instead of aborting startup. Added a regression test (`test_malformed_ipv6_config_falls_back_to_bind_derived`).

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
benmccown: Fixed in `ee811b11` — `urlparse` now fails soft (`try/except ValueError` → `None`) rather than aborting startup on malformed bracketed IPv6. Regression test added (`test_malformed_ipv6_config_falls_back_to_bind_derived`).

packages/nmp_platform_runner/tests/test_config.py
```diff
@@ -210,3 +210,60 @@ def test_derives_base_url_from_effective_host_port(self):
...
+        env: dict[str, str] = {}
+        apply_run_environment(_make_config(host="0.0.0.0", port=59007, config_path=config_path), env=env)
+        assert env["NMP_BASE_URL"] == "http://172.17.0.1:59007"
+        assert env["NMP_AUTH_POLICY_DECISION_POINT_BASE_URL"] == "http://172.17.0.1:59007"
+
+    def test_config_base_url_without_port_gets_bind_port(self, tmp_path):
+        config_path = self._write_config(tmp_path, "platform:\n  base_url: http://172.17.0.1\n")
+        env: dict[str, str] = {}
+        apply_run_environment(_make_config(host="0.0.0.0", port=9090, config_path=config_path), env=env)
+        assert env["NMP_BASE_URL"] == "http://172.17.0.1:9090"
```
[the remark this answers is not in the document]
benmccown: Fixed in `ee811b11`. We now carry the parsed scheme (`_config_file_base_url_parts` returns `(scheme, host)`) and build `{scheme}://{host}:{port}`, so a configured `https://` is preserved rather than downgraded to `http://`. Added `test_preserves_https_scheme_from_config`.

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
benmccown: Fixed in `ee811b11` — rather than narrowing the docstring, we made the behavior match it: the configured scheme is now preserved (not forced to `http://`), so "scheme and host are honored" is accurate.

packages/nmp_platform_runner/src/nmp/platform_runner/config.py
benmccown: Fixed in `ee811b11`: a configured `https://` base URL is now preserved instead of downgraded to `http://`. Added an HTTPS regression test (`test_preserves_https_scheme_from_config`).

packages/nmp_platform_runner/tests/test_config.py
benmccown: Fixed in `ee811b11`. The config-derived host now goes through the same wildcard→loopback normalization as the bind host, so the bundled `local.yaml` (`http://0.0.0.0:8080`) seeds `http://127.0.0.1:<port>` and internal PDP/readiness clients stay connectable. Added `test_wildcard_host_in_config_normalized_to_loopback` and `test_bundled_default_config_seeds_loopback` (which exercises the real `default_config_path()`).

PR #693, opened 2026-07-14
fix(studio): Fix empty UI in DD details

steramae-nvidia:
Screenshot 2026-07-14 at 3 19 44 PM

fix(studio): Fix empty UI in DD details

Signed-off-by: Sean Teramae <steramae@nvidia.com>

fix TZ in details

Signed-off-by: Sean Teramae <steramae@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Added the ability to edit complete text files in-browser up to an 8 MB limit, with clear errors when files are too large.
  * Display dataset profiling job logs when analysis can’t be loaded or no results are generated (including terminal job states).
* **Bug Fixes**
  * Fixed the “Save File” workflow to allow a second edit/save cycle after a prior successful save.
  * Improved job timestamp formatting for timezone-free date strings.
* **UI Improvements**
  * Refined template card sizing for better wrapping.
  * When a file exceeds the editing limit, users are prompted to download it for local editing.

PR #685, opened 2026-07-14
feat(studio): Support seeding DD

steramae-nvidia:
Screenshot 2026-07-14 at 11 59 00 AM
Screenshot 2026-07-14 at 11 58 59 AM

Signed-off-by: Sean Teramae <steramae@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Added seed-dataset column configuration with fileset/file selection, sampling strategy, and automatic “available columns” discovery.
  * Data Designer job building now incorporates seed configuration and dependency relationships for seed-based outputs.
  * Added automatic column detection for CSV, JSONL, and JSON content.
  * Column-type options in the add-column palette can now be disabled with user-visible explanations.
* **Bug Fixes**
  * Prevented adding more than one seed-dataset column per recipe.

PR #661, opened 2026-07-13
fix(studio): DD embedding template fix

steramae-nvidia:
This PR fixes the embedding template. 

Screenshot 2026-07-14 at 10 02 27 AM

Signed-off-by: Sean Teramae <steramae@nvidia.com>

## Summary by CodeRabbit

- **New Features**
  - Added `scores` configuration support for LLM judge columns, with validation that expects a JSON array.
  - Enabled embedding inference parameter support for embedding models (used by the Semantic search dataset).
- **Improvements**
  - Log viewers now stay pinned to the newest entries while auto-scrolling is enabled, without disrupting manual browsing; the same behavior applies to preview logs.
  - Updated default model selections to NVIDIA/Nemotron variants.
- **UI Updates**
  - Unified the Data Designer icon across navigation and empty states.
- **Tests**
  - Added coverage for stick-to-bottom behavior and expanded validation tests for LLM judge/preference-pairs.

PR #647, opened 2026-07-10
feat(studio): More templates and visual fixes

steramae-nvidia:
Add templates
Screenshot 2026-07-13 at 4 04 08 PM

Signed-off-by: Sean Teramae <steramae@nvidia.com>

## Summary by CodeRabbit

* **New Features**
  * Added new fileset templates (code generation/validation, structured data extraction, reward modeling, semantic search, and expression transforms).
  * Added validator parameter configuration with JSON guidance.
  * Improved Data Designer job building by incorporating served model names and applying default inference settings when generating model configs.

* **UI Improvements**
  * Updated template selection to use a grid layout for better responsiveness.
  * Enhanced template cards by clipping overflow, improving description sizing, and truncating long descriptions with hover/title support.

PR #711, opened 2026-07-15
fix(studio): Disable templates if no provider (w/ hint)

steramae-nvidia:
Disable templates when missing inference provider for build.nvidia.com since it won't be able to make inference requests for the LLM columns. 

Screenshot 2026-07-16 at 12 23 53 PM

Signed-off-by: Sean Teramae <steramae@nvidia.com>


steramae-nvidia:

- #528
    - #529
        - #530
            - #537
                - #546
                    - #630
                        - #647
                            - #661
                                - #685
                                    - #693
                                        - #711 ◀

nv-odrulea (changes requested):
Can we use a more generic validation for the models that does not force the user to depend on build.nvidia.com  or other NVIDIA endpoints?

steramae-nvidia:
Closing to support feedback. Will instead, allow the user to go through to the dag view without any providers added. We will instead show an error immediately on mount to alert the user that they need to set the model to a usable one.

PR #714, opened 2026-07-15
feat(experiments): Add support for multi-field sort on Experiments

shanaiabuggy:
https://github.com/user-attachments/assets/e1c1225e-821a-45fa-81ac-5968b6a00a96

## Summary by CodeRabbit

* **New Features**
  * Added multi-column evaluation sorting using ordered, comma-separated `sort` fields with optional `-` for descending.
  * Updated default sort configuration in the UI to support an ordered list of sort fields (add/remove, tie-breakers) and to keep sorting in sync with URLs.

* **Bug Fixes**
  * Improved validation to reject empty/blank, unknown, or unsortable sort tokens consistently.
  * Enforced rollup/telemetry availability checks for metric-based rollup sorts (failing with the expected error when unavailable).

* **Documentation**
  * Clarified OpenAPI documentation for `sort` and `default_sort` to match the ordered comma-separated multi-field format, defaults, and examples.

* **Tests**
  * Added/expanded unit and endpoint coverage for multi-field parsing, ordering, whitespace handling, and error cases.


web/packages/studio/src/components/DefaultSortControl/index.tsx
```diff
@@ -69,55 +71,125 @@ export const DefaultSortControl: FC<DefaultSortControlProps> = ({
...
-            <SelectTrigger
-              className="w-48"
-              placeholder="Select field"
-              aria-label="Sort field"
-              renderValue={(v) => (typeof v === 'string' && v ? labelForOption(v) : undefined)}
+        <Stack gap="density-sm">
+          {rows.map((row, index) => (
+            <SortFieldRow
+              // Rows are fully controlled by props (no internal state), so an index key is stable here.
+              key={index}
```
walston, line 180: The comment that WAS here wasn't correct (so you were right to remove it) BUT...
You'd want to use an identifier, like a `row.id`, so the DOM node could be moved by React instead of recreated. We're not really hitting the number of rows where that is critical, I think, but that comment was so wrong it said the literal opposite of what it was assuring the next dev of.

walston (commented):
very minor change but maybe use `row.id` instead of `index`

BrianNewsom (approved):

walston (approved):

PR #715, opened 2026-07-15
test(nemo-agents): add docker-mode agent deployment e2e test

benmccown:
## What

Adds the first e2e test that deploys an agent as a real **Docker container** (via the nemo-deployments plugin) and invokes it through the agents gateway — covering the container-mode path the existing subprocess-mode agent e2e tests don't.

Exercises the full chain end to end:

```
sdk.agents.invoke -> agents gateway proxy (container-mode endpoint resolution)
  -> docker agent container (nat start fastapi)
  -> Inference Gateway -> mock provider short-circuit
  -> response back through the gateway
```

## How

- **Reuses the `nmp-api` image** as the agent runtime. It already ships the NAT runtime (`nvidia-nat-core`/`-langchain` via `nemo-agents-plugin`), so `nat` is on `PATH` and the `chat_completion`/`openai` components resolve. The deployments docker executor overrides the entrypoint with `nat start fastapi`, so the image's own entrypoint is irrelevant — no bespoke agent image is built.
- **Image supplied prebuilt** via the existing `NMP_E2E_IMAGE_REGISTRY` / `NMP_E2E_IMAGE_TAG` convention (deploys `{registry}/nmp-api:{tag}`).
- **Hermetic:** the agent uses a deterministic single-LLM `chat_completion` workflow served by the e2e mock inference provider — no `NVIDIA_API_KEY` or model egress; asserts the exact mocked completion round-trips.
- **Networking:** runs the platform as a local process (subprocess harness). A new `container_base_url_host` harness option binds the platform on all interfaces and rewrites `platform.base_url` to the docker bridge, so the injected Inference Gateway URL is reachable from inside the agent container while the platform's own in-process clients still reach it.

## CI

New **`python-e2e-image-test`** job (sibling of `python-e2e-test`): depends on `build-cpu-smoke-images`, logs in to GHCR, pulls `nmp-api` (timed), and runs only `-m needs_nmp_api_image`. Gated like `kind-cpu-e2e` (`publish_images`) and, like it, left out of the required aggregate gate since it skips on forks. The `needs_nmp_api_image` marker skips the test in the plain subprocess e2e job and the kind cluster job (which don't provide the image).

## Changes

- `e2e/test_nemo_agents_docker.py` — the test.
- `e2e/configs/local-docker-agents.yaml` — docker runtime + nemo-deployments docker executor.
- `e2e/services_pool.py` — `container_base_url_host` subprocess-harness option.
- `conftest.py` + `pytest.ini` — `needs_nmp_api_image` marker + skip logic.
- `.github/workflows/ci.yaml` — the `python-e2e-image-test` job.

### DIFF
--- a/e2e/services_pool.py
+++ b/e2e/services_pool.py
@@ -61,6 +61,19 @@ class E2EHarnessConfig(TypedDict, total=False):
     lifecycle: Literal["fresh", "reuse"]
     compose_project_prefix: str
     env: dict[str, str]
+    # Subprocess backend only. When set, the harness (a) binds the platform on all
+    # interfaces (``--host 0.0.0.0``) and (b) rewrites ``platform.base_url`` in the
+    # rendered config to ``http://<this host>:<the free port it chose>`` right
+    # before launch. The platform still runs as a normal local process, but the
+    # base URL it injects into container-mode agent deployments (docker/k8s) then
+    # points at a host reachable from *inside* those containers (e.g. the docker
+    # bridge ``172.17.0.1``) instead of a loopback the container can't reach.
+    # Binding all interfaces is required so the platform's own in-process clients,
+    # which also use ``NMP_BASE_URL``, can still reach it at that host. The runner
+    # seeds ``NMP_BASE_URL`` from the config ``platform.base_url`` host (paired
+    # with the real bind port), so the rewritten host takes effect instead of the
+    # bind-derived loopback default.
+    container_base_url_host: str
 
 
 @dataclass
@@ -345,6 +358,8 @@ def _resolve_e2e_harness_config_from_node(node: Node) -> E2EHarnessConfig:
     if backend not in {"subprocess", "docker", "docker_compose"}:
         raise pytest.UsageError(f"unsupported e2e harness backend: {backend}")
     normalized["backend"] = backend
+    if "container_base_url_host" in normalized and backend != "subprocess":
+        raise pytest.UsageError("container_base_url_host is only supported with the 'subprocess' harness backend")
     if backend == "docker_compose":
         required = {"compose_file", "service_url"}
         missing = sorted(required - set(normalized))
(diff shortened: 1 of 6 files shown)


e2e/services_pool.py
```diff
@@ -61,6 +61,19 @@ class E2EHarnessConfig(TypedDict, total=False):
     lifecycle: Literal["fresh", "reuse"]
     compose_project_prefix: str
     env: dict[str, str]
+    # Subprocess backend only. When set, the harness (a) binds the platform on all
```
mckornfield: hmmm should we just leave this comment or its effect below? seems weird all the attributes are here are uncommented except this one

e2e/test_nemo_agents_docker.py
```diff
@@ -0,0 +1,342 @@
...
+    sdk.agents.invoke (gateway proxy, container-mode endpoint resolution)
+      -> docker agent container (nat start fastapi)
+      -> Inference Gateway /openai (base_url injected at deploy time)
+      -> mock provider short-circuit (no real upstream / no API key)
+      -> response back through the gateway
+
+How it runs, and where:
+
+- The deployed agent runs from the platform's own ``nmp-api`` image, which
+  already ships the NAT runtime (``nvidia-nat-core`` / ``nvidia-nat-langchain``,
```
mckornfield, line 19: will this always be true? I guess when it's not, this test will fail lol

e2e/test_nemo_agents_docker.py
```diff
@@ -0,0 +1,342 @@
...
+
+        response = sdk.agents.invoke(
+            workspace=workspace,
+            agent=agent_name,
+            input="What is 12 multiplied by 8?",
+        )
+        content = response["choices"][0]["message"]["content"]
+        assert _TEST_AGENT_RESPONSE in content, response
+    finally:
+        _run_cleanup_steps(
```
mckornfield: lol this fn is interesting. Could also just do a `with suppress(Exception)` around each method call instead of passing a lambda

mckornfield (approved):
we discussed in slack, it's probably fine to depend on the nmp-api image, yay more ci checks lol

e2e/test_nemo_agents_docker.py
benmccown: Yeah, this won't always be true :/. But when this test fails it should be straight forward for the implementer to pivot to an alternative (build a bespoke image just for this or use a test/example agent that has its deps in `nmp-api` still).

e2e/test_nemo_agents_docker.py
benmccown: Went with your proposal 👍

e2e/services_pool.py
benmccown: Trimmed it down to a one line comment and moved the longer explanation to the `_start_services_subprocess` func where this attribute is used.

crookedstorm (approved):
LGTM from devops side

PR #717, opened 2026-07-15
feat(files): profile a fileset via a job

albcui:
## Summary

Wires the profiler into the platform:

- `POST .../filesets/{name}/profile` submits a CPU job that runs the dataset profiler over the FileSet and writes the resulting DatasetProfile onto the fileset's dataset metadata.
- `GET .../filesets/{name}/profile` retrieves the stored profile with the current profiling job state.
- The profile is stored as its own entity rather than inside fileset metadata, and the task reads the fileset through range requests instead of downloading it.

## Summary by CodeRabbit

* **New Features**
  * Added fileset profiling through the API and CLI.
  * Start or reuse background profiling jobs and retrieve job status or stored dataset profiles.
  * Added optional per-partition row budgets for profiling.
  * Profiling now supports reading filesets directly and publishing computed profiles.
  * Added permissions for viewing and running fileset profiles.

* **Documentation**
  * Updated CLI and authorization references with fileset profiling commands and permissions.


docs/cli/reference.mdx
```diff
@@ -1154,6 +1156,39 @@ nemo files filesets delete [OPTIONS] NAME
 
 * `--help, -h`: Show this message and exit.
 
+##### nemo files filesets get-profile
+
+Get the stored dataset profile for a fileset, with the current profiling state.
+
+`state` is `ready` (a profile is available), `running` (a job is in flight),
+`failed` (the last job errored or was cancelled and no profile exists), or
```
anubhutivyas: for my understanding - why did we choose to categorize "cancelled" as "failed" and not a separate status?

services/core/auth/src/nmp/core/auth/assets/static-authz.yaml
```diff
@@ -20,6 +20,12 @@ authz:
       update:
         description: "Update entities"
         has_role: false
+    files:
+      filesets:
```
anubhutivyas: I think we should rename this as something else, may be "profile"? We have a separate list of permissions under "filesets" already, see the next line. It might become confusing to read.

services/core/files/src/nmp/core/files/api/v2/filesets/endpoints.py
```diff
@@ -438,6 +537,17 @@ async def update_fileset_metadata(
...
 
+    # The dataset profile is a server-managed, derived artifact that reads strip out (see
+    # fileset_output_from_entity), so a client never sees it — and a wholesale metadata PATCH
+    # would null it back out. Preserve the stored profile unless this request explicitly carries
+    # one (the profiler task writes a fresh profile through this same PATCH path).
+    if "metadata" in request.model_fields_set and request.metadata is not None:
+        stored_dataset = fileset.metadata.dataset if fileset.metadata else None
+        stored_profile = stored_dataset.profile if stored_dataset else None
+        incoming_dataset = request.metadata.dataset
+        if stored_profile is not None and incoming_dataset is not None and incoming_dataset.profile is None:
```
anubhutivyas: A request `PATCH {"metadata": {...}}` with the dataset key omitted leaves `incoming_dataset=request.metadata.dataset` as None, it would result in a None profile.

PR #723, opened 2026-07-16
feat(evaluator): Harbor runner target + agent-eval job list & create-serialization fixes

SandyChapman:
## Summary

Three related changes to the evaluator plugin's agent-eval path:

### 1. Harbor runner target (feature)
- Adds `HarborRunnerTarget` to the agent-eval runner-target union and wires it through `_resolve_target` (native `HarborAgentTaskRunner`, `jobs_dir` injected from job storage) and result-entity trait extraction (also fills in the previously-missing Fabric trait branch).
- Declares a **marker-gated** `harbor` extra on `nemo-evaluator-sdk`: `harbor = ["harbor>=0.16.1; python_version >= '3.12'"]`. Harbor requires Python ≥3.12; the marker scopes it to the 3.12 slice of the universal lock so `pip install nemo-evaluator-sdk[harbor]` works on 3.12 while the workspace floor stays 3.11 (no floor bump). `harbor` stays lazily imported, so importing the SDK / resolving a Harbor target is valid on 3.11 — only *executing* a run needs the extra.

### 2. Mixed-job list-endpoint 500 (bug fix — reported by Studio)
The evaluator plugin owns two job types (`evaluate` + `agent-evaluate`) that shared one `source` tag. The list endpoints filter by `source`, so in a workspace with both, each collection returned the other type's records and 500'd rendering the foreign spec (dozens of validation errors). Fix: give `AgentEvalJob` a distinct source via `add_job_routes(service_name=AGENT_EVAL_JOB_SOURCE)`. `EvaluateJob` keeps its derived `nemo-evaluator` source so existing records still match.

> **Pre-GA caveat:** agent-evaluate records persisted *before* this change keep the old shared source and will still cross-render into the `evaluate` list until re-sourced. agent-evaluate is pre-GA with negligible persisted data, so no migration ships here (documented in `service.py`).

### 3. Plugin job create serialization for bytes-bearing specs (latent bug fix)
The shared job factory coerced the transformed spec with `model_dump()` (python mode), leaving raw `bytes` (e.g. an inline cloudpickle metric bundle) that the typed jobs client couldn't JSON-encode — `PydanticSerializationError: invalid utf-8` on the `\x80` pickle marker. Fix: dump `mode="json"` so nested payload models' `ser_json_bytes="base64"` runs and the blob round-trips. This is in the shared `api_factory`, so it fixes create for any plugin/job type whose spec carries bytes. (Latent because the opt-in E2E suite that exercises it isn't in CI.)

## Testing

- **Unit:** Harbor resolve/compile/`run_local` (faked evaluator) across target types; harbor/fabric result traits; distinct-source list-filter forwarding; `nemo-evaluator-sdk` source distinctness; create-path bytes-serialization regression driving the real `POST /jobs` route.
- **Integration (opt-in):** mixed-job list partitioning against a live subprocess platform; a real Docker-backed Harbor e2e via `NemoJobScheduler().run_local(...)` (gated on `harbor` + Docker; oracle solves hello-world, reward 1.0 persisted).
- Validated on both Python 3.11 (CI's default) and 3.12.13; `uv lock --check` clean; ruff + ty clean for changed files.

## Follow-up

- **AALGO-343** (related) — Harbor targets under containerized job backends (docker/k8s). Harbor works via the host daemon (in-process `run_local` / subprocess backend) but not inside the `cpu-tasks` container; direction is to route Harbor at a remote/k8s sandbox rather than plugin-side DinD.

## Summary by CodeRabbit

* **New Features**
  * Added agent-evaluation support for Harbor runner targets (agent selection, model overlays, execution/retry/concurrency settings, artifacts/traces, and reward scoring).
  * Updated OpenAPI to expose Harbor runner targets in both the input and canonical spec.
* **Bug Fixes**
  * Improved job-spec JSON transport for create requests containing binary payloads (safe encoding for round-tripping).
  * Isolated agent-evaluation job listing to prevent cross-type validation/rendering.
* **Tests**
  * Added unit and integration coverage for Harbor target resolution, persistence mapping, mixed listing behavior, and local Harbor execution.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/api_factory.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/api_factory.py
@@ -873,7 +873,13 @@ def job_route_factory(
             # the request body's ``spec`` is a plain dict on the wire. The
             # Stainless SDK serialized models implicitly — the typed client
             # validates the body first, so coerce to a dict here.
-            # JSON mode so binary fields (e.g. cloudpickle blobs) serialize as base64, not raw bytes.
+            #
+            # Dump ``mode="json"`` (not python): the spec is bound for JSON transport and JSON storage,
+            # and the request body's ``spec`` is an opaque ``dict`` that has lost the nested models'
+            # serializers. A python-mode dump leaves non-JSON-native values (e.g. a cloudpickle metric
+            # bundle's raw ``bytes``, whose ``ser_json_bytes="base64"`` never runs) that then fail to
+            # JSON-encode on the wire (invalid utf-8 on the pickle marker). ``mode="json"`` runs each
+            # nested model's own JSON serializer, so bytes base64-encode and round-trip on read.
             spec_dict = job_spec.model_dump(mode="json") if isinstance(job_spec, BaseModel) else job_spec
 
             # Only include optional fields when they have values — passing None
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -1634,6 +1634,7 @@ components:
           - $ref: '#/components/schemas/AgentTargetInput'
           - $ref: '#/components/schemas/CodexRunnerTarget'
           - $ref: '#/components/schemas/FabricRunnerTarget'
+          - $ref: '#/components/schemas/HarborRunnerTarget'
           title: Target
           description: 'What generates trials online: a Model or Agent endpoint, or
             an agent runner (e.g. Codex CLI). Endpoint targets carry their own request
@@ -1786,6 +1787,7 @@ components:
           - $ref: '#/components/schemas/AgentTargetOutput'
           - $ref: '#/components/schemas/CodexRunnerTarget'
           - $ref: '#/components/schemas/FabricRunnerTarget'
+          - $ref: '#/components/schemas/HarborRunnerTarget'
           title: Target
           description: 'What generates trials online: a Model or Agent endpoint, or
             an agent runner (e.g. Codex CLI). Endpoint targets carry their own request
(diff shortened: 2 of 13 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/service.py
```diff
@@ -42,7 +49,10 @@ class EvaluatorPluginService(NemoService):
     def get_routers(self) -> list[RouterSpec]:
         router = APIRouter()
         jobs_router = add_job_routes(EvaluateJob, authz=scope)
-        agent_jobs_router = add_job_routes(AgentEvalJob, authz=scope)
+        # A distinct ``service_name`` gives agent-evaluate jobs their own ``source`` tag so this
+        # collection's list endpoint doesn't return (and 500 rendering) row EvaluateJob specs, and
+        # vice versa. EvaluateJob keeps its derived ``nemo-evaluator`` source. See AGENT_EVAL_JOB_SOURCE.
+        agent_jobs_router = add_job_routes(AgentEvalJob, service_name=AGENT_EVAL_JOB_SOURCE, authz=scope)
```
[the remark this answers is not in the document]
SandyChapman: This hasn't hit a release yet, so we won't be migrating erroneous data.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -108,9 +108,45 @@ class FabricRunnerTarget(BaseModel):
...
+    tree into SDK trials. The dataset Harbor runs against is recovered from each task's
+    ``harbor_dataset_path`` metadata, so it is not configured here. The runtime-only jobs directory is
+    injected from the job's storage at run time; only the harness-selection and run knobs live here.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    kind: Literal["harbor"] = "harbor"
+    agent_name: str | None = Field(
+        default="oracle",
```
ngoncharenko, line 125: why `oracle` is default? Is it Harbor convention?

ngoncharenko (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/api_factory.py
```diff
@@ -873,7 +873,14 @@ async def create_job(
...
             # validates the body first, so coerce to a dict here.
-            spec_dict = job_spec.model_dump() if isinstance(job_spec, BaseModel) else job_spec
+            #
+            # Dump ``mode="json"`` (not python): the spec is bound for JSON transport and JSON storage,
+            # and the request body's ``spec`` is an opaque ``dict`` that has lost the nested models'
+            # serializers. A python-mode dump leaves non-JSON-native values (e.g. a cloudpickle metric
+            # bundle's raw ``bytes``, whose ``ser_json_bytes="base64"`` never runs) that then fail to
+            # JSON-encode on the wire (invalid utf-8 on the pickle marker). ``mode="json"`` runs each
+            # nested model's own JSON serializer, so bytes base64-encode and round-trip on read.
+            spec_dict = job_spec.model_dump(mode="json") if isinstance(job_spec, BaseModel) else job_spec
```
arpitsardhana, line 883: I think, I merged this fix as part of QA bug fix recently

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
arpitsardhana: yup, oracle is harbor default agent name

arpitsardhana (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/api_factory.py
SandyChapman: I'll rebase and check.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -3140,6 +3142,83 @@ components:
...
+          items:
+            type: string
+          type: array
+          title: Artifacts
+          description: Harbor artifact sources to collect per trial.
+        trace_dir:
+          title: Trace Dir
+          description: Container path of agent traces to collect as the 'traces' artifact
+            (e.g. '/app/traces').
+          type: string
```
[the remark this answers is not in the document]
SandyChapman: Skipping: openapi.yaml is generated, and the generator renders every `str | None` field as plain `type: string` (see sibling `CodexRunnerTarget.model`). Hand-adding `type: null` to only these fields would be inconsistent with the rest of the spec and reverted on the next `make refresh-openapi`.

PR #732, opened 2026-07-16
ci: harmonize test harness a little with GPU test setup

crookedstorm:
E2E tests running in other repos occasionally have a problem with missing kubernetes resource profiles that aren't surfaced here. Adding them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes-based model job execution by using the appropriate GPU execution profile during CPU preparation.
  * Ensured CPU preparation and GPU training steps use consistent execution settings while selecting the correct CPU resources.

* **Documentation**
  * Clarified execution profile behavior for model update jobs.

### DIFF
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/config.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/config.py
@@ -84,6 +84,20 @@ def get_default_executor_profiles_for_runtime(
                     backend="kubernetes_job",
                     config=defaults.kubernetes_job,
                 ),
+                # Customizer uses the gpu profile for both CPU preparation and
+                # GPU training steps. The provider selects the requested resources.
+                KubernetesJobExecutionProfile(
+                    provider="cpu",
+                    profile="gpu",
+                    backend="kubernetes_job",
+                    config=defaults.kubernetes_job,
+                ),
+                KubernetesJobExecutionProfile(
+                    provider="gpu",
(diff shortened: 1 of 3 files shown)


services/core/jobs/src/nmp/core/jobs/controllers/backends/config.py
```diff
@@ -84,6 +84,20 @@ def get_default_executor_profiles_for_runtime(
                     backend="kubernetes_job",
                     config=defaults.kubernetes_job,
                 ),
+                # Customizer uses the gpu profile for both CPU preparation and
+                # GPU training steps. The provider selects the requested resources.
+                KubernetesJobExecutionProfile(
+                    provider="cpu",
+                    profile="gpu",
```
mckornfield, line 91: lol so funny lookin

mckornfield (approved):

ironcommit:
So this is a profile called "gpu", which has providers for cpu and gpu, which both run on a gpu.  I have to admit that this datamodel is a bit difficult.  But I understand the need here.

ironcommit (approved):

svvarom (approved):

crookedstorm:
> So this is a profile called "gpu", which has providers for cpu and gpu, which both run on a gpu. I have to admit that this datamodel is a bit difficult. But I understand the need here.

I'm not the biggest fan, but it will be easier to change if we need to once we have everything working :)

PR #734, opened 2026-07-16
feat: add local services lifecycle SDK with Unix domain socket support

ironcommit:
## Summary

Extracts the local services process lifecycle out of the CLI into a standalone SDK module (`local.services`, `local.process`, `local.transport`) so that Python callers can programmatically start, stop, connect to, and manage NeMo Platform service instances without importing CLI internals.

The key addition is first-class **Unix domain socket (UDS) transport** support across the entire stack — from the platform runner and uvicorn server binding, through inter-service HTTP clients (auth, health checks, SDK factory), to the jobs-launcher Go sidecar.

### New local services SDK (`nemo_platform_ext.local` / `nemo_platform.local`)
- `ServiceRunConfig` — declarative config for transport (UDS or TCP), socket paths, instance naming, and readiness settings
- `ServiceHandle` — returned after start; provides `client()`, `async_client()`, `stop()`, `wait_until_ready()`, and `is_running()`
- `run_services()` — foreground blocking start with flock-based liveness tracking
- `daemonize_services()` / `daemonize_services_async()` — background daemon start with readiness polling
- `connect_services()` / `connect_services_async()` — get-or-start convenience that returns a wired `NeMoPlatform` SDK client
- `ensure_services()`, `get_service_handle()`, `list_service_handles()`, `stop_services()` — full lifecycle management
- Old `cli/commands/services/_process.py` reduced to a compatibility re-export alias

### Platform endpoint resolution (`nmp.common.platform_endpoint`)
- New `PlatformEndpoint` dataclass with typed `transport` field (`"tcp"` | `"uds"`) and `sync_http_client()` / `async_http_client()` factory methods
- `parse_platform_endpoint()` — parses `http://`, `https://`, and `unix://` URLs into the typed model
- `resolve_platform_endpoint()` / `resolve_service_endpoint()` — resolve from `NMP_BASE_URL` or per-service `NMP_<SERVICE>_URL` env vars
- Integrated into `sdk_factory.py`, `auth/client.py`, `auth/middleware.py`, `service/base.py`, and `service/api/health.py` so all inter-service HTTP clients correctly use UDS transport when configured

### Platform runner UDS support
- `run_platform()`, `resolve_run_configuration()`, and `run_server()` accept `socket_path` parameter
- `apply_run_environment()` sets `NMP_BASE_URL=unix://<path>` when a socket path is present
- uvicorn binds to UDS via `uds=` parameter instead of `host`/`port`

### Jobs launcher Go changes
- New `nmpclient/endpoint.go` — parses `unix://` endpoints and configures Go HTTP clients with Unix socket dialer
- OTel exporter updated to route through UDS when `NMP_BASE_URL` uses `unix://`
- Job backends (`docker`, `kubernetes`, `subprocess`) propagate `NMP_BASE_URL` from the environment

### Tests
- Unit tests for `ServiceRunConfig` validation, `ServiceHandle` client wiring, `PlatformEndpoint` parsing, SDK factory UDS integration, runner config, and server socket binding
- Go tests for endpoint parsing and UDS client construction

## Summary by CodeRabbit

* **New Features**
  * Added programmatic local service lifecycle management (foreground/daemon), readiness probing over TCP or Unix sockets, instance discovery, and log rotation, including a dedicated daemon child entrypoint.
  * Introduced typed platform/service endpoint resolution with UDS support and optional UDS socket binding in the platform runner.
  * Added configurable OTEL log exporting (stdout/none/OTLP) with launcher-specific endpoint/headers/protocol/timeout/compression, plus updated secret fetching over UDS.

* **Bug Fixes**
  * Standardized PDP URL/client setup and made readiness polling more robust across transports.

* **Tests**
  * Expanded coverage for local lifecycle, UDS transport/status behavior, endpoint parsing, and OTEL/secret propagation.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/local/process.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/local/process.py
@@ -3,8 +3,13 @@
 
 """Local process lifecycle for ``nemo services``.
 
-Uses per-instance scoped directories under ``$XDG_STATE_HOME/nmp/instances/``
-with flock-based liveness tracking.  Each instance directory contains:
+In this module, "instance" is a local services process/resource, and "scope"
+is the stable key used for that instance's lock, descriptor, socket, and log
+paths.  The CLI exposes this key as ``--instance`` for compatibility, but
+internal code should use "scope" when referring to the key.
+
+Uses per-scope directories under ``$XDG_STATE_HOME/nmp/instances/``
+with flock-based liveness tracking. Each scope directory contains:
 
 - ``services.lock`` -- exclusive flock held for the process lifetime
 - ``instance.json``  -- descriptor with PID, port, services, etc.
@@ -23,7 +28,6 @@ import hashlib
 import json
 import logging
 import os
-import re
 import shutil
 import signal
 import socket
@@ -34,10 +38,16 @@ import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Literal
+from typing import Literal, Self
 
 import psutil
-from pydantic import BaseModel, Field
+from nmp.platform_runner.config import (
+    DEFAULT_LOCAL_SERVICES_BIND_HOST,
+    PlatformAppConfig,
+    default_state_root,
+    validate_scope,
+)
+from pydantic import BaseModel, Field, model_validator
 
 logger = logging.getLogger(__name__)
 
@@ -45,10 +55,10 @@ LOCK_FILENAME = "services.lock"
 DESCRIPTOR_FILENAME = "instance.json"
 LOG_FILENAME = "services.log"
 
-DEFAULT_SERVICES_BIND_HOST = "127.0.0.1"
 SUGGESTED_ALT_PORT = 9090
 
 _SIGTERM_POLL_INTERVAL = 0.25
+_SIGKILL_WAIT_TIMEOUT = 5.0
 _DEFAULT_STOP_TIMEOUT = 30.0
 
 
@@ -62,10 +72,7 @@ def _pause(seconds: float) -> None:
 
 
 def _base_state_dir() -> Path:
-    xdg = os.environ.get("XDG_STATE_HOME")
-    if xdg:
-        return Path(xdg) / "nmp"
-    return Path.home() / ".local" / "state" / "nmp"
+    return default_state_root()
 
 
 def _instances_dir(*, base_dir: Path | None = None) -> Path:
@@ -73,7 +80,7 @@ def _instances_dir(*, base_dir: Path | None = None) -> Path:
 
 
 def _find_git_root() -> str:
-    """Walk up from cwd looking for a ``.git`` directory.  Falls back to cwd."""
+    """Walk up from cwd looking for a ``.git`` directory. Falls back to cwd."""
     cur = Path.cwd().resolve()
     for parent in (cur, *cur.parents):
         if (parent / ".git").exists():
@@ -84,24 +91,19 @@ def _find_git_root() -> str:
 _scope_prefix_cache: str | None = None
 
 
-_SCOPE_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
+def compute_scope(*, port: int, explicit_scope: str | None = None) -> str:
+    """Compute the local services scope.
 
+    The default scope is ``sha1(git_toplevel_or_cwd)[:8]-<port>``.  Including
+    the port is intentional: it lets two local services instances from the same
+    checkout use different TCP ports without sharing a lock, descriptor, or log
+    directory.
 
-def _validate_scope(scope: str) -> str:
-    """Ensure *scope* is safe to use as a directory name."""
-    if not _SCOPE_RE.fullmatch(scope):
-        raise ValueError(f"Invalid instance scope: {scope!r}")
-    return scope
-
-
-def compute_scope(*, port: int, instance_name: str | None = None) -> str:
-    """Compute a scope identifier for this working directory + port.
-
-    Default: ``sha1(git_toplevel_or_cwd)[:8]-<port>``.
-    Override with an explicit *instance_name*.
+    Explicit scopes are validated and returned as-is, so they do not encode the
+    port.  Callers that pass an explicit scope own its uniqueness.
     """
-    if instance_name:
-        return _validate_scope(instance_name)
+    if explicit_scope:
+        return validate_scope(explicit_scope)
     global _scope_prefix_cache  # noqa: PLW0603
     if _scope_prefix_cache is None:
         root = _find_git_root()
@@ -110,7 +112,8 @@ def compute_scope(*, port: int, instance_name: str | None = None) -> str:
 
 
 def instance_dir(scope: str, *, base_dir: Path | None = None) -> Path:
-    d = _instances_dir(base_dir=base_dir) / _validate_scope(scope)
+    """Return the state directory for *scope*, creating it if needed."""
+    d = _instances_dir(base_dir=base_dir) / validate_scope(scope)
     d.mkdir(parents=True, exist_ok=True)
     return d
 
@@ -222,7 +225,7 @@ def _instance_owns_listener(
     desc = read_descriptor(scope, base_dir=base_dir)
     if desc is None:
         return False
-    return desc.port == port and _normalize_bind_host(desc.host) == _normalize_bind_host(host)
+    return desc.config.port == port and _normalize_bind_host(desc.config.host) == _normalize_bind_host(host)
 
 
 def is_port_bindable(host: str, port: int) -> bool:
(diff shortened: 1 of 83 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/local/process.py
```diff
@@ -0,0 +1,792 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Local process lifecycle for ``nemo services``.
```
mckornfield, line 4: most of this is a move then yeah? from services/_process.py?

packages/nemo_platform_ext/src/nemo_platform_ext/local/services.py
```diff
@@ -0,0 +1,574 @@
...
+    data_dir: str | Path | None = None
+    readiness_timeout: float = 60.0
+    readiness_poll_interval: float = 0.5
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "services", _as_tuple(self.services))
+        object.__setattr__(self, "controllers", _as_tuple(self.controllers))
+        object.__setattr__(self, "sidecars", _as_tuple(self.sidecars))
+
+        if self.services and self.service_group:
```
mckornfield, line 165: should we just use a pydantic model then or no?

packages/nemo_platform_ext/src/nemo_platform_ext/local/services.py
```diff
@@ -0,0 +1,576 @@
...
+            controllers=list(cfg.controllers) if cfg.controllers is not None else None,
+            service_group=cfg.service_group,
+            controller_group=cfg.controller_group,
+            sidecars=list(cfg.sidecars) if cfg.sidecars is not None else None,
+            config_path=str(cfg.config_path) if cfg.config_path is not None else None,
+            log_path=str(log_path),
+        )
+        process.write_descriptor(desc, base_dir=cfg.state_root)
+        run_platform(
+            services=list(cfg.services) if cfg.services is not None else None,
```
mckornfield: wtb a `_list_or_none` little method

packages/nmp_common/src/nmp/common/service/api/health.py
```diff
@@ -6,15 +6,54 @@
...
+    services = data.get("services") or {}
+    if not isinstance(services, Mapping):
+        return None
+
+    ready = _status_names(services.get("ready") or [])
+    if service_name in ready:
+        return True
+
+    not_ready = _status_names(services.get("not_ready") or [])
+    return service_name not in not_ready
```
mckornfield: so this could happen if the service being checked is a typo yeah? should we error if it's not in ready AND not in not_ready?

packages/nmp_common/src/nmp/common/platform_endpoint.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Literal
+
+import httpx
+from httpx._types import TimeoutTypes
+from nmp.common.config import PlatformConfig
+
+UDS_BASE_URL = "http://nemo-platform.local"
```
mckornfield, line 17: is this like a docker related URL then? or does it require dns mucking? or some uds thing I don't understand lol

packages/nmp_platform_runner/src/nmp/platform_runner/server.py
```diff
@@ -210,12 +216,86 @@ async def root_handler() -> Response:
...
+    run_funcs: dict[str, Callable[[threading.Event], object]] = {}
+    for name in names:
+        value = registry[name]
+        run_funcs[name] = value if callable(value) else load_controller_run_func(name, value)
+    return run_funcs
+
+
+def build_platform_app(
+    *,
+    services: list[str] | None = None,
```
mckornfield: would love a data structure instead of such large kwarg lists

services/core/jobs/jobs-launcher/cmd/run.go
```diff
@@ -10,17 +10,22 @@ import (
 	"fmt"
```
mckornfield, line 10: confused, is this work to the jobs-launcher also in support of the domain socket stuff?

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess_runtime.py
```diff
@@ -55,6 +63,79 @@ def close(self) -> None:
...
+    def __init__(self, socket_path: str) -> None:
+        super().__init__()
+        self._client = httpx.Client(
+            transport=httpx.HTTPTransport(uds=socket_path),
+            follow_redirects=True,
+        )
+
+    def request(
+        self,
+        method: Any,
```
mckornfield, line 76: lol Any all the way

mckornfield (approved):

services/core/jobs/jobs-launcher/cmd/run.go
ironcommit: yeah, needed to fix a few things so that requests could go over TCP and UDP.  This also solves some issues related to permissions on the logs, which could go in another PR, but handled them here.

packages/nmp_common/src/nmp/common/platform_endpoint.py
ironcommit: for UDS the domain is ignored.

services/core/jobs/src/nmp/core/jobs/controllers/backends/subprocess_runtime.py
ironcommit: yeah, this is just mirroring the type from the requests package, not great.

packages/nmp_platform_runner/src/nmp/platform_runner/server.py
ironcommit: I moved a bunch of fields into a PlatformAppConfig struct, which makes sense as a grouping.

packages/nmp_common/src/nmp/common/service/api/health.py
ironcommit: Logic improved.

packages/nemo_platform_ext/src/nemo_platform_ext/local/services.py
ironcommit: improved

packages/nemo_platform_ext/src/nemo_platform_ext/local/services.py
ironcommit: yeah, good call.

packages/nemo_platform_ext/src/nemo_platform_ext/local/process.py
ironcommit: removed

packages/nemo_platform_ext/src/nemo_platform_ext/local/transport.py
```diff
@@ -0,0 +1,154 @@
...
+        timeout=timeout,
+    )
+
+
+def build_sync_http_client(socket_path: Path, *, timeout: HttpxTimeout = _DEFAULT_TIMEOUT) -> httpx.Client:
+    return httpx.Client(
+        transport=httpx.HTTPTransport(uds=str(socket_path)),
+        follow_redirects=True,
+        timeout=timeout,
+    )
```
mikeknep, line 59: I think this poses a serious problem for Data Designer. The Data Designer plugin passes Inference Gateway URLs to the upstream library, where the library constructs its own client—we **do not** pass an entire client to the library. When the library creates its client, it will not have the UDS transport, and so it will try normal TCP/DNS and fail.

This has actually already been an issue with the ASGI-transport client setup in our tests. The DD plugin integration tests test real interactions with the service using an ASGI-transport client, but none of the configs we supply in those tests require/run inference (they are all "technically valid but in practice fairly useless" configs that only include deterministic, non-inference-based columns) because the ASGI transport does not get passed to the upstream library. We **do** have some e2e tests that make inference requests and leverage the IGW test helpers that allow setting predefined responses, but those only work in the e2e setup where we aren't using ASGI.

ironcommit: UDS is planned only for the services sdk.  Access to IGW will require TCP for a number of reasons.

packages/nmp_common/src/nmp/common/auth/middleware.py
```diff
@@ -166,7 +166,10 @@ def _get_client(self, request: Request) -> httpx.AsyncClient:
             An async HTTP client configured with auth.policy_decision_point_request_timeout_seconds
         """
         if self._client is None:
-            self._client = httpx.AsyncClient(timeout=self.config.policy_decision_point_request_timeout_seconds)
+            from nmp.common.platform_endpoint import parse_platform_endpoint
```
mikeknep: is there a particular reason this import is lazy? I see it lazily imported like this in a few modules, not just here.

mikeknep (approved):

packages/nmp_common/src/nmp/common/auth/middleware.py
ironcommit: Unfortunately we have some module cycles, which this works around.  But some of them can be removed.

PR #737, opened 2026-07-16
fix(customizer): namespace backend schemas to fix OpenAPI name collision

albcui:
Automodel and Unsloth each define models named `TrainingSpec` (plus LoRAParams, DatasetSpec, ScheduleSpec, BatchSpec, OptimizerSpec, OutputRequest, OutputResponse). Because both backends mount into one `/apis/customization` FastAPI app, Pydantic's module-qualified `$defs` keys were collapsed by the platform's schema-name normalizer, so both `*JobInput.training` ended up `$ref`-ing a single `TrainingSpec` — Unsloth's request contract silently took Automodel's shape.

Layer 1 (the fix): add a NamespacedModel metaclass base. Pydantic freezes a model's JSON-schema name at class creation, so per-backend subclasses (AutomodelSchema / UnslothSchema) prefix each owned model there, emitting AutomodelTrainingSpec / UnslothTrainingSpec. Top-level names already starting with the prefix (AutomodelJobInput, ...) are left unchanged.

Layer 2 (backstop): `tweak_spec` / `_normalize_refs_and_schema_keys` gain a `strict_collisions` flag. It raises on a differing-content name collision so a future un-namespaced backend fails the build with an actionable message. The flag is enabled only for the self-contained plugin specs (where the customization app lives); platform/service specs keep the legacy warn-and-collapse behavior, since they carry pre-existing collisions.

While enabling the gate it surfaced a pre-existing duplicate: two identical BackendFormat enums (core-models vs nemo_platform_plugin.inference_middleware). Dedupe by having core-models re-export the plugin's canonical enum — the plugin is the lower-level package core-models already depends on, and IGW already treats it as canonical. Behavior-preserving; the platform spec is byte-identical.

Known follow-ups (left as warnings, not fixed here): DeleteResponse and GenericSortField each collide across common / core-entities / guardrails with genuinely different content.

## Summary by CodeRabbit

* **New Features**
  * Added namespaced OpenAPI schemas for customization backends (Automodel, Unsloth, RL) to reduce component name collisions when apps are merged.
  * Expanded generated job-spec schemas with backend-specific variants (training/dataset/schedule/batch/optimizer/parallelism/output, plus RL/Unsloth families).
  * Added configurable strict handling for OpenAPI schema-name collisions (warn vs fail).

* **Bug Fixes**
  * Improved schema collision detection with deduplication, consistent `$ref` rewriting, and clearer warning/error behavior.

* **Tests**
  * Added coverage for strict vs non-strict collision scenarios and namespace-based schema naming.

### DIFF
--- a/packages/nmp_customization_common/src/nmp/customization_common/schema.py
+++ b/packages/nmp_customization_common/src/nmp/customization_common/schema.py
@@ -1,0 +1,59 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Namespaced Pydantic base for customization-backend schemas.
+
+Every backend (automodel, unsloth, …) is merged into one FastAPI app under
+``/apis/customization``, so two backends that each define e.g. ``TrainingSpec``
+collide in the generated OpenAPI. Pydantic freezes a model's JSON-schema name at
+class-creation time, so the prefix has to be applied there — this metaclass does
+it. ``class TrainingSpec(AutomodelSchema)`` is emitted as ``AutomodelTrainingSpec``.
+
+Usage: declare a per-backend subclass that sets ``__schema_namespace__`` and
+inherit *that* from every model the backend owns::
+
+    class AutomodelSchema(NamespacedModel):
+        __schema_namespace__ = "Automodel"
+
+    class TrainingSpec(AutomodelSchema):  # emitted as ``AutomodelTrainingSpec``
+        ...
+
+A model whose class name already starts with the prefix (e.g.
+``AutomodelJobInput``) is left unchanged, so top-level request/response names stay
+stable.
+"""
+
+from __future__ import annotations
+
+from typing import ClassVar
+
+from pydantic import BaseModel, ConfigDict
+
+# ``type(BaseModel)`` is pydantic's ``ModelMetaclass``; deriving from it this way
+# avoids importing pydantic internals.
+_ModelMeta = type(BaseModel)
+
+
+class _NamespacedMeta(_ModelMeta):
+    def __new__(mcs, name, bases, ns, **kw):
+        prefix = ns.get("__schema_namespace__") or next(
+            (p for b in bases if (p := getattr(b, "__schema_namespace__", None))), None
+        )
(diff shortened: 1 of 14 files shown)


soluwalana:
Should we update the RL parameters so that they are also namespaced for consistency even though there is no failing collisions there?

albcui:
> Should we update the RL parameters so that they are also namespaced for consistency even though there is no failing collisions there?

Yea good point, RL should get the same treatment.

packages/nmp_customization_common/src/nmp/customization_common/schema.py
```diff
@@ -0,0 +1,59 @@
...
+
+# ``type(BaseModel)`` is pydantic's ``ModelMetaclass``; deriving from it this way
+# avoids importing pydantic internals.
+_ModelMeta = type(BaseModel)
+
+
+class _NamespacedMeta(_ModelMeta):
+    def __new__(mcs, name, bases, ns, **kw):
+        prefix = ns.get("__schema_namespace__") or next(
+            (p for b in bases if (p := getattr(b, "__schema_namespace__", None))), None
```
mckornfield, line 40: oh boy lol

plugins/nemo-customizer/pyproject.toml
```diff
@@ -33,6 +33,10 @@ build-backend = "hatchling.build"
 packages = ["src/nemo_customizer"]
 
 [tool.nemo.openapi]
+# nemo-customizer merges every customization backend (automodel, unsloth, …)
+# into one /apis/customization app, so enforce that no two backends emit the
+# same OpenAPI schema name with differing content (see NamespacedModel).
+strict_schema_collisions = true
```
mckornfield, line 39: I do wonder if this should be the default, or if we should do a follow up ticket if there are lots of issues with making this the default?

mckornfield (approved):
personally I'd be for simplifying and just making exploding the default, but maybe that's a follow up?

soluwalana (approved):

plugins/nemo-customizer/pyproject.toml
albcui: Yup, there are existing collisions in the services directory (Guardrails and Models I think) that need to be resolved first, before we can flip this to default. I have this as a follow-up.

packages/nmp_customization_common/src/nmp/customization_common/schema.py
albcui: yup thanks claude

PR #738, opened 2026-07-16
test(e2e): run agent container-deploy e2e on kubernetes (kind) too

benmccown:
Extends the docker-mode agent deployment e2e to also run on the kind (Kubernetes) CI job, proving the container-mode chain (gateway → agent container → Inference Gateway → mock provider → response) on the k8s deployments backend.

## What

- **Shared core:** extracted the backend-agnostic deploy/invoke/assert flow into `e2e/agents_deploy_helpers.py` (mock-backed agent config, deploy → wait running → assert container-mode endpoint shape → invoke → assert mocked round-trip → cleanup). A `reap_backend_resources` hook lets a backend reap leaked resources (docker container) during teardown.
- **Docker module** (`test_nemo_agents_docker.py`) refactored to call the shared core. Behavior and markers (`subprocess_only`, `needs_nmp_api_image`, Linux-only) unchanged.
- **New k8s module** (`test_nemo_agents_k8s.py`): `deployment_mode="k8s"`, marked `container_only` (runs only against an external cluster — the kind e2e job) + `needs_nmp_api_image`.
- **Kind Helm values** (`e2e/k8s/values/kind.yaml`): wire a nemo-deployments `k8s` executor. The `nmp-core` controller Role already grants the needed deployments/services/configmaps/pods permissions; `default_namespace` is the release namespace so the agent lands beside the platform and the injected in-cluster IGW URL is reachable.
- **CI** (`kind-cpu-e2e`): export `NMP_E2E_IMAGE_REGISTRY`/`NMP_E2E_IMAGE_TAG` so the k8s test composes the (node-pre-pulled) `nmp-api` image ref.

## Why not one dual-marked test function

The docker and k8s variants have mutually-exclusive markers (`subprocess_only` vs `container_only`) and different platform sources (self-driven subprocess platform + docker executor vs external Helm cluster) and networking (docker-bridge base_url juggling vs in-cluster DNS). So they are two thin wrappers over one shared core rather than one function with two markers.

## Validation

- Local: both modules collect; k8s skips without `NMP_BASE_URL` ("requires NMP_BASE_URL"), docker skips on non-Linux; ruff/ty/pre-commit clean.
- `helm template` with `kind.yaml` renders the expected `agents.deployments` + `deployments.executors[k8s]` config into the platform ConfigMap.
- Full k8s path runs in the `Kind CPU e2e` CI job on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes mode agent end-to-end test deploying a Deployment + Service and invoking via the gateway.
  * Introduced shared container-mode agent deploy/invoke/teardown helpers for e2e runs.
  * Wired CPU-only e2e Kubernetes executor configuration and CI-provided agent image registry/tag.

* **Tests**
  * Docker and Kubernetes agent tests now validate mocked gateway round-trips using the shared flow.
  * Added unit coverage for effective Kubernetes namespace resolution.

* **Bug Fixes**
  * Improved Kubernetes deployment/job operations to consistently use the effective target namespace.

### DIFF
--- a/e2e/agents_deploy_helpers.py
+++ b/e2e/agents_deploy_helpers.py
@@ -1,0 +1,255 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared helpers for container-mode agent deployment e2e tests.
+
+Both the Docker (``test_nemo_agents_docker.py``) and Kubernetes
+(``test_nemo_agents_k8s.py``) modules deploy a real agent **container** through
+the nemo-deployments plugin and invoke it through the agents gateway. The end-to
+-end chain they prove is identical apart from the backend::
+
+    sdk.agents.invoke (gateway proxy, container-mode endpoint resolution)
+      -> agent container (nat start fastapi) on docker | kubernetes
+      -> Inference Gateway /openai (base_url injected at deploy time)
+      -> mock provider short-circuit (no real upstream / no API key)
+      -> response back through the gateway
+
+This module holds the backend-agnostic core: the mock-provider-backed agent
+config, the create -> wait-running -> assert-container-shape -> invoke -> assert
+flow, and cleanup. The per-backend modules own only what genuinely differs
+(pytest markers, how the deployment image ref is resolved, and best-effort
+container cleanup).
+"""
+
+import time
+import uuid
+from collections.abc import Callable
+from typing import Any
+
+import httpx
+import pytest
+from nemo_platform import NeMoPlatform
+from nmp.testing import MockProviderResponse, add_mock_provider
+
+# The mocked completion the deployed agent must round-trip back to the caller.
+TEST_AGENT_RESPONSE = "The answer to your question is 42."
+
+
+def unique_name(prefix: str) -> str:
+    return f"e2e-{prefix}-{uuid.uuid4().hex[:8]}"
+
+
+def _chat_completion_response(content: str, model: str) -> dict[str, Any]:
+    return {
+        "id": "chatcmpl-agents-container-e2e",
+        "object": "chat.completion",
+        "model": model,
+        "choices": [
+            {
+                "message": {"role": "assistant", "content": content},
+                "finish_reason": "stop",
+                "index": 0,
+            }
+        ],
+        "usage": {"prompt_tokens": 1, "completion_tokens": 1, "total_tokens": 2},
+    }
+
+
+def _mock_backed_agent_config(model_name: str) -> dict[str, Any]:
+    """A deterministic single-LLM workflow pointed at the mock model.
+
+    ``base_url``/``api_key`` are intentionally omitted so the deployment injects
+    the Inference Gateway URL (and the mock provider short-circuits the call).
+    """
+    return {
+        "llms": {
+            "main": {
+                "_type": "openai",
+                "model_name": model_name,
+            }
+        },
+        "workflow": {
+            "_type": "chat_completion",
+            "llm_name": "main",
+        },
+    }
+
+
+def _page_data(page: Any) -> list[dict[str, Any]]:
+    if isinstance(page, dict):
+        data = page.get("data", [])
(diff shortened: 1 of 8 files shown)


e2e/agents_deploy_helpers.py
```diff
@@ -0,0 +1,255 @@
...
+from collections.abc import Callable
+from typing import Any
+
+import httpx
+import pytest
+from nemo_platform import NeMoPlatform
+from nmp.testing import MockProviderResponse, add_mock_provider
+
+# The mocked completion the deployed agent must round-trip back to the caller.
+TEST_AGENT_RESPONSE = "The answer to your question is 42."
```
mckornfield, line 35: lol

e2e/k8s/values/kind.yaml
```diff
@@ -33,3 +33,23 @@ platformConfig:
       backends:
         nim_operator:
           enabled: false
+  # Wire a nemo-deployments k8s executor so deployment_mode=k8s agent
+  # deployments (test_nemo_agents_k8s.py) run as real Deployment+Service
+  # workloads. The nmp-core controller ServiceAccount Role already grants
+  # apps/deployments, services, configmaps, and pods (see
+  # k8s/helm/templates/core/controller-role.yaml). We omit default_namespace so
```
mckornfield: wonder if this comment overexplains, especially for an omitted field

e2e/test_nemo_agents_k8s.py
```diff
@@ -0,0 +1,102 @@
...
+    mocked completion round-trips from inside the pod back to the caller.
+    """
+    run_container_agent_deploy_and_invoke(
+        sdk,
+        workspace=workspace,
+        deployment_mode="k8s",
+        image=agent_deployment_image,
+        # Pod scheduling + (node-local) image resolution can take longer than the
+        # docker path's local container start.
+        running_timeout_seconds=420,
```
mckornfield, line 101: is this time one that can flake I wonder

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/config.py
```diff
@@ -19,11 +29,16 @@ class K8sExecutorConfig(BaseModel):
...
-        default="default",
+    default_namespace: str | None = Field(
+        default=None,
         min_length=1,
         max_length=63,
-        description="Namespace for resources when entity backend_config.k8s.namespace is unset.",
+        description=(
+            "Namespace for resources when entity backend_config.k8s.namespace is unset. When this is "
+            "itself unset, the executor defaults to its own pod namespace (POD_NAMESPACE, injected via "
+            "the downward API), so a deployed platform places agents beside itself; if that is also "
```
mckornfield, line 39: downward api?

mckornfield (approved):
minor wording stuff

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/config.py
benmccown: TWBL (today we both learned) about [the kubernetes downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) 😆

e2e/k8s/values/kind.yaml
benmccown: Trimmed it down!

e2e/test_nemo_agents_k8s.py
benmccown: Do you mean like "could things take longer than this timeout and could it cause successful runs to fail"? Or something else?

mckornfield: Yeah that's all, ie would something be slow enough to cause this to fail but pass on retry

benmccown: Got it. I think the test was taking ~30 seconds when I measured locally, so I _think_ 7 mins should give us enough wiggle room. Probably more likely reason to flake is transient issues with docker socket availability again or something like that, which would happen irrespective of max duration timeout.

svvarom (approved):

PR #741, opened 2026-07-16
fix(evaluator): normalize Docker Codex evidence permissions

ngoncharenko:
## Summary

- keep the Dockerized Codex CLI running as root so its auth mount and tool-install behavior continue to work
- return every workspace and evidence tree readable and writable by the invoking host user without granting access to unrelated local users
- support rootful Docker, rootless Docker, and Docker Desktop ownership mappings without assuming container UID/GID values equal host values
- preserve the original Codex failure status while treating an unsafe or unreadable artifact tree as a trial-production failure
- synchronize the generated Python SDK mirror
- added a simple demo [example.py](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/packages/nemo_evaluator_sdk/examples/codex_docker/example.py) about how to evaluate codex in docker container mode

## Problem reported in AALGO-346

`CodexDockerCliAgentRuntime` runs `node:22-alpine` as root and bind-mounts the host workspace and evidence directories. Agent processes could create root-owned, mode-restricted files such as `__pycache__` entries. After Docker returned, the host evaluator, filesystem verifiers, CI artifact collection, and Allure result persistence could fail with `Permission denied` even though Codex completed successfully.

## Additional problems found during review

- **The first readability fallback overexposed evidence.** Recursive `chmod -R u+rwX,go+rX` changed restrictive files and directories to group/world-readable and traversable modes. On a shared POSIX host, another UID could read prompts, task inputs, seeded source, logs, generated files, and final output.
- **Raw host UID/GID values are not portable inside containers.** `chown <os.getuid()>:<os.getgid()>` can fail or map to subordinate host IDs under Docker Desktop and rootless user namespaces. Making that `chown` authoritative caused otherwise successful tasks to fail on Docker Desktop.
- **Some host-created evidence bypassed container cleanup.** Task metadata and prompts existed before Docker ran; stdout, stderr, fallback output, and errors were written after it returned. Ordinary write-then-`chmod` behavior left an umask-dependent exposure window and did not cover all failure paths.
- **Permissive or reused output directories weakened the boundary.** A private ancestor was not enforced before task inputs and workspace seeds were written.
- **Cleanup failures could mask or compound Codex failures.** Failed executions still contain sensitive evidence, and an inaccessible `error.json` path could raise while constructing a failed trial and abort sibling tasks.
- **The generated SDK mirror could retain the old permission behavior.** The source and vendored runtime must change together.

## Fix

- Create or repair the unmounted run root, task evidence directory, and workspace as host-owned `0700` directories. Reject a leaf symlink, non-directory, or directory owned by a different host UID before writing sensitive data.
- Publish task, prompt, stdout, stderr, fallback-output, and error files atomically through same-directory `0600` temporary files and `os.replace`.
- Capture the evidence mount's UID:GID as observed inside the container before Codex starts. Use that namespace-correct identity for best-effort recursive ownership repair.
- Require `chmod -R u+rwX,go-rwx /workspace /evidence`. This restores owner access, preserves owner execute where appropriate, and removes all group/other access.
- Validate the complete tree on the host with no-follow traversal after every completed Docker wrapper outcome:
  - every non-symlink artifact must be owned by the invoking host UID
  - directories must provide owner `rwx`
  - files must provide owner `rw`, with owner execute allowed
  - no group or other permission bits may remain
- Keep a Codex nonzero status authoritative. If permission validation also fails, retain the Codex error and add `permission_cleanup_error`; if Codex succeeds but the postcondition fails, fail the trial.
- Make `error.json` persistence best-effort so an inaccessible evidence directory still produces a task-local failed trial instead of aborting concurrent work.
- Set `PYTHONDONTWRITEBYTECODE=1` as defense in depth and regenerate the Python SDK mirror.

## Impact

Successful Docker Codex trials now return private, host-consumable workspace and evidence trees across the tested Docker Desktop ownership model. Failed trials retain their original Codex status while exposing cleanup failures separately. Verifiers and recursive persistence can consume the artifacts without making them readable to unrelated host users.

## Validation

- `uv run --frozen pytest packages/nemo_evaluator_sdk/tests/agent_eval/test_codex_runtime.py packages/nemo_evaluator_sdk/tests/agent_eval/test_codex_runtime_live.py -q -rs` — **22 passed**
  - two successful tasks run concurrently
  - restrictive container-created files are normalized
  - the real filesystem verifier and symlink-preserving recursive copy succeed
  - escaping symlinks are not dereferenced by the verifier
  - fake Codex status `23` remains authoritative
  - unsafe ownership/mode and failed error-artifact paths are covered
- original cross-UID reproduction after the fix — directory `0700`, file `0600`, distinct UID in the same GID denied
- scoped Ruff lint and format checks — passed
- scoped `ty` checks — passed
- `make vendor` — passed twice; only the expected SDK mirror changed
- `git diff --check` — passed

Linear: [AALGO-346](https://linear.app/nvidia/issue/AALGO-346/agent-evalcodex-docker-docker-codex-runtime-creates-unreadable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced host-private, owner-only evidence/workspace handling for Codex-backed runs, with safer persistence of prompt/output artifacts.
  * Improved trial failure reporting by separating Codex execution errors from artifact persistence and permission cleanup issues, while preserving the original failure details.
  * Made final output retrieval more resilient using `final_output.txt` when present, with clearer failure outcomes when reading/republishing fails.
  * Disabled Python bytecode generation during Docker-based Codex runs.

* **Tests**
  * Expanded local and live Docker regression coverage for strict ownership/permission enforcement, including `error.json`, final output, and negative cases (e.g., symlink/special-file attempts).

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
@@ -3,6 +3,8 @@
 
 """Codex-backed agent-eval runtimes."""
 
+# ruff: noqa: I001, T201 - the vendored SDK mirror uses different import-order and print settings.
+
 from __future__ import annotations
 
 import asyncio
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
@@ -289,11 +289,26 @@ class CodexDockerCliAgentRuntime(CodexCliAgentRuntime):
         if self._model is not None:
             inner_command.extend(["--model", self._model])
         inner_command.append("-")
+        # Codex intentionally runs as root: the container mounts its auth under /root and coding tasks
+        # may need to install tools. Repair the bind-mounted trees before Docker returns so the host can
+        # score and persist every artifact the agent created. Keep the Codex failure authoritative; only
+        # surface the chmod status when Codex itself succeeded.
+        shell_command = (
+            f"{shlex.join(inner_command)}; "
+            "codex_status=$?; "
+            f"chown -R {os.getuid()}:{os.getgid()} /workspace /evidence 2>/dev/null || true; "
+            "chmod -R u+rwX,go+rX /workspace /evidence; "
+            "permissions_status=$?; "
+            'if [ "$codex_status" -ne 0 ]; then exit "$codex_status"; fi; '
+            'exit "$permissions_status"'
+        )
         return [
             self._docker_bin,
             "run",
             "--rm",
             "-i",
+            "-e",
+            "PYTHONDONTWRITEBYTECODE=1",
             "-v",
             f"{self._auth_path.resolve()}:/root/.codex/auth.json:ro",
             "-v",
@@ -303,7 +318,7 @@ class CodexDockerCliAgentRuntime(CodexCliAgentRuntime):
             self._image,
             "sh",
             "-lc",
-            shlex.join(inner_command),
+            shell_command,
         ]
 
 
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
@@ -13,7 +13,9 @@ import json
 import os
 import shlex
 import shutil
+import stat
 import subprocess
+import tempfile
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from enum import StrEnum
 from pathlib import Path
@@ -94,8 +96,16 @@ class CodexCliAgentRuntime:
     async def _run_task(self, index: int, task: AgentEvalTask, config: AgentEvalRunConfig) -> AgentEvalTrial:
         evidence_dir = self._evidence_dir(index, task, config)
         workspace_dir = evidence_dir / "workspace"
-        evidence_dir.mkdir(parents=True, exist_ok=True)
-        workspace_dir.mkdir(parents=True, exist_ok=True)
+
+        try:
+            # The task directory is mounted into Docker, but its private parent is not. Keeping that
+            # parent host-owned and 0700 preserves the local confidentiality boundary even when a
+            # container is interrupted before its recursive cleanup completes.
+            _ensure_private_directory(evidence_dir.parent)
+            _ensure_private_directory(evidence_dir)
+            _ensure_private_directory(workspace_dir)
+        except Exception as exc:
+            return _failed_codex_trial(task, evidence_dir, exc, runtime_name=self._runtime_name)
 
         prompt_path = evidence_dir / "prompt.txt"
         task_path = evidence_dir / "task.json"
@@ -107,7 +117,10 @@ class CodexCliAgentRuntime:
         # this evidence dir into the sandbox (danger-full-access), so serializing `intent` (desired
         # behavior) or `reference` (held-out ground truth) here would let the agent read them back out
         # of `/evidence/task.json` — the same reward-hacking leak the intent-free prompt closes.
-        task_path.write_text(task.model_dump_json(indent=2, exclude={"intent", "reference"}), encoding="utf-8")
+        try:
+            _write_private_text(task_path, task.model_dump_json(indent=2, exclude={"intent", "reference"}))
+        except Exception as exc:
+            return _failed_codex_trial(task, evidence_dir, exc, runtime_name=self._runtime_name)
 
         command = self._command(workspace_dir=workspace_dir, final_output_path=final_output_path)
         process: Any | None = None
@@ -120,13 +133,15 @@ class CodexCliAgentRuntime:
             # Build the prompt after seeding and inside the guarded block: an instruction-less task
             # raises here, failing just this task instead of aborting the run (and seeding wins if both).
             prompt = self._prompt_builder(task)
-            prompt_path.write_text(prompt, encoding="utf-8")
+            _write_private_text(prompt_path, prompt)
             process = await self._process_factory(
                 *command,
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
             )
+            if process is None:
+                raise RuntimeError("process factory failed to create a process")
             stdout, stderr = await asyncio.wait_for(
                 process.communicate(prompt.encode("utf-8")),
                 timeout=self._timeout_s,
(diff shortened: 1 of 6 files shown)


ngoncharenko:
@coderabbitai review the PR

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
```diff
@@ -289,11 +291,26 @@ def _command(self, *, workspace_dir: Path, final_output_path: Path) -> list[str]
...
         inner_command.append("-")
+        # Codex intentionally runs as root: the container mounts its auth under /root and coding tasks
+        # may need to install tools. Repair the bind-mounted trees before Docker returns so the host can
+        # score and persist every artifact the agent created. Keep the Codex failure authoritative; only
+        # surface the chmod status when Codex itself succeeded.
+        shell_command = (
+            f"{shlex.join(inner_command)}; "
+            "codex_status=$?; "
+            f"chown -R {os.getuid()}:{os.getgid()} /workspace /evidence 2>/dev/null || true; "
+            "chmod -R u+rwX,go+rX /workspace /evidence; "
```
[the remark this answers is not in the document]
ngoncharenko: validate the fix

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
```diff
@@ -398,30 +470,122 @@ def _failed_codex_trial(
...
+    descriptor, temporary_name = tempfile.mkstemp(prefix=f".{path.name}.", suffix=".tmp", dir=path.parent)
+    temporary_path = Path(temporary_name)
+    try:
+        os.fchmod(descriptor, 0o600)
+        with os.fdopen(descriptor, "w", encoding="utf-8") as temporary_file:
+            temporary_file.write(content)
+        os.replace(temporary_path, path)
+    finally:
+        with contextlib.suppress(OSError):
+            os.close(descriptor)
```
arpitsardhana: nit: Can it lead to double close in rare scenario where os.replace fails since

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
```diff
@@ -398,30 +470,122 @@ def _failed_codex_trial(
...
+
+    _write_private_text(path, output_text)
+    return output_text
+
+
+def _validate_private_tree(root: Path) -> None:
+    """Require a host-owned, owner-only tree without following agent-created symlinks."""
+    expected_uid = os.getuid()
+    pending = [root]
```
arpitsardhana, line 573: nit: Is it better idea to work with fd rather than paths? since paths are more vulnerable to race conditions
Open a root directory as readonly and strictly not allowing symlinks. os.scandir can take fd as arguments

arpitsardhana (approved):
Minor nits to consider

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
ngoncharenko: Resolved in 755f184. `_write_private_text` now transfers descriptor ownership to `os.fdopen` and immediately marks the raw descriptor as unavailable; cleanup closes it only when `os.fdopen` itself fails. Regression tests cover both an `os.replace` failure (no second raw close) and an `os.fdopen` failure (the original descriptor is closed).

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/codex/runtime.py
arpitsardhana: something like

 ```python
def _validate_private_tree(root: Path) -> None:
    """Require a host-owned, owner-only tree without following agent-created symlinks."""
    expected_uid = os.getuid()
    
    # Open the root directory safely without following symlinks
    root_fd = os.open(root, os.O_RDONLY | os.O_DIRECTORY | os.O_NOFOLLOW)
    pending_fds = [root_fd]
    
    try:
        while pending_fds:
            fd = pending_fds.pop()
            try:
                fd_stat = os.fstat(fd)
                if fd_stat.st_uid != expected_uid:
                    raise PermissionError("Artifact is not owned by the invoking host user.")
                
                mode = stat.S_IMODE(fd_stat.st_mode)
                if mode & 0o077:
                    raise PermissionError(f"Artifact grants group or other access ({mode:o}).")
                
                if stat.S_ISDIR(fd_stat.st_mode):
                    if mode & 0o700 != 0o700:
                        raise PermissionError(f"Directory lacks critical owner permissions ({mode:o}).")
                    
                    # Modern Python allows passing a file descriptor directly to os.scandir!
                    # This completely neutralizes path-based TOCTOU symlink swaps.
                    with os.scandir(fd) as entries:
                        for entry in entries:
                            # Use entry.stat(follow_symlinks=False) to safely check for symlinks
                            entry_stat = entry.stat(follow_symlinks=False)
                            if stat.S_ISLNK(entry_stat.st_mode):
                                continue
                            
                            if stat.S_ISDIR(entry_stat.st_mode):
                                # Open subdirectory relative to the parent or via full path without following leaf symlinks
                                sub_fd = os.open(entry.path, os.O_RDONLY | os.O_DIRECTORY | os.O_NOFOLLOW)
                                pending_fds.append(sub_fd)
                            elif stat.S_ISREG(entry_stat.st_mode):
                                if stat.S_IMODE(entry_stat.st_mode) & 0o600 != 0o600:
                                    raise PermissionError(f"File is not owner-readable/writable: {entry.path}")
                            else:
                                raise PermissionError(f"Unsupported artifact type at {entry.path}")
            finally:
                os.close(fd)
    except Exception:
        # Clean up any remaining open file descriptors on error
        for open_fd in pending_fds:
            with contextlib.suppress(OSError):
                os.close(open_fd)
        raise```

Caveat:
- Generated example with gemini
- It is a nit, make your judgement if complexity is worth it

ngoncharenko: Good point. Converting only the validator to fd-based traversal would be incomplete because final-output handling and CandidateEvidence references remain path-based.
Validation runs after `docker run --rm` completes, and the evidence parent is host-owned with mode 0700, so the remaining race requires another process running as the same host UID. We should handle this as a follow-up in my view: anchor the entire post-run flow to an evidence-root fd using openat/dir_fd, O_NOFOLLOW, and fstat, including traversal, final-output reads, and artifact publication

ngoncharenko: Thanks for the suggestion. I agree the fd-anchored direction is useful defense in depth, but the generated sketch needs corrected `dir_fd` child opens, `fstat()` validation for every opened object, and preservation of the existing ownership and group/other permission checks. I created [AALGO-356](https://linear.app/nvidia/issue/AALGO-356/agent-evalcodex-docker-harden-evidence-traversal-against-path-toctou) to track the complete follow-up. The story links this thread and the proposed sketch, and includes the threat boundary, implementation details, acceptance criteria, deterministic race tests, and validation commands. Keeping it as follow-up work because the current validation runs after `docker run --rm` and beneath a host-owned `0700` parent.

PR #742, opened 2026-07-16
fix(evaluator): emit evaluation_context in publish-to-intake

shanaiabuggy:
## Summary

The evaluator's publish-to-intake path (added in #456) still emits the deprecated `experiment_context` on ATIF ingest. As part of the **Experiment → Evaluation** rename, intake deprecated `experiment_context` in favor of `evaluation_context`. This migrates the boundary mapping to the canonical shape.

Per `mapping.py`'s own design note ("a later rename is a one-file change"), the wire migration lives entirely in that module:

- `ExperimentContextParam` → `EvaluationContextParam`
- `run_task_to_experiment_context` → `run_task_to_evaluation_context`, now emitting `{"evaluation_id": …, "test_case_id": …}` under the `"evaluation_context"` ATIF body key
- tests updated to match (unit `test_mapping.py`/`test_publish.py` + the integration trace read)

### Left for the evaluator team
The public `publish_to_intake(experiment_id=…)` parameter and `PublishReport.experiment_id` field are **intentionally unchanged** — renaming that external surface to `evaluation_id` is your call. The value already flows through to `evaluation_context.evaluation_id`, so functionally this is complete.

### Not urgent / non-breaking
`experiment_context` still works as a deprecated alias in intake (`evaluation_context` takes precedence when both are sent), so publish isn't broken today — this is a forward migration before the alias is removed. No SDK version bump needed: the plugin uses the in-repo workspace SDK, which already exposes `EvaluationContextParam`.

## Validation
- In-repo SDK confirmed to export `EvaluationContextParam` (TypedDict: `evaluation_id`, `test_case_id`).
- `mapping.py` compiles; no stale `experiment_context`/`ExperimentContextParam` references remain in the changed files.
- Unit + integration test assertions updated in lockstep. (Full suite runs in CI — the plugin needs the whole workspace built.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the evaluation data sent to Intake to use evaluation context.
  - Evaluation records now include the evaluation ID and test case ID for clearer traceability.
  - Updated published trajectory and score metadata to consistently reflect the new evaluation context.
  - Improved integration coverage to verify evaluation context is preserved end-to-end through the publishing flow.

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
@@ -37,7 +37,7 @@ from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEval
 from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial
 from nemo_platform.types.intake.evaluator_result_create_params import EvaluatorResultCreateParams
 from nemo_platform.types.intake.evaluator_result_data_type import EvaluatorResultDataType
-from nemo_platform.types.intake.experiment_context_param import ExperimentContextParam
+from nemo_platform.types.intake.evaluation_context_param import EvaluationContextParam
 from nemo_platform.types.intake.ingest.atif_agent_param import AtifAgentParam
 from nemo_platform.types.intake.ingest.atif_create_params import AtifCreateParams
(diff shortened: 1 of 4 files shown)


arpitsardhana (approved):

SandyChapman (approved):
Thanks for this @shanaiabuggy ! I'll follow up with the `experiment_id` -> `evaluation_id` swap once this lands.

PR #753, opened 2026-07-17
feat(experiments): test-case-weighted rollups (aggregate per test case before pooling)

shanaiabuggy:
Screenshot 2026-07-17 at 2 06 18 PM

## What

Two-level ("test-case-weighted") rollups for the Experiments leaderboard — [ASE-561](https://linear.app/nvidia/issue/ASE-561).

When a test case is run k times, metrics are now rolled up **per test case first** (pass@1 = mean over its attempts), then aggregated **across test cases** — so a heavily-retried test case counts once instead of dominating the pool. This also corrects medians/percentiles, which were previously computed over the wrong distribution (all attempts vs. per-test-case values).

## Changes

**Backend** (`evaluation_rollup_repository.py`)
- Group key = `test_case_id` when set, else `session_id` — so unlabeled runs pool per-attempt exactly like the old flat rollup (no empty-string bucket collapse), and there's no hard requirement on `test_case_id`.
- Scores: pass@1 (mean over a test case's attempts) → distribution across test cases.
- Cost/latency: **avg per attempt** within a test case (so the number doesn't scale with k) → distribution across test cases; attempts missing a value are excluded, not counted as zero.
- New **`test_case_count`** rollup field (distinct test cases) on the evaluation response, sortable.

**Frontend** (Studio)
- Leaderboard **"Run Count" → "Total test cases"** (`test_case_count`), sortable.
- Avg-cell tooltip now states the calculation transparently, e.g. `Mean cost over 2 test cases — each test case averaged over its attempts.`

**SDK** regenerated for the new `test_case_count` field.

## Testing
- Unit + **integration (real ClickHouse)**, including an unbalanced-k case that proves test-case-weighting (score mean **0.5**, not the pooled **0.75**; cost avg-per-attempt) and a missing-`test_case_id` degrade case (pools per attempt).

## Note
`run_count` (total attempts) stays on the API and still feeds the tooltip's context; only the leaderboard column moved to test cases.

## Summary by CodeRabbit

* **New Features**
  * Added `test_case_count` to evaluation and rollup responses, exposed as a sortable “Total test cases” metric in experiment group tables.
  * Updated API sorting to support ordering by total test cases.

* **Bug Fixes**
  * Updated rollup aggregation to compute statistics using test-case-level weighting (not raw attempts).
  * Evaluations with no `test_case_id` no longer contribute to test-case-weighted metrics (`test_case_count` returns 0).

* **Tests**
  * Extended integration and repository tests to validate hydration, sorting support, and the new rollup semantics.

### DIFF
--- a/services/intake/src/nmp/intake/spans/evaluation_rollup_repository.py
+++ b/services/intake/src/nmp/intake/spans/evaluation_rollup_repository.py
@@ -29,6 +29,7 @@ class ScoreRollup:
 class EvaluationRollup:
     evaluation_id: str
     run_count: int = 0
+    test_case_count: int = 0
     model_names: list[str] = field(default_factory=list)
     agent_names: list[str] = field(default_factory=list)
     agent_versions: list[str] = field(default_factory=list)
@@ -62,6 +63,7 @@ class EvaluationRollupRepository:
             )
         ):
             rollups[row["evaluation_id"]].run_count = int(row["run_count"])
+            rollups[row["evaluation_id"]].test_case_count = int(row["test_case_count"])
 
         for row in result_rows(
             await self._client.query(
@@ -116,7 +118,7 @@ def _evaluation_id_parameters(evaluation_ids: list[str]) -> tuple[str, dict[str,
 
 def _scoped_sessions_sql(trace_index_table: str, evaluation_names_sql: str) -> str:
     return f"""
-        SELECT workspace, evaluation_id, session_id, latency_ms
+        SELECT workspace, evaluation_id, session_id, test_case_id, latency_ms
         FROM {trace_index_table} FINAL
         WHERE workspace = %(workspace)s
             AND is_deleted = 0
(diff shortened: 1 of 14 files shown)


services/intake/src/nmp/intake/spans/evaluation_rollup_repository.py
```diff
@@ -126,14 +128,25 @@ def _scoped_sessions_sql(trace_index_table: str, evaluation_names_sql: str) -> s
...
 
 
+def _test_case_key_sql(sessions_alias: str = "") -> str:
+    """Per-test-case grouping key for the two-level rollup: the test case (``test_case_id``) when set,
+    else the individual attempt (``session_id``) so unlabeled runs pool per-attempt exactly like a flat
+    rollup. Namespaced so a ``test_case_id`` can never collide with a ``session_id``. Counting the
+    distinct values of this key over an evaluation's sessions gives ``test_case_count``.
+    """
+    prefix = f"{sessions_alias}." if sessions_alias else ""
+    return f"if({prefix}test_case_id != '', concat('t:', {prefix}test_case_id), concat('s:', {prefix}session_id))"
```
shanaiabuggy: Somewhere else in the parent query should be dropping blank test_case_ids instead of handling the else here

BrianNewsom (approved):

PR #755, opened 2026-07-17
fix(studio): Add a model dropdown for Base Model filtering on the Virtual Model Grid

htolentino-nvidia:
https://linear.app/nvidia/issue/ASTD-307/virtual-models-default-model-filter-is-a-free-text-input-with-no

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated “Default Model” filter with single-select search behavior.

* **Bug Fixes**
  * Improved filtering accuracy by matching the selected default model exactly.
  * Ensured model-related column displays stay synchronized with the current workspace.


steramae-nvidia (approved):
why the change from $like to $eq?

htolentino-nvidia:
Good question @steramae-nvidia, I do think when the user selects a specific model on the dropdown they wouldn't expect to see models that would be also matched by $like, since they clicked llama-8b for example (and not any other llama variant on the dropdown) I wouldn't want to display base models like llama-8b-custom-ft. Now, this was requested by QA, I'm not super sure the dropdown for this type of filtering is the more useful pattern here. The user can't really filter by all nemotron models if they wanted so, but at least here we're matching the existing behaviour on in the custom models page.

PR #756, opened 2026-07-17
fix(customizer): resolve metadata drift in model upload

soluwalana:
## Summary by CodeRabbit

- **New Features**
  - Upload metadata now preserves model tool-calling settings, including chat templates and parser configuration.
  - Metadata is derived consistently from the model configuration across supported job types.

- **Bug Fixes**
  - Improved fileset metadata validation and handling when model configuration is missing or incomplete.
  - Prevented incorrectly structured tool-calling metadata from being accepted.

### DIFF
--- a/packages/nmp_customization_common/tests/tasks/test_file_io_metadata.py
+++ b/packages/nmp_customization_common/tests/tasks/test_file_io_metadata.py
@@ -5,17 +5,43 @@
 
 from types import SimpleNamespace
 
+import pytest
+from nemo_platform_plugin.files.metadata import (
+    FilesetMetadata,
+    ModelMetadataContent,
+    ToolCallingMetadataContent,
+)
 from nmp.customization_common.tasks.file_io_metadata import (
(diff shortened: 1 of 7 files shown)


albcui (approved):

services/automodel/src/nmp/automodel/app/jobs/compiler.py
albcui: stale comment

```suggestion
    # separately via fileset metadata (see build_output_fileset_metadata_from_model_entity),
```

packages/nmp_customization_common/tests/tasks/test_file_io_metadata.py
albcui, line 1: Can we harden this test to cover the original root cause of the bug -- we should validate against the `FilesetMetadata` schema itself. Here we just asserting on dicts.

soluwalana: Added additional tests

PR #759, opened 2026-07-17
feat(experiments): Allow an evaluation to belong to multiple experiment groups

shanaiabuggy:
https://github.com/user-attachments/assets/8d605ede-5a44-4086-a943-bc0d6e78b3cf

## Summary

Lets a single evaluation belong to **multiple** experiment groups instead of exactly one, and adds the UI to curate evaluations across groups. Membership becomes many-to-many while the old single-group field keeps working as a deprecated alias. (ASE-559)

## Backend

**Entities**
- Add a `$contains` array-membership filter operator (SQL + in-memory parity) so lists can match rows whose array field contains a value.
- Fix `eq(field, None)` to also match an **absent** JSON key — on PostgreSQL a missing key extracts to SQL `NULL`, which the previous `== "null"` check missed (added an `IS NULL` branch).

**Intake**
- `EvaluationRequest`/`Response` gain `experiment_ids: list[str]` (≥1) as the canonical membership field. `experiment_group_id` stays as a deprecated request alias (coalesced into `experiment_ids`) and response alias (first of `experiment_ids`).
- The `experiment_group_id` list filter is rewritten to a membership match spanning `experiment_ids` (new rows) and the legacy scalar (old rows), so listing a group returns every member.
- Deleting a group cascades: members drop that membership; a member whose *only* membership was that group is soft-deleted.
- **New `PATCH /evaluations/{name}`** — a partial update (only fields present in the request change), mirroring the models service's PATCH. The add-to-group flow sends the merged `experiment_ids`; any new group must exist and the set must be non-empty (the ≥1-group invariant). Rejects empty `experiment_ids` (400) and unknown group ids (400).

**Design note:** membership is a field on the evaluation, so we curate it via the resource's `PATCH` — additive and non-breaking, matching the house "PATCH = partial" convention — rather than adding dedicated add/remove endpoints. The full-body `PUT` is unchanged.

## Frontend

- New bulk **"Add to experiment group"** action on the evaluations list: select rows, add them to another group (via `PATCH` with the merged `experiment_ids`).
- Modal picks the target group from a dropdown, with an inline **"Create new group"** option (name / description / default sort) that creates the group and then adds the selected evaluations.
- Groups every selected evaluation already belongs to are excluded; adds are best-effort and surface a warning on partial failure.

## Generated artifacts / follow-ups

- Regenerated OpenAPI specs, web SDK, and static authz mappings.
- Python SDK (Stainless) regen is a separate key-gated step — run `make stainless` before merge.

## Summary

* **New Features**
  * Partial update API for evaluations (PATCH) with `experiment_ids` to replace group memberships.
  * Multi-group ownership for evaluations via `experiment_ids`.
  * Studio bulk **“Add to experiment group”** (add to existing or create new).
  * Array **“contains”** filtering support.
* **Improvements**
  * `experiment_group_id` remains as a deprecated alias; APIs now primarily use `experiment_ids`.
  * Group deletion and listing behavior now respects multi-group memberships.
* **Bug Fixes**
  * Reject empty `experiment_ids` and unknown group IDs.
  * Deleting a group won’t remove evaluations that belong to other groups.


openapi/ga/openapi.yaml
```diff
@@ -3552,6 +3552,104 @@ paths:
...
+        '404':
+          description: Evaluation not found
+        '409':
+          description: Cannot remove the evaluation from its only experiment
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
```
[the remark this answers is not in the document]
shanaiabuggy: /groups/{group_id} is new in this PR and has never been released to main, so there are no existing clients to break — oasdiff flags it only because an earlier commit in this same PR added /groups before it was renamed to /experiments. Keeping a deprecated alias would ship a dead path for an endpoint that never existed publicly, which is exactly what this rename is meant to avoid ahead of the upcoming ExperimentGroup → Experiment domain rename. Declining.

services/core/entities/src/nmp/core/entities/app/repository/sqlalchemy/filter.py
```diff
@@ -178,6 +178,26 @@ def nin(self, field: str, values: List[Any]) -> Any:
...
+        Intended for array-valued JSON fields (e.g. ``data.experiment_ids``); values are
+        assumed to be JSON scalars without embedded double quotes (entity ids qualify).
+        """
+        column, is_json = self._get_column(field)
+        if not is_json:
+            raise ValueError(f"$contains requires a JSON array field, got non-JSON field '{field}'")
+        needle = str(value)
+        for ch in ("\\", "%", "_"):
+            needle = needle.replace(ch, f"\\{ch}")
+        return cast(column, String).like(f'%"{needle}"%', escape="\\")
```
shanaiabuggy: hmm why do we have to do this?

web/packages/studio/src/components/dataViews/ExperimentGroupDataView/AddToGroupModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+              }}
+            />
+            <SelectContent className="w-(--radix-popper-anchor-width)">
+              <SelectListbox>
+                <SelectItem value={CREATE_NEW}>
+                  <Flex gap="density-sm" align="center">
+                    <Plus className="size-4" />
+                    Create new group
+                  </Flex>
+                </SelectItem>
```
nakolean, line 269: **Nitpick**: Looking at the existing 'workspace' dropdown which includes a 'create new workspace' the Create new option appears at the bottom of the list, so maybe we drop this down below {groups.map}?

web/packages/studio/src/components/dataViews/ExperimentGroupDataView/AddToGroupModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+      setSelectedGroupId('');
+      setDefaultSort(DEFAULT_SORT);
+      reset();
+    }
+  }, [open, reset]);
+
+  // Only fetch groups while the modal is open. A single large page covers any realistic group count.
+  const { data: groupsPage, isLoading } = useListExperimentGroups(
+    workspace,
+    { page_size: 1000 },
```
nakolean: **Suggestion**: We have a const for page_size of 1000. Could use that here as well `DEFAULT_LARGE_PAGE_SIZE` from `studio/src/constants/constants`

nakolean (approved):

services/core/entities/src/nmp/core/entities/app/repository/sqlalchemy/filter.py
shanaiabuggy: 1. The text-LIKE approach → Instead of writing engine-specific JSON queries, we dump the array to text (["g1","g2"]) and string-match it. That one expression works on both SQLite and Postgres identically. That's the whole reason we're not using a proper JSON operator.
  2. The escaping → this is not about supporting both engines. It's a side effect of using LIKE. LIKE treats `%` and `_` as wildcards, so we escape any `%` `_` `\` in the id — otherwise an id could accidentally act as a wildcard pattern. This would be needed even if we only supported one engine.

So: we string-match because we support both engines (SQLite and Postgres); we escape because string-matching uses LIKE.

cc: @BrianNewsom

services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -640,6 +646,80 @@ async def unpin_evaluation(
...
+@router.post(
+    "/v2/workspaces/{workspace}/evaluations/{name}/experiments/{experiment_id}",
+    response_model=EvaluationResponse,
+    tags=[EVALUATIONS_TAG],
+    responses={
+        400: {"description": "Experiment does not exist or is deleted"},
+        404: {"description": "Evaluation not found"},
+    },
+)
+async def add_evaluation_to_experiment(
```
shanaiabuggy: remove this, we can just use evaluation patch

services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -640,6 +646,80 @@ async def unpin_evaluation(
...
+@router.delete(
+    "/v2/workspaces/{workspace}/evaluations/{name}/experiments/{experiment_id}",
+    response_model=EvaluationResponse,
+    tags=[EVALUATIONS_TAG],
+    responses={
+        404: {"description": "Evaluation not found"},
+        409: {"description": "Cannot remove the evaluation from its only experiment"},
+    },
+)
+async def remove_evaluation_from_experiment(
```
shanaiabuggy: remove this, we can just use evaluation patch

PR #761, opened 2026-07-17
feat(nemo-agents): update fabric dep to use published wheel and fabric validation helpers

mmogallapalli:
## Summary

This PR implements[ AIRCORE-897](https://linear.app/nvidia/issue/AIRCORE-897/use-fabric-sdk-planning-and-diagnostics-for-translated-fabricconfig) by wiring the NeMo Agents Fabric config translation path into Fabric SDK planning and diagnostics.

The validation flow established here is:

```text
Platform agent config dict / AgentConfig
  -> translated FabricConfig
  -> Fabric plan
  -> Fabric doctor
```

This keeps the Platform-owned `nemo-agents-spec-v1` config as the persisted contract while using Fabric SDK APIs to validate the translated execution config before later register/deploy/invoke wiring.

## Changes

- Added Fabric SDK/runtime dependency wiring for `nemo-agents-plugin`.
- Removed the temporary lazy Fabric import workaround now that Fabric wheels are available.
- Updated the translator to use the typed Fabric SDK config models directly.
- Added `validate_platform_agent_config(...)` as the Platform-facing validation helper.
- Kept `validate_fabric_config(...)` as the lower-level FabricConfig validation primitive.
- Runs Fabric `plan()` off the event loop.
- Wraps Fabric `doctor()` with a validation timeout.
- Normalizes Platform config, translation, plan, doctor, timeout, and preflight failures into Fabric validation errors.
- Added focused unit coverage for translated config validation and Fabric preflight handling.

## Notes

- `nemo-fabric` is installed by default for SDK config translation, planning, and diagnostics.
- `nemo-fabric-runtime` is intentionally not installed by default because its native install path currently requires `nvidia-smi`, which is not available during service image builds.
- Relay is not added as a Fabric extra yet because the current workspace dependency graph still has a `nemo-relay` pin conflict.
- Harness adapter packages are still intentionally outside this ticket’s default dependency wiring.

## Out Of Scope

- CLI/API register wiring.
- Deploy semantics.
- Invoke/runtime lifecycle behavior.
- Durable runtime/session management.
- Harness adapter installation strategy.
- Stable Fabric version pinning.

Those are covered by follow-up AIRCORE-895 implementation tickets.

## Validation

Passed:

```bash
uv run --frozen pytest plugins/nemo-agents/tests/unit/test_agent_config.py plugins/nemo-agents/tests/unit/test_fabric_translator.py plugins/nemo-agents/tests/unit/test_fabric_validation.py
```

Result: `37 passed`

## Summary by CodeRabbit

- **New Features**
  - Added platform-level validation that translates an agent config into Fabric config and runs plan/doctor diagnostics, returning both translated config and validation results.

- **Bug Fixes**
  - Updated Fabric integration behavior to require the Fabric SDK up front (missing dependency fails immediately).
  - Improved validation behavior, including acceptance of doctor reports provided as dictionaries.

- **Chores**
  - Updated bundled dependency extras and constraints, refreshed third-party license records, and regenerated the main requirements lock output.
  - Refreshed unit tests to match the updated Fabric integration flow.


plugins/nemo-agents/src/nemo_agents_plugin/fabric/translator.py
```diff
@@ -5,9 +5,10 @@
 
 from __future__ import annotations
 
-import importlib
 from typing import Any
 
+# Fabric is a plugin dependency but may be absent from ty's lint environment.
```
mikeknep: why/how? Isn't the agents plugin a workspace member and effectively auto-installed like the others today?

plugins/nemo-agents/src/nemo_agents_plugin/fabric/validation.py
```diff
@@ -22,6 +29,16 @@ class FabricValidationResult:
     doctor_report: Any
 
 
+@dataclass(frozen=True, slots=True)
+class PlatformFabricValidationResult:
+    """Result of Platform config translation and Fabric validation."""
+
+    agent_config: AgentConfig
+    fabric_config: Any
```
mikeknep: can this be a `FabricConfig` instead of `Any`?

plugins/nemo-agents/src/nemo_agents_plugin/fabric/validation.py
```diff
@@ -22,6 +29,16 @@ class FabricValidationResult:
...
+    plan: Any
+    doctor_report: Any
```
mikeknep: Should these two collapse into this?
```python
fabric_validation_result: FabricValidationResult
```

plugins/nemo-agents/tests/unit/test_fabric_translator.py
```diff
@@ -40,14 +39,18 @@ def enable_relay(
...
+        translator,
+        "fabric",
+        SimpleNamespace(
+            EnvironmentConfig=_FabricObject,
+            FabricConfig=_FakeFabricConfig,
+            HarnessConfig=_FabricObject,
+            MetadataConfig=_FabricObject,
+            ModelConfig=_FabricObject,
+        ),
+    )
```
mikeknep: Given how directly the production code under test works with these types, it feels like we should just use the real types in the tests instead of these mocks? Especially now that we can import them sanely. For example, if Fabric makes a breaking change to any of these types, I think we'd want the tests to have a chance at catching it.

mmogallapalli: Good call, I updated these tests to use the real models. Only left the faked ones where we need to isolate plan/doctor behavior in validation tests.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/validation.py
mmogallapalli: Updated to use `FabricConfig`

plugins/nemo-agents/src/nemo_agents_plugin/fabric/validation.py
mmogallapalli: I like that, updated to reflect the consolidation,

plugins/nemo-agents/src/nemo_agents_plugin/fabric/translator.py
mmogallapalli: The underlying issue appears to be a repo type-lint environment mismatch, not a missing runtime dependency from `nemo-agents`.

`nemo-agents-plugin` is already wired as a workspace member / enabled plugin, so normal plugin installs should include its dependency set. The issue seems to be specific to `lint-python-types.sh`: it runs `ty` from a narrower lint environment, while `ty` still scans `plugins/nemo-agents/src` via extra paths. That means `ty` can see files in `nemo-agents` without necessarily having all `nemo-agents-plugin` dependencies installed in that specific lint env.

I first tried the cleaner dependency-based fix by adding `nemo-fabric` to root `dev`, following the existing type-check helper pattern, e.g.:

```toml
"nemo-platform-plugin",          # Optional runtime dep; included here for type checking
"nvidia-nat-atif>=1.8.0,<1.9",   # ATIF schema models for agentic-use trajectory validation
```

But CI still failed because this job appears to run with the narrower `insights` type-check env, so the root `dev` addition did not solve the actual failing environment.

Given that, I went back to the earlier patch, add targeted `ty: ignore[unresolved-import]` comments with a short explanation. 

Longer term, I think the cleaner repo-level fix could be:
- create a dedicated repo-wide `typecheck` dependency group for all source paths included by `ty`;
- change `lint-python-types.sh` so plugin source is checked from an env that installs those plugin deps;

mikeknep (approved):

PR #768, opened 2026-07-17
fix(auth): forward synthesized HF principal to PDP check

anastasia-nesterenko:
## Summary

- Pass preprocessed principal headers into the PDP authorization check.
- Preserve existing behavior for requests that use their original headers.
- Add regression coverage verifying that HF-compatible Bearer authentication authorizes as the synthesized service principal.

## Problem

The HF-compatible authentication flow synthesizes principal headers from service Bearer tokens without mutating `request.headers`. The PDP authorization path rebuilt its headers from the original request, discarding the synthesized service principal.

## Testing

- `uv run --frozen pytest packages/nmp_common/tests/auth/test_middleware.py -q`
- 27 tests passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization handling for HF-compatible requests by supporting preprocessed headers in the authorization flow.
  * Service principals synthesized from Bearer tokens (e.g., `service:<name>`) are now passed correctly to authorization checks.
* **Tests**
  * Added unit coverage to confirm authorization succeeds and that the derived service principal ID is correctly provided to the authorization client.


packages/nmp_common/src/nmp/common/auth/middleware.py
```diff
@@ -568,7 +573,8 @@ async def _handle_auth_check(self, request: Request, call_next: Callable) -> Res
             - 500 Internal Server Error: For unexpected PDP errors
             - Response from downstream: If authorized
         """
-        headers_dict = dict(request.headers)
+        if headers_dict is None:
```
mckornfield: should we also check `{}` empty dict?

mckornfield (approved):

PR #780, opened 2026-07-20
feat(intake): add column sorting to evaluation sessions endpoint [ASE-591]

walston:
## Summary

Adds multi-field column sorting to the `GET /v2/workspaces/{workspace}/evaluations/{name}/sessions` endpoint. Sort composes globally with pagination (ORDER BY runs in ClickHouse before LIMIT/OFFSET).

### Endpoint (`endpoints.py`)
- New `_SESSION_SORT_FIELDS` whitelist constant (module-level, per codebase convention)
- New `_parse_session_sort_keys(sort)` parser: comma-separated `-field` format, 422 on invalid field
- Added `sort: str | None = Query(default=None)` param to `list_evaluation_sessions`
- Parsed `sort_keys` threaded to `list_sessions` (raw string never reaches the repository)

### Repository + SQL (`evaluation_session_repository.py`)
- `list_sessions` accepts `sort_keys: list[tuple[str, bool]] | None = None`
- `_build_order_by`: builds a comma-separated ORDER BY clause from sort keys + tiebreaker; NULLS LAST on all user keys
- `_pre_page_metrics_cte_sql`: computes cost/tokens for **all** sessions before pagination when sorting by those fields — necessary because those values are not in `trace_index` and would otherwise only be computed for the current page
- `_list_sql` now accepts `sort_keys` and generates dynamic ORDER BY in both `page_sessions` and the final SELECT; conditionally injects `pre_page_metrics` CTE for cost/tokens sorts
- Default (no sort param): preserves original `start_time ASC, root_span_id ASC` behaviour

### Sortable fields
`test_case_id`, `started_at`, `ended_at`, `latency_ms`, `status`, `cost_total_usd`, `tokens`

### Tests
- Updated existing payload-mode tests to pass the now-required `sort_keys` arg
- Added: default order, single-field asc/desc, multi-field, pre_page_metrics injection for cost/tokens, tiebreaker presence, `_build_order_by` unit tests

### OpenAPI
- `make refresh-openapi` run; `ListEvaluationSessionsParams` gains `sort`
- `make stainless` (Python SDK sync) requires `STAINLESS_API_KEY` — run separately

## Out of scope
Frontend wiring — separate PR on a separate branch (ASE-591 FE)

Closes ASE-591 (partial — BE only)

## Summary by CodeRabbit

* **New Features**
  * Added an optional `sort` query parameter for evaluation sessions, supporting comma-separated multi-field ordering with `-` for descending and default ordering by `started_at` ascending when omitted.
  * Sorting supports specific session fields and metric-based ordering (for example, cost and token totals).
  * Updated the endpoint’s labeling to align with the “Experiments” category.
* **Bug Fixes**
  * Ensured deterministic, stable ordering across pagination with consistent tie-breaking.
  * Expanded validation and error behavior for unsupported/empty `sort` (400) and requests too large to sort by metrics (413), while retaining `503` for backend unavailability.

### DIFF
--- a/services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
+++ b/services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
@@ -51,7 +51,7 @@ from nmp.intake.spans.evaluation_rollup_repository import (
     EvaluationRollupRepository,
     ScoreRollup,
 )
-from nmp.intake.spans.evaluation_session_repository import EvaluationSessionRepository
+from nmp.intake.spans.evaluation_session_repository import EvaluationSessionRepository, MetricSortTooLargeError
 from nmp.intake.spans.storage import make_pagination
 
 logger = logging.getLogger(__name__)
@@ -657,10 +657,11 @@ async def unpin_evaluation(
 @router.get(
     "/v2/workspaces/{workspace}/evaluations/{name}/sessions",
     response_model=Page[EvaluationSessionResponse],
-    tags=[EVALUATIONS_TAG],
     responses={
-        400: {"description": "Invalid filter value"},
+        400: {"description": "Invalid filter value or sort field"},
         404: {"description": "Evaluation not found"},
+        413: {"description": "Too many sessions to sort by cost or tokens"},
+        422: {"description": "Unsupported or empty sort field"},
         503: {"description": "ClickHouse unavailable"},
     },
     openapi_extra=generate_openapi_extra_params(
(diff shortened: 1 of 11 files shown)


services/intake/src/nmp/intake/api/v2/experiments/endpoints.py
```diff
@@ -671,8 +684,10 @@ async def list_evaluation_sessions(
             "300 characters; detailed returns full root-span payloads."
         ),
     ),
+    sort: str | None = Query(default=None, description="Comma-separated sort fields..."),
```
shanaiabuggy: Update description:

```
sort: str | None = Query(
    default=None,
    description=(
        "Comma-separated list of fields to sort by, applied in order (the first field dominates); "
        "prefix a field with '-' for descending — e.g. '-cost_total_usd,latency_ms'. Fields: "
        "test_case_id, started_at, ended_at, latency_ms, status, cost_total_usd, tokens. When omitted, "
        "sessions are ordered by started_at ascending."
    ),
),
```

shanaiabuggy (approved):
approving with 1 comment! also tagging in @BrianNewsom to do a quick review as the clickhouse connoisseur

BrianNewsom:
Performance concern for cost_total_usd/tokens sorts: ClickHouse 24.3 inlines CTEs rather than materializing them. page_sessions is referenced by current_page_spans, session_metrics, session_scores, and the
  final SELECT; because page_sessions depends on pre_page_metrics, the all-session span aggregation may execute once per reference (it will do the expensive queries 4x)

  Suggested direction:

  Query 1: count scoped sessions

  Query 2: compute sorted page
  scoped sessions
      → aggregate token/cost metric across their spans once
      → ORDER BY + LIMIT/OFFSET
      → ordered page of session IDs

  Query 3: hydrate page
  ordered session IDs (≤ 1000)
      → fetch session fields
      → aggregate page spans
      → fetch evaluator scores
      → restore Query 2 ordering in Python

  This should preserve global metric sorting while ensuring the expensive all-session aggregation runs once per request.

BrianNewsom (approved):

PR #784, opened 2026-07-20
feat(evaluator-sdk): allow a prebuilt image for FabricContainerRuntime

SandyChapman:
## What

`FabricContainerRuntime` always provisioned the harness-agnostic Fabric image (build-if-missing). This adds an optional keyword-only `image` parameter so a caller can supply a prebuilt image tag when the sandbox needs extra tooling — e.g. a document-processing image (pandoc/libreoffice), a pinned base, or an image from an air-gapped registry.

## Behavior

- **Backward-compatible.** `image` defaults to `None`; when unset, behavior is unchanged (`ensure_fabric_image()` build-if-missing on first run).
- When provided, the tag is used verbatim and the build step is skipped. Image *building* remains a client concern, as before.

## Tests

Two new cases in `test_fabric_container_runtime.py` (existing fake-provider harness):
- a supplied image reaches the `SandboxSpec` and short-circuits `ensure_fabric_image`;
- the default path still provisions the build-if-missing image.

`pytest` green (17/17 in the file); `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

* **New Features**
  * Added an optional custom Fabric container image override for evaluation runs.
  * Caller-provided image tags are used as-is and stored in `trial` metadata.
  * When no image is provided, the runtime provisions the default image automatically.
* **Bug Fixes**
  * Failed trials now consistently include the selected image in `trial` metadata.
* **Tests**
  * Added coverage for supplied-image usage, default image provisioning, and early-failure metadata behavior.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
@@ -93,6 +93,7 @@ class FabricContainerRuntime:
         provider: SandboxProvider,
         profiles: Sequence[FabricProfileConfig | Mapping[str, Any]] = (),
         secrets: Mapping[str, SecretRef] = {},
+        image: str | None = None,
     ) -> None:
         # The Fabric agent is fully described by its ``FabricConfig`` (harness + model + runtime); it is
         # consumed structurally as a mapping to cross the sandbox boundary as JSON.
@@ -105,8 +106,9 @@ class FabricContainerRuntime:
         self._secrets = dict(secrets)
         self._resolved_env: dict[str, str] = {}
         self._secrets_resolved = False
-        # Provisioned opaquely on first run (build-if-missing from the config's harness).
-        self._image: str | None = None
+        # Optional prebuilt image: the trial runs inside it, so it must contain the Fabric CLI + adapter.
+        # None -> stock harness-agnostic image built on first run.
+        self._image: str | None = image
 
     async def resolve_secrets(self, secret_resolver: SecretResolver) -> None:
         """Resolve declared ``SecretRef``\\ s to values, keyed by the env var each harness reads.
@@ -143,7 +145,8 @@ class FabricContainerRuntime:
             # Provision the harness-agnostic Fabric image once, build-if-missing (a first build compiles
             # nemo-fabric — minutes); keep the blocking build off the shared event loop. Inside the guard
             # so the provider is disposed even if provisioning or secret resolution raises.
-            self._image = await asyncio.to_thread(ensure_fabric_image)
+            if self._image is None:
+                self._image = await asyncio.to_thread(ensure_fabric_image)
             if self._secrets and not self._secrets_resolved:
                 # No orchestrator resolved our secrets (standalone run) — fall back to local env resolution.
                 await self.resolve_secrets(LocalSecretResolver())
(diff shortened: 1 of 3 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -105,8 +106,10 @@ def __init__(
         self._secrets = dict(secrets)
         self._resolved_env: dict[str, str] = {}
         self._secrets_resolved = False
-        # Provisioned opaquely on first run (build-if-missing from the config's harness).
-        self._image: str | None = None
+        # A caller-supplied image tag is an escape hatch for sandboxes that need extra tooling (e.g. a
+        # document-processing image) — image *building* stays a client concern. When None, the harness-
+        # agnostic Fabric image is provisioned opaquely on first run (build-if-missing).
+        self._image: str | None = image
```
ngoncharenko, line 111: if non-fabric image is provided, does it mean that this is the container the trial is going to happen or the agent still ends up running in the fabric's container?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: The provided image *is* the container the trial runs in: the sandbox provider starts a container from `image` and execs `fabric run` inside it, so the harness + agent run in your image — there's no separate Fabric-managed container. The one constraint is that the image must still contain the Fabric CLI/runtime + the target harness adapter, so build it *from* the stock Fabric image and layer your extra tooling (e.g. pandoc/libreoffice) on top. With `image=None` the stock harness-agnostic image is built as before. I've clarified this in the param's comment in 66990e6.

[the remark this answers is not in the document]
SandyChapman: Addressed in 66990e6: factored out `_base_metadata()` and threaded it through the early-failure path in `_run_task`, so trials that fail before `_to_trial` (sandbox startup, seeding, download) now carry the runtime + selected image. Added a regression test (`test_early_failure_trial_carries_image_metadata`) covering a download-time failure.

ngoncharenko: thanks, we will need to clarify the constraints on the image in the docs and would be nice to validate that the layers contain Fabric base

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -105,8 +106,12 @@ def __init__(
         self._secrets = dict(secrets)
         self._resolved_env: dict[str, str] = {}
         self._secrets_resolved = False
-        # Provisioned opaquely on first run (build-if-missing from the config's harness).
-        self._image: str | None = None
+        # A caller-supplied image tag is an escape hatch for sandboxes that need extra tooling (e.g. a
+        # document-processing image with pandoc/libreoffice). The trial runs *inside* this image — the
+        # sandbox execs ``fabric run`` in it — so it must still contain the Fabric CLI/runtime + the target
```
arpitsardhana: nit: Too much commentary by claude/codex makes code too verbose. Preferably a single line would be enough

arpitsardhana (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: Good call — cut to 1–2 lines in 521ee45.

PR #785, opened 2026-07-20
fix(guardrails): Ensure self-check config with retains custom base URL

JashG:
## Summary

Fixes a bug where a guardrail config's `main` model entry was silently dropped if it didn't specify a `model` name, even when it set other fields like `parameters.base_url` to point the main LLM at a custom
endpoint.

This addresses an edge case for configs relying on custom base URLs: requests would fail with a 503 "Failed to run input rails" instead of routing requests to the configured URL.

## Root cause

Under the IGW Plugin model, a `main` entry's `model` name is never actually used; the real model always comes from the request body (see `rails.build_main_llm`). To avoid tripping nemoguardrails' `Model` validator (which requires a non-empty `model` on every entry), `stabilize()` used to strip out any `main` entry that didn't name a
model in `model`, `parameters.model`, or `parameters.model_name`, including ones that only existed to pin `engine`/`parameters.base_url`. Stripping those entries silently discarded the `parameters.base_url` override.

## Fix

Instead of dropping nameless `main` entries, `stabilize()` now fills in a placeholder (`MAIN_MODEL_REQUEST_PLACEHOLDER = "<request>"`) on the `model` field before handing the config to the upstream validator. The entry (with its real `engine`/`parameters`) is kept and surfaced as `main_model_template`, and the placeholder itself is never used for an actual call. `build_main_llm` still populates the real model name from the incoming request body.

**NOTE:** Only the top-level `model` field is treated as canonical for naming a `main` entry going forward. The code previously supported alternative ways of populating the model `parameters.model` / `parameters.model_name`, but
these were an artifact of the legacy codebase and were never documented or formally supported, so I removed support for these.

## Testing

- Added/updated unit tests in `test_llmrails_cache.py` covering: a
  nameless `main` entry with only `base_url` surviving as
  `main_model_template`, all "missing model name" wire shapes (absent,
  `None`, empty string, whitespace), duplicate-`main` detection, and
  content-hash stability.
- Removed a pre-existing `"model": "rail-main-placeholder"` workaround
  from `test_self_check_rails.py` and `test_middleware_config_caching.py`.
  Both configs only set `parameters.base_url` on their `main` entry,
  so they now exercise the real fix end-to-end instead of masking the
  underlying bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Guardrails configurations where the `main` entry omits a model name, ensuring it’s retained and normalized using a request placeholder during validation.
  * Preserved `main` templates and non-model overrides (such as connection/base URL settings) instead of dropping nameless `main` entries.
  * Aligned missing-model-name validation and error messaging with the updated normalization flow.
  * Kept middleware config caching behavior consistent, including entity-backed updates and duplicate `main` detection.

### DIFF
--- a/plugins/nemo-guardrails/src/nemo_guardrails_plugin/llmrails_cache.py
+++ b/plugins/nemo-guardrails/src/nemo_guardrails_plugin/llmrails_cache.py
@@ -205,8 +205,7 @@ def _fill_main_model_placeholder(model_config: dict[str, Any]) -> dict[str, Any]
 
     parameters = model_config.get("parameters")
     if isinstance(parameters, dict) and (
-        _has_nonempty_model_name(parameters.get("model"))
-        or _has_nonempty_model_name(parameters.get("model_name"))
+        _has_nonempty_model_name(parameters.get("model")) or _has_nonempty_model_name(parameters.get("model_name"))
     ):
         return model_config
 
--- a/plugins/nemo-guardrails/src/nemo_guardrails_plugin/llmrails_cache.py
+++ b/plugins/nemo-guardrails/src/nemo_guardrails_plugin/llmrails_cache.py
@@ -179,10 +179,20 @@ body's model, so this placeholder is to satisfy the schema validator.
 """
 
 
+def _has_nonempty_model_name(value: object) -> bool:
+    """``True`` when ``value`` is a non-empty (after strip) model name string."""
+    return isinstance(value, str) and bool(value.strip())
+
+
 def _fill_main_model_placeholder(model_config: dict[str, Any]) -> dict[str, Any]:
     """Given a model config from a GuardrailConfig, populates the ``model`` field with
     a placeholder value if the config represents the main model and doesn't already have
     a model name.
+
+    The upstream ``Model`` schema accepts a name in exactly one of three places:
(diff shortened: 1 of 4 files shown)


plugins/nemo-guardrails/src/nemo_guardrails_plugin/llmrails_cache.py
```diff
@@ -171,30 +171,30 @@ def extract_output_rails_streaming_config(
...
+    a model name.
+
+    This is required to satisfy the nemoguardrails library schema validator.
+    We always populate the actual model name using the incoming request
+    body's model, so this placeholder is to satisfy the schema validator.
+    """
+    if model_config.get("type") != MAIN_MODEL_TYPE:
+        return model_config
+    model = model_config.get("model")
+    if model is not None and not (isinstance(model, str) and not model.strip()):
```
albcui: In addition to `model`, we should probably also check `parameters`, and whether `model`/`model_name` exists because the library honours those as well. Right now, if we fill `model` in addition to the existing `parameters.model`, then the library would complain.

JashG: Good call, fixed

albcui (approved):

PR #786, opened 2026-07-20
fix(openapi):  schema name collision hardening

albcui:
Summary:
- The issue is that when there's a name collision in the OpenAPI `$ref` namespace AND the actual schema differs, we accept the first one and throw away the rest. The result is an incorrect OpenAPI spec post merge. This PR hardens the OpenAPI spec processing to strictly error when collisions are detected. This revealed the following issues, which are fixed as part of this PR.
- The name `DeleteResponse` and `GenericSortField` had collisions between Guardrails and Core Entities services.
  - Guardrails: drop the local `DeleteResponse` and `GenericSortField`, and re-use the ones in `nmp-common`
  - Core Entities Service: re-use `GenericSortField` from `nmp-common`. Re-named `DeleteResponse` to `EntityDeleteResponse` for now because it has an extra `deleted_count` field. However, ideally, we can leverage `nmp-common`'s `DeleteResponse` as well.

## Summary by CodeRabbit

* **New Features**
  * Added `updated_at` / `-updated_at` sorting to multiple list commands (including workspaces, filesets, guardrail configs, and related resources).
  * Standardized list sorting to use a shared, entity-style sort contract.
  * Updated delete responses to return the deleted resource’s `id` (removing `deleted_at`).

* **Documentation**
  * Updated CLI reference and OpenAPI specifications to reflect new `--sort` options and the updated delete response schema.

* **Bug Fixes**
  * OpenAPI schema-name collisions now fail by default during spec generation.


JashG (approved):

script/openapi_helper/plugin_config.py
```diff
@@ -22,12 +22,11 @@ class PluginConfig:
...
-    # nemo-customizer mounts every customization backend under
-    # /apis/customization). Then two backends defining a same-named model with
-    # differing content is always a real bug, and spec generation should fail
-    # loudly instead of silently collapsing them. Off => warn-and-collapse.
-    strict_schema_collisions: bool = False
+    # Schema-name collisions fail spec generation by default. Set to false to
+    # opt a plugin out (warn-and-collapse) if its spec must tolerate a known
+    # pre-existing collision — collapsing silently ships a wrong contract, so
+    # opting out should be rare and deliberate.
+    strict_schema_collisions: bool = True
```
mckornfield: my only nit... should we just delete this flag... or is there a reason to ALLOW schema collisions?

mckornfield (approved):

script/openapi_helper/plugin_config.py
albcui: I agree. My initial thought was to leave a backdoor open to devs who urgently need to merge into main without thinking through a collision fix. However, that would result in a net worse code quality, and should be avoided as much as possible.

Removed.

PR #799, opened 2026-07-20
feat(client): add platform NemoClient provider

maxdubrinsky:
## Summary

- Add a plugin-side `NemoClientProvider` protocol with explicit override, entry-point discovery, and standalone environment fallback.
- Add the platform provider backed by service URL routing, shared HTTP transports, request principal headers, and OTEL propagation.
- Register provider entry points in `nmp-common` and the bundled `nemo-platform` package.
- Fail closed for broken or conflicting installed providers and prevent OTEL context from replacing authoritative identity headers.

## Validation

- 81 focused provider and client-factory tests passed after rebasing onto current `main`.
- Ruff, formatting, focused `ty`, wheel entry-point inspection, and `git diff --check` passed.
- Independent adversarial review found no blocker.

## Follow-up

Directly owned `NemoClient` transport teardown is pre-existing and intentionally outside this PR. AIRCORE-933 tracks ownership-aware close and context-manager support.

Linear: https://linear.app/nvidia/issue/AIRCORE-883


maxdubrinsky:
Folded into #800, which is now retargeted to `main` and carries the full NemoClient foundation (provider AIRCORE-883 + request-scoped DI AIRCORE-904). Closing to consolidate the two stacked drafts into one reviewable PR. No content lost: #800's branch already contained every commit from this branch.

PR #705, opened 2026-07-15
feat(models): deployments_plugin cutover, drop legacy backends and k8s-nim-operator (AIRCORE-759)

tylersbray:
## Summary

- Completes AIRCORE-759 P2: models controller uses `deployments_plugin` exclusively; removes legacy `docker`, `k8s_nim_operator`, and `none` backends (~18k lines deleted).
- Ports NIM parity from `nimservice_compiler` (tool-call plugin, multi-LLM env, `k8s_nim_operator_config`, `override_config`) into `deployments_plugin/compiler`.
- Drops bundled `k8s-nim-operator` Helm subchart and NIM CR RBAC; adds required kind smoke e2e (`e2e/test_models_deployments_plugin.py`) wired into `kind-cpu-smoke` CI.
- Implements non-blocking idempotent delete with `deleting_timeout_seconds` on models + plugin reconciler (PR #663 carryover).

## Test plan

- [x] `uv run pytest services/core/models/tests/unit/ -q` (834 passed)
- [x] `uv run pytest services/core/models/tests/integration/ -q` (236 passed, 7 skipped)
- [x] `uv run pytest services/core/models/tests/unit/controllers/backends/deployments_plugin/ -v`
- [x] `uv run pytest plugins/nemo-deployments/tests/unit/reconciler/ -k delete -q`
- [x] Helm template renders without k8s-nim-operator (`helm template` + `kind.yaml`)
- [ ] Manual smoke (pre-merge): docker NIM+vLLM, k8s vLLM, LOST drift, delete ordering on dev pod
- [ ] `kind-cpu-smoke` CI (runs on this PR)

## Enablement

`deployments_plugin` is now the default backend for docker/k8s/none runtimes. Helm `basePlatformConfig` and e2e values configure `local-k8s` executor.

## Notes

- Builds on merged PR #663 (`deployments_plugin` backend introduction).
- Implements the models-deployment slice of AIRCORE-834 early via kind smoke; remaining plugin entity CRUD e2e stays on that ticket.
- `k8s_nim_operator_config` and `override_config` are deprecated in OpenAPI but still honored on k8s NIM paths.
- Clusters upgraded from operator-based installs may retain orphan NIMService CRs; stale READY deployments surface as LOST.

## Summary by CodeRabbit

* **New Features**
  * Added deployments-plugin support for model lifecycles across Docker and Kubernetes.
  * Added generic CPU model deployment validation, including readiness and cleanup checks.
[...]

PR #804, opened 2026-07-20
fix(models): fix default model_puller image for NIMCache

anubhutivyas:
**Issue:**

- Creating a deployment on k8s backend fails during NIMCache init with below error because the `NIMCache` pod runs the platform `nmp-api` image, which does not contain `download-to-cache`.
```
Error: failed to start container "nim-cache-ctr": OCI runtime create failed:
exec: "download-to-cache": executable file not found in $PATH
```
- `compile_nimcache` builds a NIMCache CR where the HF source modelPuller is set from the models service's `huggingface_model_puller` whose value is set to `nmp-api` image by default

**Change:**
Following how we have a separate field in `DockerBackendConfig` for docker backend, create a new field `huggingface_model_puller` in `K8sNimOperatorConfig`  and set the default as `nds-v2-huggingface-cli:25.10`, which ships download-to-cache. 

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration option for the Hugging Face model-puller image used by NIMCache, with a pinned default image.
* **Documentation**
  * Updated the configuration reference to reflect the new default and clarify the cache-population entrypoint behavior.
* **Bug Fixes**
  * Ensured NIMCache downloads for Files-service–backed deployments consistently use the configured puller image.
* **Tests**
  * Updated and added unit tests to validate the default puller image behavior and related NIMCache generation.

### DIFF
--- a/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
+++ b/services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
@@ -358,7 +358,6 @@ class NimOperatorReconciler(Reconciler):
                 model_namespace=model_namespace,
                 model_name=model_name,
                 pvc_size=pvc_size,
-                huggingface_model_puller=self._huggingface_model_puller,
                 model_revision=resolved.model_revision,
             )
             await self._create_nimcache(nimcache)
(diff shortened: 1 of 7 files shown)


services/core/models/src/nmp/core/models/controllers/backends/k8s_nim_operator/reconcilers/nim_operator.py
albcui: Since you added the model puller to the `K8sNimOperatorConfig`, this is redundant information. We should just rely on the `backend_config.huggingface_model_puller`.

anubhutivyas: This was actually left as is intentionally because I didn't want to touch the `tool-call-plugin`'s container image, but thinking through this I feel it won't break anything as we were using the nds image earlier too. Removed `_huggingface_model_puller` from contructor and now using `backend_config.huggingface_model_puller` everywhere.

anubhutivyas:
Closing it because we have removed the k8s-nim-operator in this PR - https://github.com/NVIDIA-NeMo/nemo-platform/pull/705

PR #813, opened 2026-07-21
fix(guardrails): Propagate mutated text to proxied request/response

JashG:
## Summary
- Write input-rail masked/transformed text back onto the **last user message** in the proxied request (string content only). This matches the library behavior I observe - i.e. only the last user message is modified by rails that might redact/modify content (such as PII redaction).
- Write output-rail masked text back onto `choices[0].message.content`.
- Mutate `request.body` in place (sanitize + messages replace) instead of reconstructing `InferenceRequest`, so fields like `typed_body` are preserved; IGW re-derives `typed_body` after the plugin.

## Summary by CodeRabbit

* **New Features**
  * Input rail transformations now sanitize and forward updated upstream messages, rewriting the last `role=user` message content when text-only.
  * Output rail transformations are propagated into the assistant content returned to clients.
* **Bug Fixes**
  * Prevents masking/write-back for non-text (multimodal-like) user inputs and avoids updates when rail output matches the original content.
  * Improves safety by avoiding unintended request/response mutations.
* **Documentation**
  * Clarified the OpenAI-shaped generation response format used for rail-modified content.
* **Tests**
  * Expanded unit coverage for input masking behavior, multimodal skipping, and output write-back/update detection.

### DIFF
--- a/plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
+++ b/plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
@@ -382,15 +382,18 @@ def build_output_response_body(
 
     choices = list(response.get("choices", [])) if isinstance(response.get("choices"), list) else []
     response["choices"] = choices
+
     # Output rails validate choices[0].message, so return only the choice that was checked.
-    # If output rails masked ``$bot_message``, write that text back onto the choice.
     if generation_response is not None and choices:
         choice = {**choices[0], "index": 0}
         message = dict(choice.get("message") or {}) if isinstance(choice.get("message"), dict) else {}
         original_content = message.get("content")
+
+        # If the output rail modified the message content, write it back onto the choice.
         if isinstance(original_content, str):
             message["content"] = apply_output_rail_modifications(original_content, generation_response)
             choice["message"] = message
+
         response["choices"] = [choice]
 
     guardrails_data = build_guardrails_data(
--- a/plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
+++ b/plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
@@ -144,8 +144,12 @@ def is_blocked_generation_response(generation_response: GenerationResponse) -> b
 
 
 def extract_response_content(generation_response: GenerationResponse) -> str:
-    """
-    Extract the last assistant message content from a GenerationResponse.
+    """Return the post-rail text from a ``GenerationResponse``.
+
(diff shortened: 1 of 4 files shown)


plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
```diff
@@ -158,6 +162,85 @@ def extract_response_content(generation_response: GenerationResponse) -> str:
...
+# as GenerationResponse.response with role=assistant (not role=user).
+# Example: request user "Hi John" → response [{"role": "assistant", "content": "Hi <PERSON>"}].
+# If the message content is redacted, we need to write it back onto the last user message
+# in the request.
+# ---------------------------------------------------------------------------
+
+
+def resolve_rail_message_content(*, original: str, processed: str) -> str:
+    """Return the text to forward after rails: redacted if changed, else original."""
+    return processed if processed != original else original
```
gabwow: if processed == original, it returns processed (same as original). Otherwise this returns processed, so this function and the related checks aren't doing anything.

gabwow (changes requested):

plugins/nemo-guardrails/tests/unit/test_middleware.py
```diff
@@ -162,13 +162,19 @@ def _make_entity(
     )
 
 
-def _make_generation_response(*, is_blocked: bool = False) -> GenerationResponse:
+def _make_generation_response(
+    *,
+    is_blocked: bool = False,
+    content: str = "I'm sorry, I can't help with that.",
+    output_data: dict[str, Any] | None = None,
```
gabwow: can we say You Aren't Going to Need It here, and assume that if we ever want to start using output_data in the future we could add it in? Currently we don't use it.

plugins/nemo-guardrails/src/nemo_guardrails_plugin/responses.py
JashG: Good catch, this was redundant. Removed in https://github.com/NVIDIA-NeMo/nemo-platform/pull/813/commits/a8ea1bd924e2523d52ee2bf6ff976634dc87940d
We just need to compare `processed` vs. `original` and return `processed` if it differs.

gabwow (approved):
thanks, one additional optional nit.

plugins/nemo-guardrails/tests/unit/test_middleware.py
JashG: I removed `output_data`. Was using it in a previous iteration, but it's not needed at all, so removed.

PR #816, opened 2026-07-21
feat(evaluator-sdk): inject multiple skills into FabricAgentRuntime

SandyChapman:
## What

`FabricAgentRuntime` (the Fabric **host** runtime) held a single skill and `with_skill` *replaced* it, so one runtime could not inject more than one agent skill. This is gap identified from the LAB-examples analysis: LAB hands the agent all three of its skills (docx/pptx/xlsx) on every task, and a single-skill runtime can only mirror it by running three separate single-skill arms — which is semantically wrong (LAB is not a per-skill benchmark).

This lets one runtime stage a **set** of skills.

## Changes

- **`runtime.py`** — replace `self._skill` with `self._skills: list[AgentSkill]`. The constructor takes a single `skills=` argument (a lone skill is just a one-element set). `with_skills(skills)` derives a configured clone; `with_skill` is kept as a thin wrapper for the single-skill A/B pattern the skill-eval example uses. A/B semantics preserved: baseline = `with_skills([])`, treated = the set. Duplicate skill names are rejected up front (before any task runs) since two bundles claiming the same `<name>/` would collide.
- **`skills.py`** — add `install_skills(...)`, which loops the existing `install_skill` per skill (each bundle is already namespaced under `<name>/`) and, for the **native** mode, merges the per-skill paths into **one** `skills` overlay. Fabric applies profile `skills.paths` last-wins, so a per-skill overlay would silently drop all but the last. Codex self-injection stages and later removes every bundle. Also adds `require_unique_skill_names`.
- **Trial metadata** — new `skills` list (one provenance per injected skill). The historical `skill` field stays the lone provenance for a one-skill run (else `None`), so `SkillUsedMetric` and existing single-skill trials/tests are byte-for-byte unchanged.

## Coordination with L1 (container runtime)

Skill injection for `FabricContainerRuntime`, staging into the sandbox seed set has **not landed** on `main` (yet). Per the task's fallback, this is scoped to the **host** runtime; applying the same multi-skill shape to the container runtime is a **follow-up** once the new work for fabric in a container merges.

## Tests

Extended `test_fabric_runtime.py`:
- multiple native skills → each staged under its own `<name>/`, all listed in a single merged `eval_skill` overlay, in order;
- multiple codex skills → each discoverable at `.agents/skills/<name>/` during the run, all removed before the workspace is exposed as evidence;
- duplicate names rejected at construction and via `with_skills`;
- existing single-skill tests updated to `skills=[skill]`.

`pytest` green (56/56 across `test_fabric_runtime.py` + `test_fabric_skills.py`); `ruff check` / `ruff format --check` clean; vendored SDK mirror regenerated and committed alongside the source.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring and injecting multiple skills in a single evaluation run.
  * Introduced chainable helpers to add one skill or multiple skills to the runtime, without altering the original instance.
  * Updated trial metadata to include per-skill provenance and an aggregated skills list for multi-skill runs.
* **Bug Fixes**
  * Rejects duplicate skill names during staging to avoid collisions.
  * Improved multi-skill overlay merging and Codex skill staging/cleanup, including when runs fail.
* **Tests**
  * Expanded unit and async integration coverage for multi-skill behavior, duplicate-name handling, and failure cleanup.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
@@ -41,7 +41,8 @@ from nemo_evaluator_sdk.agent_eval.runtimes.fabric.skills import (
     SKILL_MODE_CODEX_SKILLS_DIR,
     AgentSkill,
     SkillProvenance,
-    install_skill,
+    install_skills,
+    require_unique_skill_names,
     resolve_skill_mode,
 )
 from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig, AgentEvalTask
@@ -128,7 +129,7 @@ class FabricAgentRuntime:
         timeout_s: int = DEFAULT_FABRIC_TIMEOUT_S,
         capture_trajectory: bool = True,
         runtime_name: str = _RUNTIME_NAME,
-        skill: AgentSkill | None = None,
+        skills: Sequence[AgentSkill] | None = None,
     ) -> None:
         self._config = config
         self._profiles = list(profiles or [])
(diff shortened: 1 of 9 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -273,27 +286,28 @@ async def _run_task(
...
+                    install_skills,
+                    skills=self._skills,
                     adapter_id=agent_config.harness.adapter_id,
                     mode=skill_mode,
                     workspace_dir=workspace_dir,
                     skill_stage_dir=(evidence_dir / _SKILL_SUBDIR).resolve(),
                     existing_skill_paths=self._existing_skill_paths(),
                 )
-                skill_provenance = installation.provenance
+                skill_provenances = installation.provenances
```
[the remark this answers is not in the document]
SandyChapman: I think this was addressed in the latest push. Please confim.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -316,17 +330,31 @@ async def _run_task(
...
+
+    @staticmethod
+    def _skill_metadata(provenances: list[SkillProvenance]) -> dict[str, Any]:
+        """Trial-metadata fields describing the injected skill set (the A/B provenance).
+
+        ``skills`` is the full list of injected-skill provenances (empty = baseline). ``skill`` keeps the
+        historical single-provenance field — the lone provenance for a one-skill run, else ``None`` — so
+        single-skill consumers (e.g. ``SkillUsedMetric``) and existing trials keep working unchanged.
+        """
+        return {"skill": provenances[0] if len(provenances) == 1 else None, "skills": provenances}
```
ngoncharenko, line 357: SkillUsedMetric (packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/metrics.py) needs to be also updated to recognize these two options.

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: Good call out. I've updated.

PR #827, opened 2026-07-21
fix(guardrails): Ensure input rail logs are not dropped from final response

JashG:
## Summary
Stash the input-rail `GenerationResponse` in plugin state before returning an `ImmediateResponse`, so response middleware can still rebuild `guardrails_data.log` correctly.

IGW still runs response middleware after an input is blocked, which is where the `guardrails_data` field gets populated from the input and output rail results. Without this change, the blocked input's logs would be silently not included in the response.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved input-rail details when a request is blocked.
  * Ensured requested guardrails logging remains available in blocked responses.
  * Improved response metadata to include the relevant input and output rail results.

### DIFF
--- a/plugins/nemo-guardrails/tests/unit/test_middleware.py
+++ b/plugins/nemo-guardrails/tests/unit/test_middleware.py
@@ -811,8 +811,9 @@ class TestProcessRequest:
         guardrails_data = final.response_body_annotations["guardrails_data"]
         activated_rails = (guardrails_data.get("log") or {}).get("activated_rails") or []
         activated_by_name = {rail["name"]: rail for rail in activated_rails}
-        assert activated_by_name["self check input"]["stop"] is True
-        assert activated_by_name["custom check output"]["stop"] is False
+
+        # Verify the input rail's log fields are surfaced in the response.
+        assert "self check input" in activated_by_name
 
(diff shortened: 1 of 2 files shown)


plugins/nemo-guardrails/tests/unit/test_middleware.py
```diff
@@ -700,6 +702,62 @@ async def test_blocked_rail_returns_immediate_response(self, middleware: Guardra
         # for diagnostics, but the wire format keeps the legacy shape so
         # downstream consumers don't have to track per-revision IDs.
         assert result.response_body_annotations["guardrails_data"]["config_ids"] == ["ws/my-config"]
+        # Must stash even on the blocked path — response middleware still runs
+        # after ImmediateResponse and rebuilds guardrails_data from this state.
+        assert ctx.state(PLUGIN_NAME).get(STATE_KEY_INPUT_GENERATION_RESPONSE) is generation_response
+
+    async def test_blocked_input_preserves_requested_log_through_response_middleware(
```
gabwow, line 764: non-blocking: can we simplify this unit test by asserting only that the blocked input’s activated_rails survives process_response?

gabwow (approved):

PR #838, opened 2026-07-21
feat(nemo-agents): one shot local invocation support for fabric backed agents

mmogallapalli:
## Summary

This PR adds the first usable Platform flow for Fabric-backed NeMo Agents across AIRCORE-899, AIRCORE-900, and AIRCORE-901.

The main flow is:

```text
Platform agent.yaml -> AgentConfig -> FabricConfig -> Fabric runtime invocation
```

The Platform still owns the persisted `nemo-agents-spec-v1` config shape. Fabric remains the execution layer used after the plugin translates the Platform-owned config into a typed in-memory `FabricConfig`.

## Changes

- Added Fabric SDK dependency wiring while deferring Fabric runtime image wiring until packaging compatibility is settled.
- Added `nemo agents create --agent-config ...` support for `nemo-agents-spec-v1`.
- Added API create support for Platform-owned Fabric agent configs.
- Added validation through `AgentConfig -> FabricConfig -> Fabric plan/doctor`.
- Added deploy support for Fabric-backed configs as validation/preparation only.
- Added local one-shot invocation through `nemo agents invoke --agent-config ...`.
- Added local workspace directory preparation before Fabric invocation.
- Added normalized Fabric invocation output including response, error, artifacts, events, metadata, and Fabric IDs.
- Added unit coverage for CLI/API create, deploy preparation, one-shot invocation, and runner behavior.

## Scope Notes

Fabric-backed deploy does not start or retain a durable runtime in this PR. For this milestone, deploy means the Platform config can be translated and validated through Fabric planning/preflight.

`nemo agents invoke --agent <registered-agent-name>` is also out of scope for Fabric-backed agents in this PR. That path still relies on a deployed server/chat-completions endpoint. Follow-up work will define the remote runtime/session model where a Platform connection maps to a Fabric runtime.

## Out Of Scope

- Registered-agent Fabric invocation with `nemo agents invoke --agent ...`.
- Durable Fabric runtime/session lifecycle.
- Multi-user remote server design.
- Richer Fabric deployment/session status fields.
- Automatic harness adapter installation.
- Default Fabric runtime installation in Platform images.
- Final stable Fabric version pinning.

## Manual Validation

Platform setup and create:

```bash
make bootstrap-python
nemo setup --auto --start-services --install-skills --no-deploy-agent
nemo agents create \
  --name fabric-test \
  --agent-config plugins/nemo-agents/examples/nemo-agent-config/agent.yaml
```

Local macOS Fabric runtime setup used for smoke validation:

```bash
# macOS only, until Fabric runtime macOS wheels exist.
uv pip install -e /Users/mmogallapall/workspace/NeMo-Fabric/python
```

Fabric runtime is not included in the default Platform image yet because the current alpha runtime wheels target `manylinux_2_39`, while the Platform Docker base is `manylinux_2_36`. Runtime dependency/image wiring is deferred until Fabric publishes a compatible wheel or Platform updates its base image.

Hermes harness setup used by the example config:

```bash
uv pip install --prerelease=allow --index https://pypi.nvidia.com \
  "nemo-fabric-adapters-hermes==0.1.0a20260720"

# Temporary local smoke workaround: Fabric docs imply hermes-agent>=0.17.0,
# but that range currently failed dependency resolution in this environment.
uv pip install "hermes-agent==0.16.0"
```

Local one-shot invocation:

```bash
nemo agents invoke \
  --agent-config plugins/nemo-agents/examples/nemo-agent-config/agent.yaml \
  --input "Say hello in one sentence."
```

Result: succeeded. The command returned a normalized Platform result with `response`, artifacts, events, metadata, `runtime_id`, `invocation_id`, and `request_id`.

## Automated Validation

Passed:

```bash
tools/lint/lint-python-style.sh plugins/nemo-agents/src/nemo_agents_plugin/fabric/invocation.py plugins/nemo-agents/tests/unit/test_fabric_invocation.py
uv run --frozen pytest plugins/nemo-agents/tests/unit/test_fabric_invocation.py
```

Also previously passed focused unit coverage for Fabric runtime, Fabric invocation, CLI, runner, controller, deployments API, and runner in-memory behavior.

## Summary by CodeRabbit

* **New Features**
  * Added support for platform-owned agent configs across agent creation, deployment, and local one-shot invoke, including validation/normalization.
  * Enabled Fabric-backed local one-shot execution with structured JSON output and automatic local workspace setup.
* **Bug Fixes**
  * Invalid or unsupported configuration formats now return clear HTTP 400 errors.
  * Deployment status/startup timing now reflects backend-reported status values.
* **Documentation**
  * Updated CLI help and API docs to consistently use “agent config” terminology.
* **Tests**
  * Expanded unit and e2e coverage for the new config formats and Fabric one-shot behavior.


plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -8,11 +8,11 @@
...
 **Local commands (no platform required):**
 
-These wrap NAT's runtime directly and work without a running NeMo Platform
+These run against a local agent config and work without a running NeMo Platform
 instance.
 
-- ``invoke``   — single invocation (wraps ``nat run``)
-- ``run``      — start a persistent local FastAPI server (wraps ``nat serve``)
+- ``invoke``   — single invocation
+- ``run``      — start a persistent local FastAPI server for NAT configs
```
mikeknep, line 15: Much of the docstring-related diff in this PR appears to be removing references to NAT; do we want this here or some other language?

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/deployments.py
```diff
@@ -103,6 +111,22 @@ async def create_deployment(
     return saved
 
 
+def _resolve_deployment_config(agent: Agent, *, workspace: str) -> dict[str, Any]:
```
mikeknep, line 114: Between this and `_validate_agent_config_for_create` in api/v2/agents.py, it feels like some sort of protocol might be emerging, but maybe it's too soon to refactor towards that?

plugins/nemo-agents/src/nemo_agents_plugin/fabric/invocation.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+async def invoke_agent_config_once(
+    config: dict[str, Any],
+    inputs: Sequence[Any],
+    *,
+    base_dir: Path,
+) -> list[FabricRuntimeResult]:
+    """Translate a Platform agent config and run each input through Fabric once."""
+    agent_config = AgentConfig.model_validate(config)
```
mikeknep: Wondering if this function should take an `AgentConfig` and push the validation and conversion from dict to that type to the caller(s).

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -1222,6 +1238,15 @@ def _local_invoke(
...
         raise typer.Exit(code=1)
 
+    config_dict = _load_yaml(agent_config)
+    config_format = config_dict.get("config_format", NAT_WORKFLOW_CONFIG_FORMAT)
+    if config_format == NEMO_AGENTS_SPEC_CONFIG_FORMAT:
+        _local_fabric_invoke(config_dict, queries, base_dir=agent_config.parent)
+        return
+    if config_format != NAT_WORKFLOW_CONFIG_FORMAT:
+        typer.echo(f"Error: unsupported config_format {config_format!r}", err=True)
+        raise typer.Exit(code=1)
```
mikeknep, line 1248: This second if-then-raise check seems a little strange to me. It'll only raise if the user provides a yaml file that **does** have a `config_format` key present, but with a value that is **not** `NEMO_AGENTS_SPEC_CONFIG_FORMAT`.

If `config_format` is not present, we default to `NAT_WORKFLOW_CONFIG_FORMAT`, even if the rest of the yaml file is some completely unrelated/unexpected shape.

I'm not very familiar, but it seems like the second scenario would happen way more often than the first?

mikeknep: Similar thing going on in `create` above in this file/diff

mmogallapalli: For context, I set it up this way to preserve the existing NAT config behavior: if `config_format` isn’t specified, we treat it as `nat-workflow-v1 `by default. For the new `nemo-agents-spec-v1 `format, `config_format` is required explicitly.

I agree this means a random YAML without `config_format` can still enter the NAT path, but that matches the previous behavior and should fail later during NAT validation/execution rather than changing compatibility in this MR.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mmogallapalli: for now this MR enables single invocation (invoke equivalent) for Fabric based configs, I kept NAT based wording for run/package since we don't have equivalent support for Fabric based configs yet.

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/deployments.py
mmogallapalli: i think you're right, i can look into standing up some sort of shared protocol/registry in the next PR since that will entail supporting persistent runtime/more lifecycle operations.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/invocation.py
mmogallapalli: Thats valid, made that change

mikeknep (approved):

PR #841, opened 2026-07-22
feat(evaluator-sdk): agent-skill injection for FabricContainerRuntime

SandyChapman:
## What

Adds agent-skill injection to the sandboxed `FabricContainerRuntime`, so an **isolated (Docker) Fabric eval can run WITH an agentskills.io skill** — previously only the host `FabricAgentRuntime` supported skills (limitation **L1** from the LAB-examples gap analysis). It mirrors the host runtime's skill design and extends the same two-mode contract to the container.

## How

- `__init__` gains a `skill: AgentSkill | None`; new **`with_skill()`** clone method so an A/B eval derives baseline (`with_skill(None)`) and treated (`with_skill(skill)`) runtimes from one instance.
- The **injection mode is resolved once per run** (the adapter is constant across the taskset) via Fabric's capability planner (`_resolve_skill_mode`). `nemo_fabric` is imported lazily on the host **only when a skill is set**, so the no-skill container path stays dependency-free.
- New **`stage_skill_seed()`** in `skills.py` — the sandboxed sibling of `install_skill`. Instead of a host `copytree` (the container has no host workspace), it renders the bundle into the sandbox **seed set** (`SandboxSpec.files`) at the harness's in-sandbox discovery path:
  - **native** (hermes / any adapter that `accepts: ["skills"]`) → `/in/skills/<name>/` + a `skills.paths` profile overlay (preserving any pre-configured skill paths, last-wins);
  - **codex** → `<workspace>/.agents/skills/<name>/` (self-discovery), no overlay.
- The codex bundle is scrubbed from the downloaded `/out/workspace` before it's exposed as evidence (so injected files don't skew workspace-reading metrics). Native staging lives under `/in` and is never downloaded, so it needs no cleanup.
- Introduces **`SkillMode = Literal["native", "codex_skills_dir"]`**, threaded through both runtimes and `SkillProvenance.mode`.

Text bundles only: the seed set is `dict[str, str]`, so a binary file raises `SkillInjectionError` (the host `install_skill` copytree path still handles binary bundles).

## Tests

- 8 new unit cases in `test_fabric_container_runtime.py` mirroring the host skill-injection suite: native seed-set + overlay, preconfigured-path preservation, runtime-discovered adapter, codex seed + evidence exclusion, fail-fast on an unsupported adapter, no-skill-no-probe, and `with_skill` copy. Full fabric suite (container + host + skills) = **76 passing**; ruff + format clean; `ty` clean; SDK mirror re-vendored (`make vendor`).
- **Live-validated (native / hermes):** ran one task twice through the runtime against a real Docker sandbox — baseline vs. a skill mandating an unguessable marker token. The marker appeared only in the treated output, confirming the agent genuinely picks up the injected skill. Provenance (`mode=native`, staged path, content hash) stamped correctly; no leaked containers; native skill correctly absent from workspace evidence.
- **Codex mode is unit-tested only.** A full sandboxed codex run needs the codex CLI in the image, which the Fabric sandbox image does not provision yet (hermes+relay only) — separate from this change.

## Notes

- The **codex branch is a workaround for a gap in Fabric's codex adapter**, not a missing abstraction: the codex `fabric-adapter.json` declares `accepts: ["models", "telemetry"]` (no `skills`), so Fabric routes codex skills `unsupported`, while the codex CLI itself self-discovers `.agents/skills/`. When the codex adapter adds `skills`, this branch can be deleted and codex flows through the same native path.

## Summary by CodeRabbit

* **New Features**
  * Added multi-skill injection for Fabric agent and container runtimes, supporting native and Codex staging.
  * Introduced additive skill configuration via `with_skill` / `with_skills`, and enhanced trials to record multi-skill provenance.

* **Bug Fixes**
  * Added fail-fast validation when injected skills can’t be routed for the selected harness adapter.
  * Improved Codex workspace cleanup and ensured skill provenance metadata is included on both successful and failed trials.

* **Tests**
  * Expanded coverage for overlays, provenance stamping (single vs multi), staging rollback, skill deduplication, and collision behavior.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
@@ -29,16 +29,29 @@ Relay writes ATIF **inside the image** (no host gateway), which removes the bare
 from __future__ import annotations
 
 import asyncio
+import copy
 import json
 import logging
 import shlex
+import shutil
 import tempfile
 from collections.abc import Mapping, Sequence
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING, Any, cast
 
 from nemo_evaluator_sdk.agent_eval.runtimes.fabric import _common
 from nemo_evaluator_sdk.agent_eval.runtimes.fabric.image import ensure_fabric_image
+from nemo_evaluator_sdk.agent_eval.runtimes.fabric.skills import (
+    CODEX_SKILLS_DIR,
+    SKILL_MODE_CODEX_SKILLS_DIR,
+    AgentSkill,
+    SkillInjectionError,
+    SkillMode,
+    SkillProvenance,
+    SkillSet,
+    resolve_skill_mode,
+    stage_skills_seed,
+)
 from nemo_evaluator_sdk.agent_eval.runtimes.sandbox.api import AsyncSandbox
 from nemo_evaluator_sdk.agent_eval.runtimes.sandbox.base import SandboxExecResult, SandboxProvider, SandboxSpec
 from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig, AgentEvalTask
@@ -67,6 +80,11 @@ if TYPE_CHECKING:
 # AgentEvalTask); until then it is an internal default rather than a runtime-construction knob.
 DEFAULT_FABRIC_TIMEOUT_S = 600
 _RUNTIME_NAME = "fabric_container"
+_MISSING_FABRIC_MSG = (
+    "FabricContainerRuntime skill injection requires the `nemo-fabric` package (native NeMo Fabric SDK) "
+    "on the host to resolve how a skill reaches the selected adapter; the container otherwise runs Fabric "
+    "only inside the sandbox."
+)
 
 # Fixed in-container layout. The runtime seeds ``/in`` (agent config, profiles, input), execs Fabric's
 # CLI, and reads the produced ``/out`` subtree back across the boundary.
@@ -81,6 +99,13 @@ _FABRIC_STDERR = f"{_LOGS_DIR}/fabric-stderr.txt"
 _AGENT_PATH = f"{_IN_DIR}/agent.yaml"
 _INPUT_PATH = f"{_IN_DIR}/input.txt"
 _WORKSPACE_PROFILE_NAME = "eval_workspace"
+# In-sandbox root for a natively-injected skill bundle. It lives under ``/in`` (not ``/out``), so it is
+# never part of the downloaded ``/out`` evidence — only codex-mode skills, which must sit in the workspace
+# for the harness to self-discover them, need post-download cleanup.
+_SKILLS_DIR = f"{_IN_DIR}/skills"
+# Sentinel skill path attached only to probe Fabric's capability planner for the selected adapter's skills
+# routing (mirrors the host runtime). Never staged and need not exist on disk.
+_SKILL_PROBE_PATH = "nemo-eval-skill-capability-probe"
 
 
 class FabricContainerRuntime:
@@ -94,6 +119,7 @@ class FabricContainerRuntime:
         profiles: Sequence[FabricProfileConfig | Mapping[str, Any]] = (),
         secrets: Mapping[str, SecretRef] = {},
         image: str | None = None,
+        skills: Sequence[AgentSkill] | None = None,
     ) -> None:
         # The Fabric agent is fully described by its ``FabricConfig`` (harness + model + runtime); it is
         # consumed structurally as a mapping to cross the sandbox boundary as JSON.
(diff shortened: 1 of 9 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -109,6 +133,37 @@ def __init__(
...
         # None -> stock harness-agnostic image built on first run.
         self._image: str | None = image
+        # Optional agent skills injected per task (A/B: baseline vs. treated via ``with_skills``). How they
+        # reach the harness is resolved once per run (the adapter is constant across the taskset) in
+        # ``run_tasks``; only touched when a skill is set, so the no-skill path stays dependency-free. Names
+        # must be unique — each stages to its own ``<name>/`` bundle, so a repeat would collide.
+        self._skills: list[AgentSkill] = list(skills or [])
+        require_unique_skill_names(self._skills)
+
+    def with_skills(self, skills: Sequence[AgentSkill]) -> FabricContainerRuntime:
```
ngoncharenko, line 144: one idea to reduce some duplication and also re-use for other runtimes in the future: store skills state in a standalone class

```python
# skills.py
@dataclass(frozen=True)
class SkillSet:
    skills: tuple[AgentSkill, ...] = ()

    def __post_init__(self) -> None:
        require_unique_skill_names(self.skills)

    def with_skills(self, skills: Sequence[AgentSkill]) -> SkillSet:
        combined = (*self.skills, *skills)
        require_unique_skill_names(combined)
        return SkillSet(skills=combined)

    def with_skill(self, skill: AgentSkill) -> SkillSet:
        return self.with_skills([skill])

class XYZRuntime:
   # runtime.py / container_runtime.py
   def __init__(..., skills: Sequence[AgentSkill] | None = None) -> None:
       ...
       self._skill_set = SkillSet(skills=tuple(skills or ()))
   def with_skills(self, skills: Sequence[AgentSkill]) -> FabricAgentRuntime:  # or FabricContainerRuntime
      """Return a copy with ``skills`` added; ``self`` is not modified."""
      clone = copy.copy(self)
      clone._skill_set = self._skill_set.with_skills(skills)
      return clone
  def with_skill(self, skill: AgentSkill) -> FabricAgentRuntime:
      return self.with_skills([skill])
```

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
```diff
@@ -150,22 +205,36 @@ async def run_one(index: int, task: AgentEvalTask) -> AgentEvalTrial:
...
         out_dir = evidence_dir / "out"
         evidence_dir.mkdir(parents=True, exist_ok=True)
 
         # The whole per-task flow — framing input, seeding, exec, download, and parsing the result — is
         # guarded so any failure (bad seed, sandbox crash, unreadable result) fails only this task's
         # trial rather than aborting the gathered batch.
+        skill_provenances: list[SkillProvenance] = []
         try:
-            seed_files, profile_paths = self._seed_files(task)
+            seed_files, profile_paths, skill_provenances = self._seed_files(task, skill_mode)
```
ngoncharenko, line 236: Task files can overwrite potentially Codex injected skill files if their path matches:

skills files:
```
myskills/code-review/
└── SKILL.md
```

task files:
```json
task.inputs = {
    "instruction": "Review this change",
    "files": {
        ".agents/skills/code-review/SKILL.md": """<inline_skill>"""
```

Is it something we should guard against? Hash will also point to the injected skill, but the content will be different

Example:
```python
reserved = PurePosixPath(".agents/skills") / skill.name

for rel_path in task.inputs.get("files", {}):
    seed_path = PurePosixPath(rel_path)
    if seed_path == reserved or reserved in seed_path.parents:
        raise SkillInjectionError(
            f"task seed collides with injected skill path {reserved}"
        )
```

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
```diff
@@ -328,6 +333,120 @@ def _provenance_stage_root(provenance: SkillProvenance, workspace_dir: Path) ->
...
+    src = directory.expanduser()
+    if not (src / PRIMARY_SKILL_DOC).is_file():
+        raise SkillInjectionError(f"skill directory {str(directory)!r} has no {PRIMARY_SKILL_DOC}")
+    bundle: dict[str, str] = {}
+    for path in sorted(candidate for candidate in src.rglob("*") if candidate.is_file()):
+        rel = path.relative_to(src).as_posix()
+        try:
+            bundle[rel] = path.read_text(encoding="utf-8")
+        except UnicodeDecodeError as exc:
+            raise SkillInjectionError(
```
ngoncharenko, line 411: if we injected a binary file (image, for example, as a part of skills assets), this will error out. Should we allow passing binary assets?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/skills.py
SandyChapman: Good catch — I opened [AALGO-423](https://linear.app/nvidia/issue/AALGO-423) to track adding binary asset support to containerized skill injection. The current text-only seed path is intentional for this PR (avoids encoding complexity at the seam), but a follow-up can add binary support via base64 or a separate upload path.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: Done — implemented exactly as you sketched. `SkillSet` is now a frozen dataclass in `skills.py`; both runtimes hold `_skill_set` and delegate `with_skills`/`with_skill` to it.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/container_runtime.py
SandyChapman: Good catch. Both are valid paths into `.agents/skills/` — runtime `skills` for A/B, task `files` for skills a task always ships — so tasks can still seed their own. The conflict is only the same `<name>/` from both: the task upload lands second, so the provenance hash describes content the agent never saw. Now fails that task; different names are unaffected.

PR #845, opened 2026-07-22
fix(auth): run generic entity E2E tests as a service principal

anastasia-nesterenko:
## Summary

Updates the direct Entities API E2E tests to authenticate as an internal service principal, matching how feature services access the entity store in production.

This is a test-only behavior correction. It does **not** change runtime authorization: generic Entities API CRUD remains service-only, and authenticated user principals continue to receive `403 Forbidden`.

### Changes

- Adds a module-scoped `entity_store_sdk` fixture using `as_service_for`.
- Uses the service-authenticated SDK for entity CRUD, listing, filtering, sorting, renaming, and cleanup.
- Keeps user-facing project operations on the existing caller SDK.
- Uses the SDK's absolute base URL for the `/cluster-info` E2E request.
- Clarifies the existing Rego comments describing workspace sub-resource authorization. No Rego rule or permission behavior changes.

## Why

These E2E tests were calling the internal generic Entities API with user credentials. Because that API is intentionally service-only, the calls correctly returned `403`.

Feature services access the entity store using a `service:*` principal. The tests now use the same authentication model so they validate entity-store behavior without weakening access controls.

## Runtime impact

None. This PR does not change authorization policies, role definitions, or endpoint behavior.

## Validation

To rerun on the final PR head:

- `uv run --frozen pytest e2e/test_entities.py -v --run-e2e`
- `uv run --frozen ruff check e2e/test_entities.py`
- `git diff --check`

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for entity creation, updates, deletion, project associations, standalone entities, listing, sorting, filtering, renaming, and automatic naming.
  * Improved cluster information checks to work reliably across configured endpoint formats.
* **Documentation**
  * Clarified the distinction between internal entity-store testing and user-facing feature-service testing.
  * Documented authorization behavior for workspace sub-resources and cross-workspace queries.


services/core/auth/src/nmp/core/auth/app/policies/authz.rego
```diff
@@ -396,9 +396,13 @@ entities_workspace_object_path(base_path, method) if {
 	lower(method) in ["get", "put", "delete"]
 }
 
-# Workspace-scoped sub-resources (members, projects, entities/...) are user-facing CRUD, not
-# internal-only nested APIs. Without this, 6+ segment paths only hit nested_entities_internal_only
-# (403 for non-service users). Cross-workspace queries use workspace "-"; exclude that.
+# All workspace sub-resources bypass the blanket nested-Entities hard deny and go through the
```
mckornfield, line 399: so what was the fix, if this is all just a comment and the above is test code?

mckornfield (approved):
is the fix just a test only fix, specifically using a different principal sdk?

services/core/auth/src/nmp/core/auth/app/policies/authz.rego
anastasia-nesterenko: yes, correct

ironcommit (changes requested):
Need some clarification on what is changing in the PR

ironcommit (approved):
LGTM

PR #853, opened 2026-07-22
fix(models): make docker weights volume writable for the HF weight-puller

benmccown:
## Summary

The deployments-plugin **docker** backend created the model weights volume as a plain docker named volume, which the daemon creates **root-owned (0755)**. The HF weight-puller runs as its image's default **non-root** user (e.g. `nvs`, uid 1000), so `hf download --local-dir /model-store` failed creating `/model-store/.cache` with `PermissionError [Errno 13]`, and the deployment errored with `Prerequisite '<dep>-puller' failed`.

Docker has no `fs_group` equivalent (that is k8s-only), so the volume stayed unwritable. This blocked **all** docker vLLM/NIM deployments that pull weights from the files service.

## Fix

Docker analogue of a k8s `fsGroup`, keeping the puller **non-root** (least privilege):

- **`entities.py`** — add `DockerVolumeConfig.init_chmod` / `init_image`.
- **`backends/docker/volumes.py` + `backend.py`** — in `create_volume`, when `init_chmod` is set, run a one-shot **busybox** container that `chmod`s the freshly created volume so non-root workloads can write it. Idempotent, so re-running on a reconcile is harmless.
- **`deployments_plugin/compiler.py`** — on the **docker runtime only**, set the weights volume's docker backend config to `initChmod=0777` with the configured busybox image. On k8s the volume is unchanged and pod `securityContext`/`fs_group` continues to handle it.

busybox is already a first-class dependency of this plugin (used by the existing `lora-cache-init` container), so this introduces no new concept.

## Tests

New unit tests (full deployments-plugin unit suite: **273 passed**; models deployments_plugin suite green):
- docker backend: `test_create_volume_runs_init_chmod_container`, `test_create_volume_without_init_chmod_skips_container`
- compiler: `test_docker_weights_volume_requests_init_chmod`, `test_k8s_weights_volume_has_no_docker_init_chmod`

## Manual verification (dev pod, docker backend)

- **docker + vLLM (base):** busybox chmods the volume → puller runs as its default **non-root** user (`nvs`) and exits 0 (all files downloaded) → server READY (http probe 200) → gateway chat-completion works. Previously blocked.
- **docker + NIM (base):** same puller path succeeds → NIM selects a vLLM profile, loads weights → READY → gateway chat-completion works.

k8s vLLM/NIM (base + LoRA) were already working and are unaffected (no `initChmod` emitted on k8s).

## Notes

- docker + LoRA remains a deliberate guardrail (ERROR "LoRA serving is not supported on the docker runtime yet …") — unchanged.
- Branch name is a leftover from the exploration session; the change itself is the puller/volume fix above.

## Summary by CodeRabbit

- **New Features**
  - Added optional Docker volume initialization settings to support permission mode and an init helper image (including `initChmod` / `initImage` aliases).
  - Docker deployments now initialize permissions for newly created weighted-puller volumes during volume creation.

- **Bug Fixes**
  - Ensures permission initialization runs only for newly created volumes and is skipped when the volume already exists.
  - Kubernetes behavior for weighted-puller volume initialization remains unchanged.

- **Tests**
  - Added unit tests covering Docker init-container behavior, skip logic, and Docker-vs-Kubernetes compile-time expectations.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
@@ -170,16 +170,15 @@ class DockerVolumeConfig(BaseModel):
         default=None,
         alias="initChmod",
         description=(
-            "When set, the docker backend runs a one-shot init container to `chmod` "
-            "the freshly created (root-owned) named volume to this mode (e.g. '0777'). "
-            "This is the docker analogue of a k8s fsGroup: it lets a non-root "
-            "container (e.g. the HF weight-puller) write to the volume. No-op on k8s."
+            "When set, an init container chmods the freshly created (root-owned) "
+            "named volume to this mode (e.g. '0777') so a non-root workload (e.g. "
+            "the HF weight-puller) can write to it — the docker analogue of a k8s fsGroup."
         ),
     )
     init_image: str | None = Field(
         default=None,
         alias="initImage",
-        description="Image used for the init_chmod one-shot container (e.g. busybox).",
+        description="Image used for the init_chmod container (e.g. busybox).",
     )
 
     model_config = {"populate_by_name": True}
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
@@ -166,6 +166,23 @@ class DeploymentBackendConfig(BaseModel):
 class DockerVolumeConfig(BaseModel):
     driver: str = "local"
     mount_point: str | None = None
+    init_chmod: str | None = Field(
+        default=None,
+        alias="initChmod",
+        description=(
+            "When set, the docker backend runs a one-shot init container to `chmod` "
+            "the freshly created (root-owned) named volume to this mode (e.g. '0777'). "
+            "This is the docker analogue of a k8s fsGroup: it lets a non-root "
+            "container (e.g. the HF weight-puller) write to the volume. No-op on k8s."
+        ),
+    )
+    init_image: str | None = Field(
+        default=None,
+        alias="initImage",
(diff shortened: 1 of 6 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -166,6 +166,23 @@ class DeploymentBackendConfig(BaseModel):
...
     driver: str = "local"
     mount_point: str | None = None
+    init_chmod: str | None = Field(
+        default=None,
+        alias="initChmod",
+        description=(
+            "When set, the docker backend runs a one-shot init container to `chmod` "
+            "the freshly created (root-owned) named volume to this mode (e.g. '0777'). "
+            "This is the docker analogue of a k8s fsGroup: it lets a non-root "
+            "container (e.g. the HF weight-puller) write to the volume. No-op on k8s."
```
mckornfield: could shorten this description. esp the no-op on k8s, would make sense since this is the DockerVolumeConfig

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -166,6 +166,23 @@ class DeploymentBackendConfig(BaseModel):
...
+        ),
+    )
+    init_image: str | None = Field(
+        default=None,
+        alias="initImage",
+        description="Image used for the init_chmod one-shot container (e.g. busybox).",
```
mckornfield: lol `one-shot` seems like an agent thing, is it more just an "init" or "on start" container

mckornfield (approved):

tylersbray (approved):
+1 approving, see DM for loose ends.  Thanks!

PR #860, opened 2026-07-23
fix(safe-synthesizer): fail fast with a clear error on non-tabular input

marcusds:
## Summary

A Safe Synthesizer job whose input fileset has **nested list/dict columns** (e.g. function-calling JSONL like `xlam-function-calling-60k`) crashed deep inside the library's preflight with an opaque, unactionable error:

```
preflight.check_crash: TypeError: unhashable type: 'list'
nemo_safe_synthesizer.errors.ParameterError: Pre-flight check failed with 1 error(s)
```

Root cause: the library's `ConstantColumnCheck` runs `data[col].dropna().nunique()`, which can't hash list-valued cells. Safe Synthesizer is a **tabular** synthesizer, so nested columns are never valid input — but nothing told the user that.

## Change

Validate the loaded DataFrame in `run_config()` (the common chokepoint for both platform and local runs) **before** handing it to `SafeSynthesizer`, and raise a `ParameterError` that names the offending columns and explains the requirement:

```
Input data is not flat tabular: column(s) tools, metadata contain nested
list/dict values. Safe Synthesizer expects one scalar value per cell —
flatten these columns or serialize them to strings (e.g. JSON) before running.
```

Only `object`-dtype columns are scanned (numeric/bool are skipped), so the check is cheap.

## Notes

- Raises the library's own `ParameterError` so it surfaces the same way as other bad-input errors.
- Does **not** attempt to synthesize nested data — that's out of scope for a tabular synthesizer; this just replaces a cryptic crash with a clear, actionable message.
- The underlying `ConstantColumnCheck` crash is a latent bug in `nemo_safe_synthesizer` 0.1.7 (it should guard unhashable columns like its sibling checks do) — worth a separate upstream report.
- Unrelated to the runtime-setup dependency fix in #856.
- 3 pre-existing `ty` diagnostics in this file (lines 220/416/440) are untouched by this change and present on `main`.

## Test

`test_validate_flat_tabular_data_rejects_nested_columns` (raises naming the columns) and `test_validate_flat_tabular_data_allows_flat_columns` (flat data passes). Full `test_local_run.py` suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added early validation for tabular input data.
  * Inputs containing nested lists, dictionaries, or sets now receive a clear error identifying the affected columns and recommending flattening or serialization.
  * Flat, scalar-based tabular data continues to process normally.


plugins/nemo-safe-synthesizer/src/nemo_safe_synthesizer_plugin/tasks/safe_synthesizer/__main__.py
```diff
@@ -286,6 +287,29 @@ def _setup_classify_endpoint():
         logger.info("Configured column classification endpoint: %s", full_url)
 
 
+def _validate_flat_tabular_data(data: pd.DataFrame) -> None:
```
mckornfield, line 290: lol I saw this the other day, nice

mckornfield: though wonder if we already have this in the upstream, is this a big improvement?

mckornfield (approved):

PR #870, opened 2026-07-23
feat(models): support LoRA deployments on the docker backend

benmccown:
## Summary

The deployments-plugin **docker** backend was single-container: it rejected `init_containers` and multi-container `DeploymentConfig`s, so LoRA deployments (server + adapters sidecar + `lora-cache-init`) could not run there. The models controller fast-failed docker + LoRA with a "single-container" ERROR and told users to deploy on kubernetes.

This adds **targeted multi-container orchestration** to the docker backend so it can run the LoRA shape the models compiler already emits, then lifts the guardrail.

## Changes

- **`backends/docker/containers.py`** — replace the single-container validator with `build_docker_plan()`, which splits a `DeploymentConfig` into init containers, a primary (server) container, and sidecars. Rejects sidecars that declare host ports (they share the primary's netns).
- **`backends/docker/backend.py`**
  - `create_deployment`: run init containers to completion in order → primary (publishes host ports, owns GPUs) → sidecars that join the primary's network namespace (`network=container:<server>`) and share volumes, with no ports/GPUs of their own. Group image pulls reuse the existing `_pull_image` (NGC auth) and fall back to a locally present image (e.g. the local-only `nmp-api` adapters sidecar) instead of failing.
  - `read_status`: gate READY on sidecar health (a stopped sidecar keeps the deployment STARTING; its Always restart policy recreates it).
  - `delete_deployment`: tear down the whole labeled container group (primary + sidecars), not just the primary.
- **`backends/labels.py`** — add `companion_container_name` and a container-role label so group members are discoverable.
- **models `deployments_plugin/backend.py`** — remove the docker + LoRA fast-fail guardrail.

## Tests

New/updated unit tests (deployments-plugin: 302 passed; models + reconciler suites green; lint clean):
- `test_containers.py` — plan builder splits init/primary/sidecars; rejects sidecar ports / empty config.
- docker backend mocked — multi-container create runs init→server→sidecar with `network=container:<server>` and no sidecar ports; init non-zero fails before the server starts; delete removes the whole group; pull-fallback (local image present vs absent).
- Updated the two tests that asserted the old guardrail (docker + LoRA now creates substrate; generalized the reconciler backend-error-persistence test).

## Manual verification (dev pod, docker backend)

docker vLLM + LoRA (Qwen3-1.7B + `linear-algebra` adapter):
- deploy accepted (no guardrail) → `lora-cache-init` runs to completion → **server + `lora-adapters` sidecar** come up as a group (sidecar shares the server netns) → **READY**.
- sidecar loads `default--linear-algebra` into vLLM over shared localhost; provider `/v1/models` lists base + adapter.
- gateway chat-completion works for **both** the base model and the adapter.
- `delete` tears down the full group and frees the GPU.

## Notes

- Builds on #853 (docker weights-volume puller fix), now merged to `main`.
- Branch name is a leftover convention; the change is the docker multi-container LoRA support above.
- NIM + LoRA on docker should work through the same multi-container path but was not separately exercised in the pod.
- **Local sidecar image requirement:** the LoRA adapters sidecar runs the `nmp-api` image. On the docker backend this image must already be present on the host (built/loaded locally, e.g. `nmp-api:local`) — it is a local-only tag and not pullable from a registry. The pull step tolerates this via a local-copy fallback, but a fresh host without the image will fail at sidecar start. This is inherent to the local docker dev/test flow (k8s pulls the platform image normally).

## Review feedback addressed

Follow-up to the review on PR #870 (tylersbray non-blocking approval + CodeRabbit nits):

- `_sidecars_healthy` now derives the expected sidecar set from the `DeploymentConfig`, so a *removed* (not just exited) sidecar is detected as not-ready; single-container deployments short-circuit the `containers.list` entirely.
- Narrowed the image-pull local-fallback `except` to docker `ImageNotFound`/`NotFound` so unexpected client errors aren't masked.
- Added a known-limitation comment on the sidecar `network=container:<server>` netns coupling across a server recreate.
- Added a comment that init containers are intentionally CPU-only.
- Strengthened the docker-LoRA backend test to assert the full multi-container contract (server + adapters sidecar, sidecar has no ports/GPU, server owns port + GPU, `lora-cache-init` present).
- Added `read_status` tests for sidecar readiness gating: running sidecar → READY; stopped sidecar → STARTING; removed sidecar → STARTING.
- **Deferred (follow-up):** gating READY on the sidecar having actually *loaded* the adapter into vLLM (a sidecar readiness probe / "loaded" marker), rather than just the process running.

## Summary by CodeRabbit

* **New Features**
  * Docker deployments now support a multi-container setup: init containers, a primary server, and sidecars.
  * Sidecars run as companions to the server (shared networking; they can’t publish host ports).
  * Deployment readiness now also checks that expected sidecars are present and running.
* **Bug Fixes**
  * Image pull failures fall back to locally available images when possible; otherwise deployments fail cleanly.
  * Init failures or sidecar startup failures prevent partial deployments.
  * Deleting a deployment now removes the entire container group, not just the primary container.
  * Docker-based LoRA deployments are no longer rejected during model deployment creation.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
@@ -20,6 +20,7 @@ from nemo_deployments_plugin.backends.docker import volumes as volume_ops
 from nemo_deployments_plugin.backends.docker.config import DockerExecutorConfig
 from nemo_deployments_plugin.backends.docker.containers import (
     DeploymentConfigError,
+    build_docker_plan,
     build_port_bindings,
     build_volume_bindings,
     device_requests_for_gpus,
@@ -28,7 +29,6 @@ from nemo_deployments_plugin.backends.docker.containers import (
     merged_volume_mounts,
     parse_docker_backend_config,
     restart_policy_kwargs,
-    validate_config_for_docker,
 )
 from nemo_deployments_plugin.backends.docker.gpu import GPUAllocationError, get_shared_gpu_pool
 from nemo_deployments_plugin.backends.docker.ports import find_available_port
@@ -42,10 +42,13 @@ from nemo_deployments_plugin.backends.docker.status import (
 from nemo_deployments_plugin.backends.labels import (
     BACKOFF_LIMIT_LABEL,
     CONFIG_NAME_LABEL,
+    CONTAINER_ROLE_LABEL,
+    CONTAINER_ROLE_SERVER,
     DEPLOYMENT_NAME_LABEL,
     DEPLOYMENT_WORKSPACE_LABEL,
     MANAGED_BY_KEY,
     RESTART_POLICY_LABEL,
+    companion_container_name,
     container_name,
     deployment_identity_labels,
     deployment_key,
@@ -160,7 +163,7 @@ class DockerDeploymentBackend(DeploymentBackend):
         try:
             config = await self._load_deployment_config(workspace, config_name)
             config = await resolve_deployment_config_secrets(self._sdk, config)
-            container_spec = validate_config_for_docker(config)
+            plan = build_docker_plan(config)
         except DeploymentConfigError as exc:
             return BackendStatusUpdate(status="FAILED", status_message=str(exc))
         except SecretResolutionError as exc:
@@ -174,6 +177,7 @@ class DockerDeploymentBackend(DeploymentBackend):
             logger.exception("Failed to load deployment config %s/%s", workspace, config_name)
             return BackendStatusUpdate(status="FAILED", status_message=f"Failed to load deployment config: {exc}")
 
+        container_spec = plan.primary
         docker_cfg = parse_docker_backend_config(backend_config)
         if config.backend_config.docker is not None:
             docker_cfg = config.backend_config.docker
@@ -209,17 +213,27 @@ class DockerDeploymentBackend(DeploymentBackend):
                 )
             host_ports[port_spec.container_port] = host_port
 
+        # Pull all images in the group (init + primary + sidecars) up front.
         if self._executor_config.pull_images:
-            pull_error = await self._pull_image(
-                container_spec.image,
-                ngc_api_key=env_dict(container_spec).get("NGC_API_KEY"),
-            )
-            if pull_error is not None:
-                if gpu_ids and gpu_pool is not None:
-                    gpu_pool.release_gpu(dep_key)
-                return BackendStatusUpdate(status="FAILED", status_message=pull_error)
+            for container in [*plan.init_containers, container_spec, *plan.sidecars]:
+                pull_error = await self._pull_image(
+                    container.image,
+                    ngc_api_key=env_dict(container).get("NGC_API_KEY"),
+                )
+                if pull_error is not None:
+                    # A pull failure is only fatal if the image is not already
+                    # present locally. Locally-built/loaded images (e.g. the LoRA
+                    # adapters sidecar `nmp-api:local`) are not pullable from a
+                    # registry, so fall back to the local copy when it exists.
+                    try:
+                        await asyncio.to_thread(self._client.images.get, container.image)
+                        logger.info("Image %s not pullable but present locally; using local copy", container.image)
+                    except Exception:
+                        if gpu_ids and gpu_pool is not None:
+                            gpu_pool.release_gpu(dep_key)
+                        return BackendStatusUpdate(status="FAILED", status_message=pull_error)
 
-        all_labels = {
+        base_labels = {
             **labels,
             **config.labels,
             **deployment_identity_labels(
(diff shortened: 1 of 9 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -387,17 +541,77 @@ async def read_status(self, *, workspace: str, name: str) -> BackendStatusUpdate
...
+            logger.warning("Failed to list sidecars for %s/%s", workspace, name, exc_info=True)
+            return True, "sidecar check skipped"
+
+        for container in containers:
+            role = (container.labels or {}).get(CONTAINER_ROLE_LABEL, "")
+            if not role or role == CONTAINER_ROLE_SERVER or role.startswith("init-"):
+                continue
+            if container.status != "running":
+                return False, f"sidecar '{role}' is {container.status}"
+        return True, "sidecars running"
```
tylersbray, line 612: `_sidecars_healthy` only inspects containers that are currently present, so it can't detect a sidecar that has been *removed* (as opposed to exited). If the adapters sidecar is deleted entirely, it drops out of this list, the loop finds nothing, and the function returns `(True, "sidecars running")` — so the deployment reports **READY with no adapters sidecar running**.

The docstring says a "missing sidecar is not [healthy]", but the code can't enforce that because it never compares against the *expected* sidecar set from the config. Consider deriving the expected sidecar roles from the `DeploymentConfig` and flagging any that are absent as not-ready.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -387,17 +541,77 @@ async def read_status(self, *, workspace: str, name: str) -> BackendStatusUpdate
...
+        except Exception:
+            # If we cannot enumerate the group, do not block readiness on it.
+            logger.warning("Failed to list sidecars for %s/%s", workspace, name, exc_info=True)
+            return True, "sidecar check skipped"
+
+        for container in containers:
+            role = (container.labels or {}).get(CONTAINER_ROLE_LABEL, "")
+            if not role or role == CONTAINER_ROLE_SERVER or role.startswith("init-"):
+                continue
+            if container.status != "running":
```
tylersbray, line 605: Sidecar health is `status == "running"`, so the deployment can flip to READY the moment the adapters process starts — before it has actually registered the LoRA adapter(s) into vLLM. For LoRA that registration is the whole point, so there's a window where READY is reported but the adapter isn't yet queryable.

A sidecar readiness probe (or a "loaded" marker the sidecar/server exposes) would make READY mean what callers expect.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -230,12 +244,89 @@ async def create_deployment(
...
+            sidecar_name = companion_container_name(workspace, name, sidecar.name)
+            sidecar_run_kwargs = self._build_run_kwargs(
+                workspace=workspace,
+                config=config,
+                container=sidecar,
+                name=sidecar_name,
+                labels={**base_labels, CONTAINER_ROLE_LABEL: sidecar.name},
+                host_ports={},
+                gpu_ids=[],
+                network=f"container:{c_name}",
```
tylersbray, line 306: Sidecars join `network=container:<c_name>`. Docker's `Always` policy restarts the *same* server container in place (netns preserved), so ordinary crashes are fine. But `create_deployment`'s recreate path (remove exited primary + re-run) produces a *new* container that this `container:` reference can't follow, and nothing re-runs the sidecars afterward — the group can end up half-broken after a server recreate.

Worth at least a known-limitation comment; ideally recreate the whole group together when the primary is recreated.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -209,17 +213,27 @@ async def create_deployment(
...
+                    # present locally. Locally-built/loaded images (e.g. the LoRA
+                    # adapters sidecar `nmp-api:local`) are not pullable from a
+                    # registry, so fall back to the local copy when it exists.
+                    try:
+                        await asyncio.to_thread(self._client.images.get, container.image)
+                        logger.info("Image %s not pullable but present locally; using local copy", container.image)
+                    except Exception:
+                        if gpu_ids and gpu_pool is not None:
+                            gpu_pool.release_gpu(dep_key)
+                        return BackendStatusUpdate(status="FAILED", status_message=pull_error)
```
tylersbray, line 237: This fallback means docker LoRA only works where the sidecar image (e.g. `nmp-api:local`) has already been built/loaded locally — a fresh host without it will fail at sidecar start. That's inherent to the local dev flow, but worth calling out explicitly in the docs/PR description so users aren't surprised.

Also, `except Exception:` here is broad; narrowing to docker's `NotFound` / `ImageNotFound` would avoid masking unexpected client errors as "image present locally".

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -244,39 +335,95 @@ async def create_deployment(
...
+        except Exception:
+            logger.warning("Failed to remove stale init container %s", init_name, exc_info=True)
+
+        run_kwargs: dict[str, Any] = {
+            "image": init.image,
+            "name": init_name,
+            "detach": True,
+            "labels": {**base_labels, CONTAINER_ROLE_LABEL: f"init-{init.name}"},
+            "environment": env_dict(init),
+        }
```
tylersbray, line 411: Nit: init container run kwargs never set `device_requests`, so init containers always run without GPUs. Fine for `lora-cache-init`, but if the models compiler ever emits a GPU-needing init container it would silently run without a GPU. A short comment noting init containers are intentionally CPU-only would help.

plugins/nemo-deployments/tests/unit/backends/docker/test_backend_mocked.py
```diff
@@ -73,6 +76,184 @@ async def test_create_deployment_maps_command_to_entrypoint(
...
+    sidecar.name = companion_container_name("default", "srv", "lora-adapters")
+    mock_docker_client.containers.list.return_value = [server, sidecar]
+
+    update = await docker_backend.delete_deployment("default", "srv")
+
+    assert update.status == "SUCCEEDED"
+    server.remove.assert_called_once()
+    sidecar.remove.assert_called_once()
+
+
```
tylersbray, line 256: Nit / test gaps: the multi-container tests cover the happy path and delete well, but a few resilience paths added in this PR are untested:

- a *removed* sidecar still reporting READY via `_sidecars_healthy`,
- the server-recreate / netns break for sidecars,
- a `read_status` case where a present sidecar is `stopped` and READY correctly drops back to STARTING (exercises the new readiness gating).

Adding at least the last one would lock in the new readiness-gating behavior.

tylersbray (approved):
Non-blocking approval, thanks for tackling this.  The comments in the main body of the code are really helpful.  GPU accounting looks good.  New test coverage looks good despite the nit for one more and I appreciate the lora fixture.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
benmccown: Good catch — fixed in 97fd6a8. `_sidecars_healthy` now derives the expected sidecar set from the `DeploymentConfig` and flags any expected role that is not present-and-running (so a fully removed sidecar drops the deployment out of READY, not just an exited one). Added `read_status` tests for the removed and stopped cases.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
benmccown: Deferring this one to a follow-up. Agreed that container-running != adapter-registered-into-vLLM, so there is a window where READY is reported before the adapter is queryable. Doing it right needs the sidecar (or server) to expose a "loaded" signal we can gate READY on; tracking it separately rather than in this PR. Noted in the PR description.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
benmccown: Added a known-limitation comment here (97fd6a8). In-place Always restarts preserve the netns so ordinary crashes are fine; a full recreate goes through `delete_model_deployment` (whole-group teardown) then re-create, so sidecars are re-run together. A partial/primary-only recreate would need a whole-group recreate — called out in the comment for follow-up.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
benmccown: Fixed in 97fd6a8: narrowed the `images.get` fallback to `ImageNotFound`/`NotFound` so unexpected client errors surface instead of being masked as "present locally". Also documented the local sidecar-image (`nmp-api:local`) requirement explicitly in the PR description.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
benmccown: Added the clarifying comment (97fd6a8): init containers are intentionally CPU-only; if the compiler ever emits a GPU-needing init container we would plumb `device_requests` through here.

plugins/nemo-deployments/tests/unit/backends/docker/test_backend_mocked.py
benmccown: Added `read_status` tests in 97fd6a8 covering the readiness gating: stopped sidecar → STARTING and removed sidecar → STARTING (plus the running → READY baseline). The server-recreate/netns break is captured as a known-limitation comment rather than a test for now.

PR #877, opened 2026-07-23
ci: stage nightly wheels without publishing

svvarom:
Nightly wheels should stop after staging for now. This sets the publishing flag to `false`, skips public-index polling, and reports wheel staging as dispatched. Stable wheels, containers, and Helm are unchanged.

Validation:
- `actionlint`
- `act`: wheel-only nightly sends `false`, skips public wheel polling, and reports wheel staging
- focused pre-commit hooks


crookedstorm (approved):
This looks right. I presume you tested?

svvarom:
> This looks right. I presume you tested?

yep tested it locally

PR #886, opened 2026-07-24
fix(docker): align Python base builds and auth policy packaging

svvarom:
CI built the current Python 3.13 base while Deploy could select an older Python 3.11 base. That caused `policy.wasm` to be installed into the wrong environment and auth crashed at startup.

This PR makes both build modes use the same Python base recipe, packages `policy.wasm` before `uv sync`, and validates it with the installed venv.

Testing:
- policy packaging tests: 4 passed
- normal and `USE_PREBUILT_BASES=1` both resolve `nmp-python-base-builder`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,6 +111,10 @@ jobs:
       - name: Print bake groups
         shell: bash
         run: |
+          default_python_base_target="$(docker buildx bake --print docker-cpu | jq -r '.target["nmp-python-base"].target')"
+          prebuilt_python_base_target="$(USE_PREBUILT_BASES=1 docker buildx bake --print docker-cpu | jq -r '.target["nmp-python-base"].target')"
+          test "${default_python_base_target}" = "nmp-python-base-builder"
+          test "${prebuilt_python_base_target}" = "${default_python_base_target}"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -111,6 +111,10 @@ jobs:
       - name: Print bake groups
         shell: bash
         run: |
+          default_python_base_target="$(docker buildx bake --print docker-cpu | jq -r '.target["nmp-python-base"].target')"
+          prebuilt_python_base_target="$(USE_PREBUILT_BASES=1 docker buildx bake --print docker-cpu | jq -r '.target["nmp-python-base"].target')"
+          test "${default_python_base_target}" = "nmp-python-base-builder"
+          test "${prebuilt_python_base_target}" = "${default_python_base_target}"
```
crookedstorm: This new bit here is a strangely specific regression test. I think it probably should be removed.

crookedstorm (approved):
Approved but for one comment in the ci.yaml

.github/workflows/ci.yaml
svvarom: Removed the four workflow-specific assertions. The Python base build-path fix remains in docker-bake.hcl.

PR #889, opened 2026-07-24
feat(nemo-agents): add Fabric MCP calculator example

AnuradhaKaruppiah:
## Summary

- add a standalone `fabric-calculator-agent` example whose only runtime code is a toy stdio MCP server under `mcps/`
- provide two complete Platform-owned configs under `configs/`: `agent.yaml` without MCP and `agent_with_mcp.yaml` with the calculator MCP server
- enable Relay ATIF and ATOF telemetry and document how to compare the runs and verify calculator tool calls
- store generated runtime state and telemetry under the example's ignored `.tmp/` directory
- rely on adapter descriptors from the `0.1.0a20260724` Fabric release rather than bundling descriptors

## Dependency

This branch includes and depends on #890 for the shared `skills`, `mcp`, and `tools` AgentConfig-to-FabricConfig translation. The example reuses #890's MCP contract (`transport` plus `url`) instead of adding a parallel implementation. Once #890 merges, this branch can be rebased so this PR contains only the example diff.

## Why

The Fabric invocation path added in #865 needs a small end-to-end example that demonstrates optional MCP configuration without mixing Fabric installation and runtime instructions into the existing NVIDIA NeMo Agent Toolkit calculator example.

Separate full configs keep the comparison explicit without introducing a profile concept: users run one config without tools, run the second with the calculator MCP server, and inspect telemetry to confirm whether the tool was called.

## Validation

- `46 passed` across the focused agent config, CLI, Fabric translation, invocation, and calculator example tests
- verified a real MCP stdio initialize/list-tools/call-tool exchange; `multiply([12, 8])` returns `96.0`
- Ruff lint and formatting checks passed for all changed Python files
- `uv lock --check` passed
- verified both live configs start the released Codex adapter and NeMo Relay 0.6.0 with the expected adapter plan; the authenticated Codex account returned `Quota exceeded` before the model turn, so the model-driven tool call could not complete in this environment

The README includes the commands to rerun both variants and inspect `steps[].tool_calls[]` in the generated ATIF trajectory.


AnuradhaKaruppiah:
This PR was created for discussion-only. It has served that purpose so closing it

PR #895, opened 2026-07-24
fix(inference-gateway): preserve backend authentication errors

anastasia-nesterenko:
## Summary

Preserve backend `401 Unauthorized` and `403 Forbidden` responses in the inference gateway instead of wrapping them in `502 Bad Gateway`.

Backend authentication failures now include:

~~~http
X-NeMo-Error-Source: model-provider
~~~

This lets the CLI, SDK, and API consumers distinguish ModelProvider authentication failures from authentication or authorization failures produced by NeMo Platform itself.

The CLI uses this provenance header to display provider-specific guidance instead of suggesting `nemo auth login`.

## Behavior

| Failure source | Status | Provenance header | CLI guidance |
|---|---:|---|---|
| NeMo Platform authentication | `401` | Absent | Authenticate with `nemo auth login` |
| NeMo Platform authorization | `403` | Absent | Check platform roles and permissions |
| ModelProvider authentication | `401` | `X-NeMo-Error-Source: model-provider` | Check the provider API-key secret and authentication configuration |
| ModelProvider authorization | `403` | `X-NeMo-Error-Source: model-provider` | Check credential permissions and model access |
| ModelProvider model/path not found | `404` | N/A | Continues to be mapped to `502 Bad Gateway` |

### Inference gateway

- Propagate backend `401` and `403` status codes.
- Preserve the complete backend response body without JSON parsing, rewriting, or truncation.
- Preserve upstream response metadata, including:
  - `Content-Type`
  - Repeated `WWW-Authenticate` challenges
  - Repeated `Set-Cookie` headers
- Add `X-NeMo-Error-Source: model-provider` to ModelProvider `401` and `403` responses.
- Strip stale body-framing headers and recalculate `Content-Length` from the buffered body:
  - `Content-Length`
  - `Content-Encoding`
  - `Transfer-Encoding`
- Preserve the authentication status if reading the upstream response body fails.
- Bypass model rewriting and response/post-response middleware for authentication failures.
- Apply the same behavior to real and mock ModelProviders.
- Cover direct provider, model-entity, and OpenAI-compatible routes.
- Keep backend `404` responses mapped to `502`.
- Consolidate backend-error handling to keep proxy paths consistent.

### CLI

- Detect ModelProvider failures using `X-NeMo-Error-Source`.
- Display distinct messages for:
  - ModelProvider `401` authentication failures
  - ModelProvider `403` permission failures
- Avoid suggesting `nemo auth login` when NeMo Platform authentication succeeded and the configured backend credentials were rejected.
- Apply the behavior to both the source CLI and vendored SDK CLI implementation.

## Example

A backend response such as:

~~~http
HTTP/1.1 401 Unauthorized
Content-Type: application/json
WWW-Authenticate: Bearer realm="provider"
X-NeMo-Error-Source: model-provider

{"error":{"message":"Invalid API key"}}
~~~

is returned to the caller with the original status, body, and authentication challenge.

The CLI reports:

~~~text
Model provider authentication error: (401) Invalid API key
Hint: The ModelProvider rejected its configured credentials.
Check the provider's API key secret and authentication configuration.
~~~

## Test coverage

Added coverage for:

- Backend `401` and `403` responses.
- Platform-generated `401` and `403` responses without the provenance header.
- Provider, model-entity, OpenAI-compatible, and virtual-model proxy paths.
- Real and mock ModelProvider responses.
- Router-level explicit mock-response fast paths.
- JSON, HTML, empty, and large response bodies.
- Failed or truncated upstream body reads.
- Repeated `WWW-Authenticate` and `Set-Cookie` headers.
- Incorrect mock response framing headers.
- JSON and streaming mock responses.
- Response and post-response middleware bypass.
- Provider-specific CLI messages.
- Existing non-authentication error behavior.

## Verification

- Full inference-gateway suite: `572 passed, 5 skipped`
- Focused proxy and router suite: `198 passed`
- Auth-enabled integration suite: `35 passed`
- CLI error-handling suite: `43 passed`
- Ruff lint and formatting: passed
- `git diff --check`: passed

## Linear

[AIRCORE-345 — Inference gateway wraps backend 401 Unauthorized in 502](https://linear.app/nvidia/issue/AIRCORE-345/inference-gateway-wraps-backend-401-unauthorized-in-502-preventing)

### DIFF
--- a/services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
+++ b/services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
@@ -7,7 +7,7 @@ import logging
 import uuid
 from collections.abc import AsyncIterable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, AsyncIterator, Union
+from typing import TYPE_CHECKING, Any, AsyncIterator, NoReturn, Union
 
 import aiohttp
 from aiohttp import ClientSession
@@ -309,6 +309,8 @@ def _close_response(response: aiohttp.ClientResponse | None):
 _MAX_ERROR_BODY_LEN = 2048
 _BODY_FRAMING_HEADERS_TO_STRIP = frozenset({"content-length", "content-encoding", "transfer-encoding"})
 _BACKEND_AUTH_ERROR_STATUSES = frozenset({401, 403})
+_ERROR_SOURCE_HEADER = "X-NeMo-Error-Source"
+_MODEL_PROVIDER_ERROR_SOURCE = "model-provider"
 
 
 async def _read_error_body(response: aiohttp.ClientResponse) -> tuple[bytes, str]:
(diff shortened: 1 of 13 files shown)


services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
```diff
@@ -354,24 +367,31 @@ async def proxy_request(http_client: ClientSession, next_request_info: NextReque
         )
 
         if response.status >= 400:
-            error_body = await _read_error_body(response)
+            response_status = response.status
+            response_headers = (
+                _filter_response_headers(response.headers)
+                if response_status in _BACKEND_AUTH_ERROR_STATUSES
+                else CIMultiDict()
```
mckornfield: confused, if we're NOT in error states, don't we still want the headers?

services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
```diff
@@ -492,24 +513,31 @@ async def fetch_proxy_response(
...
```
mckornfield: same q as above

benmccown (changes requested):
I like your fix with the header you added and the human readable CLI response, but I'm wondering if we could take it a step further and mutate the error response body directly. Most times, folks would be using a 3rd party inference library (like the openai sdk) so they won't get the helpful error message, and things could still be confusing.

Including a claude response below with some details on the rough idea.

---

**Yes, mutating the body would work.** The OpenAI client builds the error message directly from the response body in `_make_status_error_from_response` (`_base_client.py:407`):

```python
err_text = response.text.strip()
body = err_text
try:
    body = json.loads(err_text)
    err_msg = f"Error code: {response.status_code} - {body}"
except Exception:
    err_msg = err_text or f"Error code: {response.status_code}"
```

So if the response body is valid JSON, the user sees:
```
Error code: 401 - {'error': {'message': 'Authentication failed'}, ...}
```
i.e. the full dict repr is included in the exception message string. **Adding a field to the JSON body would appear in that message.**

If the body is plain text, the text itself becomes the message. Either way, the body content is surfaced — the client does **not** display a static message based solely on status code.

**Practical implication**: the gateway could mutate the backend's error body to prepend or inject a hint, e.g.:
```json
{"error": {"message": "Model provider authentication failed (backend rejected its API key). Original: ...", ...}}
```
and it would show up verbatim in the `openai.AuthenticationError` message the user sees. This is simpler and more universally visible than the `X-NeMo-Error-Source` header approach, which requires users to know to inspect `error.response.headers`.

anastasia-nesterenko:
closed in preference of https://github.com/NVIDIA-NeMo/nemo-platform/pull/2158

PR #899, opened 2026-07-24
fix(agents): make deployed agent inference URL container-reachable

benmccown:
## TL;DR

Deployed agents were calling the wrong URL for inference. Fixed by only ever handing an agent a URL we know is reachable from inside its container.

## The bug

`inject_gateway_url()` bakes the **API pod's own base URL** into each agent's `llms.*.base_url` at create time. That URL is wrong from inside an agent container:

- **k8s + embedded-PDP auth:** the API pod's `NMP_BASE_URL` is `http://localhost:8080` (correct for its own PDP self-call). Baked into an agent, `localhost` is the agent itself → calls never reach the platform.
- **Envoy front door:** if the base URL pointed at the Envoy auth proxy, agents looped forever on 503s (OpenAI client retry) and took down Envoy + Studio.

The one mode-aware rewrite that existed only wrote `NMP_GATEWAY_BASE_URL` — a **dead env var nothing reads** — so it never reached the config the agent runs.

## The fix

Enforce one invariant: **an agent is only ever handed a known-good, container-reachable URL — never the raw platform base URL.**

- `resolve_agent_gateway_url()` emits only known-good values per mode:
  - **k8s** → the in-cluster API Service DNS (`NMP_INTERNAL_BASE_URL`, set by Helm), or an explicit override. No internal URL → **fail fast** instead of propagating something unverified.
  - **docker** → loopback rewritten to `host.docker.internal`; other hosts pass through.
- Rebases the agent's `llms.*.base_url` onto that URL (preserving the IGW path).
- Adds `NEMO_INTERNAL_BASE_URL` / `NMP_INTERNAL_BASE_URL` (and `agents.deployments.k8s_internal_base_url`); Helm sets it on the api + controller pods.
- Deletes the dead `NMP_GATEWAY_BASE_URL`.

## Verify

`helm template --set platformConfig.auth.enabled=true` on the API pod:
```
NMP_BASE_URL:          http://localhost:8080                 # loopback — was leaking into agents
NMP_INTERNAL_BASE_URL: http://<release>-nemo-platform-api:8080  # correct target now used
```

ruff + ty + `make docs-check` clean; unit tests for docker/k8s resolution, the fail-fast path, and config rebasing all pass.

## Notes

- Native-Linux docker may still need `gateway_url_override` (documented).
- Pre-existing unrelated test failures in the nemo-agents suite (missing `nemo_fabric`, CLI table assertions) reproduce on `main`.

## Summary by CodeRabbit

- **New Features**
  - Improved deployed-agent connectivity across Docker and Kubernetes environments.
  - Kubernetes deployments now use an in-cluster API address when available.
  - Added automatic inference endpoint adjustment for container-reachable URLs.
  - Deployments now fail with a clear error when Kubernetes connectivity cannot be determined.

- **Documentation**
  - Expanded guidance for Docker and Linux networking, including configuration examples and troubleshooting steps.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/config.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/config.py
@@ -77,9 +77,17 @@ class DeploymentsRunnerConfig(BaseModel):
     gateway_url_override: str | None = Field(
         default=None,
         description=(
-            "Optional container-reachable platform base URL. When unset, docker mode rewrites "
-            "loopback hosts to host.docker.internal; k8s mode leaves the host base URL as-is "
-            "(in-cluster IGW DNS is AIRCORE-863)."
+            "Platform base URL baked into deployed agents as their inference endpoint, used verbatim "
+            "for both docker and k8s. When unset, the deploy path derives a container-reachable URL: "
+            "docker rewrites loopback hosts to host.docker.internal; k8s uses k8s_internal_base_url."
+        ),
+    )
+    k8s_internal_base_url: str | None = Field(
+        default=None,
+        description=(
+            "In-cluster platform base URL (the API Service DNS, e.g. http://<release>-nmp-api:8080) "
+            "used as the inference endpoint for k8s-mode agents. Set automatically by the Helm chart. "
+            "Read from NEMO_INTERNAL_BASE_URL, then NMP_INTERNAL_BASE_URL, when unset."
         ),
     )
     plugin_wheels_init_image: str | None = Field(
--- a/plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
@@ -14,9 +14,11 @@ AgentRun (Razvan RFCs), not AgentDeployment.
 from __future__ import annotations
 
 import asyncio
+import copy
 import logging
 import time
 from typing import Any
+from urllib.parse import urlsplit
 
 import yaml
 from nemo_agents_plugin.config import AgentsConfig, DeploymentsRunnerConfig
@@ -27,7 +29,7 @@ from nemo_agents_plugin.entities import (
     Endpoint,
 )
 from nemo_agents_plugin.runner.backend import DeploymentInfo, ExternalLog, LogLocation, RunnerBackend
-from nemo_agents_plugin.utils import get_base_url
+from nemo_agents_plugin.utils import get_base_url, get_internal_base_url
 from nemo_deployments_plugin.entities import (
     ConfigFile,
     Container,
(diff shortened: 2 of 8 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -79,21 +81,77 @@ def map_status(backend_status: str) -> DeploymentStatus:
...
+    explicit third-party ``base_url`` are left unchanged.
+    """
+    reachable = urlsplit(gateway_url.rstrip("/"))
+    reachable_origin = f"{reachable.scheme}://{reachable.netloc}"
+    config = copy.deepcopy(nat_config)
+    for llm_cfg in config.get("llms", {}).values():
+        if not isinstance(llm_cfg, dict) or llm_cfg.get("_type") not in ("openai", "nim"):
+            continue
+        current = llm_cfg.get("base_url")
+        if not isinstance(current, str) or "/apis/inference-gateway/" not in current:
```
mckornfield, line 150: wonder if there's a case where this is a false positive

plugins/nemo-agents/src/nemo_agents_plugin/config.py
```diff
@@ -77,9 +77,17 @@ class DeploymentsRunnerConfig(BaseModel):
     gateway_url_override: str | None = Field(
         default=None,
         description=(
-            "Optional container-reachable platform base URL. When unset, docker mode rewrites "
-            "loopback hosts to host.docker.internal; k8s mode leaves the host base URL as-is "
-            "(in-cluster IGW DNS is AIRCORE-863)."
+            "Platform base URL baked into deployed agents as their inference endpoint, used verbatim "
+            "for both docker and k8s. When unset, the deploy path derives a container-reachable URL: "
+            "docker rewrites loopback hosts to host.docker.internal; k8s uses k8s_internal_base_url."
```
mckornfield, line 82: might be good to make clear that `k8s_internal_base_url` is a config value, not a literal string (maybe putting the host.docker.internal in quotes)

mckornfield (approved):

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -79,21 +81,77 @@ def map_status(backend_status: str) -> DeploymentStatus:
...
+        if internal_base_url:
+            return internal_base_url.rstrip("/")
+        raise UnreachableGatewayURLError(
+            f"No container-reachable inference base URL for k8s deployment: platform base URL "
+            f"{base_url!r} is not usable from an agent pod and no internal API Service URL is set. "
+            "Set NEMO_INTERNAL_BASE_URL / NMP_INTERNAL_BASE_URL (or deployments.k8s_internal_base_url), "
+            "or deployments.gateway_url_override."
+        )
+
+    parts = urlsplit(base_url.rstrip("/"))
```
tylersbray, line 127: Kind of dumb question, but is this the implicit `if mode == "docker":` branch?  Can it be explicit just 'cuz?

tylersbray (approved):

tylersbray:
Reviewed and satisfied with the fix — approving. Root cause and the "only ever hand a container-reachable URL" invariant look right, fail-fast + no orphaned config is clean, and the `[::1]` case is handled via `urlsplit`.

Not blockers for the release, just some adjacent edge cases worth tracking as follow-ups:

1. **Reachability vs. authorization.** This bug only manifested with auth enabled, and this PR fixes *connectivity* only. The deployed agent's LLM `api_key` is `"not-used"` and `build_deployment_config` injects no principal/bearer token into the agent container. Worth confirming end-to-end that a deployed agent's IGW call via the internal Service DNS is actually *authorized* (200), not just reachable, on k8s + embedded-PDP auth — i.e. that the internal DNS path bypasses the front-door PDP. That's the exact gap the e2e tests missed.

2. **Only `llms` are rewritten.** Both `inject_gateway_url` and `rewrite_config_base_urls` touch only `config["llms"]` with `_type in (openai, nim)`. An `embedders`/`retrievers` block manually pointed at a loopback IGW URL would keep `localhost` and be unreachable — same class of bug, out of scope here but worth a follow-up issue.

3. **`gateway_url_override` is origin-only.** `rewrite_config_base_urls` uses only the resolved gateway's scheme+netloc and re-applies each LLM's own path, so an override that includes a path prefix (e.g. a reverse proxy at `/nemo`) would be silently dropped. Minor, but worth a one-line note on the config field.

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
benmccown: Correct. Can defs make it explicit. Will pull this into a follow up PR.

plugins/nemo-agents/src/nemo_agents_plugin/config.py
benmccown: Will add quotes in follow up PR, just to get this merged.

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
benmccown: Will address this in a follow up to see if we can include host or something, to make it more explicit

PR #904, opened 2026-07-27
feat(studio): Anonymizer Rewrite strategy params [ASTD-334]

marcusds:
Closes [ASTD-334](https://linear.app/nvidia/issue/ASTD-334/studio-anonymizer-rewrite-strategy-params).

Unlocks **Rewrite** in the anonymization strategy select and adds its parameter UI, the last of the five strategies. Follows [#898](https://github.com/NVIDIA-NeMo/nemo-platform/pull/898) (ASTD-333).

## What's here

New `RewriteParamsSection`, rendered below the strategy dropdown when Rewrite is selected:

| Field | Control | Maps to |
|---|---|---|
| Privacy Goal | Segmented (Default / Custom) | — |
| Protect / Preserve | Textareas, shown only in Custom | `rewrite.privacy_goal` |
| LLM Instructions | Textarea | `rewrite.instructions` |
| Risk Tolerance | Slider, Minimal → High | `rewrite.risk_tolerance` |
| Max Repair Rounds | Number, min 0 | `rewrite.max_repair_iterations` |
| Strict Entity Protection | Checkbox | `rewrite.strict_entity_protection` |

Default mode omits `privacy_goal` entirely so the server auto-populates it, rather than sending empty strings that would 422. Per the ASTD-333 precedent there are no client-side length rules on protect/preserve — the backend's 422s map to inline field errors, and `apiErrors.ts` now covers the rewrite locs.

## Notable decisions

**Risk Tolerance uses the raw KUI `Slider`, not `ControlledSlider.`** `ControlledSlider` writes a number into the form field unconditionally, which would clobber the enum. Raw `Slider` + `useController` keeps `riskTolerance` as `'minimal' | 'low' | 'moderate' | 'high'` in form state with the index mapping local to the component; `stepFormatFn` renders the tick labels.

**`RISK_TOLERANCE_LABELS` is `Record<RiskTolerance, string>`** — exhaustive against the SDK enum, so a new preset fails the build rather than silently rendering without a label.

## Fix: rewrite jobs must map the replacement generator

Rewrite jobs were failing preflight with:

```
InvalidConfigError: Selected model aliases not found in model pool:
replace.replacement_generator='gpt-oss-120b'. Known aliases: ['model-1']
```

Rewrite reuses the replacement generator internally, so the library validates that role for rewrite too:

```python
# anonymizer/interface/anonymizer.py:648
check_substitute=isinstance(config.replace, Substitute) or config.rewrite is not None,
```

We were sending `selected_models` with only `detection` + `rewrite`, so the server kept its own default (`default_model_configs/replace.yaml` → `gpt-oss-120b`), which isn't in the pool we send. `activeRolesForStrategy('rewrite')` now includes `REPLACE_ROLE` and the builder emits `selected_models.replace`. Verified by calling `validate_model_alias_references` directly with both payload shapes — old fails with the exact error above, new passes.

## Drive-by fixes

Both pre-existing from ASTD-327, surfaced while testing this form:

- **Select triggers rendered raw values** (`rewrite`, `dataset`, `sha256`). KUI's `SelectEntry` is `{ children, value }` — `label` isn't a prop, so it was dropped and the trigger fell back to the value. Fixed across the three option lists in this route. Typecheck didn't catch it because the extra key was structurally compatible.
- **Cancel / Full Run floated over the form on scroll.** `overflow-auto` was on the Panel root, but `.nv-panel-root` is the flex column holding the footer, so the footer scrolled with the content. Moved the scroll to the content slot (`flex-1 min-h-0 overflow-auto`).

## Screenshots

01-rewrite-default
02-rewrite-custom-privacy-goal
04-footer-pinned-on-scroll

## Testing

- 21 unit tests pass (5 new: rewrite config shape, privacy-goal modes, instruction trimming, param carry-through, replacement-generator mapping)
- typecheck, eslint, prettier clean
- Verified in a running Studio against a live platform: form renders, Custom reveals protect/preserve, Model Settings resolves all rewrite roles including the new Replacement Generator, footer stays pinned through scroll

## Out of scope

[ASTD-336](https://linear.app/nvidia/issue/ASTD-336/studio-anonymizer-entity-detector-silently-falls-back-to-an-llm-when) filed separately: the Entity Detector role falls back to a general LLM when the workspace has no GLiNER model, and the library injects GLiNER-only params into whatever that role points at, so jobs die at detection with a 400. Pre-existing and affects all five strategies.

## Summary by CodeRabbit

- **New Features**
  - Added rewrite strategy support with privacy goals, instructions, risk tolerance, repair rounds, and strict entity protection settings.
  - Added a two-tab anonymizer builder with Source and Model Settings sections.
  - Improved automatic model selection for required anonymizer roles.
  - Rewrite configurations are now included when submitting jobs.

- **Bug Fixes**
  - Rewrite-related API validation errors now map to the appropriate form fields.
  - Improved selected-model handling for rewrite and substitute strategies.


web/packages/studio/src/routes/AnonymizerBuilderRoute/components/RoleModelDefaults.tsx
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useDefaultRoleModels } from '@studio/routes/AnonymizerBuilderRoute/useDefaultRoleModels';
+import { FC } from 'react';
+
+/** Seeds role model defaults from inside the form provider, independent of the active tab. */
+export const RoleModelDefaults: FC = () => {
+  useDefaultRoleModels();
+  return null;
+};
```
steramae-nvidia: Instead of this shape, you could extract the `<form>` under `<FormProvider>` and rename this AnonymizerBuilderForm? Taking the UI out of the Route component

web/packages/studio/src/routes/AnonymizerBuilderRoute/index.tsx
```diff
@@ -45,6 +46,8 @@ const PANEL_TABS = [
   { value: TAB_MODEL_SETTINGS, children: 'Model Settings' },
 ];
 
+const PANEL_ATTRIBUTES = { PanelContent: { className: 'flex-1 min-h-0 overflow-auto' } };
```
nakolean: **Question**: Does this really need to be moved into a const?

web/packages/studio/src/routes/AnonymizerBuilderRoute/components/RoleModelDefaults.tsx
marcusds: Yeah definitely, my fault for not claude being silly on this.

nakolean (approved):

PR #909, opened 2026-07-27
feat(nemo-agents): support for nemo agents run and nemo agents deploy for Fabric backed agents

mmogallapalli:
## Summary

This PR implements the first Platform-managed serving lifecycle for Fabric-backed NeMo Agents as part of AIRCORE-932.

It adds a local FastAPI serving process that creates one Fabric runtime per logical session. The same server can be started directly through `nemo agents run` for foreground development or through `nemo agents deploy` for Platform-managed serving:

```text
Platform-owned agent.yaml
  -> nemo agents run
  -> foreground Fabric serving process

Platform Agent entity
  -> nemo agents deploy
  -> managed Fabric serving process

Either Fabric serving path
  -> logical session
  -> translated FabricConfig
  -> FabricRuntime
  -> ordered invoke calls
  -> runtime stop
```

Fabric continues to own harness execution and the runtime `start` / `invoke` / `stop` lifecycle. NeMo Platform owns the multi-user server, logical session identity, runtime registry, request routing, concurrency policy, expiration, and cleanup.

## Changes

- Added an OpenAI-compatible Fabric serving application with:
  - `GET /health`
  - `POST /v1/chat/completions`
  - `DELETE /v1/sessions/{session_id}`
- Updated `nemo agents run` to dispatch by agent config format:
  - `nemo-agents-spec-v1` starts the Platform-owned Fabric server.
  - `nat-workflow-v1` continues to run `nat start fastapi`.
  - Unsupported formats fail before starting a server.
- Added typed request and response models for the chat-completions boundary.
- Added a runtime session registry that maps opaque Platform session IDs to active Fabric runtimes.
- Added a session manager responsible for:
  - Lazy `FabricConfig` translation and runtime startup.
  - Reusing the same runtime for later turns in a logical session.
  - Serializing invocations within one session.
  - Limiting concurrent invocations across independent sessions.
  - Explicit session closure.
  - Idle-session expiration.
  - Draining and stopping all runtimes during server shutdown.
- Added invocation support for an already-active Fabric runtime while retaining the existing one-shot invocation path.
- Added shared local-environment preparation so configured workspaces exist before either one-shot or managed runtime startup.
- Added a shared agent-config format registry/protocol used by agent creation and deployment config resolution.
- Updated the in-memory runner to deploy `nemo-agents-spec-v1` agents by:
  - Persisting a canonical `agent.yaml` under the Platform agent workspace.
  - Running Fabric plan/doctor validation before spawning the server.
  - Starting the Fabric server as a managed local subprocess.
  - Reusing existing port allocation, readiness polling, log handling, process termination, and deployment cleanup behavior.
- Preserved the existing NAT deployment path for `nat-workflow-v1`.
- Added focused tests for format handling, HTTP routing, session lifecycle, concurrency, expiration, runtime cleanup, local deployment, and failure paths.

### Platform owns sessions; Fabric owns runtimes

A Platform session is a logical conversation rather than an HTTP connection. Platform maintains the `session_id -> FabricRuntime` mapping, while Fabric remains unaware of users, HTTP routing, and other runtimes.

This keeps the integration aligned with Fabric's public lifecycle contract instead of adding lifecycle behavior inside adapters.

### Runtimes are created lazily

Server startup loads and validates the reusable Platform agent definition but does not create a Fabric runtime. A complete `FabricConfig` is translated when the first request opens a logical session, and that config is bound to the resulting runtime.

This avoids allocating harness resources for sessions that never invoke the agent and leaves room for future per-session policy, environment, and profile resolution.

### Session identity uses a response header

The first request may omit `X-Nemo-Session-Id`. Platform generates an opaque session ID and returns it in that response header. Later requests provide the same header to reuse the runtime.

Using a header keeps the request body compatible with the OpenAI chat-completions shape. Supplying an unknown or closed session ID returns `404`; it does not silently create a replacement runtime.

### One runtime processes one turn at a time

Invocations for the same session are serialized with a per-session lock because ordered turns share harness state. Different sessions may run concurrently, subject to a server-wide semaphore. The initial default permits eight concurrent invocations.

### The runtime owns conversation state

Each HTTP request passes the current user message to the existing runtime. Prior turns are not replayed from the HTTP payload because the Fabric runtime and selected harness adapter own the session's conversation state.

### Cleanup is explicit and bounded

Clients can close sessions explicitly. The server also expires idle sessions after 30 minutes, checks every five minutes, and drains all remaining runtimes during shutdown. Runtime registration failures also stop any runtime that was already started.

### Local deployment builds on the existing runner

The first implementation uses the existing in-memory subprocess backend rather than introducing a second process-management system. Fabric and NAT deployments therefore share port allocation, health polling, logs, status transitions, termination, and filesystem cleanup.

Docker/Kubernetes runtime placement and distributed session ownership remain separate follow-up work.

### `run` and `deploy` share the Fabric server

`nemo agents run` starts the Fabric server in the foreground from a local `agent.yaml`, with the caller choosing the host and port and stopping it with `Ctrl-C`. It does not require a running Platform or create an Agent or AgentDeployment entity.

`nemo agents deploy` starts the same server through the existing Platform runner. Platform allocates the port, tracks deployment status, routes requests through the gateway, exposes logs, and owns process cleanup.

Keeping both paths on the same server implementation provides the Fabric equivalent of the existing NAT `run`/`deploy` split without duplicating runtime or session behavior.

### Agent formats share a small internal protocol

Agent creation and deployment now resolve behavior through config-format handlers. This keeps `nat-workflow-v1` as the default and adds `nemo-agents-spec-v1` without spreading format-specific branches through the API layer.

## Error Behavior

- Unknown or closed session: `404`
- Fabric runtime startup failure: `503`
- Fabric invocation failure or failed result: `502`
- Fabric invocation timeout: `504`
- Runtime shutdown failure: `502`

Errors for an existing session preserve the session ID header where appropriate.

## Out of Scope

- Durable session recovery after a Platform/server restart.
- Distributed session registries or routing across replicas.
- Docker/Kubernetes Fabric server deployment and remote runtime placement.
- Authentication and authorization inside the Fabric serving process; those remain Platform gateway concerns.
- Streaming chat completions.
- User-facing cancellation APIs.
- Per-user concurrency quotas or configurable limits through public API fields.

## Validation

Focused branch coverage:

```text
222 passed
```

This includes agent config handling, format dispatch, Fabric translation/validation, one-shot and active-runtime invocation, serving routes, session registry/manager behavior, deployment APIs, controller behavior, and the in-memory runner.

Additional nested NeMo Agents suites:

```text
104 passed
```

Focused CLI validation, including NAT and Fabric `nemo agents run` dispatch:

```text
18 passed
```

Repository Python style and formatting:

```text
All checks passed
2793 files already formatted
```

Manual end-to-end validation:

1. Registered a `nemo-agents-spec-v1` Agent from `agent.yaml`.
2. Deployed it through the Platform API/CLI.
3. Observed `pending -> starting -> running`.
4. Invoked it through the Platform agent gateway.
5. Opened a logical session and received `X-Nemo-Session-Id`.
6. Reused that session for a second turn and confirmed conversation state was preserved.
7. Closed the session and received `204`.
8. Confirmed reuse of the closed session returned `404`.
9. Deleted the deployment and verified its process and prepared base directory were removed.

## Summary by CodeRabbit

* **New Features**
  * Added a local Fabric agent serving server with OpenAI-compatible chat completions, session reuse, idle cleanup, and per-session concurrency limits.
  * Added unified agent configuration validation and deployment-resolution across NAT and Nemo spec formats (with consistent normalization).
  * Added support for launching either Fabric or NAT servers via the local CLI based on agent config type.
* **Bug Fixes**
  * Improved error handling and HTTP status mapping for timeouts, execution failures, session issues, and response conversion issues.
  * Enforced secure local workspace creation (rejecting absolute/escaping paths) and strengthened Fabric deployment directory cleanup.


plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -426,6 +428,33 @@ def _spawn(
...
+            "--agent-config",
+            str(config_path),
+            "--host",
+            "127.0.0.1",
+            "--port",
+            str(port),
+        ]
+        log_path.parent.mkdir(parents=True, exist_ok=True)
+        logger.info("Spawning: %s  (log: %s)", " ".join(cmd), log_path)
+        log_file = log_path.open("a")
```
tylersbray: "w" instead of "a" would match NAT, plus it looks like we wipe everything between runs anyway?  Do we need "a" persistence?

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
```diff
@@ -0,0 +1,112 @@
...
+        return config
+
+    def resolve_for_deployment(
+        self,
+        config: dict[str, Any],
+        *,
+        workspace: str,
+        agent_name: str,
+    ) -> dict[str, Any]:
+        resolved = inject_gateway_url(config, workspace)
```
tylersbray, line 58: Does Fabric need any similar injection?

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
```diff
@@ -0,0 +1,148 @@
...
+    FabricSessionRegistry,
+)
+from nemo_agents_plugin.fabric.translator import FabricTranslationError, translate_agent_config
+
+# CI type-checks this plugin via ty extra-paths without installing nemo-agents deps.
+from nemo_fabric import Fabric, FabricError  # ty: ignore[unresolved-import]
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_MAX_CONCURRENT_INVOCATIONS = 8
```
tylersbray, line 27: Might want all of these as tunable config settings someday?  Not today...

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -426,6 +428,33 @@ def _spawn(
...
+        try:
+            return subprocess.Popen(cmd, stdout=log_file, stderr=subprocess.STDOUT)
+        finally:
+            log_file.close()
+
     def _terminate(self, name: str, proc: subprocess.Popen[bytes]) -> None:
```
tylersbray, line 458: AI says "Deploy delete does not call DELETE /v1/sessions/*; it SIGTERMs uvicorn and hopes lifespan drains. After 5s → SIGKILL can skip graceful runtime.stop(). Acceptable for first PR; raise as Warning/follow-up, not blocker, unless you’ve seen Fabric leave orphan harnesses."  See also server.py ~158–163 close_all_sessions()

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        """Drain the registry and stop every remaining runtime."""
+        sessions = await self._session_registry.drain()
+
+        async def stop_session(session: FabricRuntimeSession) -> None:
+            try:
+                await self._stop_session(session)
+            except FabricSessionStopError:
+                logger.exception("Failed to stop Fabric session %s during shutdown.", session.session_id)
+
+        await asyncio.gather(*(stop_session(session) for session in sessions))
```
tylersbray: Add return_exceptions=True here, you only catch FabricError in _stop_session, anything else will wreck this otherwise.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
```diff
@@ -0,0 +1,148 @@
...
+class FabricSessionManager:
+    """Create Fabric runtimes lazily from one reusable Platform agent definition."""
+
+    def __init__(
+        self,
+        agent_config: AgentConfig,
+        *,
+        base_dir: Path,
+        session_registry: FabricSessionRegistry,
+        fabric: Any | None = None,
```
tylersbray: You might be able to tighten this typing?  Or does your CI note with the import mean you can't?

tylersbray (approved):
Nice foundation, looks good.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/invocation.py
```diff
@@ -23,27 +24,17 @@ async def invoke_agent_config_once(
 ) -> list[FabricRuntimeResult]:
     """Translate a Platform agent config and run each input through Fabric once."""
     fabric_config = translate_agent_config(agent_config)
-    await asyncio.to_thread(_ensure_local_workspace_dir, agent_config, base_dir)
+    await asyncio.to_thread(ensure_local_workspace_dir, agent_config, base_dir)
```
mikeknep, line 27: why the `await asyncio.to_thread` wrapper here? Could you just call the function normally, or am I missing something? 🤔

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
```diff
@@ -0,0 +1,282 @@
...
+    *,
+    max_concurrent_invocations: int = DEFAULT_MAX_CONCURRENT_INVOCATIONS,
+    idle_session_timeout_seconds: float = DEFAULT_IDLE_SESSION_TIMEOUT_SECONDS,
+    session_cleanup_interval_seconds: float = DEFAULT_SESSION_CLEANUP_INTERVAL_SECONDS,
+) -> FastAPI:
+    """Create a serving app that validates its agent definition at startup."""
+    if idle_session_timeout_seconds <= 0:
+        raise ValueError("idle_session_timeout_seconds must be greater than zero.")
+    if session_cleanup_interval_seconds <= 0:
+        raise ValueError("session_cleanup_interval_seconds must be greater than zero.")
```
mikeknep: style nit: what if these additional kwargs were on some settings model/dataclass? Could put the validation on that object, and only need to pass one optional argument in. Can take it or leave it.

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
```diff
@@ -0,0 +1,112 @@
...
+        self,
+        config: dict[str, Any],
+        *,
+        workspace: str,
+        agent_name: str,
+    ) -> dict[str, Any]:
+        del workspace, agent_name
+        return self._normalize(config)
+
+    @staticmethod
```
mikeknep, line 78: Is there a particular reason this is a staticmethod? Doesn't appear necessary AFAICT

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
```diff
@@ -0,0 +1,112 @@
...
+
+class UnsupportedAgentConfigFormatError(AgentConfigFormatError):
+    """Raised when no handler exists for an agent config format."""
+
+
+class InvalidAgentConfigError(AgentConfigFormatError):
+    """Raised when an agent config does not satisfy its format contract."""
+
+
+class AgentConfigFormatHandler(Protocol):
```
mikeknep, line 33: 🙌

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mmogallapalli: Can be tightened, adding that as well.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mmogallapalli: added

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
mmogallapalli: tracking as a followup

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mmogallapalli: Agreed, will make them configurable as a followup

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
mmogallapalli: thanks for flagging this, adding a helper method for fabric

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
mmogallapalli: Good call, changing to "w"

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
mmogallapalli: I made this a static method because it’s used by both the validation and deployment-resolution methods and doesn’t depend on instance state, so I wanted to avoid an unused `self` parameter.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/invocation.py
mmogallapalli: I added this because `RunnerBackend` says that blocking operations should use `asyncio.to_thread`. We still need to wait for the workspace to be created before proceeding, but this keeps the synchronous filesystem operation from blocking the event loop.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
mmogallapalli: I like this suggestion make it much cleaner, added it

plugins/nemo-agents/src/nemo_agents_plugin/agent_config_formats.py
mikeknep: Fair enough

mikeknep (approved):

PR #912, opened 2026-07-27
feat(studio): Anonymizer entity picker categories and custom labels [ASTD-335]

marcusds:
Closes [ASTD-335](https://linear.app/nvidia/issue/ASTD-335/studio-anonymizer-entity-picker-categories-custom-labels-chips).

> **Depends on [#936](https://github.com/NVIDIA-NeMo/nemo-platform/pull/936) (ASTD-344)** — merge that first. It fixes `ControlledCombobox` so a multi-select can read typed input; this branch is stacked on it and the entity picker now uses the shared component instead of a hand-rolled KUI `Combobox`. Until #936 lands, the diff here includes its commit.

Upgrades the lean entity picker from ASTD-327 to the full design.

## What's here

| Before | Now |
|---|---|
| Flat list of 65 labels | Grouped under 13 categories, in design order |
| Default labels only | Type any label; offered as a **Custom label** item |
| Plain multi-select | Chips below the field, coloured by category |
| Include-defaults **hid** the picker | Both together; request merges defaults + picks |

## Notable decisions

**The category map is curated in Studio.** `/entity-labels` returns a flat array of 65 bare strings with no category field (verified against a live service), so grouping has to live client-side. The map is keyed to the exact label ids the API returns and covers all 65. Anything the API adds later that isn't mapped lands in an `Other` section rather than vanishing from the menu — `buildEntitySections` is driven by what the API returns, not by the map.

**Custom labels are a synthetic menu item.** KUI's `Combobox` has no creatable/free-solo mode, and the shared `ControlledCombobox` isn't usable here: it pins `inputValue` to `''` in multi-select, and its `freeForm` path writes the raw input *string* into what is a `string[]` field (`ControlledCombobox/index.tsx:75-82` — a latent bug for any multi+freeForm caller, left alone here to keep this PR off shared components). So this builds directly on KUI `Combobox` + `useController`, and surfaces whatever is typed as a one-item **Custom label** section.

**Trigger uses `multipleMode="count"`.** Rendering chips below the field while the combobox also renders its own tags in the trigger duplicated every selection. `renderValue` doesn't suppress them — it only applies when closed, and returning `null` reads as "no custom renderer". `count` gives a compact "2 selected" summary and leaves the coloured chips as the single source of truth.

**Include-defaults merge.** `entity_labels` *replaces* the default set server-side, so "include defaults" has to transmit them explicitly. `buildDetectConfig` unions defaults with the picks, and omits `detect` entirely when the selection adds nothing (empty, or a subset of the defaults) so the server stays on its own defaults instead of receiving a redundant 65-item list.

**No Strict/Permissive tabs.** The ticket flagged these as "consider". Pulling the Figma Entities frame shows `Strict`/`Permissive` are the *variant names of the design component*, whose two states render the existing **Custom** / **Auto-detect** segmented control — already shipped in ASTD-327. Nothing to build; the descriptive copy was updated to match the design's wording.

## Testing

- 30 unit tests (9 new): category coverage and uniqueness, section grouping order, `Other` fallback, custom-candidate rules, and the four detect-merge cases
- typecheck, eslint, prettier clean
- Driven in a running Studio against a live platform: category headings render, a typed `ice_cream_flavor` is offered and selectable, chips appear coloured with working removal, and the checkbox shows the live count of 65

## Screenshots

01-category-menu
02-custom-label-candidate
03-chips-and-defaults-merge

## Summary by CodeRabbit

* **New Features**
  * Improved anonymizer entity selection with categorized labels and color-coded tags.
  * Added support for entering and selecting custom entity labels.
  * Default entity labels are now loaded automatically for the current workspace.
  * Added validation to require default entities or at least one custom label.
  * Full Run is disabled while entity labels are loading.

* **Bug Fixes**
  * Improved job configuration to correctly combine default and custom entity selections.

### DIFF
--- a/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/AnonymizerBuilderForm.tsx
+++ b/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/AnonymizerBuilderForm.tsx
@@ -50,7 +50,8 @@ export const AnonymizerBuilderForm: FC = () => {
   const [submitError, setSubmitError] = useState<string | undefined>(undefined);
 
   const { isLoading: isLoadingModels } = useDefaultRoleModels();
-  const { data: defaultEntityLabels } = useAnonymizerListEntityLabels(workspace, { query: {} });
+  const { data: defaultEntityLabels, isLoading: isLoadingEntityLabels } =
+    useAnonymizerListEntityLabels(workspace, { query: {} });
 
   const createJob = useAnonymizerCreateRunJob({
     mutation: {
@@ -116,7 +117,7 @@ export const AnonymizerBuilderForm: FC = () => {
                 kind="primary"
                 color="brand"
                 type="submit"
-                disabled={createJob.isPending || isLoadingModels}
+                disabled={createJob.isPending || isLoadingModels || isLoadingEntityLabels}
               >
                 Full Run
               </Button>
--- a/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/AnonymizerBuilderForm.tsx
+++ b/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/AnonymizerBuilderForm.tsx
@@ -1,7 +1,10 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useAnonymizerCreateRunJob } from '@nemo/sdk/generated/anonymizer/api';
+import {
+  useAnonymizerCreateRunJob,
+  useAnonymizerListEntityLabels,
+} from '@nemo/sdk/generated/anonymizer/api';
 import type { RunJob } from '@nemo/sdk/generated/anonymizer/schema';
 import {
   Banner,
@@ -47,6 +50,7 @@ export const AnonymizerBuilderForm: FC = () => {
   const [submitError, setSubmitError] = useState<string | undefined>(undefined);
 
   const { isLoading: isLoadingModels } = useDefaultRoleModels();
+  const { data: defaultEntityLabels } = useAnonymizerListEntityLabels(workspace, { query: {} });
 
   const createJob = useAnonymizerCreateRunJob({
     mutation: {
@@ -76,7 +80,10 @@ export const AnonymizerBuilderForm: FC = () => {
   const onSubmit = form.handleSubmit(
     (values) => {
       setSubmitError(undefined);
-      createJob.mutate({ workspace, data: buildAnonymizerJobRequest(values) });
+      createJob.mutate({
+        workspace,
+        data: buildAnonymizerJobRequest(values, defaultEntityLabels?.data ?? []),
+      });
     },
     (errors) => {
       const onlyModelErrors = Object.keys(errors).every((key) => key === 'roleModels');
--- a/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
+++ b/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
@@ -100,7 +100,7 @@ export const EntitiesSection: FC = () => {
                   onClick={() => removeLabel(label)}
                 >
                   {label}
-                  <X fontSize="small" />
+                  <X size={14} />
                 </Tag>
               ))}
             </Flex>
--- a/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
+++ b/web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
@@ -2,31 +2,55 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { ControlledCheckbox } from '@nemo/common/src/components/form/ControlledCheckbox';
-import { ControlledCombobox } from '@nemo/common/src/components/form/ControlledCombobox';
 import { ControlledSegmentedControl } from '@nemo/common/src/components/form/ControlledSegmentedControl';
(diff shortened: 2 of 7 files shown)


web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
```diff
@@ -39,27 +63,55 @@ export const EntitiesSection: FC = () => {
...
-            slotLabel: 'Entity Labels',
-            slotInfo: 'Only these entity types will be detected and replaced.',
-          }}
-        />
+        <Stack gap="density-md">
+          <FormField
+            slotLabel="Entity Labels"
+            slotInfo="Pick from the detected entity types, or type your own label and select it."
+          >
+            <Combobox
```
steramae-nvidia: ControlledCombobox?

web/packages/studio/src/routes/AnonymizerBuilderRoute/constants.ts
```diff
@@ -103,6 +103,132 @@ export const ENTITY_MODE_OPTIONS: { value: EntityMode; children: string }[] = [
   { value: ENTITY_MODE_AUTO, children: 'Auto-detect' },
 ];
 
+export type EntityTagColor = 'blue' | 'gray' | 'green' | 'purple' | 'red' | 'teal' | 'yellow';
```
steramae-nvidia: Isn't this already defined by Tag? Something like `ComponentsProps<typeof Tag>['color']`

steramae-nvidia (approved):

web/packages/studio/src/routes/AnonymizerBuilderRoute/components/AnonymizerBuilderForm.tsx
```diff
@@ -47,6 +50,7 @@ export const AnonymizerBuilderForm: FC = () => {
   const [submitError, setSubmitError] = useState<string | undefined>(undefined);
 
   const { isLoading: isLoadingModels } = useDefaultRoleModels();
+  const { data: defaultEntityLabels } = useAnonymizerListEntityLabels(workspace, { query: {} });
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in b02c6d8d7b. `Full Run` is now disabled on `isLoadingEntityLabels` as well, so a submit cannot land while `defaultEntityLabels` is still `[]`. You are right about the consequence — with `includeDefaultEntities` on, the merge would have produced just the custom picks and sent those as the complete `entity_labels`, narrowing detection rather than widening it."

web/packages/studio/src/routes/AnonymizerBuilderRoute/schema.ts
```diff
@@ -168,18 +168,39 @@ const buildRewriteConfig = (form: AnonymizerFormData): Rewrite => {
...
+  defaultEntityLabels: string[]
+): AnonymizerConfigInput['detect'] => {
+  if (form.entityMode !== ENTITY_MODE_CUSTOM) return undefined;
+
+  const labels = form.includeDefaultEntities
+    ? [...new Set([...defaultEntityLabels, ...form.entityLabels])]
+    : form.entityLabels;
+
+  if (!labels.length) return undefined;
+  if (form.includeDefaultEntities && labels.length === defaultEntityLabels.length) return undefined;
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in b02c6d8d7b. Rather than adjusting `buildDetectConfig`, the form schema now rejects the state outright: `superRefine` raises on `entityLabels` when Custom mode has no labels and defaults are excluded, so the user gets an inline error instead of a silent fallback to the server defaults. Tests cover both the rejection and the two valid combinations."

web/packages/studio/src/routes/AnonymizerBuilderRoute/schema.ts
```diff
@@ -168,18 +168,39 @@ const buildRewriteConfig = (form: AnonymizerFormData): Rewrite => {
...
+
+  return { entity_labels: labels };
+};
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in b02c6d8d7b, using `new Set(defaultEntityLabels).size` as you suggested. Added a test that passes a defaults list containing a duplicate to confirm a genuine custom pick still survives."

web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
```diff
@@ -39,27 +63,55 @@ export const EntitiesSection: FC = () => {
...
+                  key={label}
+                  color={entityTagColor(label)}
+                  kind="outline"
+                  aria-label={`Remove ${label}`}
+                  onClick={() => removeLabel(label)}
+                >
+                  {label}
+                  <X fontSize="small" />
+                </Tag>
+              ))}
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in b02c6d8d7b — `size={14}`. Worth noting the `fontSize=\"small\"` came from copying the existing pattern in `components/common/FilterToolbar/components/FilterTags.tsx`, which has the same issue and renders its dismiss icons at 24px."

web/packages/studio/src/routes/AnonymizerBuilderRoute/constants.ts
marcusds: Yup

web/packages/studio/src/routes/AnonymizerBuilderRoute/components/EntitiesSection.tsx
marcusds: There was limitations why I didn't use it, I didn't want to make changes that could break it elsewhere, but let me dig a bit more into it.

marcusds: Done

PR #914, opened 2026-07-27
feat(agents): authenticate deployed agents via service-principal sidecar

benmccown:
## TL;DR

When platform auth is enabled, a deployed agent's Inference Gateway calls were rejected with **401** — the agent (a NAT runtime whose OpenAI HTTP client we don't control) carries no platform credential (`api_key: "not-used"`). This injects a loopback **auth-proxy sidecar** that stamps a service-principal identity so those calls authenticate. Works for both k8s and docker agent deployments.

## The problem

Follows #899 (which fixed *connectivity* — the agent reaching the internal API Service). This fixes *authorization*: with auth on, the agent's IGW call had no principal and got a 401 → OpenAI-client retry loop.

Per @ironcommit, static `service:*` identity is the intended fallback while workload token exchange is unavailable, and it's what the docker monolith already depends on. So this uses that mechanism rather than the (out-of-scope) RFC 8693 token exchange.

## The fix

**nemo-deployments owns the sidecar; nemo-agents just requests it.**

- **New auth-proxy sidecar** (`nmp_common/auth/workload_proxy`): a loopback FastAPI forwarder run via `nemo services run --sidecars auth-proxy` (registered in `AVAILABLE_SIDECARS`). It strips inbound auth/principal headers and stamps `X-NMP-Principal-Id: service:<identity>`, which OPA authorizes via the ServiceSystem role. Streams responses.
- **DeploymentConfig flags**: `auth_proxy_sidecar: bool` + `auth_proxy_sidecar_identity: str | None`. The **deployments plugin** compiles and injects the sidecar from these flags (`nemo_deployments_plugin.auth_proxy`), for both backends:
  - **k8s** — native init-container sidecar (`restartPolicy: Always`); upstream is the in-cluster API Service DNS.
  - **docker** — sidecar sharing the primary's network namespace (agent reaches it on localhost); upstream is rewritten to a docker-reachable host (e.g. `host.docker.internal` / bridge addr) via `determine_loopback_override()`, mirroring #899.
  - Injection is a **no-op when platform auth is disabled**.
- **nemo-agents** sets the two flags (identity `agents`) and points the agent's `llms.*.base_url` at the proxy (`http://127.0.0.1:8090`) when auth is on. Unchanged when auth is off (direct Service DNS per #899). It does not build the sidecar or know its mechanics.
- **Plugins don't import `nmp_common`**: added `nemo_platform_plugin.auth.platform_auth_enabled()` as the wrapper both plugins consult.
- **Fix latent k8s compiler bug**: exec probes failed because `V1Probe` needs the `_exec` kwarg (`exec` is a Python keyword). The sidecar uses an exec readiness probe since it binds loopback only (a pod-IP httpGet probe is refused).

The NIM/vLLM model images are never touched. The LoRA adapters sidecar already self-injects `service:models` via `get_platform_sdk(as_service=...)` and is unaffected — the proxy is only for workloads that can't set the header themselves.

## Verified end-to-end (dev pod, auth enabled)

- **k8s** (kind + Helm): agent deployment reaches `2/2 Running` (`auth-proxy: ready=true` via the exec probe); invoke returns **HTTP 200 `{"value":"...96..."}`** through the sidecar + a real nvidia-build model.
- **docker** (local `nemo services run`): deploy `--mode docker` yields the agent + `auth-proxy` containers (`Up (healthy)`), sidecar `network=container:<agent>` (shared netns), env `NMP_AUTH_PROXY_PRINCIPAL=agents` / `NMP_BASE_URL=http://172.17.0.1:8080`. Agent httpx log shows it calling `127.0.0.1:8090`; the IGW logs `status_code=404` (model resolution) — **not 401**, confirming the `service:agents` principal authenticated.

## Tests

- Sidecar forwarder: principal stamping, passthrough, `/healthz`.
- Deployments-side `build_auth_proxy_container`: requested / auth-off no-op / identity default / docker loopback-upstream rewrite; docker `build_docker_plan` injection.
- Agents-side: flag-setting + `base_url` wiring (auth on/off).
- k8s exec-probe compiler serialization (`_exec`).

## Summary by CodeRabbit

* **New Features**
  * Added an auth-proxy sidecar to forward platform-bound requests while stamping a service identity and preventing header spoofing.
  * When platform auth is enabled, inference traffic is routed through the local auth-proxy for both Kubernetes and Docker deployments (configurable image and port).
  * Added `/healthz` for health/readiness checks.
* **Bug Fixes**
  * Improved Kubernetes probe handling for exec-based probes.
* **Tests**
  * Expanded coverage for principal stamping/normalization, status passthrough, sidecar injection, docker URL rewriting, probe serialization, and proxy shutdown cleanup.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/auth.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/auth.py
@@ -1,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Auth helpers exposed to plugins.
+
+Plugins must not import ``nmp_common`` directly. This module wraps the pieces of
+the platform auth configuration that plugins need. The underlying auth config
+lives in ``nmp_common`` (only present in the platform process image), so it is
+imported lazily and failures degrade to "disabled" rather than raising in
+environments without it.
+"""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def platform_auth_enabled() -> bool:
+    """Return whether platform authentication is enabled."""
+    try:
+        from nmp.common.config import get_auth_config
+
+        return bool(get_auth_config().enabled)
+    except Exception:
+        logger.debug("Could not resolve auth config; assuming auth disabled", exc_info=True)
+        return False
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/auth.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/auth.py
@@ -18,7 +18,22 @@ logger = logging.getLogger(__name__)
 
 
 def platform_auth_enabled() -> bool:
-    """Return whether platform authentication is enabled."""
+    """Return whether platform authentication is enabled.
+
+    Returns ``False`` on any failure to resolve the auth config. The realistic
+    failure is ``ImportError``: ``nmp_common`` ships only in the platform process
+    image, so when this package is used standalone (outside the platform) there
+    is no auth config and "disabled" is the correct answer.
+
+    Other failures are effectively unreachable in the context that matters here
+    (the deployment controller, which runs *inside* the platform image): a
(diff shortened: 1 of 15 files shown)


ironcommit (approved):
LGTM

PR #919, opened 2026-07-27
fix(studio): validate entity names against the entity-store pattern

marcusds:
## Problem

Creating an inference provider named `Sparl` passed client-side validation and then failed on submit:

```
{"detail": "[{'type': 'string_pattern_mismatch', 'loc': ['body', 'name'],
  'msg': \"String should match pattern '^[a-z](?!.*--)[a-z0-9\\\\-@.+_]{1,62}(?<!-)$'\", ...}]"}
```

The forms validated against the SDK-generated regex `^[\w\-.]+$` (max 255) — that's what the service DTOs advertise in the OpenAPI spec. The entity store enforces a stricter RFC-1035-ish `NAME_PATTERN` downstream (`packages/nmp_common/src/nmp/common/entities/constants.py` on `main`; moving to `nemo_platform_plugin/entity_naming.py` in #978), so the spec is looser than reality and invalid names got through to a confusing 422 toast.

Three places had already worked around this independently, each with its own copy of the pattern or its own wording of the same rule.

## Change

New `web/packages/common/src/utils/entityName.ts` — mirrors `NAME_PATTERN` and reports the **specific** rule a value breaks, with a repaired-name suggestion where one exists:

| input | message |
|---|---|
| `Sparl` | `Name must be lowercase. Try "sparl".` |
| `invalid name!` | `Name cannot contain spaces, "!". Use lowercase letters, numbers, and - _ . @ + only. Try "invalid-name".` |
| `1provider` | `Name must start with a lowercase letter. Try "provider".` |
| `my--provider` | `Name cannot contain consecutive hyphens. Try "my-provider".` |
| `myprovider-` | `Name cannot end with a hyphen. Try "myprovider".` |
| 64 chars | `Name must be 63 characters or fewer (currently 64). Try "…".` |

Adopted in every form that names an entity:

- `CreateInferenceProviderSidePanel` — replaced the SDK regex; also pre-empts the duplicate-name 409 using the provider list it already fetches
- `CreateSecretModal` — dropped its duplicate `SECRET_NAME_REGEXP` (closes the `TODO #4082` there)
- `FilesetNewRoute` / `FilesetCreateModal` — dropped two more hand-written wordings
- `SubmitEvaluationModal` — replaced a manual `.test()` inside a `superRefine`

`filesetName.ts` now delegates to the shared helper and keeps only `toValidFilesetName` (domain-specific `'fileset'` fallback) and `FILESET_NAME_MAX_LENGTH`.

The regex itself is no longer written here at all — `ENTITY_NAME_REGEXP` is `entitiesCreateEntityBodyNameRegExp` from the generated zod, taken from the generic entity-store endpoint rather than one of the per-resource copies. That only works because of #978 (see below).

`ENTITY_NAME_MAX_LENGTH` stays local: the entity-store schema declares no `maxLength`, so there is nothing generic to import. `entityName.test.ts` pins it against the fileset and secret create schemas that do declare one, and asserts all four generated schemas — entity, fileset, secret, model provider — still share a single pattern. Drift between them fails CI. Before #978 the model-provider assertion would have failed against `^[\w\-.]+$`.

What stays hand-written is the part codegen can't express: the per-rule error messages and the sanitizer's character classes.

## Stacked on #978 — merge that first

> [!IMPORTANT]
> This PR targets `astd-349-entity-name-pattern-dtos/mschwab`, not `main`. **#978 must merge first**; GitHub will retarget this to `main` automatically when it does.

@steramae-nvidia asked whether the spec itself was the problem. It was, and #978 ([ASTD-349](https://linear.app/nvidia/issue/ASTD-349)) fixes it:

- `CreateModelProviderRequest`, `CreateFilesetRequest`, and `PlatformSecretCreateRequest` now declare `NAME_PATTERN` instead of `^[\w\-.]+$`, with `max_length` 255 → 63 to match. The secrets DTO enforced its rule in a `field_validator`, so it published *no* pattern at all; that moved to `pattern=`.
- The four Python copies of `NAME_PATTERN` collapse into one `nemo_platform_plugin/entity_naming.py`, re-exported by `nmp_common`. `nmp_common` depends on `nemo-platform-plugin`, so the plugin is the only side both can import from.

Originally this PR shipped its own mirrored regex as a stopgap. Stacking on #978 removes that: the generated zod now carries the strict pattern, so there is no hand-written copy of `NAME_PATTERN` left anywhere in `web/`.

## Testing

- `packages/studio` — 2657/2657 pass
- `packages/common` — 1359/1359 pass
- Added `entityName.test.ts` (per-rule messages, sanitizer, zod schema, generated-schema agreement) and 7 cases to the provider panel test
- `tsc --noEmit`, eslint, prettier clean on both packages

One existing assertion changed, and it's an upgrade: `FilesetNewRoute` typing `tiny-gpt2-A` previously matched the generic blob (which happened to contain "must start with a lowercase letter"); it now asserts `Name must be lowercase. Try "tiny-gpt2-a".`

## Summary by CodeRabbit

* **New Features**
  * Applied consistent name validation across filesets, inference providers, secrets, and evaluation submissions, including shared guidance and automatic sanitization/fallback.
  * Prevent duplicate inference provider names.
  * Disable the “Add Provider” action while providers are still loading.
* **Refactor**
  * Centralized entity-name validation and help text, replacing feature-specific regex rules.
* **Bug Fixes**
  * Updated validation messaging to be more specific (e.g., lowercase/capitalization suggestions).
* **Tests**
  * Added unit tests for entity-name rules and updated existing validation test expectations.


web/packages/studio/src/routes/InferenceProvidersListRoute/CreateInferenceProviderSidePanel/index.tsx
```diff
@@ -130,14 +123,23 @@ export const CreateInferenceProviderSidePanel: FC<CreateInferenceProviderSidePan
...
+
   const {
     control,
     reset: resetForm,
     handleSubmit,
     setValue,
     formState: { errors },
   } = useForm({
-    resolver: zodResolver(createProviderFormSchema),
+    resolver: zodResolver(formSchema),
```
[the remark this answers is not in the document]
marcusds: Claude says: "Gated submission on the providers query in 6153573 — `Add Provider` is now disabled while the list loads, so the refinement never runs against an empty set. Regression test added. Skipping the all-pages fetch / server-side availability check: past 100 providers the client check degrades to the server 409, which the panel already surfaces inline and which an existing test covers. Not worth an N-page fetch on every panel open."

marcusds:
@coderabbitai Added the explicit return type in 6153573 — `entityNameSchema` now declares `z.ZodEffects<z.ZodString, string, string>` (zod ^3.25.76).

web/packages/studio/src/routes/FilesetNewRoute/types.ts
```diff
@@ -1,30 +1,16 @@
...
 import { z } from 'zod';
 
-/**
- * Override the SDK-generated name validation. The generated zod uses the Files
- * service DTO's loose pattern (`^[\w\-.]+$`, max 255); the entity store
- * downstream enforces a stricter RFC-1035-ish pattern. Validate against the
- * strict pattern here so the user sees a useful inline error instead of a 422.
- */
+// Override the SDK-generated name validation, which uses a looser pattern than the
+// entity store enforces.
```
steramae-nvidia: Is this a problem for the openapi schema? Anyone relying on the openapi schema should expect that the regex pattern in the spec should be valid.

marcusds: I was just kicking this issue down the road, but yeah I'll do it now.

marcusds: Done

steramae-nvidia (approved):

PR #920, opened 2026-07-27
fix(jobs): jobs-launcher authorization without token exchange

ironcommit:
# Fix Kubernetes job log upload auth without service bearer spoofing

## Summary

This fixes Kubernetes job log upload and task-side SDK auth when platform auth is enabled but workload token exchange is disabled.

In that configuration, jobs do not receive `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`, so they cannot exchange a workload identity token for a real access token. The previous fallback sent:

```text
Authorization: Bearer service:jobs
```

That is not valid auth for the Files OTLP log upload endpoint, so job logs failed with `401 Unauthorized`:

```text
failed to send logs to http://nemo-platform-api:8080/apis/files/v2/workspaces/<workspace>/filesets/<fileset>/otlp/v1/logs: 401 Unauthorized
```

## Problem

The failing Kubernetes setup has:

- `auth.enabled=true`
- workload token exchange disabled
- no workload token file mounted into the job
- job runtime URLs pointed directly at `http://nemo-platform-api:8080`
- job log upload endpoint set to `/apis/files/v2/workspaces/{workspace}/filesets/{fileset}/otlp/v1/logs`

The jobs launcher correctly detected that no workload token file was configured, but it fell back to a raw service bearer token. That raw bearer shortcut is only supported by the legacy HF-compatible Files path. It is intentionally not supported across normal API endpoints because an external caller that can reach the API directly could spoof a service principal with `Authorization: Bearer service:<name>`.

The same configuration also exposed a task-side SDK issue: individual tasks should not need to know whether workload token exchange is enabled. Task code should call a single SDK helper and let the SDK/factory layer choose the right auth mechanism.

## Root Cause

The no-workload-token path conflated two different auth mechanisms:

- Real bearer auth: `Authorization: Bearer <access-token>` from OIDC or workload token exchange.
- Internal service identity: `X-NMP-Principal-Id: service:<name>` inside the trusted service boundary.

When workload exchange is disabled, `Bearer service:jobs` is not a real access token. The Files OTLP endpoint goes through normal middleware auth, so it returned 401.

## Solution

Use internal service-principal headers for the no-workload-token fallback, and keep real bearer auth only for real tokens.

Changes included:

- Jobs launcher log upload:
  - If `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` is present, exchange the workload token and send `Authorization: Bearer <access-token>`.
  - If no workload token file is present, send `X-NMP-Principal-Id: service:jobs`.
  - Rename the fallback log mechanism from service bearer token auth to service principal header auth.

- Workload auth source:
  - Expose a `Headers(ctx)` method so token-backed auth and header-backed auth share one exporter path.

- Task SDK behavior:
  - Centralize the workload-token-vs-service-header decision in `nmp.common.sdk_factory`.
  - `get_task_sdk(as_service=...)` now chooses workload identity when `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` is set.
  - If no workload token file is set, `get_task_sdk(as_service=...)` uses `X-NMP-Principal-Id: service:<name>` plus on-behalf-of headers from `NMP_PRINCIPAL`.
  - Plain task code no longer has to branch on workload token env vars.

- Plugin SDK provider:
  - Mirror the same task SDK behavior in `nemo_platform_plugin.sdk_provider.DefaultSDKProvider` so plugin task authors also use one helper.

- Hello-world workload task:
  - The task now simply calls `get_task_sdk(as_service="jobs")`.

## Summary by CodeRabbit

- **New Features**
  - Added workload-identity token-file support for platform and task SDK requests, including internal-marker-only headers and streamlined sync/async SDK creation.
  - Added platform endpoint helpers to retrieve SDK-specific HTTP clients for TCP vs UDS endpoints.
- **Bug Fixes**
  - Updated OTLP job log authentication to emit per-request header sets; service identity now uses `X-NMP-Principal-Id` when workload identity isn’t enabled.
- **Tests**
  - Expanded workload-identity and OTLP auth middleware coverage across SDK factories, plugins, jobs launcher, and hello-world integration.
- **Documentation**
  - Refreshed OTLP exporter option comments to reflect the new header behavior.

### DIFF
--- a/packages/nmp_common/src/nmp/common/sdk_factory.py
+++ b/packages/nmp_common/src/nmp/common/sdk_factory.py
@@ -9,6 +9,7 @@ from typing import Any, Callable, Optional, TypeVar, cast
 
 import httpx
 from nemo_platform import AsyncNeMoPlatform, NeMoPlatform
+from nemo_platform_plugin.client.constants import is_workload_identity_token_file_set
 from nmp.common.auth import Principal, get_principal_auth_headers, principal_from_env
 from nmp.common.config import Configuration, PlatformConfig
 from nmp.common.http_clients import shared_async_http_client, shared_sync_http_client
@@ -156,6 +157,26 @@ def _async_http_client_for_endpoint(
     return shared_async_http_client()
 
 
+def _should_bootstrap_workload_identity(
+    *,
+    as_service: str | None,
+    on_behalf_of: str | Principal | None,
+    http_client: httpx.Client | httpx.AsyncClient | None,
+    endpoint: PlatformEndpoint,
+) -> bool:
+    return (
+        as_service is None
+        and on_behalf_of is None
+        and http_client is None
+        and endpoint.transport != "uds"
+        and is_workload_identity_token_file_set()
+    )
+
+
+def _workload_identity_extra_headers(*, internal: bool) -> dict[str, str]:
+    return MARK_INTERNAL_REQUEST_HEADERS.copy() if internal else {}
+
+
 def _get_default_headers(
     as_service: str | None = None, internal: bool = False, on_behalf_of: str | Principal | None = None
 ) -> dict[str, str]:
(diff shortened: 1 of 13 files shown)


benmccown (approved):

packages/nmp_common/src/nmp/common/sdk_factory.py
```diff
@@ -156,6 +158,30 @@ def _async_http_client_for_endpoint(
     return shared_async_http_client()
 
 
+def _workload_identity_token_file_configured() -> bool:
```
mckornfield: feel like `_is_set` or something is clearer

packages/nmp_common/tests/sdk_factory/test_sdk.py
```diff
@@ -20,6 +22,17 @@
 )
 
 
+def _workload_oidc_config() -> NMPOIDCConfig:
+    return NMPOIDCConfig(
```
mckornfield, line 28: lmao all the acronyms

mckornfield (approved):

anastasia-nesterenko (approved):

packages/nmp_common/src/nmp/common/sdk_factory.py
```diff
@@ -241,8 +267,21 @@ def get_platform_sdk(
...
+        http_client=http_client,
+        endpoint=endpoint,
+    ):
+        headers = _workload_identity_extra_headers(internal=internal)
+        sdk = NeMoPlatform(
+            base_url=base_url or endpoint.connect_base_url,
+            default_headers=headers if headers else None,
+        )
+        return attach_platform_request_router(sdk)
+
```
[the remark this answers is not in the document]
ironcommit: Handling as part of a future PR.

PR #868, opened 2026-07-23
fix(deployments): handle Docker restart races

ironcommit:
## Summary
Docker deployment status reads now reconcile from Docker inspect state instead of mutating containers in the read path. This preserves one-shot and retry semantics across controller restarts and avoids releasing Docker GPU allocations before a deployment reaches a terminal status.

This PR no longer carries the Authentik/Envoy keep-alive template changes; that work has already merged into `main` via #1101.

## Changes
- Read Docker `State.ExitCode` and `RestartCount` from inspect attrs for create-time status mapping and `read_status()`.
- Keep stopped `Never` containers observable and report their terminal status without removing them from `read_status()`.
- Preserve retryable `OnFailure` deployments by comparing Docker's persisted restart count with the deployment backoff limit.
- Require positive `backoffLimit` / `backoff_limit` values in `DeploymentConfig`, the create request schema, and generated OpenAPI.
- Guard Docker `on-failure` run kwargs against `MaximumRetryCount=0`, and treat legacy zero-backoff Docker labels as unlimited retry and non-terminal so GPU allocations are not released while Docker would keep retrying.
- Release Docker GPU allocations only when status handling reaches a terminal deployment result.
- Map missing containers through restart-policy-aware status so `Always` stays recoverable as `LOST` while one-shot policies report terminal failure when the backend resource is gone.
- Treat foreign resource-scope containers like missing containers, including GPU release for terminal one-shot status results.
- Add Docker unit and integration coverage for terminal containers, retryable and exhausted `OnFailure` exits, missing one-shot containers, lost `Always` containers, foreign-container mismatch handling, zero-backoff legacy handling, API/entity validation, and GPU-release behavior.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [x] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [x] Documentation updated for user-visible behavior
- [ ] Documentation not applicable — justification:

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [x] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

[...]

PR #925, opened 2026-07-27
fix(nemo-deployments): observe one-shot Docker status at create time

tylersbray:
## Summary
- For `restart_policy=Never`, wait for the primary container to exit at create time and return `SUCCEEDED`/`FAILED` with the exit code (bounded by `docker_timeout`).
- For `OnFailure`, immediately inspect after create; map already-exited containers with the same backoff logic as `read_status`.
- Shares exit mapping via `_status_from_exited_container` so create and read stay aligned; no die-event replay and no e2e sleep workaround.

Addresses the fast-exit flake discussed for `e2e/test_nemo_deployments_docker.py` (and related to the approach in #868) by observing live container state instead of reconstructing status from Docker events.

## Test plan
- [x] Unit: `plugins/nemo-deployments/tests/unit/backends/docker/` (73 passed)
- [x] Integration against a real Docker daemon: `plugins/nemo-deployments/tests/integration/backends/docker/` (3 passed)
- [ ] Optional: e2e `test_docker_job_deployment_reaches_succeeded` with the existing fast alpine workload (no sleep)

## Summary by CodeRabbit

* **New Features**
  * One-shot deployments can now return `SUCCEEDED`/`FAILED` immediately after container creation when the container has already exited.
  * Added `oneshot_observe_timeout_seconds` (default **5**, min **1**) to limit the initial observe window before deferring to later status polling.
* **Bug Fixes**
  * Improved `Never`/`OnFailure` status mapping using exit code, restart count, and backoff-related retry decisions.
  * More robust handling of init/one-shot termination timeouts or communication issues, including improved cleanup on unexpected monitoring failures.
* **Tests**
  * Expanded Docker backend unit/integration and idempotency coverage for the above scenarios.

### DIFF
--- a/e2e/configs/local-docker-agents.yaml
+++ b/e2e/configs/local-docker-agents.yaml
@@ -49,8 +49,8 @@ deployments:
         # local Docker daemon (pulled by the test's CI job), so disable the
         # per-run pull and use the image as-is.
         pull_images: false
-        port_range_start: 9000
-        port_range_end: 9100
+        port_range_start: 49152
+        port_range_end: 49251
 
 models:
   controller:
--- a/packages/nmp_platform/config/local.yaml
+++ b/packages/nmp_platform/config/local.yaml
@@ -128,8 +128,8 @@ deployments:
       backend: docker
       config:
         pull_images: false
-        port_range_start: 9000
-        port_range_end: 9100
+        port_range_start: 49152
+        port_range_end: 49251
   default_executor: local-docker
 
 models:
--- a/packages/nmp_platform_runner/src/nmp/platform_runner/config/local.yaml
+++ b/packages/nmp_platform_runner/src/nmp/platform_runner/config/local.yaml
@@ -78,8 +78,8 @@ deployments:
       backend: docker
       config:
         pull_images: false
-        port_range_start: 9000
-        port_range_end: 9100
+        port_range_start: 49152
+        port_range_end: 49251
   default_executor: local-docker
 
 inference_gateway: {}
--- a/plugins/nemo-deployments/README.md
+++ b/plugins/nemo-deployments/README.md
@@ -30,8 +30,8 @@ Host port allocation for published container ports is configured per named docke
 executor (not on `DeploymentConfig.backend_config.docker`). Set the inclusive
 `port_range_start` / `port_range_end` bounds on the executor `config` block in
 platform YAML. The allocator scans every host port from `port_range_start`
-through `port_range_end`, including both endpoints (for example, 9000–9100
-allows 101 ports):
+through `port_range_end`, including both endpoints (for example, 49152–49251
+allows 100 ports):
 
 ```yaml
 deployments:
@@ -39,8 +39,8 @@ deployments:
     - name: local-docker
       backend: docker
       config:
-        port_range_start: 9000
-        port_range_end: 9100  # inclusive
+        port_range_start: 49152
+        port_range_end: 49251  # inclusive
   default_executor: local-docker
 ```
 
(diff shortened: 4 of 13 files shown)


benmccown (approved):

benmccown (changes requested):
Docker SDK timeout is 600 seconds by default. If you have a longer running one-shot container (like the model weights puller) this is going to block the reconciler loop and it also might take longer than that 600 second timeout.

tylersbray:
@benjamin-mccown Good catch — you're right that using `docker_timeout` (600s default) for the Never `container.wait()` would block the serial reconciler on long one-shots like model weight pullers.

Pushed a follow-up: added `DockerExecutorConfig.oneshot_observe_timeout_seconds` (default **5s**, aligned with controller `interval_seconds`). Never jobs still get a short create-time wait for fast exits (alpine/echo); if they're still running after that bound we return `STARTING` and finish via existing `read_status` polling + retention. `docker_timeout` stays for pulls/init; OnFailure remains inspect-only.

benmccown (approved):

tylersbray:
Added one more commit (`a80002c70`) that fixes the host-port collision behind the integration-test failures we saw on this PR. Bundling it here rather than in a separate PR because it lands in the same Docker create path this PR already changes.

**What was breaking.** `test_lost_detection_for_always` failed with `Bind for 0.0.0.0:9000 failed: port is already allocated`. The deployments Docker tests allocate from the 9000-9100 default range, and the intake/evaluator suites publish ClickHouse on 9000 on the same daemon. Under `pytest-xdist` those run concurrently. Reproducible on `main`, not caused by this PR.

**Three defects, all in the allocator:**

1. `is_port_free` bound `127.0.0.1` with `SO_REUSEADDR`. Both are wrong for checking whether Docker can publish: Docker binds `0.0.0.0`, and `SO_REUSEADDR` lets the probe succeed against a port a wildcard publisher already holds. It now binds `0.0.0.0` with no `SO_REUSEADDR`.
2. `find_available_port` listed containers filtered by our `managed-by` label, so a ClickHouse container was invisible to it. Every container on the daemon competes for host ports; `resource_scope` governs ownership and cleanup, not port safety.
3. Docker's own reservations aren't observable until you attempt the publish, so a concurrent create can still win the race no matter how good the probe is. The publish failure is now retried on a different port (bounded, 3 attempts). Note `containers.run` creates *then* starts, so a failed start leaves a container holding the name — that gets removed before the retry. That leak existed before this fix too.

Fix 1 is what actually closes the observed failure; 2 and 3 close the remaining windows.

**Verification.** With a container publishing `0.0.0.0:9000`, the old probe reports the port free and the new one doesn't:

```
old probe says 9000 free: True
new probe says 9000 free: False
allocator picks: 9001
```

The Docker integration tests pass with 9000 occupied. Unit coverage added for the wildcard probe, the foreign-container scan, and all three retry outcomes (reallocate and succeed, exhaust attempts, don't retry unrelated failures.

Separately, the deployments Docker integration tests now use 21000-21100 instead of the 9000-9100 product default, so they don't contend with ClickHouse even when the allocator is right. The product default is unchanged.
EOF
)

tylersbray:
Follow-up to the host-port allocator fix: after discussing the product default with Ben, commit `b732f70c89` moves Docker deployment allocation from `9000-9100` to `49152-49251` (100 ports in the IANA dynamic/private range).

This avoids the standing ClickHouse/native-port collision by default while preserving roughly the same capacity. The allocator scan/retry fixes remain necessary for arbitrary collisions. Updated the executor default, platform/e2e configs, docs, and default-value test; existing deployments and explicit executor overrides are unaffected.

Validation: focused executor-config tests passed, and the full pre-commit/pre-push hooks passed.

PR #934, opened 2026-07-27
fix(nemo-agents): strip workspace qualifier in eval preflight

walston:
Fixes [ASTD-271](https://linear.app/nvidia/issue/ASTD-271/eval-preflight-rejects-workspace-qualified-default-model-should-strip)

# Summary

`validate_llm_models` preflight rejected workspace-qualified default models (e.g. `default/nvidia-nemotron`) with a "not registered as VirtualModel" error, even when the VM existed — breaking `nemo agents evaluate`/`optimize` whenever `${NEMO_DEFAULT_MODEL}` resolved to the qualified form that `nemo setup` stores by default.

Root cause: the VM route (`GET /v2/workspaces/{workspace}/virtual-models/{name}`) takes `workspace` as its own path segment, so passing `default/foo` as `name` double-specifies the workspace and 404s. Preflight was the only consumer doing a raw retrieve-by-name; `nemo chat` and IGW's OpenAI proxy already handle the qualified form.

Fix: strip a leading `{workspace}/` from `model_name` with `str.removeprefix(f"{workspace}/")` before retrieve, mirroring IGW's OpenAI proxy. `removeprefix` strips only the exact prefix, so bare names and legitimate inner slashes (`nvidia/nemotron`, foreign-workspace prefixes) pass through untouched.

# Test plan
- [ ] `uv run --frozen ty check` (CI type lint) passes — verified locally, EXIT 0
- [ ] `python -m pytest plugins/nemo-agents/tests/unit/test_utils.py::TestValidateLLMModels` — 14 pass, incl. `test_strips_workspace_qualifier_before_lookup` (qualified → bare reaches SDK) and `test_bare_and_inner_slash_names_are_untouched` (bare/vendor/foreign-ws names verbatim)
- [ ] `nemo agents evaluate run` with `${NEMO_DEFAULT_MODEL}` resolving to a workspace-qualified value passes preflight and runs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and lookup for workspace-qualified LLM model names.
  * Routes qualified names to the correct workspace, removes redundant prefixes, and avoids duplicate lookups.
  * Provides clearer workspace-qualified error and warning messages.
  * Skips model names that normalize to an empty value.

* **Tests**
  * Added coverage for workspace routing, normalization, deduplication, and empty-name handling.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/utils.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/utils.py
@@ -335,6 +335,11 @@ def validate_llm_models(
         # prefix, leaving bare names and slashes inside the name (e.g.
         # "nvidia/nemotron") untouched.
         model_name = model_name.removeprefix(f"{workspace}/")
+        # A bare "{workspace}/" (e.g. "default/") normalizes to "" — re-check the
+        # non-empty invariant, which was validated before stripping. An empty
+        # name is not a routable model and would otherwise reach the SDK.
+        if not model_name:
+            continue
         to_check.setdefault(model_name, llm_key)
 
     if not to_check:
--- a/plugins/nemo-agents/src/nemo_agents_plugin/utils.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/utils.py
@@ -272,9 +272,11 @@ def validate_llm_models(
     """Pre-flight check that every IGW-routed LLM in *config* exists as a VirtualModel.
 
     Iterates ``config["llms"]`` and, for each block whose ``_type`` is in
-    :data:`_IGW_LLM_TYPES`, calls
-    ``sdk.inference.virtual_models.retrieve(model_name, workspace=workspace)``.
-    Names are deduplicated before lookup so the same model declared under
+    :data:`_IGW_LLM_TYPES`, strips a leading ``{workspace}/`` qualifier from
+    ``model_name`` (mirroring IGW's OpenAI proxy — the VM route takes
+    workspace as its own path segment, so the bare name is what it expects)
+    then calls ``sdk.inference.virtual_models.retrieve(name, workspace=workspace)``.
+    Names are deduplicated *after* stripping so the same model declared under
     multiple LLM keys (e.g. agent + judge) costs one network call.
 
     LLM blocks whose ``model_name`` still contains an unexpanded ``$VAR`` /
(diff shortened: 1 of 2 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/utils.py
```diff
@@ -325,6 +327,19 @@ def validate_llm_models(
                 model_name,
             )
             continue
+        # Strip a leading "{workspace}/" qualifier before lookup. The VM route
```
marcusds: I would trim these comments

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
```diff
@@ -325,6 +327,19 @@ def validate_llm_models(
...
+        # (GET /v2/workspaces/{workspace}/virtual-models/{name}) takes workspace
+        # as its own path segment, so a qualified name like "default/foo" would
+        # double-specify it and 404. This mirrors IGW's OpenAI proxy
+        # (removeprefix(f"{workspace}/")). removeprefix only strips this exact
+        # prefix, leaving bare names and slashes inside the name (e.g.
+        # "nvidia/nemotron") untouched.
+        model_name = model_name.removeprefix(f"{workspace}/")
+        # A bare "{workspace}/" (e.g. "default/") normalizes to "" — re-check the
```
marcusds: And here

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
```diff
@@ -325,6 +327,19 @@ def validate_llm_models(
...
             )
             continue
+        # Strip a leading "{workspace}/" qualifier before lookup. The VM route
+        # (GET /v2/workspaces/{workspace}/virtual-models/{name}) takes workspace
+        # as its own path segment, so a qualified name like "default/foo" would
+        # double-specify it and 404. This mirrors IGW's OpenAI proxy
+        # (removeprefix(f"{workspace}/")). removeprefix only strips this exact
+        # prefix, leaving bare names and slashes inside the name (e.g.
+        # "nvidia/nemotron") untouched.
+        model_name = model_name.removeprefix(f"{workspace}/")
```
marcusds: Could use `from nemo_platform_plugin.entities import parse_qualified_name` which other plugins seem to use, but this seems OK

marcusds (approved):

PR #936, opened 2026-07-27
fix(studio): let ControlledCombobox back a typed multi-select [ASTD-344]

marcusds:
Closes [ASTD-344](https://linear.app/nvidia/issue/ASTD-344/studio-controlledcombobox-unusable-for-multi-select-with-typed-input).

Split out of [#912](https://github.com/NVIDIA-NeMo/nemo-platform/pull/912) review feedback — @steramae-nvidia asked why the Anonymizer entity picker hand-rolled a KUI `Combobox` instead of using this shared component. It couldn't, for two reasons. **Merge this first**; #912 then migrates onto it.

## 1. The typed text was unreachable in multi-select

`inputValue` and `onInputValueChange` were both in the `Omit<>`, and the value was pinned:

```ts
inputValue: kind === 'multiple' ? '' : dispValue || '',
```

So no caller could supply or observe the search text — ruling out anything keyed off what the user typed, such as offering a "create this label" entry. Both props are now accepted; when a caller passes neither, behaviour is byte-for-byte what it was.

## 2. `freeForm` corrupted an array field

```ts
if (freeForm) {
  handleSelectedValueChange(newValue);  // string -> onChangeControl
}
```

For `kind="multiple"` the field holds a `string[]`, so this replaced the entire selection with a bare string on every keystroke. Multi-select no longer takes that path; callers wanting custom values surface them as items.

**No existing caller was affected.** Of the five call sites, `ZodFormField` uses `kind="multiple"` without `freeForm`, and `MappingFields` uses `freeForm` twice but only single-select. Nothing combined them, so the corruption path was unreachable — latent until someone wrote the obvious multi-select-with-custom-entries.

## Testing

Adds `index.test.tsx` — the component had none, unlike its `MappingFields` / `ZodFormField` neighbours. Four cases: single-select `freeForm` still writes through, multi-select never receives a raw string, controlled input reaches the caller, and ordinary multi-select selection still works.

I verified the regression test earns its keep by stashing the fix and re-running: **`1 failed | 3 passed`**, with the multi-select case being the failure. Worth being precise about the controlled-input case — it happens to pass pre-fix too, because `...comboboxProps` spreads after `inputValue` in `baseProps`, so a caller's value would have won at runtime. The `Omit<>` blocked it at the type level, which is what actually made it unusable.

Regression surface for the two existing consumers is covered by their own suites: 114 tests across `packages/common/src/components/form` pass, plus typecheck on `@nemo/common` and `nemo-studio-ui`, eslint, and prettier.

## Summary by CodeRabbit

* **New Features**
  * Added support for externally controlling combobox input text, including `onInputValueChange` handling.
  * Improved free-form behavior by syncing typed input into the selection/form value only for single-select mode.
* **Bug Fixes**
  * Prevented free-form typed text from overwriting multi-select form array values.
* **Tests**
  * Added coverage for single- vs multi-select free-form behavior, including externally controlled input and option selection via user clicks.
* **Documentation**
  * Added Storybook examples for single/multi, free-form, loading, disabled, and custom typed entries.

### DIFF
--- a/web/packages/common/src/components/form/ControlledCombobox/index.tsx
+++ b/web/packages/common/src/components/form/ControlledCombobox/index.tsx
@@ -76,11 +76,8 @@ export const ControlledCombobox = <T extends 'single' | 'multiple' = 'single'>({
 
   const handleTempValueChange = (newValue: string) => {
     onInputValueChange?.(newValue);
-    // The caller owns the search text, so don't shadow it with local state.
     if (isInputControlled) return;
-    // In freeForm, we set form value to enable custom values. Multi-select holds a string[],
-    // so writing the raw input there would replace the whole selection with one string —
-    // those callers surface custom entries as items instead.
+    // Multi-select holds a string[], so raw input would clobber the whole selection.
     if (freeForm && kind !== 'multiple') {
       handleSelectedValueChange(newValue);
     } else {
--- a/web/packages/common/src/components/form/ControlledCombobox/index.tsx
+++ b/web/packages/common/src/components/form/ControlledCombobox/index.tsx
@@ -4,7 +4,7 @@
 import { TextInputSpinner } from '@nemo/common/src/components/form/TextInputSpinner';
 import { UseControllerComponentProps } from '@nemo/common/src/types';
 import { Combobox, ComboboxProps, FormField } from '@nvidia/foundations-react-core';
-import { ReactNode, useState } from 'react';
+import { KeyboardEvent, ReactNode, useState } from 'react';
 import { useController } from 'react-hook-form';
 
 // Generic type to handle both single and multiple selection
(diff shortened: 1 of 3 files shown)


web/packages/common/src/components/form/ControlledCombobox/index.tsx
```diff
@@ -72,9 +72,13 @@ export const ControlledCombobox = <T extends 'single' | 'multiple' = 'single'>({
...
 
+  const isInputControlled = inputValue !== undefined;
+
   const handleTempValueChange = (newValue: string) => {
-    // In freeForm, we set form value to enable custom values.
-    if (freeForm) {
+    onInputValueChange?.(newValue);
+    if (isInputControlled) return;
+    // Multi-select holds a string[], so raw input would clobber the whole selection.
+    if (freeForm && kind !== 'multiple') {
```
steramae-nvidia: Can we write this logic in a way that doesn't prevent `freeForm` for multiple?

marcusds: Added, originally I didn't add it cause I didn't like how it would have had to work, but added it anyway. The problem is now you have to hit enter to add one to then continue adding more.

steramae-nvidia (approved):

PR #947, opened 2026-07-28
fix(platform): resolve race condition with process cleanup

soluwalana:
Found a race condition in e2e tests stemming from process cleanup.

This PR fixes that error and addresses the CodeRabbit comments that occurred after fixing it. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved instance shutdown coordination, preserving descriptors while competing locks remain held.
  * Warmup scheduling now safely clamps out-of-range values with warnings.

* **New Features**
  * Training schedules account for sequence packing when estimating steps per epoch.
  * RL validation now enforces optimizer parameter ranges and limits output names to 255 characters.

* **Documentation**
  * Updated job APIs with RL and Unsloth workspace endpoints and expanded response details.

* **Tests**
  * Added coverage for shutdown, scheduling, validation, and orphan-cleanup scenarios.


plugins/nemo-customizer/openapi/openapi.yaml
```diff
@@ -392,12 +392,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPValidationError'
-  /apis/customization/v2/workspaces/{workspace}/unsloth/jobs:
+  /apis/customization/v2/workspaces/{workspace}/rl/jobs:
```
albcui: The nemo-rl plugin was intentionally un-registered in `plugins/nemo-rl/pyproject.toml `, and these endpoints were removed.

I think you might have an older venv with the nemo-rl plugin installed, which re-generated these endpoints. Can you run `uv sync && make refresh-openapi` to refresh the stale openapi spec?

albcui (approved):
LGTM pending OpenAPI regen

PR #953, opened 2026-07-28
fix(models): slow models controller step and inaccurate controller health check

benmccown:
### How control loops decide whether they are healthy

Background work in the platform runs as control loops. Each loop is a thread that repeats the same cycle forever: run one `step()`, sleep for its poll interval, run the next `step()`. The models controller's interval defaults to 5 seconds.

The framework needs some way to tell whether a loop is still alive, so it records a timestamp each time a step begins — a heartbeat. Readiness compares that timestamp against a deadline of `3 x interval`, which for the models controller is 15 seconds. If the loop has not produced a heartbeat within 15 seconds it is reported unhealthy, `/health/ready` returns 503, and Kubernetes takes the pod out of service.

That works as long as a step reliably finishes well inside 15 seconds. The heartbeat was only sent once per step, so the deadline was really a bet that no step would ever take longer than three intervals.

### What went wrong

Models controller steps were taking roughly 54 seconds. Since the only heartbeat came at the start of a step, the loop looked alive for the first 15 seconds of each cycle and stalled for the remaining ~44. Readiness therefore alternated between passing and failing on a roughly 59-second cycle, and the controller pod spent most of its time marked unhealthy and pulled out of the Service — even though it was working the whole time, just slowly.

There are two separate defects behind that, and this PR addresses both:

1. The health model could not tell "slow" apart from "stuck", because progress was only reported once per step.
2. The models controller step was far slower than the work actually required.

## Fix 1: report progress more than once per step

A step that has 400 items to get through legitimately takes longer than three poll intervals. What matters for liveness is not "did the step finish" but "is the loop still getting somewhere". So controllers now report progress as each unit of work completes, rather than only when a step begins.

How it is put together:

- `Heartbeat` holds a single timestamp meaning "last observed progress".
- `HeartbeatMixin` gives a controller an `emit_heartbeat()` method to call as it finishes each item.
- `TrackLastExecutionTime` owns the `Heartbeat` and hands it to the wrapped controller when it wraps it. Entering `step()` still counts as progress, so a controller that reports nothing behaves exactly as it did before and no existing call site changed.
- Nested helpers (the reconcilers) receive the bound `emit_heartbeat` method rather than the `Heartbeat` itself, so they report progress without knowing anything about how liveness is measured.
- The liveness check in `Loop` is untouched. It already read this timestamp; the timestamp is simply updated more often now.

The important property is that beats only ever follow completed work — never a timer, never around a call that might block. A single call that hangs forever still stops the clock and is still caught. Only slow-but-advancing work is now treated as healthy.

Wired into the jobs scheduler and reconciler, workspace cleanup, the models controller, and the adapters sidecar.

Separately, when a loop does go unhealthy it now says so in the log. Previously the only record was at DEBUG, so a pod failing readiness produced no visible explanation. `Loop` records why the check failed and `ControllerManager` logs the change with the loop name and the reason, on transition only, so it is not repeated on every probe.

### Where the time went

Provider reconciliation walks every model a provider serves and, for each one, makes sure two things exist: a Model Entity for it, and a passthrough VirtualModel pointing at that entity. It was doing this one model at a time:

- one `GET` to read the Model Entity, and
- one `POST` attempting to create the VirtualModel, relying on a 409 conflict to mean "already there".

In a development workspace with 4 providers each serving 102 models, that is 408 served models and **816 sequential API calls per step**, every 5 seconds. At the measured ~60 ms per call that is most of a minute. In the steady state nothing needed changing, so essentially all of it was re-confirming state that was already correct.

Reading the same data in bulk instead is dramatically cheaper. Measured against that workspace:

| | Calls | Time |
|---|---|---|
| Read 404 Model Entities one at a time | 404 | **37.7 s** |
| Read all 404 in one paginated list | 1 | **0.33 s** |
| Read all 306 VirtualModels (`page_size=200`) | 2 | **0.33 s** |

### What changed

**`ModelEntityCache`** (new) reads the Model Entities once per reconciliation phase and collects up the intended changes instead of writing them as it goes:

- each entity is written at most once, and only if something actually differs — so a converged pass performs no writes at all
- the desired state for an entity is assembled in memory, so several providers contributing to the same entity are merged into one write rather than overwriting each other
- reads see staged changes layered over the snapshot, so code that stages a change and then reads the same entity sees its own write
- `refresh()` refuses to re-read while changes are still staged, which makes the required ordering a failure rather than a convention

Both deployment and provider reconciliation read and write these entities, and the deployment phase can remove a provider link that the provider phase would otherwise put back. So the cache is refreshed at the start of each phase that writes entities and flushed at the end of that phase, meaning neither phase ever decides based on state the other has already changed.

**VirtualModels are read once per pass.** Orphan cleanup was already listing all of them on every step, so reusing that read costs nothing and removes all 408 create attempts. The existence set covers every VirtualModel rather than only auto-provisioned ones, so a name already held by a user-managed VirtualModel is still left alone. Cleanup consumes the same snapshot and still runs after every provider, because it depends on the served models resolved during the pass.

### Result

Per-step API calls no longer scale with the number of served models. In the steady state the pass drops from ~826 calls to roughly a dozen, and from ~54 seconds to a couple of seconds — comfortably inside the 15-second liveness deadline, with Fix 1 covering the cold-start case where there genuinely is a lot to create.

## Notes for reviewers

- Each provider's `update_status` now lands before its entity links are written, since the flush moved to end-of-phase. Publishing `served_models` independently of entity initialization is existing behaviour.
- Staged changes are applied even when a step bails out early: provider deletion stages the removal of its entity links and the provider itself is already gone, so discarding them would leave the links behind with nothing to trigger another attempt.
- Reporting progress means a step that advances slowly forever will not be flagged. That is intended, and is the point of the change.

## Testing

- `ModelEntityCache`: overlay reads, refresh-while-staged rejection, single write for an entity linked by two providers, no write when converged, two providers creating one entity collapsing to a single create, conflict adoption, per-entity failure isolation, and a phase-ordering test that a removed link is not reinstated by the next phase.
- Liveness: a long step reporting progress stays healthy, a step reporting none goes unhealthy, and health changes log once per edge.
- Provider reconciliation ordering: VirtualModels read once per pass, existing and user-managed names not recreated, one created mid-pass never deleted as an orphan, concurrent deletion tolerated, and unresolved providers leaving `served_models` unset so cleanup falls back to persisted values.
- 1233 tests covering the changed code pass; `ruff check`, `ruff format --check`, and `ty check` are clean against baseline.

## Summary by CodeRabbit

* **New Features**
  * Added heartbeat-based progress tracking for controllers to support more accurate health monitoring of long-running work.
  * Controllers now emit heartbeats across workspace cleanup, job reconciliation/scheduling, adapter reconciliation, and model reconciliation.

* **Bug Fixes**
  * Health monitoring now surfaces exact unhealthy reasons (including stalled progress or inactive loop threads).
  * Continued progress reporting even when per-item cleanup or reconciliation operations fail.

* **Improvements**
  * Reduced repeated unhealthy logging by only reporting health transitions; improved recovery reporting when conditions return to healthy.

### DIFF
--- a/packages/nmp_common/src/nmp/common/controller/controller.py
+++ b/packages/nmp_common/src/nmp/common/controller/controller.py
@@ -19,43 +19,31 @@ logger = getLogger(__name__)
 class Heartbeat:
     """Timestamp of the last observed progress in a control loop.
 
-    A single loop iteration may take considerably longer than the loop's poll
-    interval when there is a lot of work to do. ``Heartbeat`` records the most
-    recent point at which the loop demonstrably made progress, which lets the
-    liveness check in :class:`Loop` distinguish a loop that is working through a
-    large batch from one that is stuck.
+    Lets :class:`Loop` tell a loop working through a large batch apart from one
+    that is stuck, since a single iteration can outlast the poll interval.
     """
 
     def __init__(self) -> None:
         self._last = datetime.now(timezone.utc)
 
     def beat(self) -> None:
-        """Record that progress was just made."""
         self._last = datetime.now(timezone.utc)
 
     def last(self) -> datetime:
-        """Return the time of the most recently recorded progress."""
         return self._last
 
 
 class HeartbeatMixin:
-    """Gives a controller a way to report incremental progress.
+    """Lets a controller report progress while a single ``step()`` is still running.
 
-    Controllers whose ``step()`` iterates over a variable number of items should
-    mix this in and call :meth:`emit_heartbeat` as each item is finished. Nested
-    collaborators are handed the bound :meth:`emit_heartbeat` method rather than
-    the :class:`Heartbeat` itself, so they report progress without depending on
-    how liveness is measured.
-
-    ``_heartbeat`` is a class-level default so that mixing this in never obliges
-    a subclass to call ``super().__init__()``. :meth:`attach_heartbeat` binds the
-    instance that :class:`Loop` measures; emitting before then is a no-op.
+    Mixed into controllers whose ``step()`` iterates a variable number of items.
     """
 
-    _heartbeat: "Heartbeat | None" = None
+    # Class-level default so mixing this in never obliges a subclass to call
+    # super().__init__(); emitting before attach_heartbeat() is a no-op.
+    _heartbeat: Heartbeat | None = None
 
     def attach_heartbeat(self, heartbeat: Heartbeat) -> None:
-        """Bind the heartbeat that this controller reports progress to."""
         self._heartbeat = heartbeat
 
     def emit_heartbeat(self) -> None:
@@ -108,10 +96,9 @@ class ProvidesLastExecutionTime(ABC):
 class TrackLastExecutionTime(Controller, ProvidesLastExecutionTime):
     """Tracks when the wrapped controller last made progress.
 
-    Entering ``step()`` counts as progress. A controller that also mixes in
-    :class:`HeartbeatMixin` shares this wrapper's :class:`Heartbeat`, so progress
-    it reports part-way through a long iteration is visible here too. There is
-    only ever one timestamp, and it means "last observed progress".
+    Entering ``step()`` counts as progress; a controller mixing in
+    :class:`HeartbeatMixin` shares this wrapper's :class:`Heartbeat` so its
+    mid-iteration progress lands on the same single timestamp.
     """
 
     def __init__(self, controller: Controller):
--- a/packages/nmp_common/src/nmp/common/controller/controller.py
+++ b/packages/nmp_common/src/nmp/common/controller/controller.py
@@ -16,6 +16,59 @@ from nmp.common.observability.context import AppContext, initialize_app_ctx
 logger = getLogger(__name__)
 
 
+class Heartbeat:
+    """Timestamp of the last observed progress in a control loop.
+
+    A single loop iteration may take considerably longer than the loop's poll
+    interval when there is a lot of work to do. ``Heartbeat`` records the most
+    recent point at which the loop demonstrably made progress, which lets the
+    liveness check in :class:`Loop` distinguish a loop that is working through a
+    large batch from one that is stuck.
+    """
+
+    def __init__(self) -> None:
+        self._last = datetime.now(timezone.utc)
+
+    def beat(self) -> None:
+        """Record that progress was just made."""
+        self._last = datetime.now(timezone.utc)
+
+    def last(self) -> datetime:
+        """Return the time of the most recently recorded progress."""
+        return self._last
+
+
+class HeartbeatMixin:
+    """Gives a controller a way to report incremental progress.
+
+    Controllers whose ``step()`` iterates over a variable number of items should
+    mix this in and call :meth:`emit_heartbeat` as each item is finished. Nested
+    collaborators are handed the bound :meth:`emit_heartbeat` method rather than
+    the :class:`Heartbeat` itself, so they report progress without depending on
+    how liveness is measured.
+
+    ``_heartbeat`` is a class-level default so that mixing this in never obliges
+    a subclass to call ``super().__init__()``. :meth:`attach_heartbeat` binds the
+    instance that :class:`Loop` measures; emitting before then is a no-op.
+    """
+
+    _heartbeat: "Heartbeat | None" = None
+
+    def attach_heartbeat(self, heartbeat: Heartbeat) -> None:
+        """Bind the heartbeat that this controller reports progress to."""
+        self._heartbeat = heartbeat
+
+    def emit_heartbeat(self) -> None:
+        """Report that a unit of work finished.
+
+        Only call this once work has actually completed. Emitting on a timer, or
+        around a call that may block, would report progress that did not happen
+        and defeat the liveness check.
+        """
+        if self._heartbeat is not None:
+            self._heartbeat.beat()
+
+
 class Controller(ABC):
     """Step represents a function intended to be run in a loop."""
 
@@ -53,17 +106,25 @@ class ProvidesLastExecutionTime(ABC):
 
 
 class TrackLastExecutionTime(Controller, ProvidesLastExecutionTime):
-    """TrackLastExecutionTime"""
+    """Tracks when the wrapped controller last made progress.
+
+    Entering ``step()`` counts as progress. A controller that also mixes in
+    :class:`HeartbeatMixin` shares this wrapper's :class:`Heartbeat`, so progress
+    it reports part-way through a long iteration is visible here too. There is
+    only ever one timestamp, and it means "last observed progress".
+    """
 
     def __init__(self, controller: Controller):
-        self._last_runtime = datetime.now(timezone.utc)
         self._internal = controller
+        self._heartbeat = Heartbeat()
+        if isinstance(controller, HeartbeatMixin):
+            controller.attach_heartbeat(self._heartbeat)
 
     def last_execution_time(self) -> datetime:
-        return self._last_runtime
+        return self._heartbeat.last()
 
     def step(self):
-        self._last_runtime = datetime.now(timezone.utc)
+        self._heartbeat.beat()
         self._internal.step()
 
     @property
@@ -114,6 +175,7 @@ class Loop(threading.Thread):
         self._internal = controller
         self._stop_signal = stop_signal if stop_signal is not None else threading.Event()
         self._shutdown_func = shutdown_func
+        self._unhealthy_reason: str | None = None
 
         # Capture the current context so it can be used in the thread
         self._context = contextvars.copy_context()
@@ -144,6 +206,11 @@ class Loop(threading.Thread):
         self._stop_signal.set()
 
     @property
+    def unhealthy_reason(self) -> str | None:
+        """Why the most recent :attr:`is_healthy` evaluation failed, if it did."""
+        return self._unhealthy_reason
+
+    @property
     def is_healthy(self) -> bool:
         """Check if the internal controller is healthy.
 
@@ -152,6 +219,7 @@ class Loop(threading.Thread):
         """
         # Thread must be alive
         if not self.is_alive():
+            self._unhealthy_reason = "loop thread is not alive"
             logger.debug(f"Controller thread {self.name} is not alive")
             return False
 
@@ -163,6 +231,7 @@ class Loop(threading.Thread):
             max_delay = sleep_secs * 3  # Allow 3 sleep windows before marking unhealthy
             time_since_last = (now - last_execution).total_seconds()
             if time_since_last > max_delay:
+                self._unhealthy_reason = f"no progress for {time_since_last:.2f}s (max allowed: {max_delay:.2f}s)"
                 logger.debug(
                     f"Controller thread {self.name} has not executed in {time_since_last:.2f}s "
                     f"(max allowed: {max_delay:.2f}s)"
@@ -170,7 +239,9 @@ class Loop(threading.Thread):
                 return False
 
         if not self._internal.is_healthy:
+            self._unhealthy_reason = "controller reported itself unhealthy"
             logger.debug(f"Controller thread {self.name} internal controller is unhealthy")
             return False
 
+        self._unhealthy_reason = None
         return True
--- a/packages/nmp_common/src/nmp/common/controller/controller_manager.py
+++ b/packages/nmp_common/src/nmp/common/controller/controller_manager.py
@@ -31,6 +31,7 @@ class ControllerManager:
         if ControllerManager._instance is not None:
             raise RuntimeError("Use ControllerManager.get_instance() to get the singleton instance")
         self._loops: Dict[str, Loop] = {}
+        self._reported_health: Dict[str, bool] = {}
 
     @classmethod
     def get_instance(cls) -> "ControllerManager":
@@ -42,6 +43,7 @@ class ControllerManager:
         if cls._instance is None:
             cls._instance = cls.__new__(cls)
             cls._instance._loops = {}
+            cls._instance._reported_health = {}
         return cls._instance
 
     def register(self, name: str, loop: Loop) -> None:
@@ -75,6 +77,7 @@ class ControllerManager:
         if name not in self._loops:
             raise KeyError(f"No loop with name '{name}' is registered")
         del self._loops[name]
+        self._reported_health.pop(name, None)
         logger.info(f"Unregistered loop: {name}")
 
     def get_loop(self, name: str) -> Loop:
@@ -134,11 +137,13 @@ class ControllerManager:
                     if not is_healthy:
                         all_healthy = False
                         logger.debug(f"Loop '{name}' is unhealthy")
+                    self._log_health_transition(name, is_healthy, getattr(loop, "unhealthy_reason", None))
                 except Exception as e:
                     if detailed:
                         health_status[name] = False
                     all_healthy = False
                     logger.error(f"Error checking health of loop '{name}': {e}", exc_info=True)
+                    self._log_health_transition(name, False, f"health check raised: {e}")
             else:
                 # No is_healthy property, assume healthy
                 if detailed:
@@ -147,6 +152,25 @@ class ControllerManager:
 
         return all_healthy, health_status if detailed else {}
 
+    def _log_health_transition(self, name: str, is_healthy: bool, reason: str | None) -> None:
+        """Log only when a loop's health changes.
+
+        Health is evaluated on every readiness probe, so logging each unhealthy
+        evaluation would be far too noisy to be useful. Reporting the edges makes
+        the loop responsible for a degraded process identifiable from the logs
+        alone.
+        """
+        previous = self._reported_health.get(name)
+        if previous == is_healthy:
+            return
+
+        self._reported_health[name] = is_healthy
+        if is_healthy:
+            if previous is not None:
+                logger.info(f"Control loop '{name}' is healthy again")
+        else:
+            logger.warning(f"Control loop '{name}' is unhealthy: {reason or 'reason unavailable'}")
+
     def clear(self) -> None:
         """Clear all registered loops.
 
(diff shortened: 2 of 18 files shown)


packages/nmp_common/src/nmp/common/controller/controller.py
```diff
@@ -16,6 +16,59 @@
 logger = getLogger(__name__)
 
 
+class Heartbeat:
+    """Timestamp of the last observed progress in a control loop.
+
+    A single loop iteration may take considerably longer than the loop's poll
```
mckornfield: probably overall review comment: might be worth asking the agent to be more concise with a few class level comments. Feel like (or hope) some of this stuff is evident from the code

packages/nmp_common/src/nmp/common/controller/controller.py
```diff
@@ -53,17 +106,25 @@ def last_execution_time(self) -> datetime: ...
...
+    :class:`HeartbeatMixin` shares this wrapper's :class:`Heartbeat`, so progress
+    it reports part-way through a long iteration is visible here too. There is
+    only ever one timestamp, and it means "last observed progress".
+    """
 
     def __init__(self, controller: Controller):
-        self._last_runtime = datetime.now(timezone.utc)
         self._internal = controller
+        self._heartbeat = Heartbeat()
+        if isinstance(controller, HeartbeatMixin):
```
mckornfield, line 107: would it ever not be this type?

packages/nmp_common/src/nmp/common/controller/controller.py
```diff
@@ -163,14 +231,17 @@ def is_healthy(self) -> bool:
...
+                self._unhealthy_reason = f"no progress for {time_since_last:.2f}s (max allowed: {max_delay:.2f}s)"
                 logger.debug(
                     f"Controller thread {self.name} has not executed in {time_since_last:.2f}s "
                     f"(max allowed: {max_delay:.2f}s)"
                 )
                 return False
 
         if not self._internal.is_healthy:
+            self._unhealthy_reason = "controller reported itself unhealthy"
             logger.debug(f"Controller thread {self.name} internal controller is unhealthy")
```
mckornfield, line 230: did we raise these levels here or elsewhere or no?

packages/nmp_common/tests/controller/test_controller.py
```diff
@@ -0,0 +1,173 @@
...
+    loop = Loop(TimedLoopWaiter(0.05, stop_signal=stop_signal), wrapper, stop_signal=stop_signal)
+    loop.name = "self-unhealthy"
+
+    loop.start()
+    try:
+        _run_loop_until(loop, lambda: not loop.is_healthy)
+        assert loop.unhealthy_reason == "controller reported itself unhealthy"
+    finally:
+        stop_signal.set()
+        loop.join(timeout=5)
```
mckornfield: one thing to be wary about here are these timeouts/sleeps in these tests, maybe just worth another pass to make sure nothing flaky is introduced here (run them a lot/in parallel?)

services/core/entities/src/nmp/core/entities/controllers/workspace_cleanup.py
```diff
@@ -101,8 +101,11 @@ async def _async_step(self):
                     return
 
                 await self._cleanup_jobs(workspace)
+                self.emit_heartbeat()
```
mckornfield, line 104: haha want to buy aspect oriented programming

services/core/models/src/nmp/core/models/controllers/deployment_reconciler.py
```diff
@@ -785,21 +799,13 @@ async def _cleanup_model_entities_for_provider(
                     _model_ref = parse_entity_ref(served_model.model_entity_id)
                     model_workspace, model_name = _model_ref.workspace, _model_ref.name
 
-                    # Get the Model Entity
-                    model_entity = await self._models_sdk.models.retrieve(
-                        name=model_name,
-                        workspace=model_workspace,
-                    )
+                    model_entity = self._entity_cache.get(model_workspace, model_name)
```
mckornfield, line 813: are there any cache invalidation issues we're introducing/not considering? or are those tested for

services/core/models/src/nmp/core/models/controllers/entity_cache.py
```diff
@@ -0,0 +1,225 @@
...
+"""Batched reads and writes of Model Entities for a single reconciliation phase.
+
+Reconciling providers and deployments requires looking at the Model Entity behind
+every served model, and often needs to link or unlink a provider on it. Doing that
+one entity at a time costs two round trips per served model per pass, which grows
+with the product of providers and the size of their model catalogues.
+
+:class:`ModelEntityCache` reads the entities once per phase and accumulates the
+intended changes, so each entity is written at most once and only when something
+actually differs. Because the desired state for an entity is assembled in memory
```
mckornfield, line 13: is there an OOM concern here or no (just swag on the data size)

services/core/models/src/nmp/core/models/controllers/entity_cache.py
```diff
@@ -0,0 +1,225 @@
...
+
+        create_kwargs = {k: v for k, v in staged.create_kwargs.items() if k not in ("name", "workspace")}
+        create_kwargs.update(staged.field_updates)
+        if staged.link_providers:
+            create_kwargs["model_providers"] = list(staged.link_providers)
+
+        try:
+            created = await self._models_sdk.models.create(workspace=workspace, name=name, **create_kwargs)
+        except ConflictError:
+            # Created concurrently; adopt it and apply the staged changes instead.
```
mckornfield, line 225: how would this be? we only have one controller running yeah?

services/core/models/tests/unit/controllers/test_provider_reconciler.py
```diff
@@ -1918,6 +1976,31 @@ def _virtual_model(
...
+        "fileset": attrs.get("fileset"),
+        "api_endpoint": attrs.get("api_endpoint"),
+        "backend_format": attrs.get("backend_format"),
+    }
+    entity = MagicMock()
+    for key, value in fields.items():
+        setattr(entity, key, value)
+    entity.model_copy = MagicMock(
+        side_effect=lambda update: _existing_entity(
+            workspace, name, **{**{k: v for k, v in fields.items() if k not in ("workspace", "name")}, **update}
```
mckornfield: lol this is quite the dict comprehension (with the `**`)

mckornfield (approved):
tried to call out some high level things, mostly q, yay readiness/liveness fixes

ironcommit (approved):
LGTM

svvarom (approved):

packages/nmp_common/src/nmp/common/controller/controller_manager.py
```diff
@@ -134,11 +137,13 @@ def validate_all_healthy(self, detailed: bool = True) -> tuple[bool, Dict[str, b
                     if not is_healthy:
                         all_healthy = False
                         logger.debug(f"Loop '{name}' is unhealthy")
+                    self._log_health_transition(name, is_healthy, getattr(loop, "unhealthy_reason", None))
                 except Exception as e:
                     if detailed:
                         health_status[name] = False
                     all_healthy = False
                     logger.error(f"Error checking health of loop '{name}': {e}", exc_info=True)
+                    self._log_health_transition(name, False, f"health check raised: {e}")
```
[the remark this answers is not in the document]
benmccown: Skipping this one. That `logger.error` is unchanged from before this PR — I only added the `_log_health_transition` call next to it, so the repeated logging isn't something introduced here. It also only triggers when `loop.is_healthy` itself raises, which is pathological enough that I'd rather it stay loud than be deduplicated.

packages/nmp_common/src/nmp/common/controller/controller.py
```diff
@@ -16,6 +16,59 @@
...
+    collaborators are handed the bound :meth:`emit_heartbeat` method rather than
+    the :class:`Heartbeat` itself, so they report progress without depending on
+    how liveness is measured.
+
+    ``_heartbeat`` is a class-level default so that mixing this in never obliges
+    a subclass to call ``super().__init__()``. :meth:`attach_heartbeat` binds the
+    instance that :class:`Loop` measures; emitting before then is a no-op.
+    """
+
+    _heartbeat: "Heartbeat | None" = None
```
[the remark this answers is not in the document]
benmccown: Fixed. Concrete annotation now, which also matches the repo convention of preferring concrete type hints over string-based ones.

services/core/models/src/nmp/core/models/controllers/deployment_reconciler.py
```diff
@@ -785,21 +799,13 @@ async def _cleanup_model_entities_for_provider(
...
-                    current_providers = model_entity.model_providers or []
-                    if provider_id in current_providers:
-                        updated_providers = [p for p in current_providers if p != provider_id]
-                        await self._models_sdk.models.update(
-                            name=model_name,
-                            workspace=model_workspace,
-                            model_providers=updated_providers,
-                        )
+                    if provider_id in (model_entity.model_providers or []):
+                        self._entity_cache.stage_provider_unlink(model_workspace, model_name, provider_id)
```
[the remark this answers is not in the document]
benmccown: Fixed. Added a `loaded` guard that logs and skips rather than silently dropping the unlink. You're right that nothing enforced the ordering here — and it was inconsistent, since I already had that fallback in `_retrieve_model_entity` and not here.

services/core/models/src/nmp/core/models/controllers/entity_cache.py
```diff
@@ -0,0 +1,225 @@
...
+                else:
+                    await self._update(workspace, name, staged, existing)
+            except Exception:
+                # One entity must not stop the rest; the next pass retries.
+                logger.warning(
+                    "Failed to apply Model Entity changes for %s/%s",
+                    workspace,
+                    name,
+                    exc_info=True,
+                )
```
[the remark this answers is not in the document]
benmccown: Good catch, this was real. Fixed: a failed write now keeps its staged change so a later flush retries it, instead of being discarded. My old comment ("the next pass retries") was only true for provider links, which are recomputed from discovery — not for provider-deletion unlinks, which are derived from a provider that's deleted immediately afterwards.

One design note: naive retention collided with `refresh()`'s guard, since retained failures would make the next refresh raise. `_PendingEntity` now carries an `attempts` counter, so refresh still rejects changes no flush has attempted (staged and forgotten) while replaying ones that were attempted and failed. That's safe because staged changes are differences rather than absolute state, so replaying against a newer snapshot stays correct — there's a test covering exactly that.

services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
```diff
@@ -313,33 +329,68 @@ async def reconcile_model_providers(self, provider_contexts: list[ModelContext])
...
         now = datetime.now(timezone.utc)
 
+        # Read the VirtualModels once for the whole pass. ``existing_vm_names``
+        # answers "does this one already exist" without a create-and-catch-conflict
+        # per served model, and grows as VirtualModels are created so providers
+        # sharing a workspace do not each try the same name. ``vm_snapshot`` stays
+        # as read: orphan cleanup only ever considers VirtualModels that predate
+        # the pass, and keeping it separate means a VirtualModel created here can
+        # never be mistaken for a deletion candidate.
+        vm_snapshot, existing_vm_names = await self._load_virtual_models()
```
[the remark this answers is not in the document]
benmccown: Fixed, and you're right that this was a regression from hoisting the listing out of the cleanup's try/except. `_load_virtual_models` now returns `None` on failure, and the pass skips VirtualModel creation and orphan cleanup while letting discovery, provider status updates, and entity linking continue. Added a test asserting the entity write still happens and no VirtualModel work is attempted.

services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
```diff
@@ -981,7 +1012,12 @@ async def _ensure_passthrough_virtual_model(self, workspace: str, model_name: st
         Args:
             workspace: Workspace of the model entity.
             model_name: Name of the model entity (also used as the VirtualModel name).
+            existing_vm_names: ``(workspace, name)`` of every known VirtualModel.
         """
+        if (workspace, model_name) in existing_vm_names:
+            logger.debug("Passthrough VirtualModel %s/%s already exists", workspace, model_name)
+            return
```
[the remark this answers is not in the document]
benmccown: Fixed. Now says the passthrough is being skipped because the name is already in use, since the existence set deliberately covers user-managed VirtualModels whose `default_model_entity` points elsewhere.

packages/nmp_common/src/nmp/common/controller/controller.py
benmccown: Fair, they were doing too much explaining. Trimmed `Heartbeat`, `HeartbeatMixin`, and `TrackLastExecutionTime` down to the parts that aren't evident from the code — the "only emit after work actually completed" rule and why `_heartbeat` is a class-level default. The rest is gone.

packages/nmp_common/src/nmp/common/controller/controller.py
benmccown: Yes, fairly often. `TrackLastExecutionTime` wraps arbitrary `Controller`s, and plugin controllers built through `plugin_adapter.py` don't mix in `HeartbeatMixin` — nor does anything not touched by this PR. The isinstance check is what keeps those working unchanged: they still get the step-entry timestamp, just not mid-step progress.

packages/nmp_common/src/nmp/common/controller/controller.py
benmccown: Deliberately left at DEBUG. These fire on every readiness probe (~10s), so promoting them would be constant noise. The visible logging is edge-triggered instead, over in `ControllerManager._log_health_transition`: WARNING once when a loop goes unhealthy (with the reason, e.g. `no progress for 44.20s (max allowed: 15.00s)`) and INFO once when it recovers. That's the gap that made the original problem invisible — the only signal was DEBUG.

packages/nmp_common/tests/controller/test_controller.py
benmccown: You were right to flag it, and the real answer was that those tests shouldn't have been using wall-clock time at all. Rewrote them to drive a controlled clock via `time_machine` (already a root dev dependency and an active pytest plugin) — no started threads, no sleeps, no timeouts anywhere in the file. Also reads better: the key test now walks a batch forward in explicit half-window steps and asserts the loop stays healthy the whole way. Runtime dropped ~1.6s to ~0.9s.

Note the pre-existing tests in `test_controller_manager.py` still use real threads and `wait(timeout=2)`; I left those alone as out of scope here.

services/core/entities/src/nmp/core/entities/controllers/workspace_cleanup.py
benmccown: Ha — the repeated `finally: emit_heartbeat()` is admittedly boilerplate. It's explicit on purpose: progress has to be reported only after work actually completes, and a decorator would make it easy to accidentally report progress around a call that blocks, which would defeat the check entirely.

services/core/models/src/nmp/core/models/controllers/deployment_reconciler.py
benmccown: Good question, it's the riskiest part of the change. What's in place: the cache is re-read at the start of each phase that writes entities and flushed at the end, `refresh()` refuses to run while unattempted changes are staged, and reads overlay staged changes so a read-after-write within a phase sees itself. Tests: `test_refresh_after_flush_does_not_reapply_earlier_state` (a link removed in phase one isn't reinstated by phase two — the specific bug the two-phase refresh exists to prevent) and `test_step_refreshes_the_entity_cache_for_each_writing_phase` (pins the refresh/flush ordering so it can't be "optimized" back to one refresh per step).

That said, review found two invalidation-adjacent holes I'd missed, both now fixed: flush not running when a phase raised, and failed writes being dropped rather than retried.

services/core/models/src/nmp/core/models/controllers/entity_cache.py
benmccown: Legitimate concern, and it was a conscious tradeoff. It scales with *total* platform entities, not just the ones the pass cares about. Measured against a workspace with 404 entities: one paginated list at ~0.33s and negligible memory, versus 404 individual reads at ~37.7s. So it's a large win now and fine at this size, but there is a ceiling — around 10k entities it's seconds per pass, and at 100k it'd need changing. The escape hatches if we get there are listing only workspaces that own providers, or filtering the query by the discovered names. Happy to leave a note if you'd like it tracked.

services/core/models/src/nmp/core/models/controllers/entity_cache.py
benmccown: One controller, but more than one writer. The API pods also create Model Entities from user requests and other services, and there are multiple API replicas, so the controller can race a concurrent create. The branch adopts the existing entity instead of erroring.

Worth noting what it deliberately doesn't do: it won't force our creation attributes onto the entity that won already — overwriting someone else's `description` would be wrong. The owning reconciler re-evaluates what's actually missing on a later pass. There's now a test pinning that.

services/core/models/tests/unit/controllers/test_provider_reconciler.py
benmccown: Fixed as part of consolidating the duplicated test scaffolding — `_AsyncPaginator`, the entity factory, and `seed_entity_cache` were re-implemented in three modules with subtly different `model_copy` behaviour. They're now single definitions in `conftest.py`, and that comprehension is a plain named function.

services/core/models/src/nmp/core/models/controllers/entity_cache.py
mckornfield: lol seems like a problem we'll know about when we get there

PR #956, opened 2026-07-28
feat: usage telemetry

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Added anonymous usage telemetry for CLI command outcomes, onboarding/setup steps, and terminal platform job results.
  * Added global `--no-telemetry` to disable telemetry for the current invocation.
* **Documentation**
  * Added “Telemetry and Privacy” content (and linked it from the README/navigation), including what’s collected/excluded and how to opt out.
  * Updated CLI docs/global options to include `--no-telemetry`.
  * Refreshed setup guidance with a NeMo Platform artifacts in NGC reference.
* **Tests**
  * Expanded telemetry test coverage for command, job, and onboarding events, including opt-out and help behavior.
  * Updated telemetry contract expectations to schema version 1.10.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
@@ -42,6 +42,8 @@ from nemo_platform_ext.cli.commands.skills.base import Scope, Skill
 from nemo_platform_ext.cli.commands.skills.registry import get_installer, load_skills
 from nemo_platform_ext.cli.core.context import CLIContext
 from nemo_platform_ext.cli.core.errors import handle_errors
+from nemo_platform_ext.cli.telemetry import emit
+from nemo_platform_ext.cli.telemetry.events import OnboardingStepEvent, TaskStatusEnum
 from nemo_platform_ext.client.tls import client_verify_from_env
 from nemo_platform_ext.config.config import Config
 from nemo_platform_ext.config.models import DEFAULT_BASE_URL, ConfigFile, ConfigParams, LocalServicesConfig
@@ -133,6 +135,14 @@ KNOWN_PROVIDERS: tuple[KnownProvider, ...] = (
 )
 
 _KNOWN_PROVIDERS_BY_NAME: dict[str, KnownProvider] = {p.name: p for p in KNOWN_PROVIDERS}
+
+
+def _provider_type_for_connection(name: str, host_url: str) -> str:
+    known = _KNOWN_PROVIDERS_BY_NAME.get(name)
+    if known is not None and known.host_url.rstrip("/") == host_url.rstrip("/"):
+        return known.name
+    return "custom"
+
 
 # ---------------------------------------------------------------------------
 # Onboarding paths — shown after setup completes
@@ -536,7 +546,21 @@ def _create_provider(
             kwargs["required_extra_headers"] = {header_name.strip(): header_value.strip()}
     if default_extra_headers:
         kwargs["default_extra_headers"] = default_extra_headers
-    client.inference.providers.create(**kwargs)
+    provider_type = _provider_type_for_connection(name, host_url)
+    try:
+        client.inference.providers.create(**kwargs)
+    except Exception:
+        emit.emit_event(
+            OnboardingStepEvent(
+                step="provider_connected", task_status=TaskStatusEnum.ERROR, provider_type=provider_type
+            )
+        )
+        raise
+    emit.emit_event(
+        OnboardingStepEvent(
+            step="provider_connected", task_status=TaskStatusEnum.COMPLETED, provider_type=provider_type
+        )
+    )
 
 
 def _update_provider(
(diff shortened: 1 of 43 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/help_formatter.py
```diff
@@ -47,36 +47,86 @@ def main(
...
+        import time
 
+        from nemo_platform_ext.cli.core.errors import handle_exception
+        from nemo_platform_ext.cli.telemetry import runtime
+        from nemo_platform_ext.cli.telemetry.events import TaskStatusEnum
+
+        # Only the root group (the app-level ManifestBackedNmpGroup) drives telemetry.
+        # Nested groups never have main() called during dispatch, but this guard keeps
+        # the single-emit contract explicit and cheap without an import cycle.
+        is_telemetry_root = type(self).__name__ == "ManifestBackedNmpGroup"
```
anastasia-nesterenko: ok with string name check?

anastasia-nesterenko (approved):

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -563,13 +587,61 @@ def _update_provider(
 _NON_COMPLIANT_MARKER = "Non-OpenAI compliant"
 
 
+def _bucket_model_count(count: int) -> str:
+    """Bucket a discovered-model count into a coarse range for telemetry.
+
+    Keeps the emitted value low-cardinality (no exact counts leave the machine).
+    """
+    if count <= 0:
```
tylersbray, line 609: This is the sort of thing where if it was even just two years ago everyone would still be reading this and Mike K would be all, "You could use match here."

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/waiters.py
```diff
@@ -55,6 +57,89 @@ def _status_text(status: Any) -> str:
...
+            "error": TaskStatusEnum.ERROR,
+            "cancelled": TaskStatusEnum.CANCELED,
+        }
+        task_status = status_map.get(status, TaskStatusEnum.UNDEFINED)
+
+        details = getattr(job_status, "status_details", None)
+        if not isinstance(details, dict):
+            details = {}
+
+        # job_type is the job's operation entry point(s) (for example "auditor.audit"), read
```
tylersbray: This one may need scrubbing:

[P2] Do not treat job step names as safe plugin.operation IDs

Where: [waiters.py](packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/waiters.py) _emit_job_run_event (~L112–133)

Evidence: Code/comments/tests claim step names are static plugin.operation values (auditor.audit). Real in-tree compilers use free-form names allowed by PlatformJobStepSpec (audit-job, evaluate, evaluate-suite). plugins = ep.split(".", 1)[0] then copies the full step name when there is no dot.

Why it matters: Contradicts the privacy disclosure (“no user identifiers”) and pollutes analytics. Custom jobs can encode customer/project strings in step names.

Ask: Sanitize before emit — allow-list known operations, emit resource_label / "custom", or another low-cardinality bucket. Fix the misleading comment and align test_job_events.py fixtures with real step-name shapes.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -536,7 +546,21 @@ def _create_provider(
...
+        )
+        raise
+    emit.emit_event(
+        OnboardingStepEvent(
+            step="provider_connected", task_status=TaskStatusEnum.COMPLETED, provider_type=provider_type
+        )
+    )
 
 
 def _update_provider(
```
tylersbray, line 566: [P3] Re-run setup via _update_provider skips provider_connected

Where: [setup.py](packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py) _update_provider (~L566)

Creates are instrumented; updates are not. Idempotent/reconfig paths drop out of the onboarding funnel.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -1335,6 +1417,31 @@ def _deploy_demo_agent(
...
+    deployment outcome as ``agent_deployed`` on success, ERROR (re-raised) on failure.
+    """
+    try:
+        deployed = _deploy_demo_agent_impl(base_url, workspace, config_path, default_model, headers=headers)
+    except Exception:
+        emit.emit_event(
+            OnboardingStepEvent(step="agent_deployed", task_status=TaskStatusEnum.ERROR, agent_deployed=False)
+        )
+        raise
+    emit.emit_event(
```
tylersbray: [P3] Failed demo deploy reported as COMPLETED

Where: _deploy_demo_agent (~L1433–1435)

_deploy_demo_agent_impl returns False on failed/timeout, but the wrapper emits COMPLETED with agent_deployed=False. Confirm intentional (boolean outcome) vs map those paths to ERROR.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/help_formatter.py
```diff
@@ -47,36 +49,94 @@ def main(
         standalone_mode: bool = True,
         **extra: Any,
     ) -> Any:
-        """Override main to use custom error handling."""
-        from nemo_platform_ext.cli.core.errors import handle_exception
+        """Override main to use custom error handling and emit one telemetry event.
+
+        This is the single path every command type flows through (hand-registered,
```
tylersbray, line 56: Neat comment 👍

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/help_formatter.py
```diff
@@ -47,36 +49,94 @@ def main(
...
+        # called during dispatch, but this guard keeps the single-emit contract explicit.
+        is_telemetry_root = bool(getattr(self, "_nmp_emit_command_invoked", False))
+        help_requested = False
+        if is_telemetry_root:
+            runtime.reset()
+            # Reading help (e.g. ``nemo docs --help``) is not usage, so it must not emit
+            # a command_invoked event. Detect a help request from the resolved args; the
+            # help option names are the same ones every command registers via
+            # ``_context_settings_with_help``.
+            import sys
```
tylersbray: I remember seeing an embedded 'import nmp_platform' in the last one and was like, "maybe it's lazy for a reason," it looked like it was for a really simple check and maybe it's a heavyweight import?  But now I see sys and time imported down here and I'm wondering if you can hoist any of these.

tylersbray (approved):
Nice, thanks.  The maybe step name leak is the main thing that should be double checked, the extra setup.py cleanup would be nice, added some nitpicks just to make myself read it, no pressure.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
mikeknep: This is my bat signal 😂  Although funny enough match _wouldn't_ work here—you can't use comparison operators in case expressions!

PR #959, opened 2026-07-28
chore: actions hardening

### LINKED ISSUE #82: chore(ci): Repo-wide SHA pinning of actions/*, persist-credentials: false, and Dependabot config
## Summary

Pin all `actions/*` action references across `.github/workflows/` to full commit SHAs instead of floating version tags (e.g., `@v6`, `@v8`), add `persist-credentials: false` to every `actions/checkout` step, and add a Dependabot config (`.github/dependabot.yml`) so the pinned SHAs auto-update.

## Motivation

Currently, all jobs in `ci.yaml` (and other workflows) use major-version tags for `actions/checkout`, `actions/upload-artifact`, `actions/download-artifact`, etc. zizmor flags these as `unpinned-uses` and `artipacked` violations. Fixing only a single job (e.g., the new `wheel-test` job added in #78) would create inconsistency, so the fix should be applied uniformly across the repo.

## Work items

- [ ] Audit all `.github/workflows/*.yaml` and `.github/actions/**/action.yaml` for `actions/*` references using floating tags
- [ ] Replace each with the corresponding full commit SHA
- [ ] Add `persist-credentials: false` to every `actions/checkout` step
- [ ] Add `.github/dependabot.yml` with `package-ecosystem: github-actions` so Dependabot opens PRs to keep SHAs current

## References

- Flagged by zizmor (`unpinned-uses`, `artipacked`) during review of #78
- Requested by @matthewgrossman in https://github.com/NVIDIA-NeMo/nemo-platform/pull/78#discussion_r3313503110

mckornfield:
* closes #82

## Summary by CodeRabbit

* **Chores**
  * Added automated weekly updates for GitHub Actions dependencies, with eligible updates explicitly filtered and grouped.

* **Security**
  * Improved CI and security workflow credential handling by disabling credential persistence after repository checkout, reducing exposure during subsequent steps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,19 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    allow:
+      # Keep automated action bumps limited to org-approved action owners/actions.
+      - dependency-name: "actions/*"
+      - dependency-name: "github/codeql-action*"
+      - dependency-name: "docker/*"
+      - dependency-name: "azure/setup-helm"
+      - dependency-name: "trufflesecurity/trufflehog"
     groups:
-      github-actions:
+      approved-github-actions:
+        applies-to: "version-updates"
         patterns:
-          - "*"
+          - "actions/*"
+          - "github/codeql-action*"
+          - "docker/*"
+          - "azure/setup-helm"
+          - "trufflesecurity/trufflehog"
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
```
crookedstorm: I actually had a ticket to do this, so thanks! However, we must be more picky than this. Approved actions going forward will only match specific commit shas. Some actions cannot be updated without tickets to IT, in otherwords.

any github-owned actions, docker actions (fully "Verfied" by Github--any verified actions count) and specific other ones I can look up, should be in the list of patterns. Flox ones, the ones we use for release labels and some others cannot be.

mckornfield: ok, we'll see how decent the agent is at populating this guy

crookedstorm (approved):
Looks right to me

PR #964, opened 2026-07-29
fix(evaluator-sdk): make Harbor's per-trial resume reachable with agent_dir set

ngoncharenko:
Fixes [AALGO-430](https://linear.app/nvidia/issue/AALGO-430/harboragenttaskrunner-nondeterministic-scoped-agent-import-path-makes). Refs [AALGO-427](https://linear.app/nvidia/issue/AALGO-427/close-the-harboragenttaskrunner-cache-gap-results-persist-but-there-is).

Targets `ngoncharenko/aalgo-312-wire-eval-into-optimizer`, which this branch builds on.

## The bug

`scoped_harbor_agent_import` minted the synthetic package name with a fresh `uuid4()` on every call. That string becomes `AgentConfig.import_path` and so part of Harbor's `JobConfig`, which Harbor compares field-by-field before resuming a job directory. The comparison therefore failed on *every* rerun, Harbor raised `FileExistsError` instead of resuming, and its per-trial resume was unreachable for any caller that sets `agent_dir` — which is every Experimentalist run.

AALGO-427 had papered over this by discarding the job directory on every cache miss (`force_rerun` whenever `agent_dir` was set). That removed the crash by never resuming.

 ### Example
Run 1
Agent unchanged. UUID happens to be deadbeef:
```
import_path = "_nemo_evaluator_harbor_agents.agent_deadbeef.harbor_wrapper:WrappedAgent"
```
Harbor writes jobs/resume-job/config.json with that path, finishes trial 1 of 2, then the process exits mid-run (or you only asked for 1 attempt and later bump to 2).

Run 2 - same agent files, same job dir
New UUID cafef00d:
```
import_path = "_nemo_evaluator_harbor_agents.agent_cafef00d.harbor_wrapper:WrappedAgent"
```
Harbor compares the new JobConfig to the saved one:

```
saved:  ...agent_deadbeef...
new:    ...agent_cafef00d...
```
→ configs differ → Harbor raises FileExistsError: ... resumed with a different config
→ cannot resume; completed Docker work is stranded.

Nothing about the agent changed. Only the random suffix did.

After the fix
Suffix = content digest of agent_dir:

# both runs, identical agent files:
```
_nemo_evaluator_harbor_agents.agent_3f8a91c2b4e1.harbor_wrapper:WrappedAgent
```
Same contents → same path → Harbor’s equality check passes → missing trials resume.

Edit the agent (wrapper.py changes) → digest changes → path changes → Harbor correctly refuses resume (stale results for a different agent).

Why UUID was there originally?
UUID avoided collisions if two different agent folders named agent were imported in one process. That isolation goal was valid; the bug was making the name part of a persisted config that must be stable across reruns. Content-addressing keeps isolation and stability.

Before the fix, Experimentalist always set agent_dir, so every eval with a custom agent hit this path and could never use Harbor’s per-trial resume.

## The fix

**Content-addressed import path.** The suffix now comes from a content digest of `agent_dir`, so an unchanged agent produces a stable `JobConfig` and an edited one invalidates the job dir on Harbor's own terms. Identical contents share a package name, so the `sys.modules` injection is refcounted and torn down on the *last* scope exit. The digest takes the same `jobs_dir` exclusion `_cache_stamp` uses, so a `jobs_dir` nested under `agent_dir` can't shift the import path as results accumulate.

**Handling Harbor's refusal.** Removing the unconditional discard exposed a second problem: the SDK cache stamp is deliberately looser than Harbor's `JobConfig.__eq__`. `quiet`, `n_concurrent_trials` and the `task_names` filter change the JobConfig without changing the results, and keying on them would discard every completed candidate when an experiment resumes on a box with a different core count (the Experimentalist defaults `n_concurrent_trials` to `os.cpu_count()`). So a directory that passes the stamp can still be refused — and Harbor refuses by raising.

`_build_native_job` now identifies that refusal *positively* — bare message, no `errno`, naming this job dir — discards, re-runs, and logs which field forced it (`n_concurrent_trials: 10 -> 4`). Any other `FileExistsError` propagates untouched rather than costing a job directory.

## Why not align the config instead

Mirroring Harbor's compared set would fix the crash at the source but regress two intentional behaviours: a core-count change would discard completed candidates, and keying on `task_names` would break subset-of-a-cached-job hits that `_stamp_coverage` exists to support. The stamp answers "did these inputs produce these results?"; Harbor answers "can I resume this directory?" — different questions, and the SDK's is the right one to drive a cache. The divergence is now documented and guarded rather than removed.

## Tests

- Real-Docker e2e for the resume this change exists to enable (restoring `uuid4` makes it fail with the original `FileExistsError`)
- Refusal predicate pinned against both real Harbor messages, an OS-level `EEXIST`, and a refusal naming a different job dir
- Regression: a job dir that exists plus an unrelated mid-run `FileExistsError` must propagate and leave the directory intact
- Drift guard: every `JobConfig` field must be classified as Harbor-ignored, never-set-by-the-SDK, stamp-covered, or knowingly-unkeyed-with-a-reason — a Harbor upgrade adding a compared field fails the build
- Wording pin so a Harbor reword of its refusal surfaces at upgrade time
- Import-path stability, change detection, agent isolation, overlapping-scope survival, refcount teardown

All guards mutation-checked: each fix reverted individually makes the intended test fail.

## Verification

```
uv run --frozen pytest packages/nemo_evaluator_sdk/tests/agent_eval -q   # 319 passed, 1 skipped (Docker e2e included)
uv run --frozen pytest plugins/nemo-experimentalist/tests -q             # 611 passed, 2 skipped
uv run ruff check packages/nemo_evaluator_sdk                            # clean
uv run --frozen ty check packages/nemo_evaluator_sdk                     # 122 diagnostics, unchanged from base
```

The vendored SDK is a separate commit, regenerated with `make vendor` — import-path rewrites only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Harbor job resumption when custom agent directories are used.
  * Preserves completed work instead of unnecessarily discarding cached jobs.
  * Automatically retries jobs when Harbor explicitly refuses to resume, while preserving unrelated errors and valid job data.
  * Improved handling of agent changes and partial job recovery.

* **Tests**
  * Added comprehensive coverage for caching, resumption, import isolation, error handling, and end-to-end recovery scenarios.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
@@ -47,7 +47,6 @@ from datetime import datetime, timezone
 from pathlib import Path
 from types import ModuleType
 from typing import Any
-from uuid import uuid4
 
 from nemo_evaluator_sdk.agent_eval.results import AgentEvalResult
 from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus
@@ -73,6 +72,13 @@ _TASK_TEMPLATE_DIRNAME = "task_template"
 _AGENT_IMPORT_ROOT = "_nemo_evaluator_harbor_agents"
 # Guards the sys.modules mutation while injecting/removing scoped agent packages.
 _IMPORT_LOCK = threading.Lock()
+# Open scopes per content-addressed agent package. Identical agent contents share a
+# package name, so teardown must wait for the last scope rather than the first.
+_AGENT_PACKAGE_REFCOUNTS: dict[str, int] = {}
+# Characters of the agent-content digest used to disambiguate the package name. Long
+# enough that distinct agents don't collide; short enough to keep import paths (and
+# Harbor's persisted JobConfig) readable.
+_IMPORT_DIGEST_CHARS = 12
 # Records which inputs produced a job dir, so a rerun can tell a reusable cache from
 # a stale one. A file, not a directory: Harbor rmtree's stray directories in a job dir.
 CACHE_STAMP_FILENAME = ".nemo-eval-harbor-cache.json"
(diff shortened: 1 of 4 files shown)


ngoncharenko:
@claude review the PR

ngoncharenko:
@coderabbitai review the PR

SandyChapman (approved):
lgtm

PR #976, opened 2026-07-29
fix(ci): make stable GitHub releases idempotent

svvarom:
Stable subset and custom runs are also used to finish or retry individual artifacts. The workflow previously keyed GitHub tag/release creation to `release-scope: all`, so the successful 0.3.0 custom run skipped its tag and GitHub Release.

This makes `stable` the release intent: every successful non-dry-run stable run converges on its version tag and GitHub Release, regardless of artifact scope. Retries are idempotent:

- A version tag that points to another SHA fails clearly.
- An existing GitHub Release succeeds as a no-op.
- A matching tag without a GitHub Release creates the missing release.

The extra `create-github-release` checkbox and its planner wiring are removed.

Incident: https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/30477833073/job/90703676069

## Human attention needed

- **Decision:** Treat every stable run as an instruction to create or confirm the version tag and GitHub Release.
- **Read carefully:** Stable subset and custom runs now create the tag/release for a new version; retries of an existing version are safe no-ops.
- **Safe to skim:** Documentation.
- **Proof:** `actionlint .github/workflows/release.yaml` and `git diff --check` pass.
- **Biggest risk:** Selecting `stable` for a subset with a brand-new version now cuts the version tag/release. This is intentional; `release-scope` controls artifacts only.

## Validation

- `actionlint .github/workflows/release.yaml`
- `git diff --check`
- Verified stable `all`, `wheels`, `containers`, `helm`, and `custom` scopes share the same release gate.
- Verified nightly, dry-run, and failed polling skip GitHub Release creation.
- Verified the release step no-ops for an existing release, repairs a matching tag-only state, and rejects a tag/SHA mismatch.

## Summary by CodeRabbit

* **Bug Fixes / Improvements**
  * GitHub Release creation is now consistently handled for all successful non-dry-run stable releases, regardless of release scope.
  * The “Create GitHub Release” step is more guarded: it checks that the existing tag points to the expected commit and avoids recreating a release if one already exists.

* **Documentation**
  * Updated release documentation to reflect the always-on stable release tag/GitHub Release behavior.
  * Removed guidance about the prior explicit GitHub release setting and added updated stable one-container/one-wheel examples.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -60,11 +60,6 @@ on:
         required: false
         type: boolean
         default: false
-      create-github-release:
-        description: "Create the Git tag and GitHub Release for a custom stable release. Stable all releases do this automatically."
-        required: false
-        type: boolean
-        default: false
       helm-version:
         description: "Optional exact Helm chart version override for stable Helm-only releases."
         required: false
@@ -164,7 +159,6 @@ jobs:
       helm_version_override: ${{ steps.plan.outputs.helm_version_override }}
       update_ngc_metadata: ${{ steps.plan.outputs.update_ngc_metadata }}
       send_notifications: ${{ steps.plan.outputs.send_notifications }}
-      create_github_release: ${{ steps.plan.outputs.create_github_release }}
       dry_run: ${{ steps.plan.outputs.dry_run }}
     steps:
       # This job is intentionally the first gate. It resolves the user's
@@ -195,9 +189,6 @@ jobs:
             const releaseScope = isManual ? (inputs["release-scope"] || "all") : "all";
             const updateNgcMetadata = isManual && inputs["update-ngc-metadata"] === "true";
             const sendNotifications = inputs["send-notifications"] !== "false";
-            const createGitHubReleaseRequested = (
-              isManual && inputs["create-github-release"] === "true"
-            );
             const dryRun = isManual && inputs["dry-run"] === "true";
             const helmVersionOverride = isManual ? (inputs["helm-version"] ?? "").trim() : "";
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -60,6 +60,11 @@ on:
         required: false
         type: boolean
         default: false
+      create-github-release:
```
mckornfield: each day, the parameter list grows

crookedstorm: Which is fine if the default is "do what we want most of the time"

.github/workflows/release.yaml
```diff
@@ -1100,7 +1118,7 @@ jobs:
     needs: [plan-release, poll-final-release]
     if: >-
       needs.plan-release.outputs.release_type == 'stable' &&
-      needs.plan-release.outputs.release_scope == 'all' &&
+      needs.plan-release.outputs.create_github_release == 'true' &&
```
crookedstorm: Why do we need this at all? I don't see any reason not to just remove this and not introduce the new parameter.

An improvement would be to also just output what it would have created with ACT="true", but I emphatically reject needing to explicitly ask for release when I said make a stabe release already. :)

.github/workflows/release.yaml
crookedstorm: And I now believe this does NOT do that. This is the opposite of that, imo. Commented further below. I could be wrong.

.github/workflows/release.yaml
svvarom: But we can use stable custom runs to retry failed artifacts too, so we need a way to say whether that run should also cut the tag/release.

svvarom: Otherwise it'll fail by default if the tag already exists

crookedstorm: What about checking for the tag first and making it idempotent instead? This adds a random clickbox for what it should do by default.

svvarom: Updated the PR

.github/workflows/release.yaml
svvarom: dropped the input

PR #978, opened 2026-07-29
fix(api): declare the entity-store name pattern on create DTOs [ASTD-349]

marcusds:
## Problem

The OpenAPI spec advertises a name pattern the platform does not accept.

Create-request DTOs declare `^[\w\-.]+$` (max 255), but the entity store enforces the stricter RFC-1035-ish `NAME_PATTERN` downstream:

```
^[a-z](?!.*--)[a-z0-9\-@.+_]{1,62}(?<!-)$
```

So `Sparl` or `My_Provider` validates against the published spec, passes any generated client's validation, and then fails with a 422. That hits every SDK consumer, not just Studio.

Raised by @steramae-nvidia on #919:

> Is this a problem for the openapi schema? Anyone relying on the openapi schema should expect that the regex pattern in the spec should be valid.

#919 works around this on the frontend with a hand-mirrored copy of `NAME_PATTERN`. That mirror exists only because the spec is wrong.

## Change

Two commits.

### 1. Declare the real pattern on the create DTOs

| DTO | Before | After |
|---|---|---|
| `CreateModelProviderRequest.name` | `^[\w\-.]+$`, max 255 | `NAME_PATTERN`, max 63 |
| `CreateFilesetRequest.name` | `^[\w\-.]+$`, max 255 | `NAME_PATTERN`, max 63 |
| `PlatformSecretCreateRequest.name` | `field_validator` only — no pattern in the spec at all | `pattern=`, max 63 |

Three things worth calling out:

**`regex_engine="python-re"` is required.** `NAME_PATTERN` uses lookahead and lookbehind; Pydantic's default Rust engine rejects them outright. The entity-store schemas already set this — without it the models don't build.

**`max_length` 255 → 63.** The regex caps at 63 by construction (`^[a-z]` + `{1,62}`), so the old pair contradicted itself.

**`@` is legal under `NAME_PATTERN`.** Secrets previously rejected `a@b` via its own validator; it now accepts it, matching what the entity store always allowed. A loosening in one direction alongside the tightening.

### 2. Collapse the mirrored constants

The first commit made an existing problem worse, so the second one fixes it. `NAME_PATTERN` was copy-pasted verbatim in **four** Python modules: `nmp_common`, plus `files/types.py`, `secrets/types.py`, and `jobs/spec.py` inside `nemo_platform_plugin`. Each carried a comment explaining it was inlined to avoid an `nmp_common` dependency, and `jobs/spec.py` cited `files/types.py` as the precedent — the duplication was self-propagating.

The dependency only runs one way: `nmp_common/pyproject.toml` depends on `nemo-platform-plugin`, and `nmp_common/entities/client.py` imports `EntityBase` from it. So the plugin cannot import `nmp_common` without a cycle, which makes the plugin the correct home — as the `no-nmp-common-in-plugins` pre-commit hook's own comment prescribes.

New leaf module `nemo_platform_plugin/entity_naming.py` holds the single definition and imports nothing, so modules that need to stay leaf nodes still can. `nmp_common` re-exports it, leaving every existing `constants.NAME_PATTERN` call site untouched.

Regenerating the OpenAPI spec after this commit produces **no diff** — it's a pure refactor.

## Testing

- All three DTOs verified directly: `Sparl`, `my--provider`, `myprovider-`, `1provider`, 64-char, `My_Provider` rejected; `ok-name`, `a@b` accepted
- **4895 Python tests pass** after the refactor — plugin, secrets, files, models, entities, jobs, nmp_common. A wider sweep including evaluator, data-designer, and ext also passes, aside from 8 pre-existing failures in `nemo_platform_ext` daemonize/socket tests plus 2 evaluator errors, all confirmed identical on a clean tree
- Re-export identity asserted: `nmp.common.entities.constants.NAME_PATTERN is nemo_platform_plugin.entity_naming.NAME_PATTERN`
- 44 Studio tests pass on the three name-entry forms
- `ruff`, `ruff format`, `ty` clean on changed files
- Web SDK regenerated: `filesCreateFilesetBodyNameMax = 63` and the new regex. That output is gitignored, so nothing to commit

Two secrets tests changed. `test_create_secret_with_invalid_name_returns_friendly_error` asserted a message that no longer exists and listed `a@b` as invalid — replaced with a reject-list plus an accept-list. The two uppercase regression tests in `test_secrets_with_auth.py` were failing because the shared request model now rejects locally before the HTTP call; they use `model_construct` to bypass it so the original 500-vs-422 server-side coverage stays intact.

## Known remaining drift (not fixed here)

`packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/models.py:31` carries a hand-rewritten lookaround-free variant commented "Keep this aligned with `nmp.common.entities.constants.NAME_PATTERN`". It has drifted:

| input | entity store | evaluator SDK |
|---|---|---|
| `a-b-b-…` (81 chars) | reject | **accept** |
| `a-b-b-…` (125 chars) | reject | **accept** |

Each hyphen branch consumes two characters per repetition, so `{1,62}` admits up to 125 characters instead of 63. `nemo_evaluator_sdk` has no dependency on `nemo_platform_plugin`, so fixing it needs either a new dependency or a test asserting equivalence — out of scope here, happy to file it.

On the frontend, `main` today has two copies (`filesetName.ts` and `CreateSecretModal/constants.ts`). #919 collapses both into one — `entityName.ts` — so once that merges there is a single frontend copy left. It can drop its hand-written regex in favour of the generated `filesCreateFilesetBodyNameRegExp` once this PR lands; the per-rule error messages and the sanitizer's character classes have no generated equivalent and stay.

## Follow-up

`web/packages/common/src/utils/entityName.ts` (added in #919) can be reduced to the generated schema.

## Summary by CodeRabbit

* **Validation Improvements**
  * Tightened `name` validation across filesets, model providers, platform secrets, and related entity inputs.
  * Enforced max length of **63**, with **lowercase-first** names, approved characters including **`@ . + _`**, plus existing constraints: **no consecutive hyphens** and **no trailing hyphen**.
  * Updated OpenAPI and CLI/help text to reflect the stricter, consistent rules.
* **Tests**
  * Expanded API boundary tests for broader invalid-name rejection.
  * Added/strengthened checks to ensure validation behavior results in consistent **HTTP 422** responses for invalid inputs, while allowing valid entity-store-style names.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
@@ -20,6 +20,6 @@ NAME_MAX_LENGTH = 63
 
 NAME_PATTERN_DESCRIPTION = (
     "Name must start with a lowercase letter, be 2-63 characters, "
-    "and contain only lowercase letters, digits, and hyphens "
+    "and contain only lowercase letters, digits, hyphens, and @ . + _ "
     "(no consecutive hyphens, cannot end with a hyphen)."
 )
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""The single definition of the entity name rule.
+
+This lives in ``nemo_platform_plugin`` rather than ``nmp_common`` because the
+dependency runs that way — ``nmp_common`` depends on this package, so anything
+here is importable from both sides. ``nmp.common.entities.constants`` re-exports
+these names.
+
+Imports nothing, so modules that need to stay leaf nodes can still use it.
+
+Any Pydantic model declaring ``pattern=NAME_PATTERN`` must also set
+``model_config = ConfigDict(regex_engine="python-re")``; the pattern uses
+lookaround, which Pydantic's default Rust engine rejects.
+"""
+
+# RFC 1035 compliant pattern with temporary support for special characters.
+# Allows lowercase letters, digits, hyphens, and temporarily: @, ., +, _
+# - Must start with a lowercase letter [a-z]
+# - Length: 2-63 characters
+# - No consecutive hyphens (--)
+# - Must not end with a hyphen
+# TODO(#3530): Remove @, ., +, _ once versioning is implemented and predefined target names (e.g., llama-3.2-3b-instruct@v1.0.0+A100) are updated.
+NAME_PATTERN = r"^[a-z](?!.*--)[a-z0-9\-@.+_]{1,62}(?<!-)$"
+
+NAME_MAX_LENGTH = 63
+
+NAME_PATTERN_DESCRIPTION = (
+    "Name must start with a lowercase letter, be 2-63 characters, "
+    "and contain only lowercase letters, digits, and hyphens "
+    "(no consecutive hyphens, cannot end with a hyphen)."
+)
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
@@ -20,6 +20,6 @@ NAME_MAX_LENGTH = 63
 
 NAME_PATTERN_DESCRIPTION = (
     "Name must start with a lowercase letter, be 2-63 characters, "
-    "and contain only lowercase letters, digits, hyphens, and @ . + _ "
+    "and use lowercase letters, digits, hyphens, and dots "
     "(no consecutive hyphens, cannot end with a hyphen)."
 )
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
@@ -1,18 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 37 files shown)


packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
```diff
@@ -0,0 +1,25 @@
...
+Models using ``pattern=NAME_PATTERN`` need ``regex_engine="python-re"``.
+"""
+
+# RFC 1035 compliant pattern with temporary support for special characters.
+# Allows lowercase letters, digits, hyphens, and temporarily: @, ., +, _
+# - Must start with a lowercase letter [a-z]
+# - Length: 2-63 characters
+# - No consecutive hyphens (--)
+# - Must not end with a hyphen
+# TODO(#3530): Remove @, ., +, _ once versioning is implemented and predefined target names (e.g., llama-3.2-3b-instruct@v1.0.0+A100) are updated.
```
mckornfield, line 16: any reason not to just leave off `@.+` ? seems like a migration headache

mckornfield (commented):
feel like we should leave off those special characters if versioning is going to happen, even _eventually_

packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
```diff
@@ -0,0 +1,33 @@
...
+NAME_PATTERN = r"^[a-z](?!.*--)[a-z0-9\-@.+_]{1,62}(?<!-)$"
+
+NAME_MAX_LENGTH = 63
+
+NAME_PATTERN_DESCRIPTION = (
+    "Name must start with a lowercase letter, be 2-63 characters, "
+    "and contain only lowercase letters, digits, and hyphens "
+    "(no consecutive hyphens, cannot end with a hyphen)."
+)
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in 8715b68dad."

packages/nemo_platform_plugin/src/nemo_platform_plugin/entity_naming.py
marcusds: Dropepd `@+`, kept `.` since is used.

maxdubrinsky (approved):

PR #979, opened 2026-07-29
feat(nemo-agents): serve Fabric agents via Docker and Kubernetes deployments

tylersbray:
## Summary
- Extend `DeploymentsRunnerBackend` so `nemo-agents-spec-v1` agents compile to the Platform Fabric FastAPI shim (`python -m nemo_agents_plugin.fabric.server`) for Docker and Kubernetes, with `agent.yaml` injection/mount and `/health` readiness.
- Add `rewrite_fabric_config_base_urls()` so Fabric model/harness IGW URLs are rebased for container reachability (including auth-proxy), parallel to the existing NAT rewriter.
- Leave NAT `nat start fastapi` container behavior unchanged; add unit coverage for Fabric docker/k8s compile, rewrite, create, and auth-proxy paths.

Closes AIRCORE-947.

## Test plan
- [x] `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_runner_deployments.py -v` (44 passed)
- [x] Gateway unit tests still green with deployment suite
- [x] `ty check` / pre-commit hooks on commit
- [x] CI on this PR
- [ ] Live Fabric container deploy once AIRCORE-902 ships Fabric in the default image (full matrix: AIRCORE-960)

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying Fabric agent configurations in Docker and Kubernetes.
  * Fabric deployments now start with the correct Fabric server entrypoint and configuration wiring for the target platform.
  * Automatically rebases Fabric model and harness `base_url` values to the inference gateway host when applicable.
  * When enabled, Fabric deployments can route `base_url` rewriting through an auth proxy.
* **Bug Fixes**
  * Clarified deployment configuration descriptions for the agent server listen port and the configuration mount path behavior for Fabric vs workflow deployments.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import asyncio
 import copy
 import logging
+import shlex
 import time
 from pathlib import PurePosixPath
 from typing import Any
@@ -59,6 +60,7 @@ _HTTP_PORT_NAME = "http"
 _PLUGIN_WHEELS_VOLUME = "plugin-wheels"
 _PLUGIN_WHEELS_MOUNT = "/opt/nemo/plugin-wheels"
 _NAT_CONFIG_ENV = "NAT_CONFIG_PATH"
+_NAT_CONFIG_YAML_ENV = "NAT_CONFIG_YAML"
 _AGENT_CONFIG_YAML_ENV = "AGENT_CONFIG_YAML"
 _AGENT_CONFIG_PATH_ENV = "AGENT_CONFIG_PATH"
 _FABRIC_SERVER_MODULE = "nemo_agents_plugin.fabric.server"
@@ -224,6 +226,17 @@ def _fabric_server_cli_args(*, config_path: str, port: int) -> list[str]:
     ]
 
 
+def _materialize_config_and_exec(*, config_path: str, yaml_env: str, argv: list[str]) -> list[str]:
+    """Return ``sh -c`` args that write the config from *yaml_env*, then exec *argv*.
+
+    Docker mode needs this because the docker backend does not mount ``config_files``.
+    Paths and argv are shell-escaped so spaces/metacharacters cannot break the script.
+    """
+    quoted_path = shlex.quote(config_path)
+    quoted_argv = " ".join(shlex.quote(arg) for arg in argv)
+    return [f'mkdir -p "$(dirname {quoted_path})" && printf "%s" "${yaml_env}" > {quoted_path} && exec {quoted_argv}']
+
+
 def executor_for_mode(config: DeploymentsRunnerConfig, mode: DeploymentMode) -> str | None:
     """Resolve the named deployments-plugin executor for *mode*."""
     if mode == "docker":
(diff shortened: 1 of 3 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -311,7 +396,7 @@ def build_deployment_config(
         update={
             "init_containers": init_containers,
             "config_files": [
-                ConfigFile(path=config_mount_path, content=nat_yaml),
+                ConfigFile(path=config_path, content=config_yaml),
```
mmogallapalli, line 405: I think this is fine for an initial milestone of getting an inline config example working on docker/k8s. For configs that reference prompt files, skills paths, etc that are supported on agent.yaml, we'll need a followup that stages those artifacts to be accessible in the deployed containers/pod

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
mmogallapalli: nit: since we now use this method for both nat and fabric backed configs, should we rename `nat_config` to something more generic, i.e. `agent_config`

tylersbray: Done in e3378342d4 — renamed to `agent_config` (param, docstring, and all call sites including tests). Left `rewrite_config_base_urls(nat_config, ...)` as-is since that one really is NAT-only; Fabric goes through the separate `rewrite_fabric_config_base_urls`.

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
tylersbray: Agreed — this milestone only handles the inline config payload. Configs that reference prompt files, skills paths, or other on-disk artifacts will need those staged into the container/pod, which is not covered here. I will file a follow-up ticket for artifact staging so it does not get lost.

tylersbray: Filed as [AIRCORE-966](https://linear.app/nvidia/issue/AIRCORE-966/stage-fabric-agentyaml-referenced-artifacts-into-container-deployments) for artifact staging (prompt files, skills paths, other harness-relative assets) in docker/k8s.

mmogallapalli: sounds good, thanks for making that issue

mmogallapalli (approved):
LGTM, thanks for taking this on. I’ll handle the packaging/dependency wiring in a follow-up PR shortly.

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -158,6 +165,75 @@ def rewrite_config_base_urls(nat_config: dict[str, Any], gateway_url: str) -> di
...
+
+
+def _materialize_config_and_exec(*, config_path: str, yaml_env: str, argv: list[str]) -> list[str]:
+    """Return ``sh -c`` args that write the config from *yaml_env*, then exec *argv*.
+
+    Docker mode needs this because the docker backend does not mount ``config_files``.
+    """
+    return [
+        f'mkdir -p "$(dirname "{config_path}")" && printf "%s" "${yaml_env}" > "{config_path}" && exec {" ".join(argv)}'
+    ]
```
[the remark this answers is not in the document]
tylersbray: Fixed in the rebase onto main (`5b8f8bda74`): `_materialize_config_and_exec` now shell-escapes `config_path` and each `argv` element with `shlex.quote` before splicing into the `sh -c` script. Added a unit test that uses a spaced config mount path to lock this in.

PR #986, opened 2026-07-30
feat(evaluator): add NeMo Gym runner for agent-eval

SandyChapman:
## What

Adds `GymAgentTaskRunner`, a NeMo Gym-backed `AgentTaskRunner` that runs an **existing** NeMo Gym environment through the `gym` CLI and adapts its rollout bundle into `AgentEvalTrial` objects, so `AgentEvaluator` can score and report Gym runs through the same seam as Harbor/Fabric. Gym owns execution *and* scoring; this runner only surfaces its per-attempt reward — it does not re-derive rewards.

Includes a worked `mcqa` example (`packages/nemo_evaluator_sdk/examples/gym/`) and unit tests against captured fixtures. Verified end-to-end against a real Gym checkout (5 tasks → 10 trials).

## Mapping (Gym → Evaluator)

- One Gym dataset → one run; each **distinct row** → one `AgentEvalTask` (id = content hash of the row); each attempt (`--num-repeats`) → one `AgentEvalTrial`.
- Rollouts are attributed back to tasks by Gym's positional `_ng_task_index` (Gym mutates `responses_create_params` and drops injected fields, so a content round-trip is impossible), with a guard that **fails loud** if our row-dedup partition diverges from Gym's.
- `GymRewardMetric` surfaces Gym's per-attempt `reward`; a missing/malformed reward is left **unscored** (`None` → nan, excluded from the mean) rather than counted as a spurious `0.0`.
- The `*_failures.jsonl` sidecar is read and surfaced as `FAILED` trials (counted + diagnosed), not silently dropped.

## Execution & robustness

- Two-step Gym flow — `gym env start` then `gym eval run --no-serve --input` — which reads a dataset directly without triggering Gym's split-driven data-prep. Shells out to the `gym` executable in the caller-provided checkout; never imports `nemo_gym`, never handles secrets.
- Teardown signals the whole process group with **SIGINT** (mimics Ctrl-C so Gym's own `shutdown()` stops Ray + uvicorn cleanly), escalating to SIGKILL after a grace window.
- Readiness keys off Gym's own `All N / N servers ready!` line (no server-count guessing); bounded collection via `collection_timeout_s`; refuses to reuse a populated `output_dir` (Gym appends to its failures sidecar); sets `RAY_ENABLE_UV_RUN_RUNTIME_ENV=0` so Gym's per-server venvs don't trip Ray's `uv run` hook.

## Scope / follow-ups

- Aggregations (native pass@k cross-check + bubbling Gym's `rollouts_aggregate_metrics.json` as run diagnostics) are intentionally **out of scope** here — tracked separately.
- Service-side execution (docker/k8s, Ray provisioning) is the evaluator plugin's job, not this SDK path.

Draft for early review.

## Summary by CodeRabbit

* **New Features**
  * Added Gym-backed agent evaluation with dataset-driven task discovery, configurable execution, repeat runs, and reward scoring.
  * Added a runnable command-line example that reports task and trial counts, aggregate results, and output locations.
  * Improved handling of trial attribution, failures, incomplete results, and malformed evaluation records.

* **Documentation**
  * Added setup guidance, commands, expected outputs, and operational considerations for Gym evaluations.

* **Tests**
  * Added fixtures and comprehensive coverage for task mapping, rollouts, scoring, failures, and incomplete data.


packages/nemo_evaluator_sdk/examples/gym/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Run
+
+From the nemo-platform repo root (any Python with `nemo_evaluator_sdk` importable — the runner shells out to Gym's own venv):
+
+```bash
+python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval --gym-root /path/to/Gym
+```
+
+Useful flags: `--resources-server`, `--agent`, `--model-type` (`inference_provider` for OpenAI-compatible **chat** endpoints; `openai_model` uses the OpenAI **Responses API** and 500s against chat-only endpoints), `--num-repeats`, `--output-dir`.
```
ngoncharenko, line 37: we should add a link to live gym docs for more args, settings

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,663 @@
...
+
+
+def _content_text(content: Any) -> str:
+    """Flatten a message ``content`` (a string, or a list of ``{type,text}`` parts) to text."""
+    if isinstance(content, str):
+        return content
+    if isinstance(content, Sequence):
+        parts = [part["text"] for part in content if isinstance(part, Mapping) and isinstance(part.get("text"), str)]
+        return "\n".join(parts)
+    return ""
```
ngoncharenko, line 116: maybe add a warning for the case when not a str and not a seq of str?

packages/nemo_evaluator_sdk/examples/gym/run_gym_eval.py
```diff
@@ -0,0 +1,107 @@
...
+        help="Repo-relative agent config passed to `gym env start`.",
+    )
+    parser.add_argument(
+        "--model-type",
+        default="inference_provider",
+        help="`inference_provider` speaks OpenAI-compatible chat; `openai_model` uses the Responses API.",
+    )
+    parser.add_argument("--num-repeats", type=int, default=2, help="Attempts per task (each becomes a trial).")
+    parser.add_argument(
+        "--output-dir", type=Path, default=Path("/tmp/gym-eval"), help="Where the run bundle is written."
```
ngoncharenko: `default=Path("/tmp/gym-eval")` should use uuid since it is required to be a diff dir by the gym on each run

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,663 @@
...
+    so :class:`GymAgentTaskRunner` can recover it.
+    """
+    dataset = Path(dataset)
+    tasks: list[AgentEvalTask] = []
+    seen: set[str] = set()
+    for row in _read_jsonl(dataset):
+        task_id = _canonical_row_hash(row)
+        if task_id in seen:
+            continue
+        seen.add(task_id)
```
ngoncharenko, line 286: what if user wants to run the same task twice within the dataset? we should explicitly state this constraint and put here a warning if detected a row multiple times

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,663 @@
...
+        eval_proc = await asyncio.create_subprocess_exec(
+            *eval_cmd,
+            cwd=str(cfg.gym_root),
+            env=subprocess_env,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.STDOUT,
+            start_new_session=True,
+        )
+        try:
+            stdout, _ = await asyncio.wait_for(eval_proc.communicate(), timeout=cfg.collection_timeout_s)
```
ngoncharenko: could stream this into a file in the same output dir? could even separate stdout from stderr files. Then in the error case we can point to those files

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
ngoncharenko: also, in _collect_rollouts() we provide the path to the source dataset. If we filtered here, will it make gym running all still?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,663 @@
...
+
+
+def _ordered_task_ids(dataset: str | Path) -> list[str]:
+    """Distinct row content-hash ids in first-appearance order.
+
+    This is the order Gym assigns ``_ng_task_index`` (it dedups distinct rows by
+    content in file order), so a rollout's ``_ng_task_index`` indexes into this
+    list to recover the task it belongs to.
+    """
+    ordered: list[str] = []
```
ngoncharenko: we might get mix up of rows here

Codex
---
SDK treats `_ng_task_index` as a list position, while Gym treats it as an identifier that may already be assigned by the dataset.

### Example 1: silent task swap

Dataset:

```jsonl
{"_ng_task_index":1,"responses_create_params":{"input":"What is 2 + 2?"}}
{"_ng_task_index":0,"responses_create_params":{"input":"What is France's capital?"}}
```

The SDK builds task IDs in file order:

```text
ordered_task_ids = [
  id(math_task),     # position 0
  id(capital_task),  # position 1
]
```

Gym preserves the supplied indices and emits:

```json
{"_ng_task_index": 1, "response": "4", "reward": 1.0}
{"_ng_task_index": 0, "response": "Lyon", "reward": 0.0}
```

The adapter currently resolves them as:

```text
index 1 → ordered_task_ids[1] → capital_task
index 0 → ordered_task_ids[0] → math_task
```

The resulting Evaluator trials are therefore wrong:

```text
capital_task gets response "4" and reward 1.0
math_task gets response "Lyon" and reward 0.0
```

Nothing raises because both indices are within range. Even the aggregate mean remains `0.5`, so the corruption can be hard to notice.

The correct mapping is:

```text
gym_index_to_task_id = {
  1: id(math_task),
  0: id(capital_task),
}
```

### Example 2: equivalent JSON fails after collection

These lines contain the same JSON value but have different serialization:

```jsonl
{"responses_create_params":{"input":"Choose A"},"expected":"A"}
{ "expected": "A", "responses_create_params": { "input": "Choose A" } }
```

The SDK canonicalizes both rows—sorting keys and removing insignificant whitespace—so both produce the same hash:

```text
ordered_task_ids = [same_task_id]
```

Gym deduplicates using the raw JSONL line text. Because the strings differ, Gym assigns two indices:

```text
first line  → _ng_task_index 0
second line → _ng_task_index 1
```

When the second rollout is imported, the SDK attempts:

```text
ordered_task_ids[1]
```

But the list has only one item, so the run fails after Gym has already completed the expensive collection.

### Why existing tests pass

The fixture only exercises the safe case:

- Rows have no preassigned indices.
- Rows have distinct canonical content.
- Their Gym indices happen to be `0, 1, 2, …` in file order.

Under those conditions, list position and Gym index coincide.

### Required fix

Build an explicit `dict[int, str]` mapping:

- Preserve `_ng_task_index` when present.
- For unstamped rows, reproduce Gym’s raw-line deduplication rule.
- Resolve rollouts with `mapping[record["_ng_task_index"]]`.
- Reject ambiguous collisions before launching Gym rather than after collection.

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Confirmed: `rollout_collection.py:363` only assigns an index when the row lacks one, and its fallback dedup keys off raw line text. Fixed by inverting it — we now materialize a dataset with `_ng_task_index` stamped ourselves, so attribution is a dict lookup, not a positional guess. Both of Codex's examples are regression tests. Note we now depend on Gym honoring a pre-stamped index; the `_resolve_task_id` raise is the tripwire if that regresses upstream.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Added: duplicates now warn and point at `num_repeats`, documented in the docstring and README. Row duplication structurally cannot express repeats since task identity is the row content hash.

SandyChapman: Worse than that — we never filtered, so Gym ran the whole dataset and we discarded the extra trials. The materialization fix resolves it: Gym now only sees the requested tasks.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Done: separate `gym_eval.stdout.log` / `gym_eval.stderr.log`, errors point at them, and output is mirrored to the module logger at `DEBUG` so terminal visibility is the caller's choice. Also removed an in-memory buffer of the entire transcript.

packages/nemo_evaluator_sdk/examples/gym/run_gym_eval.py
SandyChapman: Fixed: defaults to `tempfile.mkdtemp(prefix="gym-eval-")`. The old fixed path made the second run of the example fail.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Added, scoped to populated values in an unexpected shape; a missing `content` stays quiet to avoid noise.

packages/nemo_evaluator_sdk/examples/gym/README.md
SandyChapman: Added, plus a note that `env_overrides` is the escape hatch for flags we do not expose.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,884 @@
...
+                "rollout_failures": EvidenceDescriptor(kind="filesystem", format="file", ref=str(failures_path))
+            }
+        )
+        # tolerant: a truncated line in the abnormal-termination sidecar must not sink the good trials.
+        for record in _read_jsonl(failures_path, tolerant=True):
+            task_id = _resolve_task_id(record, index_to_task_id)
+            if task_id is None:
+                # kill-shaped failures (SIGTERM/OOM/actor-death) can lack a usable index — count, don't drop.
+                unattributed_failures += 1
+                continue
```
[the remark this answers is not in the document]
SandyChapman: Good catch — this contradicted the loop's own stated intent. `_resolve_task_id` now takes `strict`: the successes file keeps raising (an unknown index there means Gym reindexed, so every reward attribution is suspect), while the failures sidecar counts it as unattributable and warns. Safe to relax because failure records carry no reward, and a task left with no trial still trips `_require_full_coverage`. Used an explicit flag rather than catching the ValueError, so the two modes are visible at the call site. Tests cover both directions.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,856 @@
...
+            unattributed_failures,
+            NG_TASK_INDEX,
+        )
+
+    missing = known_task_ids - {trial.task_id for trial in trials}
+    if missing:
+        logger.warning("No Gym rollout produced a trial for %d requested task(s): %s", len(missing), sorted(missing))
+    if not trials:
+        logger.warning("No Gym rollouts under %s matched the requested tasks; nothing will be scored.", rollouts_path)
+    return trials
```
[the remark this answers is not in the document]
SandyChapman: Addressed, but not as proposed. Synthesizing FAILED trials would defeat the evaluator's deliberate guard (see the comment above `_score_trials`) and skew the mean, since an unscored trial is excluded from it — a half-collected run would report a plausible average over an unannounced subset. Kept the hard fail, but moved it into `_require_full_coverage` in `run_tasks`, which now names the rollout bundle, the failures sidecar, the Gym logs and the unrepresented task ids. That fixes the real complaint (generic error after paying full collection cost) without weakening the guard.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -0,0 +1,901 @@
...
+    # missing/null index must not collapse multiple attempts for one task onto the same trial id.
+    synth_seq: dict[str, int] = {}
+
+    success_evidence = CandidateEvidence(
+        descriptors={"rollouts": EvidenceDescriptor(kind="filesystem", format="file", ref=str(rollouts_path))}
+    )
+    for record in _read_jsonl(rollouts_path):
+        task_id = _resolve_task_id(record, index_to_task_id)
+        if task_id is None:
+            continue
```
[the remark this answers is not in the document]
SandyChapman: Valid — the successes loop dropped an unattributable record silently while the failures loop right below it counted its own. Added `unattributed_successes` with a warning naming the count, `rollouts_path` and `_ng_task_index`; skipping behaviour unchanged.

Worth noting why this matters despite Gym always stamping the index today (`rollout_collection.py:536` sets it unconditionally): with `num_repeats > 1` a dropped record leaves the task with its *other* trials, so `_require_full_coverage` would not fire and the task would just be scored on fewer attempts than were collected — a quietly wrong mean rather than a visible failure.

PR #990, opened 2026-07-30
feat: Validate no custom columns in DD configs

mikeknep:
The upstream Data Designer library allows users define custom column types via a decorated Python function. These don't work in the platform plugin because we don't have the user-created implementation server-side, nor can we serialize the function and send it over the wire. Today if a user builds a config client-side with a custom column and submits that config to the service, they get a raw error message when Pydantic tries and fails to deserialize the config. This PR ensures we provide a clearer error message that custom columns are not supported here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject unsupported custom columns during configuration, preview, creation, and remote execution.
  * Validation reports now clearly indicate when custom columns are not supported, without interrupting other validation checks.

* **Tests**
  * Added coverage for custom-column rejection in preview, creation, and SDK validation workflows.


mmogallapalli (approved):
LGTM

PR #997, opened 2026-07-30
fix: properly handle NONE case for runtime executor

mckornfield:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when Docker is unavailable or cannot connect.
  * Unavailable Docker-backed execution profiles are now skipped instead of preventing other backends from starting.
  * Invalid default executors are cleared with a warning, allowing the platform to continue loading.
  * Advertised execution profiles now stay synchronized with successfully available backends.
  * Platform startup reports Docker-related failures more clearly and exits promptly when services stop unexpectedly.
* **Tests**
  * Added coverage for Docker availability, backend registration, profile filtering, and startup failure scenarios.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
@@ -69,7 +69,9 @@ class ExecutorRegistry:
                     # Capability missing (optional packaging extra, unreachable Docker
                     # daemon, etc.): skip just that executor so the deployments service
                     # can still boot. Resolving a skipped name later raises
-                    # ExecutorNotFoundError; a skipped default is cleared below.
+                    # ExecutorNotFoundError. A configured default_executor that failed
+                    # to register still fails fast below — silent clearing hid config
+                    # mistakes and made debugging harder.
                     logger.warning(
                         "Skipping executor '%s': backend '%s' is unavailable (%s)",
                         spec.name,
@@ -77,12 +79,10 @@ class ExecutorRegistry:
                         exc,
                     )
             if default_executor and default_executor not in executors:
-                logger.warning(
-                    "Clearing default_executor '%s' because it is not registered "
-                    "(unavailable backend or missing from executor config).",
-                    default_executor,
+                raise ExecutorNotFoundError(
+                    f"default_executor '{default_executor}' is not registered "
+                    "(unavailable backend or missing from executor config)."
                 )
-                default_executor = None
         except Exception:
             for backend in executors.values():
                 backend.shutdown()
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
@@ -66,11 +66,10 @@ class ExecutorRegistry:
                 try:
                     executors[spec.name] = classes[spec.backend](sdk, spec.config)
                 except MissingBackendDependencyError as exc:
-                    # An opt-in backend whose optional extra isn't installed (e.g.
-                    # 'openshell') must not take down the whole deployments service:
-                    # skip just that executor. Resolving it later raises
-                    # ExecutorNotFoundError, and a `default_executor` that can't be
-                    # built still fails fast via the check below.
+                    # Capability missing (optional packaging extra, unreachable Docker
+                    # daemon, etc.): skip just that executor so the deployments service
+                    # can still boot. Resolving a skipped name later raises
+                    # ExecutorNotFoundError; a skipped default is cleared below.
                     logger.warning(
                         "Skipping executor '%s': backend '%s' is unavailable (%s)",
                         spec.name,
@@ -78,7 +77,12 @@ class ExecutorRegistry:
                         exc,
                     )
             if default_executor and default_executor not in executors:
-                raise ExecutorNotFoundError(f"default_executor '{default_executor}' is not registered.")
+                logger.warning(
+                    "Clearing default_executor '%s' because it is not registered "
+                    "(unavailable backend or missing from executor config).",
+                    default_executor,
+                )
+                default_executor = None
         except Exception:
             for backend in executors.values():
                 backend.shutdown()
--- a/services/core/jobs/src/nmp/core/jobs/controllers/backends/registry.py
+++ b/services/core/jobs/src/nmp/core/jobs/controllers/backends/registry.py
@@ -157,6 +157,26 @@ class BackendRegistry:
                     executor.profile,
                     exc,
                 )
+
+        # Keep GET /v2/execution-profiles aligned with backends that registered.
+        # ``profiles`` is built at import time and shared with the jobs API in local
+        # standalone, so prune the list in place when callers pass a mutable list.
+        if isinstance(profiles, list):
+            registered = frozenset((key.provider, key.profile) for key in registry)
+            kept = []
+            for profile in profiles:
+                if (profile.provider, profile.profile) in registered:
+                    kept.append(profile)
+            skipped = len(profiles) - len(kept)
+            if skipped:
+                logger.warning(
+                    "Removed %s execution profile(s) that failed backend registration so advertised "
+                    "profiles match the jobs controller registry.",
+                    skipped,
+                )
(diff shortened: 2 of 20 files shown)


tylersbray:
## Takeover notes (capability soft-skip reshape)

Took over this PR while @mckornfield is OOO and reshaped it away from blanketing all deployments executors under `Runtime.NONE`.

### What changed vs the original approach
- **Deployments:** `DockerDeploymentBackend` raises `MissingBackendDependencyError` when the daemon/socket is unavailable; the registry already soft-skips that. Skipped `default_executor` is **cleared with a warning** instead of failing startup. Openshell and other non-Docker executors remain.
- **Jobs:** Docker profiles are skipped whenever Docker is unavailable (not only under `Runtime.NONE`). `BackendRegistry.from_config` also soft-skips Docker construction failures, then syncs the shared `profiles` list so `/v2/execution-profiles` matches what can actually run.
- **Setup UX:** `_wait_for_platform` fails immediately if the service process exits (no full timeout hang), and prints an explicit Docker hint when the daemon is missing.

### Still out of scope (follow-up)
Runtime removal / unified capability framework (Ryan’s longer-term point). This PR is a tactical fix for NVBug 6537617 that does capability detection at backend registration rather than treating `Runtime` as “has Docker.”

Please re-review when convenient.

tylersbray:
### Follow-on coherence patches (post gap review)

- **Jobs registry:** soft-skip only on Docker/daemon connection errors (`DockerException`, `RequestsConnectionError`, `Timeout`, `OSError`); `ValidationError` / programming errors still raise.
- **`validate_docker_available`:** also catches `RequestsConnectionError` + `OSError` (aligned with jobs).
- **Setup Docker hint:** only when early exit **and** evidence (services.log soft-skip/daemon markers, or daemon probe false)—not on readiness timeout while the process is still alive without Docker evidence.
- **Tests:** construction soft-skip vs fatal config error, advertised-profile sync prune, validate_docker catch set, setup `proc=` + hint gating.

Longer-term work (shared capability probe / Runtime shrink) tracked in AIRE Core Linear tickets — links coming shortly.

tylersbray:
### Follow-up Linear tickets (AIRE Core)

- **Tier 2 (Medium):** [AIRCORE-971](https://linear.app/nvidia/issue/AIRCORE-971/unify-backend-capability-detection-dockergpuk8s-and-stop-using-runtime) — Unify backend capability detection; stop using Runtime as a proxy
- **Tier 3 (Low, blocked by Tier 2):** [AIRCORE-972](https://linear.app/nvidia/issue/AIRCORE-972/shrink-or-remove-platform-runtime-make-executor-selection-capability) — Shrink or remove platform Runtime; capability-driven executor selection

Please leave design feedback on those tickets before implementation.

services/core/jobs/src/nmp/core/jobs/controllers/backends/registry.py
```diff
@@ -16,6 +19,14 @@
 )
 from nmp.core.jobs.controllers.backends.subprocess import SubprocessJobBackend
 from nmp.core.jobs.controllers.backends.test import TestE2ECPUJobBackend, TestE2EGPUJobBackend
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from requests.exceptions import Timeout as RequestsTimeout
+
+logger = logging.getLogger(__name__)
+
+# Soft-skip only daemon/connection failures after validate_docker_available() was True.
```
mckornfield: soft skip is a weird bit of phrasing?

services/core/jobs/src/nmp/core/jobs/controllers/main.py
```diff
@@ -17,6 +17,26 @@
...
+
+def _sync_advertised_profiles_to_registry(backend_registry: BackendRegistry) -> None:
+    """Keep GET /v2/execution-profiles aligned with backends that actually registered.
+
+    ``profiles`` is built at import time; ``BackendRegistry.from_config`` may later
+    soft-skip Docker backends. Mutate the shared list in place so the API process
+    (same process in local standalone) stops advertising unusable executors.
+    """
+    registered = backend_registry.registered_profile_keys()
+    kept = [profile for profile in profiles if (profile.provider, profile.profile) in registered]
```
mckornfield: nested list comps, o my

services/core/jobs/src/nmp/core/jobs/controllers/main.py
```diff
@@ -17,6 +17,26 @@
...
+    skipped = len(profiles) - len(kept)
+    if skipped:
+        logger.warning(
+            "Removed %s execution profile(s) that failed backend registration so advertised "
+            "profiles match the jobs controller registry.",
+            skipped,
+        )
+        profiles[:] = kept
```
mckornfield: hmmm what is this doing?

mckornfield (commented):
just giving feedback per our dm

services/core/jobs/src/nmp/core/jobs/controllers/backends/registry.py
tylersbray: Agreed — dropped the soft-skip phrasing. Renamed to `_DOCKER_BACKEND_INIT_SKIPPABLE_ERRORS` and updated the comment to just say we skip Docker backends for daemon/connection failures.

services/core/jobs/src/nmp/core/jobs/controllers/main.py
tylersbray: Expanded to a plain `for` loop + `append` so it is easier to read.

services/core/jobs/src/nmp/core/jobs/controllers/main.py
tylersbray: That was slice assignment to mutate the shared module-level `profiles` list in place (so API importers of the same list see the prune). Switched to `profiles.clear()` + `profiles.extend(kept)` and spelled that out in the docstring.

services/core/jobs/src/nmp/core/jobs/controllers/main.py
```diff
@@ -43,6 +62,7 @@ def run(parent_stop_signal: threading.Event | None = None):
     logger.debug("Platform SDK initialized successfully.")
 
     backend_registry = BackendRegistry.from_config(nmp_sdk=nmp_sdk, profiles=profiles)
+    _sync_advertised_profiles_to_registry(backend_registry)
```
ironcommit: since this is the only caller, why don't we move this function into the `BackendRegistry.from_config`?

ironcommit (approved):
LGTM

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
```diff
@@ -66,19 +66,23 @@ def from_config(
...
                     logger.warning(
                         "Skipping executor '%s': backend '%s' is unavailable (%s)",
                         spec.name,
                         spec.backend,
                         exc,
                     )
             if default_executor and default_executor not in executors:
-                raise ExecutorNotFoundError(f"default_executor '{default_executor}' is not registered.")
+                logger.warning(
+                    "Clearing default_executor '%s' because it is not registered "
```
mikeknep: I'm a bit removed from the details and requests here, so feel free to ignore, but: this feels "too lenient"... is this going to cause weird surprises / be harder to debug than just failing here? **Edit to add**: specifically for a **configured default**, not for the collection of executors in general

services/core/jobs/src/nmp/core/jobs/controllers/main.py
tylersbray: Good call — folded the prune into `BackendRegistry.from_config` so it mutates the shared mutable profiles list in place when Docker (or other) backends are skipped. Removed the separate helper from `main.py`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/registry.py
tylersbray: Agreed — clearing a configured default was too quiet. Restored fail-fast: unavailable non-default executors are still skipped so the service can boot, but if `default_executor` itself fails to register we raise `ExecutorNotFoundError` at startup with a clear message.

PR #998, opened 2026-07-30
feat(intake): add intake clickhouse to nemo-platform helm

BrianNewsom:
## Summary by CodeRabbit

* **New Features**
  * Added embedded single-node ClickHouse for Intake with configurable persistence, ports, and Kubernetes service/service account setup.
  * Added support for externally managed ClickHouse (optionally secure/TLS) using an existing credentials Secret.
  * Enhanced Helm notes and API connectivity with ClickHouse URL/credentials/database details.
* **Bug Fixes**
  * Updated embedded ClickHouse password default so an unset value generates a random password; tightened validation for missing external settings.
* **Documentation**
  * Expanded “Intake and ClickHouse” with embedded limitations, external Secret/keys setup, and ClickHouse sizing guidance.
* **Tests**
  * Added Helm template negative assertions for embedded/external ClickHouse credential configuration expectations.

### DIFF
--- a/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
+++ b/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
@@ -20,10 +20,11 @@ spec:
       {{- include "nemo-platform.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.clickhouse.podAnnotations }}
       annotations:
+        checksum/clickhouse-credentials: {{ include "nemo-common.clickhouse.credentialsChecksum" . }}
+        {{- with .Values.clickhouse.podAnnotations }}
         {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- end }}
       labels:
         app.kubernetes.io/component: clickhouse
         {{- include "nemo-platform.labels" . | nindent 8 }}
--- a/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
+++ b/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
@@ -76,26 +76,11 @@ spec:
               containerPort: 9000
               protocol: TCP
           startupProbe:
-            httpGet:
-              path: /ping
-              port: http
-            periodSeconds: 5
-            timeoutSeconds: 3
-            failureThreshold: 60
+            {{- toYaml .Values.clickhouse.startupProbe | nindent 12 }}
           livenessProbe:
-            httpGet:
-              path: /ping
-              port: http
-            periodSeconds: 10
-            timeoutSeconds: 5
-            failureThreshold: 3
+            {{- toYaml .Values.clickhouse.livenessProbe | nindent 12 }}
           readinessProbe:
-            httpGet:
-              path: /ping
-              port: http
-            periodSeconds: 5
-            timeoutSeconds: 3
-            failureThreshold: 3
+            {{- toYaml .Values.clickhouse.readinessProbe | nindent 12 }}
           volumeMounts:
             - name: data
               mountPath: /var/lib/clickhouse
@@ -112,7 +97,7 @@ spec:
         name: data
         labels:
           app.kubernetes.io/component: clickhouse
-          {{- include "nemo-platform.labels" . | nindent 10 }}
+          {{- include "nemo-platform.selectorLabels" . | nindent 10 }}
       spec:
         accessModes:
           - ReadWriteOnce
--- a/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
+++ b/k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
@@ -1,0 +1,126 @@
+{{- if include "nemo-common.clickhouse.enabled" . }}
+{{- $imagePullSecrets := include "nemo-common.imagepullsecrets" . }}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ include "nemo-common.clickhouse.fullname" . }}
+  labels:
+    app.kubernetes.io/component: clickhouse
+    {{- include "nemo-platform.labels" . | nindent 4 }}
+  {{- with .Values.clickhouse.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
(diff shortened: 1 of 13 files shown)


tools/lint/lint-helm.sh
```diff
@@ -36,6 +36,24 @@ envoy_autoscaling_output=$(helm template "${HELM_RELEASE_NAME}" "${HELM_FOLDER}"
...
+  <<<"${external_clickhouse_output}"
+
+external_clickhouse_secret_output=$(helm template "${HELM_RELEASE_NAME}" "${HELM_FOLDER}" \
+  --set clickhouse.enabled=false \
+  --set externalClickhouse.host=clickhouse.example.internal 2>&1) && {
+  echo "External ClickHouse accepted a missing credentials Secret" >&2
+  exit 1
+}
+grep -Fq "externalClickhouse.existingSecret is required when clickhouse.enabled=false" \
+  <<<"${external_clickhouse_secret_output}"
```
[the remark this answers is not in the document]
BrianNewsom: Addressed in fc70535847. Added the distinct external ClickHouse validation case with the Secret configured and password key omitted; it asserts the expected helper error. Verified with the full tools/lint/lint-helm.sh Helm lint and kubeconform matrix plus changed-file pre-commit hooks.

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
```diff
@@ -0,0 +1,126 @@
...
+              containerPort: 9000
+              protocol: TCP
+          startupProbe:
+            httpGet:
+              path: /ping
+              port: http
+            periodSeconds: 5
+            timeoutSeconds: 3
+            failureThreshold: 60
+          livenessProbe:
```
svvarom, line 81: I wonder if we should make the probes configurable?

svvarom: We do it for the api-deployment - https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/k8s/helm/values.yaml#L570

svvarom: not really requesting a change - just thinking out loud

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
```diff
@@ -0,0 +1,126 @@
...
+        - name: data
+          emptyDir: {}
+      {{- end }}
+  {{- if .Values.clickhouse.persistence.enabled }}
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+        labels:
+          app.kubernetes.io/component: clickhouse
+          {{- include "nemo-platform.labels" . | nindent 10 }}
```
svvarom: I think `volumeClaimTemplates` are immutable and `nemo-platform.labels` values can change with the app version.

svvarom: This might fail during upgrades

svvarom: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/validation/validation.go#L262

BrianNewsom: Good catch. Will address.

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
BrianNewsom: It seems reasonable to follow the same patterns -- seems like the principle of least surprise here.

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
```diff
@@ -0,0 +1,126 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: clickhouse
+      {{- include "nemo-platform.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.clickhouse.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
```
svvarom, line 26: Could we add a ClickHouse credential checksum to these annotations, and the same checksum to the api deployment, so secret changes roll both workloads? We do something similar for the configmap https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/k8s/helm/templates/_helpers.tpl#L148

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
BrianNewsom: Addressed in 556dfed29f. The ClickHouse volumeClaimTemplates now use only stable selector labels plus the component label; chart and app version labels were removed. Added a regression assertion that rejects mutable version labels. I also validated a targeted chart/app upgrade from 0.1.0/0.3.0 to 0.2.0/0.4.0 with persistent ClickHouse enabled; the ClickHouse StatefulSet upgrade succeeded. Embedded PostgreSQL was disabled in that targeted test because its existing StatefulSet has the analogous pre-existing label issue.

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
BrianNewsom: Addressed in 556dfed29f. ClickHouse startup, liveness, and readiness probes are now configurable under clickhouse values and rendered using the same value-to-template pattern as the API deployment. Existing probe defaults are unchanged. Verified a custom startup period renders correctly, and the full Helm lint and kubeconform matrix passes.

k8s/helm/templates/clickhouse/clickhouse-statefulset.yaml
BrianNewsom: Addressed in 7b141db5c4. The API Deployment and ClickHouse StatefulSet now share checksum/clickhouse-credentials. The helper uses an explicitly configured chart-managed password first, otherwise the live Secret value, with a deterministic first-install fallback for generated credentials. Added regression checks that both annotations match and that a configured password change changes the checksum. I also verified live lookup in a disposable namespace: both workloads rendered 44a18... before rotating the existing Secret and both moved together to a66eca... afterward. The full Helm lint and kubeconform matrix passes.

svvarom (approved):
LGTM

PR #1002, opened 2026-07-30
fix(intake): performance improvements for trace list

BrianNewsom:
Use a two phase query to hydrate the trace list view instead of the current single query approach, which requires reading _all_ input/output data across spans.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trace listing pagination and multi-step loading to keep results consistent across modes.
  * Added reconciliation when hydration omits items, including warning logs and corrected total counts.
  * Refined preview payload handling to ensure proper truncation without impacting detailed views.
  * Empty preview pages now avoid unnecessary follow-up loading.

* **Tests**
  * Expanded assertions for preview and detailed query behavior, including aggregation expressions and payload formatting.
  * Added cases for empty pages and reconciliation when hydrated results miss requested refs.

### DIFF
--- a/services/intake/src/nmp/intake/repository/clickhouse/trace.py
+++ b/services/intake/src/nmp/intake/repository/clickhouse/trace.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import json
+import logging
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime, timezone
@@ -26,6 +27,8 @@ from nmp.intake.spans.storage import (
     text_query_parameters,
     text_select_for_mode,
 )
+
+logger = logging.getLogger(__name__)
 
 TRACE_SORT_COLUMNS = {
     "started_at": "started_at",
@@ -163,6 +166,17 @@ class ClickHouseTraceRepository(TraceRepository):
                     },
                 )
             )
+            rows, dropped_refs = _reconcile_hydrated_page(page_refs, rows)
+            if dropped_refs:
+                logger.warning(
+                    "Trace page hydration omitted refs returned by the page query",
+                    extra={
+                        "workspace": filters.workspace,
+                        "page": page,
+                        "dropped_trace_refs": [ref.trace_key for ref in dropped_refs],
+                    },
+                )
+                total_results = max(offset + len(rows), total_results - len(dropped_refs))
         traces = [_row_to_trace(row) for row in rows]
         return PaginatedResult(
             data=traces,
@@ -240,6 +254,8 @@ class ClickHouseTraceRepository(TraceRepository):
 
 
 def _trace_page_sql(*, trace_index_sql: str, sort: str) -> str:
+    """Build the first-phase query that selects lightweight trace roots for one page."""
+
     return f"""
         SELECT
             {_trace_select_columns(include_aggregates=False)}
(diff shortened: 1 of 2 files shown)


services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -212,47 +239,48 @@ async def latest_trace_started_at_by_group(
...
-                ORDER BY {_order_by(sort)}
-                LIMIT %(limit)s OFFSET %(offset)s
-            ),
-            rollups AS (
-                {aggregates_sql}
-            )
+def _trace_page_sql(*, trace_index_sql: str, sort: str) -> str:
+    return f"""
+        SELECT
+            {_trace_select_columns(include_aggregates=False)}
```
asutermo, line 261: the aggregates piece is the second phase it looks like?

services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -212,47 +239,48 @@ async def latest_trace_started_at_by_group(
...
+        FROM ({trace_index_sql}) AS traces
+        ORDER BY {_order_by(sort, table_alias="traces")}
+        LIMIT %(limit)s OFFSET %(offset)s
+    """
+
+
+def _trace_hydration_sql(*, trace_index_table: str, spans_table: str, mode: TraceMode) -> tuple[str, dict[str, Any]]:
+    trace_columns, trace_parameters = _trace_index_select_columns(mode=mode)
+    aggregates_sql, aggregate_parameters = _trace_aggregates_sql(spans_table)
+    query = f"""
```
asutermo, line 273: nit: it seems like an ORM or DAL  will be desirable for readability purposes. no action here, these queries are just beefy.

asutermo (approved):

services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -113,29 +132,37 @@ async def list_traces(
...
+                ClickHouseQuery(
+                    name="traces.list.hydrate",
+                    statement=hydration_sql,
+                    parameters={
+                        "workspace": filters.workspace,
+                        **hydration_parameters,
+                        **_trace_page_parameters(page_refs),
+                    },
+                )
+            )
```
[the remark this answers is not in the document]
BrianNewsom: Addressed in 27ea1b39f5. Hydration results are reconciled against the phase-one page refs in Python; missing refs are logged, the known drops reduce total_results without taking it below the returned page boundary, and a regression test covers both missing hydration and order restoration.

services/intake/src/nmp/intake/repository/clickhouse/trace.py
BrianNewsom: Yes—the lightweight page query is phase one, and root payload plus span aggregate hydration is phase two. In 27ea1b39f5 I added phase-oriented docstrings and split the page-root SQL from the aggregate SQL to make that flow explicit.

services/intake/src/nmp/intake/repository/clickhouse/trace.py
BrianNewsom: Agreed. Without introducing an ORM or broader DAL, 27ea1b39f5 decomposes the phase-two query into page-root and aggregate builders, extracts aggregate-column construction, and moves order restoration/reconciliation into Python. The SQL stays local, but the stages are easier to read.

PR #1005, opened 2026-07-30
feat(studio): anonymizer builder live preview panel [ASTD-332]

marcusds:
Builds the right-hand **Preview panel** of the anonymizer builder (`/anonymizer/new`), replacing the "Your records preview will appear here" placeholder.

Closes [ASTD-332](https://linear.app/nvidia/issue/ASTD-332/studio-anonymizer-builder-live-preview-panel).

## What it does

**Preview** (beside the `Source | Model Settings` tabs) streams the current form config through the anonymizer `/preview` endpoint and renders:

- **Original vs Replaced** columns with inline PII tags on detected/replaced entities
- **Replacement Map** table (Label | Original | → | Replacement)
- Record pager (`‹ Record 1 of N ›`), loading skeletons, empty state, error banner
- Collapsible **Logs** panel fed by the stream's `log` frames

While a run is in flight the button becomes **Stop**. `Full Run` moves into the panel header per the Figma and still submits the create-job form — the panel is inside the same `<form>`. The left panel footer is dropped: Figma marks that frame hidden and moves its only visible button into the preview header, which left `Cancel` stranded. Exit is via the breadcrumb.

## Frame types now come from the SDK

Originally this branch hand-wrote the `PreviewFrame` union and its guards, because the spec described the streaming 200 as an empty schema under the wrong media type. #1024 (ASTD-350) fixed that at the source, and this now imports the generated `LogFrame`, `PreviewDatasetFrame`, `TraceDatasetFrame`, `FailedRecordsFrame`, `Heartbeat`, `Done` and `Error`.

The runtime guards stay deliberately. Orval emits `zod.unknown()` for the response because the media type isn't JSON, and frames are untrusted network input regardless — the SDK removed the duplicated *declarations*, not the need to validate. The streaming reader stays hand-written too; orval doesn't generate incremental readers.

## Notes for review

**Replaced-side spans are searched, not replayed.** Each entity is located in the replaced text by scanning forward for its synthetic value. Replaying the original offsets drifts the moment a replacement changes length — a bug the upstream library already hit and worked around, so `parse.ts` mirrors its `display.py` logic: the exact → value-only → case-insensitive lookup ladder, plus the build-from-replacement-map fallbacks for when detection returns no spans but the map is populated.

**PII tags use `Badge`, not `Tag`.** Code Connect on the Figma `PII Tag` component maps to `Badge` (`solid` for the value, `outline` for the label); `Tag` is the interactive pill and rendered noticeably heavier than the mock.

**Stop is real, not cosmetic — but not instant.** Aborting makes Starlette close the response generator, cancelling the drain task and unwinding the function's task group until the worker's next frame send raises `PreviewMessageDeliveryError`. Because the anonymizer runs under `abandon_on_cancel=True`, the worker thread is abandoned rather than joined, so an in-flight model call still finishes.

**Validation routing was reading empty errors.** Clicking Preview on an incomplete form jumped to Model Settings even when the failing field was on Source. `formState` is a proxy that only maintains what the render body subscribes to, and this component reads none of it — so `errors` came back `{}`, and `[].every()` is vacuously true. Errors now come from `handleSubmit`'s invalid callback, which is what the Full Run path already used.

**The shared-URL-param hazard.** `useStudioDataViewState` syncs to unprefixed params (`page`, `page_size`, `sort`, …). That bit inside this PR: with a map that overflows a page, going to `?page=2` then advancing the *record* pager to a shorter map sliced past the end and rendered an empty table — with the map's pager hidden below page size, nothing was left to recover with. Clamped, with a regression test confirmed to fail without the fix. The broader collision remains for when ASTD-330 reuses this table alongside `ResultsPreviewTable`; that wants a `paramPrefix` option on the shared hook.

**Reuse audit.** Four things this branch had reimplemented now use what Studio already ships: `isAbortError` (common's version also handles `APIUserAbortError`, mine only caught `DOMException`), the NDJSON read loop (extracted as `readLineDelimitedStream`, shared with the data designer preview and given the chunk-boundary test neither call site had), `OUTPUT_SUFFIXES` (was declared here and in `AnonymizerJobDetailRoute`), and `StackedSkeleton`.

The record view lives in `components/AnonymizerRecordView/` so the job detail page (ASTD-330) and evaluation results (ASTD-322) can reuse it.

## Screenshots
01-builder-empty
02-file-picker
02b-after-file
03-builder-configured
04-preview-running
05-preview-records
07-walkthrough-final

## Testing

- 73 unit tests across `parse`, `previewApi`, `lineStream`, the record view and the tab/heading helpers
- Verified in a real browser against a mocked NDJSON stream: empty, loading, populated, stop-mid-run, validation routing, and the rewrite-mode skeleton heading
- Full studio suite green (310 files, 2772 tests), typecheck and lint clean

Rebased onto `main` after #1024 merged.

## Summary by CodeRabbit

* **New Features**
  * Added streaming anonymizer preview panel with paging (up to 10 rows), logs, stop/full-run controls, and clear empty/error states.
  * Enhanced record viewing with entity highlighting, configurable output headings, and replacement maps (including skeleton loading).
  * Added record navigation (previous/next) and improved builder tab behavior for validation.
* **Bug Fixes**
  * Improved resilience for malformed/empty preview and replacement data, plus safer preview pagination and record selection resets.
* **Tests**
  * Expanded parsing, preview streaming/frame handling, and UI state coverage for edge cases.


web/packages/studio/src/components/AnonymizerRecordView/parse.ts
```diff
@@ -0,0 +1,231 @@
...
+};
+
+const toEntity = (entry: unknown): AnonymizerEntity | undefined => {
+  const row = asRecord(entry);
+  if (!row) return undefined;
+  const { value, label, start_position: start, end_position: end } = row;
+  if (typeof label !== 'string' || typeof start !== 'number' || typeof end !== 'number') {
+    return undefined;
+  }
+  return { value: typeof value === 'string' ? value : '', label, start, end };
```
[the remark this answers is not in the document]
marcusds: Claude says: "Declining — fractional offsets are not constructible. The producer is `nemo-anonymizer==0.3.1`: `anonymizer/engine/schemas/detection.py:19-28` declares `start_position: int` / `end_position: int`, and pydantic v2 rejects a float with a fractional part. The sole ingress is `_coerce_int` (`postprocess.py:372`, `int(value)`, truncating); every other span is derived from `str.find` / `re.Match` indices. Both trace writers (`custom_columns.py:176`, `detection_workflow.py:460-466`) round-trip through `EntitySchema`. Adding a guard here would imply the case happens."

web/packages/studio/src/components/AnonymizerRecordView/parse.ts
```diff
@@ -0,0 +1,263 @@
...
+    unitForCodePoint.push(index);
+    index += (text.codePointAt(index) ?? 0) > 0xffff ? 2 : 1;
+  }
+  unitForCodePoint.push(text.length);
+
+  const unitAt = (offset: number): number => unitForCodePoint[offset] ?? text.length;
+  return entities.map((entity) => ({
+    ...entity,
+    start: unitAt(entity.start),
+    end: unitAt(entity.end),
```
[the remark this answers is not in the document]
marcusds: Claude says: "Fixed in dbb3736331. You are right that this weakened an existing guard — an offset past the last code point mapped onto `text.length`, which is exactly the value `toSegments`' `entity.end > text.length` check lets through, so a span the old code rejected became a highlight running to the end of the text. Those entities are now dropped rather than clamped, and the replacement-map fallback in `buildAnonymizerRecord` decides on the converted list since conversion can empty it. Two regression tests, both confirmed to fail against the clamping version."

web/packages/studio/src/components/AnonymizerRecordView/HighlightedText.tsx
```diff
@@ -0,0 +1,39 @@
...
+      })}
+    </Text>
+  ) : (
+    <Text color="secondary" kind="body/regular/md">
+      {emptyMessage}
+    </Text>
+  )
+);
+
+HighlightedText.displayName = 'HighlightedText';
```
steramae-nvidia: Do you need this still? I thought FCs automatically name themselves.

web/packages/studio/src/components/AnonymizerRecordView/parse.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  biography: ORIGINAL,
+  biography_replaced: REPLACED,
+  final_entities: {
+    entities: [
+      { value: 'Bobby', label: 'first_name', start_position: 0, end_position: 5 },
+      { value: '40', label: 'age', start_position: 9, end_position: 11 },
+    ],
+  },
+  _replacement_map: { replacements },
+};
```
steramae-nvidia: Constants here seem similar to AnonymizerRecordView.test. Anything to consolidate?

web/packages/studio/src/components/AnonymizerRecordView/ReplacementMapTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const pageRows = useMemo(
+    () => replacements.slice(pageIndex * pageSize, (pageIndex + 1) * pageSize),
+    [replacements, pageIndex, pageSize]
+  );
+
+  const makeColumns = useCallback<
+    ComponentProps<typeof StudioDataView<EntityReplacement>>['makeColumns']
+  >(
+    (col) => [
+      col.display({
```
steramae-nvidia: not col.accessor? 

> col.accessor connects columns to an underlying data model so they support sorting, filtering, and grouping. Conversely, col.display creates unlinked UI columns that lack a data model, making them ideal for interactive components like action buttons, checkboxes, or row expanders.

web/packages/studio/src/components/AnonymizerRecordView/HighlightedText.tsx
marcusds: Thanks, that slipped passed me. Silly agents thinking its the past.

web/packages/studio/src/components/AnonymizerRecordView/ReplacementMapTable.tsx
marcusds: Oh nice til. I have been too disconnect from kui lately.

web/packages/studio/src/components/AnonymizerRecordView/parse.test.ts
marcusds: Yeah good call.

steramae-nvidia (approved):

PR #1006, opened 2026-07-30
chore(experimentalist): mark result models as deprecated contracts

ngoncharenko:
## Background
This is a follow up to these threads to "deprecate" the existing experimentalist models that have corresponding replacements in the evaluator sdk. 

- https://github.com/NVIDIA-NeMo/nemo-platform/pull/955#pullrequestreview-4821812070
- https://github.com/NVIDIA-NeMo/nemo-platform/pull/955#discussion_r3677424997

Solution: add metadata to models and tests. 

Alternative (stronger lock): require approval in CI from codeowner on change or runtime warning

## What changed
- Mark `TrialResult` and `EvaluationResult` as deprecated compatibility contracts through static Pydantic JSON Schema metadata, with explicit SDK migration targets and blockers.
- Add CI coverage that locks the models' public shape, required fields, serialization, warning-free runtime behavior, and persisted candidate details.
- Document why both result contracts remain required today and distinguish the `AgentEvalTrial` adapter migration from end-to-end `AgentEvalResult` convergence.

## Why

- Preserve identical behavior across native-Harbor and SDK-backed evaluator paths while making contract changes deliberate and reviewer-visible.
- Avoid runtime warning noise and premature conversion to SDK result types before their data and consumer gaps are closed.
- Follow up on the model-duplication discussion in [#955](https://github.com/NVIDIA-NeMo/nemo-platform/pull/955).

## Validation

- `uv run --frozen pytest plugins/nemo-experimentalist/tests -q` — 628 passed.
- Live plain-Harbor vs SDK-backed Harbor A/B — passed.
- Scoped Ruff lint/format, `ty`, and `git diff --check` — passed.


SandyChapman (approved):
lgtm

PR #1012, opened 2026-07-31
fix(deployments): tolerate containers removed during port enumeration

SandyChapman:
## What

Fixes an intermittent `Python integration tests` failure caused by a race in Docker host-port allocation.

## The bug

`find_available_port` enumerated containers with `client.containers.list(all=True)`. docker-py implements that as **two** API calls — enumerate, then inspect each container individually — and `ignore_removed` defaults to `False`. Its own source says so:

```python
for r in resp:
    try:
        containers.append(self.get(r['Id']))
    # a container may have been removed while iterating
    except NotFound:
        if not ignore_removed:   # <- defaults to False
            raise
```

So a container removed between the enumerate and its inspect 404s and aborts the whole listing:

```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url:
http+docker://localhost/v1.48/containers/<id>/json
```

Under a teardown-heavy workload that is routine, not exceptional.

The exception was swallowed into a `None` return, which the caller reported as **"No host ports available in configured range"** — misleading twice over: the range was not full, and it sends the reader to `port_range_*` config for what is a transient daemon-query race.

## Observed impact

Recent `main` failures in `Python integration tests`, all in the Docker deployment path:

- `test_igw_cache_removes_deleted_deployment_provider` — `Deployment not READY: ERROR`, preceded by `Failed to list containers for port allocation` + the 404 above
- `test_deployments_plugin_docker_lifecycle` — `Container not running: created`

## The fix

1. **`ignore_removed=True`.** A container that no longer exists holds no host ports, so skipping it is the semantically correct answer rather than a workaround.
2. **Separate "cannot determine" from "range is full".** `find_available_port` now raises `PortEnumerationError` when enumeration fails, and reserves `None` for genuine exhaustion. Both docker-backend call sites map it to an accurate status message. This mirrors the existing `GPUAllocationError` handling a few lines above.

The second call site is the port-conflict retry loop, where `_allocate_host_ports` sits inside an exception handler — without explicit handling the new exception would have escaped uncaught instead of becoming a graceful `FAILED` status.

## Deliberately not fixed here

`is_port_free` binds and releases a probe socket, so two concurrent allocations can still pick the same port. That is a real but separate race, already mitigated by the existing `_PORT_CONFLICT_ATTEMPTS` retry loop. Left alone to keep this change scoped to the enumeration bug — happy to file it separately.

## Testing

Three new unit tests: the removal race (asserts `ignore_removed=True` is passed), enumeration failure raising `PortEnumerationError`, and `None` returned only on true exhaustion. One existing test updated — it pinned the exact `containers.list` kwargs; its original intent (no label filter, so foreign containers are counted) is preserved and now commented.

`375 passed, 17 skipped` in `plugins/nemo-deployments/tests/unit/`. Ruff clean; `ty` reports only pre-existing diagnostics in `openshell/backend.py`.

Note the integration tests that motivated this need a Docker daemon, so the fix is verified by unit tests plus CI here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker deployment port-allocation error handling.
  * Distinguishes fully occupied port ranges from failures while checking port availability.
  * Provides clearer deployment failure messages and properly releases allocated resources when port checks fail.
  * Tolerates containers removed during port enumeration.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
@@ -31,7 +31,7 @@ from nemo_deployments_plugin.backends.docker.containers import (
     restart_policy_kwargs,
 )
 from nemo_deployments_plugin.backends.docker.gpu import GPUAllocationError, get_shared_gpu_pool
-from nemo_deployments_plugin.backends.docker.ports import find_available_port
+from nemo_deployments_plugin.backends.docker.ports import PortEnumerationError, find_available_port
 from nemo_deployments_plugin.backends.docker.probes import check_readiness_probe, host_url_for_port
 from nemo_deployments_plugin.backends.docker.status import (
     LOG_MAX_CHARS,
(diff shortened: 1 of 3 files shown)


svvarom (approved):

ironcommit (approved):
LGTM

PR #1015, opened 2026-07-31
fix(nemo-agents): Support Studio multi-turn transcript replay for Fabric agents

mmogallapalli:
## Summary

- Serialize the complete OpenAI chat transcript into Fabric invocation input for multi-message requests.
- Preserve the existing input exactly for single-message, one-shot requests.
- Retain message roles and ordering so Studio can provide conversational context without managing Fabric session IDs.

Studio already provides multi-turn chat for NAT-backed agents by retaining the conversation client-side and resending the full OpenAI `messages` transcript on every turn. NAT-backed serving uses that replayed transcript as context rather than relying on a platform-managed session.

This change gives Fabric-backed agents parity with that existing Studio behavior. The Fabric shim translates the replayed transcript into a role-aware input string, while each request remains an independent invocation. Native Fabric session management and session-ID wiring remain separate follow-up work.

## Testing

- Added coverage for unchanged single-message input.
- Added coverage for role-aware serialization of system, assistant, and user messages.
- Verified the Fabric server unit suite: `25 passed`.
- Verified Ruff and `git diff --check` pass.

Studio Before:
Screenshot 2026-07-31 at 1 04 45 PM

Studio After:
Screenshot 2026-07-31 at 1 15 38 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat interactions by preserving the full conversation history when sending requests.
  * System, assistant, and user messages are now included in the proper order, providing better context for responses.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
@@ -76,7 +76,7 @@ def _to_fabric_invocation_request(
 ) -> FabricInvocationRequest:
     """Translate the chat transcript into a Platform-owned Fabric request."""
     messages = request.messages
-    # AIRCORE-969: Interim behavior to get multi-turn in Studio chat.
+    # Interim behavior to get multi-turn in Studio chat.
     input_text = (
         messages[0].content
         if len(messages) == 1
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
@@ -74,9 +74,16 @@ def _to_fabric_invocation_request(
     *,
     session_id: str,
 ) -> FabricInvocationRequest:
-    """Translate the current chat turn into a Platform-owned Fabric request."""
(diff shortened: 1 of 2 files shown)


mikeknep (approved):

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
```diff
@@ -74,9 +74,16 @@ def _to_fabric_invocation_request(
     *,
     session_id: str,
 ) -> FabricInvocationRequest:
-    """Translate the current chat turn into a Platform-owned Fabric request."""
+    """Translate the chat transcript into a Platform-owned Fabric request."""
+    messages = request.messages
+    # AIRCORE-969: Interim behavior to get multi-turn in Studio chat.
```
tylersbray: Watch out for ticket numbers kicking around in code in perpetuity, I think we try to avoid this now.

tylersbray (approved):
lgtm

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
mmogallapalli: good call out, removed

PR #1023, opened 2026-07-31
feat(evaluator): add content-addressed revisions for tasks and tasksets

SandyChapman:
## What

Stored tasks and tasksets are now **versioned**. Creating one publishes revision 1; replacing its content publishes the next. Earlier revisions stay readable, so an evaluation can be re-run against exactly the content it ran against before.

This is phase 0 of backing a Harbor-compatible dataset publish endpoint with NeMo entities — Harbor pins dataset members by digest, which we had no way to represent.

## How it works

A revision is an **immutable child entity** named `rev.<n>`, addressed by a SHA-256 digest of its content. The head record keeps current content plus mutable pointers (`latest_revision`, and a tag → ordinal map). Digest lookups are a real `(parent, data.content_hash)` query, not a denormalized index.

References gain the platform's standard `#` sub-entity fragment — `workspace/name#<tag-or-digest>` — with an absent fragment meaning `latest`. This uses the same convention filesets already use for a contained file.

**Taskset membership is resolved to exact digests on write**, and taskset expansion at eval time loads each member's *pinned* revision. That second half matters: without it, membership was pinned in storage and ignored at runtime, so a suite looked reproducible and wasn't.

## API

| Method | Path | |
|---|---|---|
| `POST` | `/{tasks,tasksets}/{name}` | Create + publish revision 1 (409 if taken) |
| `PUT` | `/{tasks,tasksets}/{name}` | Replace + publish; **201** if a revision was cut, **200** if content was unchanged |
| `GET` | `/{tasks,tasksets}/{name}/revisions` | List published revisions, newest first (paginated) |
| `GET` | `/{tasks,tasksets}/{name}/revisions/{revision}` | Read content as of a digest or tag |
| `PUT` | `/{tasks,tasksets}/{name}/tags/{tag}?revision=` | Point a tag at an existing revision |

Publishing is idempotent against the **current** revision: re-PUTting unchanged content cuts no revision but still applies tags, which is how you tag a revision after the fact. Reverting to *older* content does publish — the record genuinely changed, and deduping onto the old revision would leave the head serving content `latest` doesn't name.

## Reviewing this

It's a large diff, but most of it isn't judgement calls:

| | Lines | |
|---|---|---|
| `openapi/openapi.yaml` | +587 | generated by `make refresh-openapi` — skim |
| tests | +2678 | read selectively |
| docs | +130 | quick |
| **source** | **+1918** | the actual review |

Within the source, `tasks.py`/`tasksets.py`, the two services, and the two SDK modules are near-symmetric pairs — read one and skim its twin.

**Suggested order, highest value first:**

1. `entities.py` — what a revision *is*, and which fields the digest covers (`REVISION_POINTER_FIELDS` / `REVISION_SELF_FIELDS`). Everything else depends on this projection.
2. `revisions.py` — publish, resolve, tag. The concurrency lives here: ordinal allocation, the forward-only `latest` rule, and what happens to head content when a pointer write loses its race.
3. `api/service/task_service.py` — where those semantics become observable (200 vs 201, the rollback, the no-op path).
4. Everything else — routes, SDK, docs are mechanical once the above is settled.

**Worth the most scrutiny:** the interleavings in `publish_revision` and `_point_tags`. Three defects found during review were concurrency or ordering cases where a comment described the intended invariant and the code handled only the common path.

## Invariants

- **Every record has at least one revision.** Create rolls back if its first publish fails, so there's no draft state for consumers to handle.
- **Published taskset membership is digest-pinned**, enforced on the field rather than in the publish path.
- **`latest` only moves forward** and can't be moved by hand. Ordinal allocation is serialized by the child create and retries on contention.
- **A plain read and a `#latest` read always agree.** The head's content is always the revision `latest` names — including when a publish fails midway, and when two publishers race and the loser's revision is older.
- **Revision content is verified on read.** Resolving a revision re-hashes its content and compares it against the digest stored beside it, so a revision mutated in place is refused rather than silently served.

## Testing

- 673 unit tests in this plugin; 2253 across the packages this touches
- 9 integration tests against a real platform (`RUN_AGENT_EVAL_INTEGRATION=1`), covering what only real persistence confirms: parent-scoped ordinals, the digest query, server-side ordering, FK cascade, cross-record isolation
- Coverage on new modules: `content_hash` 100%, `task_refs` 100%, `revisions` 96%, services 94%
- `tools/lint/lint-all.sh`: 13/13 pass

## Notes for review

- **No back-compat shims** — there are no known users of the task/taskset endpoints, confirmed before starting.
- **Deleting an individual revision is deliberately not exposed.** A pinned reference would become unresolvable. Registry convention (and Harbor's own model) is *yank*, not delete; not built, since nothing needs it.
- **Known limitation:** nothing prevents deleting a task that a taskset pins. The reference dangles and fails loudly on read. Harbor has the same hole; preventing it needs a reverse index we don't have.
- **`plugins/nemo-customizer/openapi/openapi.yaml` is deliberately not included.** `make refresh-openapi` regenerates ~1100 lines there from pre-existing drift unrelated to this change; reverted to keep the diff focused. Worth a separate cleanup PR.

## Summary by CodeRabbit

* **New Features**
  * Added immutable task and taskset revisions with content digests.
  * Added revision history, retrieval by revision or tag, and revision tagging.
  * Added idempotent replace and upsert operations.
  * Taskset memberships are pinned to exact task revisions for consistent evaluations.
  * SDKs support synchronous and asynchronous revision management.

* **Documentation**
  * Expanded guidance for revisions, tags, pinned evaluations, statuses, and validation errors.

* **Bug Fixes**
  * Improved validation for duplicate references, missing revisions, invalid digests, and concurrent updates.

### DIFF
--- a/docs/evaluator/manage-tasks-tasksets.mdx
+++ b/docs/evaluator/manage-tasks-tasksets.mdx
@@ -18,15 +18,21 @@ metrics inline every time.
 | Concept | What it is | Members |
 |---------|------------|---------|
 | **Task** | A reusable agent-eval unit: `intent`, `inputs`, and the `metrics` that score it. | References the metrics that score it. |
-| **Taskset** | A flexible grouping of tasks with a description and metadata. | References member tasks by `workspace/name`. Membership is a **set** — order is not significant and duplicate references are rejected. |
+| **Taskset** | A flexible grouping of tasks with a description and metadata. | References member tasks by `workspace/name`, each pinned to an exact revision. Membership is a **set** — order is not significant and duplicate references are rejected. |
+| **Revision** | An immutable published snapshot of a task's or taskset's content, addressed by a content digest. | Belongs to the task or taskset it snapshots. |
 
 Both are addressed by `workspace/name`. Names are unique within a workspace, limited to 255
 characters, and must match `^[\w\-\.]+$`.
 
+Every stored task and taskset is **versioned**. Creating one publishes revision 1; replacing its
+content publishes the next revision. Earlier revisions stay readable for as long as the task or
+taskset exists — deleting it removes its revisions with it — which is what lets an evaluation be
+re-run against exactly the content it ran against the first time.
+
 <Note>
-Tasks and tasksets support **create, retrieve, list, and delete** — there is no update. To change a
-stored task or taskset, delete it and create a new one, or store a new version under a different
-name.
+Publishing is **idempotent**. Replacing a task with content identical to its current revision
+publishes nothing and returns the existing revision — so a pipeline can re-submit the same
+definition freely without accumulating versions.
 </Note>
 
 ## Initialize the SDK
@@ -89,6 +95,7 @@ print(stored.id, stored.metrics)
 | `metrics` | `list[MetricRefOrInline]` | No | The metrics that score the task, as `MetricRef` references (`workspace/name`) to stored metrics. Pre-built inline metric bundles (`MetricInline`) are also accepted and are normalized to stored metrics on create. |
 | `views` | `dict[str, SemanticView]` | No | Optional reporting views mapping metric outputs into named semantic scores. |
 | `metadata` | `list[MetadataItem]` | No | Key/value annotations. Keys must be unique. |
+| `tags` | `list[str]` | No | Tags to point at the revision this request publishes. `latest` is always applied server-side. |
 
 <Note>
 A stored task holds **metric references only**. Any inline metric bundle you pass on create is stored
@@ -99,25 +106,96 @@ why `stored.metrics` always comes back as a list of `MetricRef` references.
 ### Retrieve, list, and delete
 
 ```python
-# Retrieve one task by name
+# Retrieve one task by name (its current content)
 task = tasks.retrieve("capital-of-france")
+print(task.revision, task.tags)  # e.g. 1 {'latest': 1}
 
 # List tasks in the workspace (paginated)
 page = tasks.list(page=1, page_size=100, sort="-created_at")
 for item in page.data:
     print(item.name, item.intent)
 
-# Delete a task
+# Delete a task (this also removes all of its revisions)
 tasks.delete("capital-of-france")
 ```
 
 `sort` accepts `name`, `created_at`, or `updated_at`, each optionally prefixed with `-` for
 descending order.
 
+## Revisions
+
+### Publish a new revision
+
+Use `replace` to publish new content. It creates the task if it does not exist, so a publisher needs
+no existence check.
+
+```python
+revised_task = TaskInput(
+    intent="Answer the user's geography question with the capital city.",
+    inputs=TaskInputs(instruction="Name the capital city of France."),
+    metrics=[MetricRef("default/answer-exact-match")],
+    metadata=[MetadataItem(key="suite", value="geography")],
+)
+
+updated = tasks.replace("capital-of-france", task=revised_task)
+print(updated.revision)  # 2
+```
+
+Submitting content identical to the current revision publishes nothing and returns the existing
+revision — but any tags in the body are still applied, which is how you tag a revision after the
+fact.
+
+### List revisions
+
+Each entry carries the `content_hash` used to pin a reference.
+
+```python
+page = tasks.list_revisions("capital-of-france")
+for revision in page.data:      # newest first
+    print(revision.revision, revision.content_hash, revision.tags)
+```
+
+### Read a specific revision
+
+Pass a content digest or a tag. This returns the content **as published**, not the current content.
+
+```python
+# Revisions come back newest-first and paginated, so index by ordinal rather than by position —
+# `data[-1]` is only the oldest entry on the page you happen to have fetched.
+page = tasks.list_revisions("capital-of-france")
+digest = next(revision.content_hash for revision in page.data if revision.revision == 1)
+
+original = tasks.retrieve("capital-of-france", revision=digest)
+current = tasks.retrieve("capital-of-france")
+```
+
+### Tag a revision
+
+A tag is a mutable pointer to a revision — useful for marking one as reviewed or approved after it
+has been evaluated.
+
+```python
+tasks.tag("capital-of-france", "blessed", revision=digest)
+
+blessed = tasks.retrieve("capital-of-france", revision="blessed")
+```
+
+<Note>
+`latest` is managed automatically and always names the most recently published revision; it cannot be
+moved by hand. A tag name may not be empty, and may not look like a content digest (64 hexadecimal
+characters) — such a tag could be stored but never resolved, because a digest-shaped reference is
+looked up as a digest rather than as a tag.
+</Note>
+
 ## Manage Tasksets
 
 A taskset references existing tasks by `workspace/name`. All referenced tasks must already exist when
 the taskset is created; a missing or duplicate reference is rejected.
+
+Member references are **resolved to an exact revision when the taskset is stored**. You may submit a
+bare name (`capital-of-france`), a tag (`capital-of-france#latest`), or a digest — what gets stored
+is always `workspace/name#<digest>`. This is why a stored taskset keeps naming the same content even
+after a member task publishes something new, and it is what makes a suite reproducible.
 
 ```python
 from nemo_evaluator.api.schemas import TaskRef, TasksetInput
@@ -132,6 +210,7 @@ taskset = TasksetInput(
 
 stored = tasksets.create("geography-suite", taskset=taskset)
 print(stored.tasks)
+# ['default/capital-of-france#a1b2...', 'default/capital-of-japan#c3d4...']
 ```
 
 ### `TasksetInput` fields
@@ -139,8 +218,9 @@ print(stored.tasks)
 | Field | Type | Required | Description |
 |-------|------|----------|-------------|
 | `description` | `str` | No | Human-readable description of the grouping. |
-| `tasks` | `list[TaskRef]` | No | References to member tasks (`workspace/name`, or bare `name` within the same workspace). Set semantics — duplicates rejected. |
+| `tasks` | `list[TaskRef]` | No | References to member tasks (`workspace/name`, or bare `name` within the same workspace), optionally pinned with `#<tag-or-digest>`. Each is resolved to an exact digest when stored. Set semantics — duplicates rejected. |
 | `metadata` | `list[MetadataItem]` | No | Key/value annotations. Keys must be unique. |
+| `tags` | `list[str]` | No | Tags to point at the revision this request publishes. `latest` is always applied server-side. |
 
 ### Retrieve, list, and delete
 
@@ -153,6 +233,23 @@ for item in page.data:
 
 tasksets.delete("geography-suite")
 ```
+
+Tasksets carry the same revision surface as tasks — `replace`, `list_revisions`, `tag`, and
+`retrieve(revision=...)`:
+
+```python
+# Re-resolving membership after a member task published new content cuts a new revision.
+tasksets.replace("geography-suite", taskset=taskset)
+
+for revision in tasksets.list_revisions("geography-suite").data:
+    print(revision.revision, revision.content_hash)
+```
+
+<Note>
+Re-submitting the *same* member names can still publish a new revision. Members are re-resolved on
+every write, so if a member task published in the meantime the grouping now names different content
+and genuinely differs. A taskset's identity is the exact revisions it names, not the names alone.
+</Note>
 
 Deleting a taskset does not delete its member tasks — a taskset only holds references.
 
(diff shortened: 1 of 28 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/content_hash.py
```diff
@@ -0,0 +1,91 @@
...
+      explicitly (e.g. with an optional field defaulting to ``None``).
+
+    Args:
+        entity: The entity whose content to serialize.
+        exclude: Extra field names to drop on top of ``__base_fields__``. Revisioned entities pass
+            their own revision/tag bookkeeping here — a revision's digest must not cover the
+            revision index that was assigned *because of* that digest.
+    """
+    excluded = set(entity.__base_fields__) | set(exclude or ())
+    payload = entity.model_dump(exclude=excluded, exclude_computed_fields=True, mode="json")
```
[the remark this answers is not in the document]
SandyChapman: Not actioning here. Confirmed the fact — `exclude_computed_fields` landed in 2.12.0 and the declared floors are below it — but this is pre-existing on `main` (`entities/base.py` has used it since #731), so it is not introduced by this PR. Bumping the floor also rewrites `uv.lock` by ~335 lines, pruning ppc64le/s390x/armv7l/musl/riscv64 wheels (x86_64, aarch64 and macOS unaffected). That platform narrowing needs its own PR.

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -99,26 +106,97 @@ why `stored.metrics` always comes back as a list of `MetricRef` references.
...
+Pass a content digest or a tag. This returns the content **as published**, not the current content.
+
+```python
+# Revisions come back newest-first and paginated, so index by ordinal rather than by position —
+# `data[-1]` is only the oldest entry on the page you happen to have fetched.
+page = tasks.list_revisions("capital-of-france")
+digest = next(revision.content_hash for revision in page.data if revision.revision == 1)
+
+original = tasks.retrieve("capital-of-france", revision=digest)
+current = tasks.retrieve("capital-of-france")
```
ngoncharenko: should we make it a bit more clear:
```python
original = tasks.retrieve("capital-of-france", revision=digest)  # revision 1
current = tasks.retrieve("capital-of-france")                   # revision 2 (latest)
assert original.revision == 1 and current.revision == 2
assert original.inputs.instruction != current.inputs.instruction
```

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -99,26 +106,97 @@ why `stored.metrics` always comes back as a list of `MetricRef` references.
...
+
+### Tag a revision
+
+A tag is a mutable pointer to a revision — useful for marking one as reviewed or approved after it
+has been evaluated.
+
+```python
+tasks.tag("capital-of-france", "blessed", revision=digest)
+
+blessed = tasks.retrieve("capital-of-france", revision="blessed")
```
ngoncharenko: I'd re-order fields a bit in this function signature and also add explicit tag arg. Using tag as revision is not intuitive UX to me. It's subjective. I wonder how others perceive this when reading the doc for the first time.

```python
# (1st key, 2nd key, 3rd key) 
tasks.tag("capital-of-france", revision=digest, tag="blessed")

# it's surprising to use tag as revision to me
blessed = tasks.retrieve("capital-of-france", tag="blessed")
```

In the example above, if two different revisions have the same tag (eg, both are "reviewed"), two tasks will be returned?

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -178,8 +275,10 @@ input_spec = AgentEvalInputSpec(
 )
 ```
 
-When the job runs, the taskset reference is resolved: its member tasks are loaded, and each task's
-stored metric references are hydrated into runnable metrics — exactly as if you had inlined them. The
+When the job runs, the taskset reference is resolved: each member's **pinned revision** is loaded —
+not whatever that task currently contains — and its stored metric references are hydrated into
+runnable metrics, exactly as if you had inlined them. Re-running the same taskset therefore evaluates
+the same content, even if a member task has been republished since. The
```
ngoncharenko: Suggest to replace this section with this explanation:
---

When the job runs, the taskset reference is resolved like this:

- Each member is loaded at the **revision pinned in the taskset**, not the task's current tip.
- Metric references on those members are hydrated into runnable metrics, the same as for inline tasks.
- Re-running the same taskset therefore evaluates the same content, even if a member has been republished since.

```text
# What is stored on the taskset (digests fixed at create/replace time)
geography-suite@1
  └─ tasks:
       - default/capital-of-france#aaa111...   ← revision 1 content
       - default/capital-of-japan#bbb222...

# Later: the member task publishes new content
capital-of-france tip → revision 2 (#ccc333...)

# Re-grade still uses the taskset pins, not the tip
JobRun A  ──TasksetRef("geography-suite")──► load #aaa111..., #bbb222...
JobRun B  ──TasksetRef("geography-suite")──► load #aaa111..., #bbb222...   ← same content
```

To evaluate the *new* task content, publish a new taskset revision (for example
`tasksets.replace(...)`) so membership is re-resolved to the newer digests.

Submit this spec as the agent-evaluate job input

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -227,17 +326,29 @@ The SDK resources are a thin client over the Evaluator plugin REST API, mounted
...
+| `PUT` | `/tasksets/{name}/tags/{tag}?revision=` | Point a tag at an existing revision. |
+| `DELETE` | `/tasksets/{name}` | Delete a taskset and all of its revisions. |
+
+`PUT` distinguishes its two outcomes by status: **`201`** when a new revision was published, and
+**`200`** when the submitted content was already the current revision and nothing was cut.
+
+`POST` on a name that already exists returns `409`, as does a `PUT` that loses a race with a
+concurrent write. An invalid metric reference (task), a missing or duplicate task reference
+(taskset), or a reserved or malformed tag name returns `422`. Retrieving or deleting a name — or a
+revision — that does not exist returns `404`.
```
ngoncharenko: Maybe this should be subsection `### Edge Cases` and bullet point for each case and expected status code/behavior, can organize even in a table format

ngoncharenko (approved):

plugins/nemo-evaluator/src/nemo_evaluator/revisions.py
```diff
@@ -0,0 +1,493 @@
...
+                extra={"record": f"{head.workspace}/{head.name}", "ordinal": ordinal, "attempt": attempt + 1},
+            )
+            _refresh_pointers(head, await head_client.get(type(head), name=head.name, workspace=head.workspace))
+            # The refresh above adopts the *stored* allocation state, which does not necessarily
+            # account for the ordinal we just lost: if the winner's own pointer write never landed,
+            # the stored head still names N-1 and we would recompute N and lose again, every attempt
+            # until the retries run out — and every later publish would do the same, leaving the
+            # record permanently unpublishable. The failed create is itself proof that N is taken,
+            # so step past it explicitly rather than trusting the head to say so.
+            head.latest_revision = max(head.latest_revision, ordinal)
```
ngoncharenko, line 438: [P2] Deduplicate identical contended publishes

If two identical PUTs overlap, publisher A can create rev.2 before updating the head; publisher B then conflicts on rev.2, observes the still-stale head, steps forward, and creates rev.3 with the same digest. Both requests report a new revision, violating the idempotency guarantee. Inspect the conflicting revision and adopt it when its digest matches before allocating the next ordinal.

plugins/nemo-evaluator/src/nemo_evaluator/api/service/taskset_service.py
```diff
@@ -116,41 +193,187 @@ async def _validate_tasks_exist(self, tasks: list[TaskRef], *, workspace: str) -
...
+        """
+        self._reject_duplicate_members(taskset_input.tasks, workspace=workspace)
+        limit = asyncio.Semaphore(_MEMBER_RESOLUTION_CONCURRENCY)
+
+        async def _bounded(ref: TaskRef) -> TaskRef:
+            async with limit:
+                return await self._pin_member(ref, workspace=workspace)
+
+        # ``gather`` preserves input order, so membership order survives resolution — which matters
+        # because a taskset's digest depends on member order.
```
ngoncharenko: [P2] Canonicalize taskset membership before hashing

Tasksets are documented and validated as unordered sets, but preserving request order makes [a, b] and [b, a] produce different hashes and revisions. Sort the resolved digest-pinned references into a canonical order before publishing so equivalent membership remains idempotent.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
```diff
@@ -138,18 +140,56 @@ class MetricInline(BaseModel):
...
+#: and then never appear in a reference, which is a silent dead end rather than an error.
+REF_FRAGMENT_CHARSET = r"[\w\-.]+"
+
+# A *sub-entity* reference adds an optional ``#fragment``, the platform's standard way of addressing
+# something contained within an entity (filesets address a contained file the same way:
+# ``workspace/fileset#path``). For a revisioned entity the fragment selects a revision — either a tag
+# (``#latest``, ``#candidate``) or a full 64-char content digest.
+#
+# Deliberately a sibling of ``_ENTITY_REF_PATTERN`` rather than a widening of it: that constant is
+# shared by ``MetricRef`` and ``TasksetRef``, neither of which has revisions yet, and admitting a
```
ngoncharenko, line 154: [P2] Let evaluations pin taskset revisions

Taskset revisions and digests are now exposed through REST and the SDK, but a digest-qualified TasksetRef is still rejected and resolution always loads the head. Consequently a rerun after replacing the taskset expands different membership. Accept revision fragments on TasksetRef and resolve the corresponding TasksetRevisionEntity before expanding its members.

plugins/nemo-evaluator/src/nemo_evaluator/revisions.py
SandyChapman: Fixed in c0155711e5. Confirmed the interleaving: the dedup check reads the revision `latest` names, so it misses a winner whose `_point_tags` has not landed — the loser refreshes onto a head still naming N-1, recomputes the same digest, steps past the contended ordinal, and cuts a byte-identical rev.N+1 with both callers reporting 201.

`publish_revision` now inspects the revision that actually beat it on `NemoEntityConflictError` and adopts it when the digest matches, returning `created=False` so the loser gets 200. Covered by `test_identical_contended_publish_adopts_the_winners_revision` (verified failing before the fix) plus `test_contended_publish_of_different_content_still_allocates_a_new_ordinal`, so adoption stays keyed on the digest rather than on losing the race.

plugins/nemo-evaluator/src/nemo_evaluator/api/service/taskset_service.py
SandyChapman: Fixed in c0155711e5. You are right that the field and the hash disagreed — `TasksetInput.tasks` documents set semantics and rejects duplicates, but the digest covered the stored list, so `[a, b]` and `[b, a]` were one grouping under two digests.

`_resolved_content` now sorts the resolved digest-pinned refs before publishing. Sorting the *stored* content rather than only the hash input keeps the digest a hash of exactly what is stored, which is what verification on read re-checks. Documented on the field and in the guide; `test_reordering_members_publishes_no_revision` covers the idempotency, and the old `test_member_resolution_preserves_order` is replaced by `test_membership_is_stored_in_canonical_order`.

One consequence worth naming: `GET` now echoes canonical order rather than submitted order. Cheap to take now, breaking later, and there are no existing users.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
SandyChapman: Agreed, and this is the sharpest one — members are pinned but a `TasksetRef` always resolves the head, so replacing the taskset changes what a re-submitted spec evaluates. The code already anticipated it: the comment on `_SUBENTITY_REF_PATTERN` says `TasksetRef` moves onto that pattern once taskset revisions are addressable, which as of this PR they are.

Taking it as a fast-follow rather than growing this diff further: it needs the pattern swap, `parse_subentity_ref` in the expansion path, resolving `TasksetRevisionEntity` before expanding members, tests, and an OpenAPI regen. I have added a note to the docs in c0155711e5 so the current limitation is explicit rather than implied, and will open the follow-up PR as soon as this lands.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Adopted in c0155711e5 — the example now annotates both calls and asserts `original.revision == 1 and current.revision == 2` plus the differing instruction. That last assertion holds against the `replace` example earlier on the page.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Taking the change request: `retrieve` now takes `revision=` (digest) or `tag=`, not both, and `tag()`'s two selectors are keyword-only — `tasks.tag(name, tag="blessed", revision=digest)`. Same route segment underneath, so it's call-site readability only. On your question: a tag can only ever name one revision (tags are stored as tag → ordinal on the head, so re-tagging moves the pointer); I've added that to the docs.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Adopted essentially verbatim in c0155711e5 — bullets, diagram, and the closing note about republishing the taskset. I added one note below it: `TasksetRef` can't yet carry a `#<tag-or-digest>` fragment, which is the follow-up tracked in your `api/schemas.py` thread.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Done in c0155711e5 — now an `### Edge Cases` subsection with a table covering the 201/200 split, identical and reordered content, reverts, the 409/422/404 families, and the known hole where deleting a task a taskset pins returns 204 and leaves the reference dangling.

PR #1028, opened 2026-07-31
fix(nemo-agents): move fabric version pin to 0.1.0

mmogallapalli:
## Summary

Update the NeMo Agents Fabric dependencies and wheel-install constraint from `0.1.0rc6` to the stable `0.1.0` release. This prevents the wheel smoke test from attempting to resolve the unavailable prerelease runtime dependency and keeps the Fabric runtime and adapter packages on matching versions.

[Example failing CI Job](https://github.com/NVIDIA-NeMo/nemo-platform/actions/runs/30662528530/job/91262212996)

## Verification

- Confirmed `nemo-fabric==0.1.0` is installed in the project virtual environment.
- Verified the agent plugin create, subprocess deploy, and invoke flow with a Fabric-backed Codex agent.
- Confirmed the deployment reached `running` and returned the expected `platform codex works` response.
- Verified the updated lockfile with `uv lock --check`.

### DIFF
--- a/.github/wheel-constraints/nemo-platform-services.txt
+++ b/.github/wheel-constraints/nemo-platform-services.txt
@@ -41,7 +41,7 @@ langchain-openai==1.4.0
 langchain==1.3.14
 lark==1.3.1
 nemo-anonymizer==0.3.1
-nemo-fabric==0.1.0rc6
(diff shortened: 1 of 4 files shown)


svvarom (approved):
LGTM

PR #1031, opened 2026-07-31
feat(jobs): add watch command to stream job and deployment events

ironcommit:
## Summary

Adds a `nemo jobs watch` CLI command and corresponding SDK watcher methods that stream job lifecycle events - status transitions, log lines, and warnings - directly to the terminal or programmatic consumers. The sync and async Jobs SDK clients both expose `watch_job(...)`; the async client returns an async iterator backed by `async_watch_job(...)`.

### Core watch engine (`jobs/watch.py`)
- Synchronous and async iterators that poll the Jobs API and yield typed `JobWatchEvent` objects (status, log, warning)
- Deduplication of log lines across polls using cursor tracking
- Configurable filtering by `attempt_id`, `step_id`, and `task_id`
- Timeout enforcement with graceful completion on terminal job states
- Transient-error retry on the next configured `poll_interval`

### CLI command (`nemo jobs watch`)
- Rich-rendered terminal output via the job watch renderer - color-coded status changes, streaming logs, and warning highlights
- Options: `--timeout`, `--poll-interval`, `--history/--no-history`, `--attempt-id`, `--step-id`, `--task-id`, `--limit`, `--workspace`
- Log streaming is enabled for CLI watch; `include_logs=False` is SDK-only
- Non-zero exit code on job failure

### `--watch` flag on create commands
- `nemo jobs create --watch` and `nemo inference deployments create --watch` automatically watch after creation
- Mutual exclusion with the existing `--wait` flag
- Code generator updated to emit `--watch` support for configured create commands via `cli_config.yaml`

### SDK client integration
- `client.jobs.watch_job(name, ...)` -> sync iterator of `JobWatchEvent`
- `async_client.jobs.watch_job(name, ...)` -> async iterator of `JobWatchEvent`
- Plugin SDK clients wire watch into both sync and async `NeMoPlatform` clients through the Jobs resource

### Waiters refactor
- Renamed "Wait Options" panel to "Lifecycle Options" across all commands
- Consolidated shared timeout/poll-interval options

## Testing

- Unit tests for the watch engine covering status transitions, log deduplication, timeout, retries, filtering, and pagination (`test_watch.py`)
- CLI integration tests for `nemo jobs watch` and `--watch` on create commands (`test_app.py`, `test_create_wait.py`)
- Renderer tests (`test_job_watch_renderer.py`)
- Code generator tests for watch flag injection (`test_code_generator.py`, `test_config.py`)

## Summary by CodeRabbit

* **New Features**
  * Added `jobs watch` to stream job status, logs, warnings, and completion results.
  * Added `--watch` support for job and inference deployment creation, with timeout and polling controls.
  * Added filters for attempts, steps, tasks, logs, history, and pagination.
  * Added synchronous and asynchronous job monitoring support.
  * Added validation to prevent using `--wait` and `--watch` together.
* **Documentation**
  * Updated CLI reference documentation with lifecycle options and watch command usage.
* **Bug Fixes**
  * Improved handling of transient errors, timeouts, pagination, and terminal job states.

### DIFF
--- a/docs/cli/reference.mdx
+++ b/docs/cli/reference.mdx
@@ -2116,15 +2116,16 @@ nemo inference deployments create [OPTIONS] [NAME]
 * `--input-file`: Path to JSON file (use '-' for stdin)
 * `--input-data`: Input data for the request (JSON or YAML)
 
+**Lifecycle Options:**
+
+* `--wait`: Wait for the created deployment to be up and running
+* `--watch`: Watch the created deployment until it is stable while streaming status updates
+* `--timeout <INTEGER RANGE>`: Maximum time to wait or watch in seconds [default: 1200]
+* `--poll-interval <INTEGER RANGE>`: Seconds between status checks [default: 3]
+
 **Output Options:**
 
 * `--output-format, --output, -f <CHOICE>`: Output format for an entity. [possible values: json, yaml, raw, code]
-
-**Wait Options:**
-
-* `--wait`: Wait for the created deployment to reach a terminal state
-* `--timeout <INTEGER RANGE>`: Maximum time to wait in seconds [default: 1200]
-* `--poll-interval <INTEGER RANGE>`: Seconds between status checks [default: 3]
 
 ##### nemo inference deployments delete
 
@@ -3919,6 +3920,7 @@ nemo jobs [OPTIONS] COMMAND [ARGS]...
 * `resume`: Resume a paused platform job.
 * `get`: Get a platform job by name.
 * `update-status-details`: Update the status details of a platform job.
+* `watch`: Watch a platform job until it reaches a terminal status.
 * `results`: Manage results
 * `steps`: Manage steps
 * `tasks`: Manage tasks
@@ -3994,6 +3996,13 @@ nemo jobs create [OPTIONS] [NAME]
 
 * `--input-file`: Path to JSON file (use '-' for stdin)
 * `--input-data`: Input data for the request (JSON or YAML)
+
+**Lifecycle Options:**
+
+* `--wait`: Wait for the created job to reach a terminal state without streaming logs
+* `--watch`: Watch the created job to a terminal state
+* `--timeout <INTEGER RANGE>`: Maximum time to wait or watch in seconds
+* `--poll-interval <INTEGER RANGE>`: Seconds between status checks [default: 3]
 
 **Output Options:**
 
@@ -4271,6 +4280,35 @@ nemo jobs update-status-details [OPTIONS] NAME
 **Output Options:**
 
 * `--output-format, --output, -f <CHOICE>`: Output format for an entity. [possible values: json, yaml, raw, code]
+
+#### nemo jobs watch
+
+Watch a platform job until it reaches a terminal status.
+
+**Usage:**
+
+```shell
+nemo jobs watch [OPTIONS] NAME
+```
+
+**Arguments:**
+
+* `<NAME>`: Name of the platform job to watch
+
+**Options:**
+
+* `--workspace`: Workspace containing the job
+* `--attempt-id <INTEGER>`: Filter logs to an attempt ID
+* `--step-id`: Filter logs to a step ID
+* `--task-id`: Filter logs to a task ID
+* `--limit <INTEGER RANGE>`: Maximum logs to fetch per page
+* `--timeout <INTEGER RANGE>`: Maximum watch time in seconds
+* `--poll-interval <INTEGER RANGE>`: Seconds between status checks [default: 3]
+* `--history, --no-history`: Include logs already present before watching
+
+**Help:**
+
+* `--help, -h`: Show this message and exit.
 
 #### nemo jobs results
 
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py
@@ -82,15 +82,28 @@ def create_deployments(
     wait: Annotated[
         bool,
         typer.Option(
-            "--wait", help="Wait for the created deployment to reach a terminal state", rich_help_panel="Wait Options"
+            "--wait", help="Wait for the created deployment to be up and running", rich_help_panel="Lifecycle Options"
+        ),
+    ] = False,
+    watch: Annotated[
+        bool,
+        typer.Option(
+            "--watch",
+            help="Watch the created deployment until it is stable while streaming status updates",
+            rich_help_panel="Lifecycle Options",
         ),
     ] = False,
     timeout: Annotated[
-        int, typer.Option("--timeout", min=1, help="Maximum time to wait in seconds", rich_help_panel="Wait Options")
+        int,
+        typer.Option(
+            "--timeout", min=1, help="Maximum time to wait or watch in seconds", rich_help_panel="Lifecycle Options"
+        ),
     ] = 1200,
     poll_interval: Annotated[
         int,
-        typer.Option("--poll-interval", min=1, help="Seconds between status checks", rich_help_panel="Wait Options"),
+        typer.Option(
+            "--poll-interval", min=1, help="Seconds between status checks", rich_help_panel="Lifecycle Options"
+        ),
     ] = 3,
 ) -> None:
     """Create a new ModelDeployment (version 1).
@@ -137,12 +150,17 @@ def create_deployments(
     state: CLIContext = ctx.obj
     output_format = state.get_output_format(output_format)
 
+    if wait and watch:
+        raise typer.BadParameter("Cannot combine --wait and --watch.")
+
     if handle_code_generation(
         ["inference", "deployments"],
         "create",
         all_kwargs,
         output_format,
         state,
+        watch_config={"type": "inference_deployment", "resource_label": "deployment"} if watch else None,
+        watch_options={"timeout": timeout, "poll_interval": poll_interval} if watch else None,
         wait_config={"type": "inference_deployment", "resource_label": "deployment"} if wait else None,
         wait_options={"timeout": timeout, "poll_interval": poll_interval} if wait else None,
     ):
@@ -158,11 +176,10 @@ def create_deployments(
         no_truncate=state.get_no_truncate(),
         timestamp_format=state.get_timestamp_format(),
     )
-
-    if wait:
+    if wait or watch:
         wait_name = getattr(result, "name", None) or all_kwargs.get("name")
         if not wait_name:
-            raise RuntimeError("Unable to determine created resource name for --wait")
+            raise RuntimeError("Unable to determine created resource name for --wait/--watch")
         wait_workspace = all_kwargs.get("workspace")
         if not wait_for_inference_deployment(
             client,
@@ -172,6 +189,7 @@ def create_deployments(
             poll_interval=poll_interval,
         ):
             raise typer.Exit(1)
+        return
 
 
 @app.command("delete")
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/jobs/__init__.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/jobs/__init__.py
@@ -8,6 +8,8 @@ from importlib import import_module as _importlib_import_module
 from typing import Annotated, Literal
 
 import typer
+from nemo_platform_plugin.client.adapter import client_from_platform
+from nemo_platform_plugin.jobs.client import JobsClient
 
 from nemo_platform_ext.cli.core.api import build_kwargs, merge_filter_dict
 from nemo_platform_ext.cli.core.code_generator import handle_code_generation
@@ -20,6 +22,7 @@ from nemo_platform_ext.cli.core.formatters import (
     validate_stream_output_format,
 )
 from nemo_platform_ext.cli.core.help_formatter import collect_warnings, create_typer_app
+from nemo_platform_ext.cli.core.job_watch_renderer import JobWatchRenderResult, render_job_watch_events
 from nemo_platform_ext.cli.core.pagination import PaginationType, fetch_all_pages, warn_if_more_pages
 from nemo_platform_ext.cli.core.stdin_utils import read_data_input_with_flags, read_payload, validate_required_fields
 from nemo_platform_ext.cli.core.types import (
@@ -29,6 +32,7 @@ from nemo_platform_ext.cli.core.types import (
     OutputColumnsOption,
     StreamOutputOption,
 )
+from nemo_platform_ext.cli.core.waiters import wait_for_platform_job
 
 _cli_child_results = _importlib_import_module("nemo_platform_ext.cli.commands.api.jobs.results")
 _cli_child_steps = _importlib_import_module("nemo_platform_ext.cli.commands.api.jobs.steps")
@@ -102,6 +106,30 @@ def create_jobs(
         typer.Option("--input-data", help="Input data for the request (JSON or YAML)", rich_help_panel="Input Options"),
     ] = None,
     output_format: EntityOutputFormatOption = None,
+    wait: Annotated[
+        bool,
+        typer.Option(
+            "--wait",
+            help="Wait for the created job to reach a terminal state without streaming logs",
+            rich_help_panel="Lifecycle Options",
+        ),
+    ] = False,
+    watch: Annotated[
+        bool,
+        typer.Option("--watch", help="Watch the created job to a terminal state", rich_help_panel="Lifecycle Options"),
+    ] = False,
+    timeout: Annotated[
+        int | None,
+        typer.Option(
+            "--timeout", min=1, help="Maximum time to wait or watch in seconds", rich_help_panel="Lifecycle Options"
+        ),
+    ] = None,
+    poll_interval: Annotated[
+        int,
+        typer.Option(
+            "--poll-interval", min=1, help="Seconds between status checks", rich_help_panel="Lifecycle Options"
+        ),
+    ] = 3,
 ) -> None:
     """Create a new platform job.
 
@@ -156,12 +184,26 @@ def create_jobs(
     state: CLIContext = ctx.obj
     output_format = state.get_output_format(output_format)
 
-    if handle_code_generation(["jobs"], "create", all_kwargs, output_format, state):
+    if wait and watch:
+        raise typer.BadParameter("Cannot combine --wait and --watch.")
+
+    if handle_code_generation(
+        ["jobs"],
+        "create",
+        all_kwargs,
+        output_format,
+        state,
+        watch_config={"type": "platform_job", "resource_label": "job"} if watch else None,
+        watch_options={"timeout": timeout, "poll_interval": poll_interval} if watch else None,
+        wait_config={"type": "platform_job", "resource_label": "job"} if wait else None,
+        wait_options={"timeout": timeout if timeout is not None else 1200, "poll_interval": poll_interval}
+        if wait
+        else None,
+    ):
         return
 
     client = state.get_client()
     result = client.jobs.create(**all_kwargs)
-
     format_output(
         result,
         is_list=False,
@@ -169,6 +211,37 @@ def create_jobs(
         no_truncate=state.get_no_truncate(),
         timestamp_format=state.get_timestamp_format(),
     )
+    if wait or watch:
+        wait_name = getattr(result, "name", None) or all_kwargs.get("name")
+        if not wait_name:
+            raise RuntimeError("Unable to determine created resource name for --wait/--watch")
+        wait_workspace = getattr(result, "workspace", None) or all_kwargs.get("workspace")
+        if wait_workspace is None:
+            wait_workspace = client._get_workspace_path_param()
+        jobs_client = client_from_platform(client, JobsClient)
+        if wait:
+            if not wait_for_platform_job(
+                jobs_client,
+                wait_name,
+                workspace=wait_workspace,
+                resource_label="job",
+                timeout=timeout if timeout is not None else 1200,
+                poll_interval=poll_interval,
+            ):
+                raise typer.Exit(1)
+            return
+        events = jobs_client.watch_job(
+            wait_name,
+            workspace=wait_workspace,
+            timeout=timeout,
+            poll_interval=poll_interval,
+        )
+        watch_result = render_job_watch_events(events, resource_label="job")
+        if watch_result is JobWatchRenderResult.INTERRUPTED:
+            raise typer.Exit(130)
+        if watch_result is not JobWatchRenderResult.SUCCEEDED:
+            raise typer.Exit(1)
+        return
 
 
 @app.command("delete")
@@ -601,3 +674,49 @@ def update_status_details_jobs(
         no_truncate=state.get_no_truncate(),
         timestamp_format=state.get_timestamp_format(),
     )
+
+
+@app.command("watch")
+@collect_warnings
+@handle_errors
+def watch_platform_job(
+    ctx: typer.Context,
+    name: Annotated[str, typer.Argument(help="Name of the platform job to watch")],
+    workspace: Annotated[str | None, typer.Option("--workspace", help="Workspace containing the job")] = None,
+    attempt_id: Annotated[int | None, typer.Option("--attempt-id", help="Filter logs to an attempt ID")] = None,
+    step_id: Annotated[str | None, typer.Option("--step-id", help="Filter logs to a step ID")] = None,
+    task_id: Annotated[str | None, typer.Option("--task-id", help="Filter logs to a task ID")] = None,
+    limit: Annotated[int | None, typer.Option("--limit", min=1, help="Maximum logs to fetch per page")] = None,
+    timeout: Annotated[int | None, typer.Option("--timeout", min=1, help="Maximum watch time in seconds")] = None,
+    poll_interval: Annotated[
+        int,
+        typer.Option("--poll-interval", min=1, help="Seconds between status checks"),
+    ] = 3,
+    include_history: Annotated[
+        bool,
+        typer.Option("--history/--no-history", help="Include logs already present before watching"),
+    ] = True,
+) -> None:
+    """Watch a platform job until it reaches a terminal status."""
+    state: CLIContext = ctx.obj
+    client = state.get_client()
+    jobs_client = client_from_platform(client, JobsClient)
+    if workspace is None:
+        workspace = client._get_workspace_path_param()
+
+    events = jobs_client.watch_job(
+        name,
+        workspace=workspace,
+        attempt_id=attempt_id,
+        step_id=step_id,
+        task_id=task_id,
+        limit=limit,
+        timeout=timeout,
+        poll_interval=poll_interval,
+        include_history=include_history,
+    )
+    watch_result = render_job_watch_events(events, resource_label="job")
+    if watch_result is JobWatchRenderResult.INTERRUPTED:
+        raise typer.Exit(130)
+    if watch_result is not JobWatchRenderResult.SUCCEEDED:
+        raise typer.Exit(1)
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/code_generator.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/code_generator.py
@@ -11,6 +11,11 @@ from typing import Any
 
 from nemo_platform_ext.cli.core.context import CLIContext
 
+_INFERENCE_DEPLOYMENT_LIFECYCLE = "inference_deployment"
+_PLATFORM_JOB_LIFECYCLE = "platform_job"
+_LIFECYCLE_TYPES_WITH_DEADLINES = {_INFERENCE_DEPLOYMENT_LIFECYCLE}
+_LIFECYCLE_TYPES_WITH_STATUS_ERROR_HANDLING = {_INFERENCE_DEPLOYMENT_LIFECYCLE}
+
 
 def handle_code_generation(
     resource_path: list[str],
@@ -20,6 +25,8 @@ def handle_code_generation(
     context: CLIContext,
     wait_config: dict[str, Any] | None = None,
     wait_options: dict[str, Any] | None = None,
+    watch_config: dict[str, Any] | None = None,
+    watch_options: dict[str, Any] | None = None,
 ) -> bool:
     """
     Check if in code generation mode and generate code if needed.
@@ -42,6 +49,8 @@ def handle_code_generation(
             base_url=base_url,
             wait_config=wait_config,
             wait_options=wait_options,
+            watch_config=watch_config,
+            watch_options=watch_options,
         )
         formatted_code = format_code_output(code, language="python")
         print(formatted_code)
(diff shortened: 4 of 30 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py
```diff
@@ -152,18 +170,18 @@ def create_deployments(
...
-            raise RuntimeError("Unable to determine created resource name for --wait")
+            raise RuntimeError("Unable to determine created resource name for --wait/--watch")
         wait_workspace = all_kwargs.get("workspace")
         if not wait_for_inference_deployment(
             client,
             wait_name,
             workspace=wait_workspace,
             timeout=timeout,
             poll_interval=poll_interval,
+            verbose=watch,
```
tylersbray: > **agent** (`triple-review`): R1 — `--wait` silently loses status streaming

**Priority:** P1 (blocking)
**Flagged by:** Opus, Auto

`verbose=watch` means plain `nemo inference deployments create --wait` runs with `verbose=False`. On `main`, this call omitted `verbose` and defaulted to `True`, so `--wait` previously streamed status history and printed the success line. With only `--wait`, users now get the created-entity JSON, then silence until exit. `--watch` restores verbosity, but this is a regression for existing `--wait` callers.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/waiters.py
```diff
@@ -332,60 +360,57 @@ def wait_for_platform_job(
...
+                jobs,
+                name,
+                workspace=workspace,
+                poll_interval=poll_interval,
+                timeout=timeout,
+                include_logs=False,
+            ):
+                wait_elapsed = int(time.time() - start_time)
+                polling_time = datetime.now().strftime("%H:%M:%S")
+                live.update(_make_live_display(polling_time, timeout, poll_interval, wait_elapsed))
```
tylersbray: > **agent** (`triple-review`): R3 — Live elapsed timer freezes between status changes

**Priority:** P2 (non-blocking)
**Flagged by:** Opus

`live.update(...)` only runs when `watch_job` yields an event. With `include_logs=False`, status events are yielded only on status change, so a job that sits in `active` for a long time leaves the transient Live panel frozen at the last transition. On `main`, `wait_for_platform_job` called `live.update` at the top of every poll loop.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/jobs/__init__.py
```diff
@@ -150,12 +177,56 @@ def create_jobs(
...
+        events = watch_job(
+            client,
+            wait_name,
+            workspace=wait_workspace,
+            timeout=timeout,
+            poll_interval=poll_interval,
+        )
+        if not render_job_watch_events(events):
+            raise typer.Exit(1)
+        return
```
tylersbray, line 244: > **agent** (`triple-review`): R4 — `jobs create --wait/--watch` skips `format_output`

**Priority:** P2 (non-blocking)
**Flagged by:** Opus

Both `--wait` and `--watch` branches `return` before `format_output(...)`, so `-f json` / structured output is ignored and the created job entity is never printed. The sibling `inference deployments create` path still calls `format_output` first, then waits/watches.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+        time.sleep(sleep_for)
+
+
+async def _async_sleep(poll_interval: float, deadline: _WatchDeadline) -> None:
+    sleep_for = deadline.sleep_seconds(poll_interval)
+    if sleep_for > 0:
+        await asyncio.sleep(sleep_for)
+
+
+def _status_event(status_response: PlatformJobStatusResponse, job_name: str) -> JobStatusEvent:
```
tylersbray, line 445: > **agent** (`triple-review`): R5 — Failed-job watch output omits `error_details`

**Priority:** P2 (non-blocking)
**Flagged by:** Opus

`JobStatusEvent` / `_status_event` only carries `status` and `status_details`. `PlatformJobStatusResponse` also has `error_details`, which is dropped here. The renderer then prints only `ended with status 'error'`, so `nemo jobs watch` does not surface why the job failed.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+                    name,
+                    options=options,
+                    state=state,
+                    deadline=deadline,
+                )
+                logs_drained = True
+            except _TRANSPORT_ERRORS as exc:
+                yield _transient_failure_warning("log", name, exc)
+            except _HTTP_STATUS_ERRORS as exc:
+                if not _is_retryable_http_error(exc):
```
tylersbray: > **agent** (`triple-review`): R6 — Non-transient log errors abort the whole watch

**Priority:** P2 (non-blocking)
**Flagged by:** Opus

A non-retryable HTTP error from the logs endpoint (`_TRANSIENT_STATUS_CODES` is only `{429, 502, 503, 504}`) re-raises and stops status monitoring too. Logs are advisory relative to status; downgrading log-path failures to a warning (as the transient path already does) would keep the status stream alive through a logs 500.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+
+from nemo_platform._exceptions import APIConnectionError, APIStatusError, APITimeoutError
+from nemo_platform_plugin.client.errors import NemoHTTPError, NemoTransportError
+from nemo_platform_plugin.jobs.client import AsyncJobLogsClient, AsyncJobStatusClient, JobLogsClient, JobStatusClient
+from nemo_platform_plugin.jobs.schemas import PlatformJobLog, PlatformJobStatusResponse
+from nemo_platform_plugin.jobs.types import JobLogsQueryParams
+
+_SUCCESSFUL_TERMINAL_STATUSES = {"completed"}
+_FAILED_TERMINAL_STATUSES = {"cancelled", "error"}
+_TERMINAL_STATUSES = _SUCCESSFUL_TERMINAL_STATUSES | _FAILED_TERMINAL_STATUSES
```
tylersbray, line 37: > **agent** (`triple-review`): R7 — `paused` is not terminal; watch can hang forever

**Priority:** P2 (non-blocking)
**Flagged by:** Auto

`_TERMINAL_STATUSES` is only `{completed, cancelled, error}`. `PlatformJobStatus` also includes `paused` / `pausing` / `resuming`. CLI `nemo jobs watch` and `create --watch` default `timeout=None`, so a paused job polls indefinitely until Ctrl-C.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/jobs/__init__.py
```diff
@@ -150,12 +177,56 @@ def create_jobs(
...
+                client.jobs,
+                wait_name,
+                workspace=wait_workspace,
+                resource_label="job",
+                timeout=timeout if timeout is not None else 1200,
+                poll_interval=poll_interval,
+            ):
+                raise typer.Exit(1)
+            return
+        events = watch_job(
```
tylersbray: > **agent** (`triple-review`): R8 — `create --watch` / `jobs watch` skip `job_run` telemetry

**Priority:** P2 (non-blocking)
**Flagged by:** Auto

`wait_for_platform_job` emits `_emit_job_run_event` on terminal status. The `create --watch` path (and `nemo jobs watch`) call `watch_job` + `render_job_watch_events` directly and return, so successful/failed watched runs are invisible to that telemetry path.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+
+        if not recorded_first_page:
+            state.history_seen = True
+            recorded_first_page = True
+
+        next_cursor = page.metadata["next_page"]
+        if next_cursor is None or next_cursor in visited_cursors:
+            return
+        visited_cursors.add(next_cursor)
+        state.log_cursor = next_cursor
```
tylersbray, line 541: > **agent** (`triple-review`): R9 — Offset/page cursors can permanently skip late-arriving logs

**Priority:** P2 (non-blocking)
**Flagged by:** Opus, GPT-Sol, Auto

The watcher advances `state.log_cursor` as a page/offset cursor over `ORDER BY timestamp` pages. Concurrent or delayed writers can insert lines that sort before the saved page; those lines are never revisited. Worth documenting as a known limitation of poll-based watch rather than presenting the stream as complete.

packages/nemo_platform_ext/src/nemo_platform_ext/client/enhanced.py
```diff
@@ -358,3 +392,35 @@ def __getattr__(self, name: str) -> Any:
         instance = resource_cls(self)
         self.__dict__[name] = instance
         return instance
+
+    def watch_job(
```
tylersbray: > **agent** (`triple-review`): R10 — PR/docs drift vs shipped API and CLI

**Priority:** P2 (non-blocking)
**Flagged by:** Opus, GPT-Sol, Auto

The PR body documents `client.awatch_job(...)`, CLI `--no-logs` / `--include-history`, and “retry with backoff”, but:
- `AsyncNeMoPlatform` exposes `watch_job` (wrapping `async_watch_job`), not `awatch_job`
- CLI watch uses `--history/--no-history` and has no `--no-logs` (SDK `include_logs` is not wired through)
- Transient retries use a fixed `poll_interval` with no backoff

Please align the PR description / docs with the shipped surface.

tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/cli_generator/templates/create_command.py.j2
```diff
@@ -103,38 +151,95 @@ def {{ function_name }}(
...
+            raise RuntimeError("Unable to determine created resource name for {% if has_inference_deployment_wait and has_inference_deployment_watch %}--wait/--watch{% elif has_inference_deployment_wait %}--wait{% else %}--watch{% endif %}")
         wait_workspace = all_kwargs.get("workspace")
-    {% if wait_config.type == "inference_deployment" %}
         if not wait_for_inference_deployment(
             client,
             wait_name,
             workspace=wait_workspace,
             timeout=timeout,
             poll_interval=poll_interval,
+            verbose=watch,
```
tylersbray: > **agent** (`triple-review`): R11 — Template emits `verbose=watch` without a watch-config guard

**Priority:** P3 (non-blocking)
**Flagged by:** Opus

This block is entered for wait-only *or* watch configs, but `verbose=watch` is unconditional. Today both configured resources have wait+watch, so nothing breaks; a future wait-only deployment-style resource would raise `NameError: name 'watch' is not defined` at runtime. Guard with `{% if has_watch_config %}` (or equivalent) or emit `verbose=False` when watch is absent.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+                    raise
+                yield _transient_failure_warning("log", name, exc)
+
+        if status_event.terminal and logs_drained:
+            return
+
+        _sleep(poll_interval, deadline)
+
+
+async def _async_watch_job(
```
tylersbray, line 354: > **agent** (`triple-review`): R12 — Sync/async watch engines are near-duplicates

**Priority:** P3 (non-blocking)
**Flagged by:** Auto, Opus

`_watch_job` / `_async_watch_job` and `_drain_logs` / `_async_drain_logs` mirror the same control flow. Bugfixes (history cursor handling, retry policy, etc.) must be applied twice and can drift. Consider a shared helper or generating one from the other if this keeps growing.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,600 @@
...
+    poll_interval: float,
+) -> Iterator[JobWatchEvent]:
+    while True:
+        deadline.raise_if_expired()
+        try:
+            status_response = jobs.status.get_job_status(workspace=options.workspace, name=name)
+            status_event = _status_event(status_response.data(), name)
+        except _TRANSPORT_ERRORS as exc:
+            yield _transient_failure_warning("status", name, exc)
+            _sleep(poll_interval, deadline)
```
tylersbray: > **agent** (`triple-review`): R13 — Transient retries use fixed interval, not backoff

**Priority:** P3 (non-blocking)
**Flagged by:** Opus, Auto

On transport/5xx, the loop yields a warning then `_sleep(poll_interval, ...)` with no interval growth and no dedupe/cap. Sustained 503s spam identical warnings every poll. The PR body claims “Transient-error retry with backoff,” which this path does not implement.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/code_generator.py
```diff
@@ -72,14 +83,30 @@ def generate_python_code(
...
 
-    if wait_config:
+    if _lifecycle_uses_deadline(lifecycle_type, lifecycle_mode):
         lines.append("import time")
-    if wait_type == "inference_deployment":
+    if _lifecycle_uses_status_error_handling(lifecycle_type, lifecycle_mode):
         lines.append(
             "from nemo_platform import APIConnectionError, APIStatusError, APITimeoutError, NeMoPlatform, NotFoundError"
         )
+    elif lifecycle_type == _PLATFORM_JOB_LIFECYCLE and lifecycle_mode == "wait":
```
mckornfield: conditionals seem a bit duplicative (the wait only adds the last line, yeah?)

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
```diff
@@ -68,6 +122,80 @@ class _JobsMethods:
...
+
 
 class AsyncJobsClient(_JobsMethods, AsyncNemoClient):
     """Async client for the Jobs service API."""
+
+    def watch_job(
+        self,
+        name: str,
+        *,
+        workspace: str | None = None,
```
mckornfield, line 170: probably nit throughout on all these kwargs, might be nice to have a WatchConfig/WaitConfig that's passed in but 🤷  (for here and the other parts of the code)

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,695 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class _AsyncWatchClients:
+    status: AsyncJobStatusClient
+    logs: AsyncJobLogsClient | None
+
+
+@dataclass(frozen=True, slots=True)
+class _WatchOptions:
```
mckornfield, line 60: yeah this is what my comment about too many kwargs was a bout, I think it'd be nice to externalize

pytest.ini
```diff
@@ -99,6 +99,7 @@ asyncio_mode = auto
 addopts =
     --verbose
     --strict-markers
+    --import-mode=importlib
```
mckornfield: weird, what's the default?

tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/vendor/vendor_package.py
```diff
@@ -137,6 +154,32 @@ def leave_ClassDef(self, original_node: cst.ClassDef, updated_node: cst.ClassDef
         return updated_node.with_changes(body=updated_node.body.with_changes(body=body_items))
 
 
+class _ClientStaleImportRemover(cst.CSTTransformer):
+    def __init__(self, stale_imports: dict[str, tuple[str, ...]] = _CLIENT_STALE_IMPORTS) -> None:
+        self._stale_imports = {module_name: set(names) for module_name, names in stale_imports.items()}
+
+    def leave_SimpleStatementLine(
```
mckornfield: lol love the snake vs camel case

tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/vendor/vendor_package.py
```diff
@@ -81,13 +81,20 @@ class ResourceReplacement(BaseModel):
...
 _CLIENT_METHOD_NAMES = ("__init__", "__getattr__")
+_CLIENT_STALE_METHOD_NAMES = ("watch_job",)
 _CLIENT_HELPER_FUNCTION_NAMES = ("_should_bootstrap_config",)
 _CLIENT_INIT_REQUIRED_IMPORTS: dict[str, tuple[str, ...]] = {
     "nemo_platform._base_client": ("DefaultAsyncHttpxClient", "DefaultHttpxClient"),
     "nemo_platform.client.tls": ("client_verify_from_env",),
     "nemo_platform_plugin.client.constants": ("WORKLOAD_IDENTITY_TOKEN_FILE_ENVVAR",),
     "pathlib": ("Path",),
 }
+_CLIENT_STALE_IMPORTS: dict[str, tuple[str, ...]] = {
```
mckornfield: why do we need this? would we not just delete them?

mckornfield (approved):
some questions but lg

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py
```diff
@@ -152,20 +170,21 @@ def create_deployments(
...
```
tylersbray: > **agent** (`follow-up-review`): R1 — `--wait` still silently loses status streaming

**Priority:** P1 (blocking)
**Re-checked on:** `58580df`

`verbose=watch` is still in place, so plain `nemo inference deployments create --wait` runs with `verbose=False`. On `main`, this call omitted `verbose` and defaulted to `True`, streaming status history and printing the success line.

**Suggested fix:** keep `--watch` verbose, but restore verbose `--wait`, e.g. `verbose=True` when not watching, or `verbose=watch or wait` / pass `verbose=True` for the wait-only path.

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,695 @@
...
+        yield from _new_log_events(
+            page.items,
+            state=state,
+            occurrence_counts=occurrence_counts,
+            name=name,
+            emit=emit_logs,
+        )
+
+        if not recorded_first_page:
+            state.history_seen = True
```
tylersbray: > **agent** (`follow-up-review`): R2 — `include_history=False` can leak historical logs after a mid-pagination failure

**Priority:** P1 (blocking)
**Re-checked on:** `58580df`

`emit_logs = state.history_seen` is snapshotted at drain start, then after page 1 this line sets `state.history_seen = True`. If later pages fail, the next poll starts with `history_seen=True` and continues from the advanced cursor, so remaining historical pages are emitted even though the user passed `--no-history` / `include_history=False`.

**Suggested fix:** only mark history seen after a full successful drain (or keep emit false for unread historical pages). Same pattern exists on the async twin (~line 580).

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,695 @@
...
+from nemo_platform_plugin.jobs.watch_types import (
+    JobLogEvent,
+    JobStatusEvent,
+    JobWarningEvent,
+    JobWatchEvent,
+    JobWatchTimeoutError,
+)
+
+_SUCCESSFUL_TERMINAL_STATUSES = {"completed"}
+_FAILED_TERMINAL_STATUSES = {"cancelled", "error", "paused"}
```
tylersbray, line 36: > **agent** (`follow-up-review`): R7 — confirm whether `paused` should exit as failure

**Priority:** P2 (confirm intent)
**Re-checked on:** `58580df`

The earlier hang is fixed by treating `paused` as terminal here. Because it sits in `_FAILED_TERMINAL_STATUSES`, watch reports `successful=False` and the CLI exits non-zero.

**Question:** is that intentional? Pause is often a scheduler hold rather than a job error. Alternatives: keep `paused` terminal but not failed (neutral/`successful=None`), or treat it as successful terminal. Happy to leave as-is if exit-as-failure is the product choice.

tylersbray:
> **agent** (`follow-up-review`): Remaining items after re-check of `58580df`

Fresh open inline threads (not replies on resolved conversations):
- R1: https://github.com/NVIDIA-NeMo/nemo-platform/pull/1031#discussion_r3731705534
- R2: https://github.com/NVIDIA-NeMo/nemo-platform/pull/1031#discussion_r3731705641
- R7: https://github.com/NVIDIA-NeMo/nemo-platform/pull/1031#discussion_r3731705747

## Still open
- [ ] **R1 (P1)** `packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py:184` — `verbose=watch` silences plain `--wait`. Prefer default verbose for `--wait` (e.g. `verbose=True` when not watching).
- [ ] **R2 (P1)** `packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py:538` (+ async ~580) — don’t set `history_seen=True` until a full history drain completes when `include_history=False`.
- [ ] **R7 (P2, confirm)** `packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py:34` — `paused` in `_FAILED_TERMINAL_STATUSES` exits as failure; intentional vs neutral/successful terminal?

## Fixed since prior review (no action)
R3 Live timer, R4 `format_output`, R5 `error_details`, R6 log errors non-fatal, R8 telemetry, R10 docs, R11 template guard, R13 backoff.

## Optional / debt
- R9 document poll pagination limitation on sync `watch_job` too
- R12 sync/async duplication OK as follow-up

tylersbray (approved):
+1'ing Matt's approval so you are not blocked.  Will be a nice feature to ship.

ironcommit:
Addressed the human review feedback in cc464055d:

- R1: restored verbose deployment status output for plain `--wait` by generating/passing `verbose=True`.
- R2: changed sync/async job log draining so `include_history=False` only marks the baseline seen after a full successful drain, preventing unread historical pages from leaking after mid-pagination failures.
- R7: left `paused` as failure-terminal intentionally and added a code comment explaining that a paused job is resumable but incomplete, so wait/watch should not report it as done.
- Also cleaned up the duplicated platform-job codegen import branch and documented pytest's `importlib` import mode choice.

Validation:
- `uv run --frozen pytest packages/nemo_platform_ext/tests/jobs/test_watch.py packages/nemo_platform_ext/tests/cli/commands/test_create_wait.py packages/nemo_platform_ext/tests/cli/core/test_code_generator.py tools/nemo-platform-sdk-tools/tests/sdk/cli_generator/test_generator.py -v`
- `uv run --frozen pytest sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/jobs/test_watch.py::test_watch_job_suppresses_unread_history_after_partial_history_drain_failure sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/jobs/test_watch.py::test_async_watch_job_suppresses_unread_history_after_partial_history_drain_failure sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/cli/commands/test_create_wait.py::test_inference_deployment_create_exits_when_wait_fails -v`
- `uv run ruff check packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/code_generator.py packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py packages/nemo_platform_ext/tests/jobs/test_watch.py packages/nemo_platform_ext/tests/cli/commands/test_create_wait.py tools/nemo-platform-sdk-tools/tests/sdk/cli_generator/test_generator.py`
- `uv run ruff format --check packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/code_generator.py packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py packages/nemo_platform_ext/tests/jobs/test_watch.py packages/nemo_platform_ext/tests/cli/commands/test_create_wait.py tools/nemo-platform-sdk-tools/tests/sdk/cli_generator/test_generator.py`
- `uv run pre-commit run -a`

ironcommit:
Small correction to my previous validation comment: after a cleanup pass, the final commit `2784e4bfe` removes the explicit generated `verbose=...` argument for inference deployment create wait/watch paths instead of passing `verbose=True`. That preserves the default progress output for plain `--wait` while avoiding the confusing `verbose=watch`/`verbose=True` plumbing in generated commands. Targeted command/generator and vendored mirror tests were rerun after this change.

pytest.ini
ironcommit: It “avoids duplicate test module basename collisions.”
Official pytest docs: https://docs.pytest.org/en/stable/explanation/pythonpath.html#pytest-import-mechanisms-and-sys-path-pythonpath

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/api/inference/deployments/__init__.py
ironcommit: removed

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
ironcommit: Not sure of the right answer here.  We can get feedback and see what people think.  It could go either way.

tools/nemo-platform-sdk-tools/src/nemo_platform_sdk_tools/sdk/vendor/vendor_package.py
ironcommit: Yeah, def not need.

sdk/python/nemo-platform/src/nemo_platform/cli/core/code_generator.py
```diff
@@ -240,34 +309,52 @@ def _render_inference_deployment_wait_code(
...
+                workspace={workspace},
+                timeout={timeout},
+                poll_interval={poll_interval},
+                include_logs=False,
+            ):
+                if not isinstance(event, JobStatusEvent):
+                    continue
+                if not event.terminal:
+                    continue
+                if event.successful:
```
anastasia-nesterenko: nit: want to print the event?

sdk/python/nemo-platform/src/nemo_platform/cli/core/job_watch_renderer.py
```diff
@@ -0,0 +1,139 @@
...
+        for event in events:
+            if isinstance(event, JobStatusEvent):
+                _render_status(output, event)
+                if event.terminal:
+                    terminal_event = event
+            elif isinstance(event, JobLogEvent):
+                _render_log(output, event)
+            elif isinstance(event, JobWarningEvent):
+                _render_warning(output, event)
+    except JobWatchTimeoutError as exc:
```
anastasia-nesterenko: nit: can add NotFoundError, now it prints `Unexpected error: HTTP 404: Job 'non-existing' not found in workspace 'default'.`

sdk/python/nemo-platform/src/nemo_platform/cli/core/code_generator.py
```diff
@@ -240,34 +309,52 @@ def _render_inference_deployment_wait_code(
...
+                    break
+                raise RuntimeError(f"Job {{resource_name!r}} ended with status {{event.status!r}}")
```
anastasia-nesterenko: nit: would it be useful to print what type of job failed?

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
```diff
@@ -0,0 +1,689 @@
...
+    _next_sleep_interval: float | None = None
+    _last_warning_key: tuple[Literal["status", "log"], str] | None = None
+
+    def reset(self) -> None:
+        self._next_sleep_interval = None
+        self._last_warning_key = None
+
+    def next_sleep_interval(self, poll_interval: float) -> float:
+        if poll_interval <= 0:
+            self._next_sleep_interval = 0
```
anastasia-nesterenko: nit: maybe use min interval similar to min=1 for CLI?

anastasia-nesterenko (approved):
left some nit comments, @ironcommit if they make sense I can add a commit for them

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/job_watch_renderer.py
```diff
@@ -0,0 +1,139 @@
...
+) -> bool:
+    """Render job watch events and return whether the terminal status succeeded."""
+    output = console or Console()
+    errors = error_console or Console(stderr=True)
+    started_at = time.time() if start_time is None else start_time
+    terminal_event: JobStatusEvent | None = None
+
+    try:
+        for event in events:
+            if isinstance(event, JobStatusEvent):
```
parkanzky: Small nit - seems like this if/else tree would lend itself to polymorphism with the render methods encapsulated in the event type classes.

docs/cli/reference.mdx
```diff
@@ -2019,15 +2019,16 @@ nemo inference deployments create [OPTIONS] [NAME]
 * `--input-file`: Path to JSON file (use '-' for stdin)
 * `--input-data`: Input data for the request (JSON or YAML)
 
-**Output Options:**
+**Lifecycle Options:**
 
-* `--output-format, -f <CHOICE>`: Output format for an entity. [possible values: json, yaml, raw, code]
+* `--wait`: Wait for the created deployment to be up and running
+* `--watch`: Watch the created deployment until it is stable, streaming status updates
```
parkanzky: I found this a little confusing to read, it's kind of a garden-path sentence. Maybe "until it is stable, and stream status updates" or "until it is stable while streaming status updates" ?

parkanzky (approved):
LGTM looking forward to this

docs/cli/reference.mdx
ironcommit: updated

packages/nemo_platform_ext/src/nemo_platform_ext/cli/core/job_watch_renderer.py
ironcommit: updated

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/watch.py
ironcommit: updated

sdk/python/nemo-platform/src/nemo_platform/cli/core/code_generator.py
ironcommit: updated

sdk/python/nemo-platform/src/nemo_platform/cli/core/job_watch_renderer.py
ironcommit: i[dated

sdk/python/nemo-platform/src/nemo_platform/cli/core/code_generator.py
ironcommit: that makes a bit verbose.

PR #1036, opened 2026-08-01
feat(nemo-agents): support `nemo agents pacakge` for fabric backed agents

mmogallapalli:
## Summary

This PR extends `nemo agents package` to build Docker images for Fabric-backed NeMo Agents as part of AIRCORE-970. It also includes the runtime-only inference-gateway credential binding needed by Fabric adapters from AIRCORE-943.

The package command now detects the agent configuration format and routes to a runtime-specific packaging path while retaining the existing NAT workflow:

```text
nemo agents package
  -> nat-workflow-v1
     -> NAT validation
     -> NAT Dockerfile
     -> NAT image

nemo agents package
  -> nemo-agents-spec-v1
     -> Platform schema validation
     -> Fabric translation and plan
     -> packaged-artifact validation
     -> Fabric Dockerfile
     -> Fabric serving image
```

The generated Fabric image contains the agent bundle, the release-matched NeMo Agents runtime and supported Fabric adapters, the pinned NeMo Relay CLI, and the Platform-owned Fabric server used by Docker and Kubernetes deployments.

## Changes

- Updated `nemo agents package` to detect `config_format` before rendering or building:
  - `nemo-agents-spec-v1` routes to Fabric packaging.
  - `nat-workflow-v1` and configs without `config_format` route to NAT packaging.
  - Unknown formats and malformed YAML fail with a CLI error.
- Split the previous generic image builder into explicit runtime-owned entry points:
  - `build_nat_agent_image`
  - `build_fabric_agent_image`
- Added a Fabric package validation pipeline that:
  - Loads and validates the Platform `agent.yaml` schema.
  - Translates the selected agent and harness into `FabricConfig`.
  - Runs `Fabric.plan` against the agent config directory.
  - Verifies that the config and referenced skills are present inside the Docker build context.
  - Rejects absolute, missing, unreadable, non-directory, or build-context-escaping skill paths.
  - Requires each packaged skill directory to contain `SKILL.md`.
- Split Fabric validation into reusable plan and doctor operations:
  - Packaging runs plan-only validation.
  - Existing runtime/deployment validation continues to run both plan and doctor.
- Added a Fabric-specific Dockerfile renderer that:
  - Copies the complete agent bundle into `/workspace` so paths remain relative to `agent.yaml`.
  - Installs the release-matched `nemo-platform[nemo-agents-plugin]` package.
  - Installs the packaged project in the same dependency resolution when `--pyproject` is provided.
  - Includes the plugin's supported Fabric adapter dependency set.
  - Installs the pinned NeMo Relay CLI for Codex and Claude Relay execution.
  - Creates and activates `/workspace/.venv` as the runtime environment.
  - Preserves the existing non-root and sandbox-runtime image options.
  - Exposes port `8000` and starts `nemo_agents_plugin.fabric.server`.
- Split shared and runtime-specific render parameters so NAT and Fabric reuse base-image, Python, uv, sandbox, project, and metadata behavior without sharing runtime-only fields.
- Renamed shared package environment variables from NAT-specific names to runtime-neutral names:
  - `NAT_BASE_IMAGE_URL` -> `NEMO_AGENTS_BASE_IMAGE_URL`
  - `NAT_BASE_IMAGE_TAG` -> `NEMO_AGENTS_BASE_IMAGE_TAG`
  - `NAT_PYTHON_VERSION` -> `NEMO_AGENTS_PYTHON_VERSION`
  - `NAT_UV_VERSION` -> `NEMO_AGENTS_UV_VERSION`
- Retained `NAT_VERSION` as NAT-only configuration and reject an explicit `--nat-version` for Fabric agents.
- Updated the default uv version to `0.9.14`, matching the repository's supported uv constraint.
- Added Fabric image metadata and identity support:
  - Uses the Platform agent name and description when no project metadata overrides them.
  - Sets `com.nemo.agent.framework=nemo_platform_agent`.
  - Includes the Platform contract version, Relay CLI version, base image, Python, and uv versions in image identity inputs.
  - Uses the installed `nemo-platform` distribution version as the runtime contract version.
- Added a runtime-only Platform inference-gateway credential binding:
  - Supplies a placeholder credential only for models routed through the Platform inference gateway.
  - Honors the selected harness model before the shared default model.
  - Leaves direct third-party model endpoints unchanged.
  - Does not mutate or persist the placeholder in the Platform-owned agent config.
  - Passes the binding to subprocess, Docker, and Kubernetes child runtimes.
- Preserved the existing NAT package flow and added focused coverage for format routing, rendering, validation, metadata, build behavior, generated-file cleanup, credential propagation, and deployment modes.

### The CLI owns format routing

`nemo agents package` detects the configuration format before calling a builder. Each builder therefore receives only the parameters supported by its runtime.

This keeps `build_nat_agent_image` free of Fabric branches and keeps `build_fabric_agent_image` free of NAT-only inputs such as `nat_version` and `NAT_CONFIG_FILE`.

### NAT and Fabric have separate builders and render contracts

The two image types share infrastructure such as Docker invocation, build-context selection, OCI metadata, sandbox setup, and generated-file cleanup. Their runtime dependencies, validation, environment variables, ports, and entrypoints remain separate.

This avoids a compatibility alias for the old internal `build_agent_image` helper. The CLI was its only production caller, so the call sites and tests were updated directly.

### Packaging runs Fabric plan, not doctor

`Fabric.plan` validates the translated configuration and selected adapter contract without requiring the producer machine to contain the binaries that will be installed in the image.

`Fabric.doctor` remains part of runtime/deployment validation, where the complete image environment, adapter binaries, and Relay CLI are available. Running doctor during packaging would incorrectly couple image production to the host environment.

### The complete build context is the packaged agent bundle

In config-only mode, the directory containing `agent.yaml` is the Docker build context. In project mode, the directory containing `pyproject.toml` is the build context, and `agent.yaml` must be located within it.

The image copies that complete context into `/workspace`. Referenced skill paths must stay inside the context and remain valid relative to the packaged `agent.yaml`.

### Runtime dependencies follow the Platform release contract

Fabric images install `nemo-platform[nemo-agents-plugin]` at the same version as the package command that rendered the Dockerfile. The plugin owns the supported Fabric and adapter dependency set, so users do not need to select or version individual adapters while packaging.

The Relay CLI is installed separately because Codex and Claude Relay adapters launch it as an external executable. Its version is pinned independently from the Python dependency graph.

### Project mode resolves the runtime and project together

When `--pyproject` is supplied, the release-matched NeMo Agents runtime and the user's project are installed in one resolution. Conflicting project constraints therefore fail during the image build instead of producing a partially compatible runtime.

Config-only mode installs only the release package and uses the directory containing `agent.yaml` as the bundle.

### Shared environment variables are runtime-neutral

Base image, Python, and uv settings apply to both NAT and Fabric images, so their environment variables now use the `NEMO_AGENTS_` prefix. `NAT_VERSION` remains unchanged because it is meaningful only for NAT packaging.

The CLI distinguishes an explicitly supplied `--nat-version` from an ambient `NAT_VERSION`. An explicit NAT-only flag is rejected for Fabric, while an ambient NAT setting does not interfere with Fabric packaging and remains available to the NAT resolver.

### Image identity includes the runtime contract

The default image tag remains `<agent-name>-<agent-id>:<agent-version>`, but Fabric image identity includes the Platform contract, Relay CLI, base image, Python, and uv inputs. Changing a runtime-defining input therefore produces a distinct agent ID rather than silently reusing an incompatible image identity.

### Gateway credentials remain runtime-only

Some Fabric model adapters require `api_key_env` even when requests are routed through the authenticated Platform inference gateway and no upstream API key is needed. Platform supplies a non-secret placeholder to the child runtime for that case.

The binding is derived after deployment gateway resolution, is not persisted in `agent.yaml`, and is not applied to direct third-party endpoints. This avoids creating a temporary public config convention while keeping the same behavior across subprocess, Docker, and Kubernetes runners.

## Error Behavior

- Malformed YAML or unsupported `config_format`: package command exits before rendering.
- Explicit `--nat-version` with a Fabric agent: package command exits with a NAT-only flag error.
- Invalid Platform schema, Fabric translation, or Fabric plan: package command exits before Docker build.
- Missing, invalid, unreadable, or context-escaping skill artifact: package command reports all discovered artifact errors.
- Agent config outside the selected project build context: rendering and validation fail before build.
- Existing user-owned generated files are preserved according to the existing package ownership rules.
- Docker build failures still clean up transient `Dockerfile.generated` and generated `.dockerignore` files.

## Out of Scope

- Documentation updates and migration guidance for renamed environment variables; tracked in AIRCORE-957.
- Platform fileset creation and remote artifact staging for Docker/Kubernetes deployments; tracked in AIRCORE-966.

## Validation

Affected NeMo Agents unit suites:

```text
294 passed
```

This includes NAT and Fabric package routing, Dockerfile rendering, metadata and image identity, Fabric package validation, gateway credential binding, Fabric server startup, translation and validation, and subprocess/Docker/Kubernetes runner configuration.

Repository Python style and formatting:

```text
All checks passed
2954 files already formatted
```

Manual Docker end-to-end validation:

1. Rendered a Fabric Dockerfile from a `nemo-agents-spec-v1` `agent.yaml`.
2. Built a local image from the monorepo workspace.
3. Verified the packaged Platform and Fabric modules imported inside the image.
4. Registered the Fabric agent with a local Platform.
5. Deployed the image in Docker mode and observed `pending -> starting -> running`.
6. Retrieved the running deployment through the CLI.
7. Invoked the Codex-backed agent through the Platform gateway and received the expected response.

### Local CLI flow

The branch version is not available from the package index, so local validation rendered the release Dockerfile and replaced its published-package install with a frozen workspace sync. Released builds use the generated Dockerfile directly and do not need the `perl` substitution.

```bash
cd /Users/mmogallapall/nemo-platform

export NMP_BASE_URL="http://127.0.0.1:8080"
export AGENT_CONFIG="$PWD/plugins/nemo-agents/examples/nemo-agent-config/agent.yaml"

export SUFFIX="$(date +%s)"
export AGENT_NAME="fabric-docker-${SUFFIX}"
export DEPLOYMENT_NAME="fabric-docker-deployment-${SUFFIX}"
export IMAGE="fabric-docker-${SUFFIX}:local"
export LOCAL_DOCKERFILE="/tmp/Dockerfile.fabric-local-${SUFFIX}"

uv run nemo agents package \
  --agent "$AGENT_CONFIG" \
  --pyproject "$PWD/pyproject.toml" \
  --uv-version 0.9.18 \
  --no-build \
  --no-ignore \
  --output "$LOCAL_DOCKERFILE"

perl -0pi -e \
  's#uv pip install "nemo-platform\[nemo-agents-plugin\]==[^"]+" \.#uv sync --frozen --package nemo-agents-plugin --no-dev --no-editable#' \
  "$LOCAL_DOCKERFILE"

uv run nemo agents package \
  --agent "$AGENT_CONFIG" \
  --pyproject "$PWD/pyproject.toml" \
  --dockerfile "$LOCAL_DOCKERFILE" \
  --tag "$IMAGE" \
  --skip-validation

docker image inspect "$IMAGE" >/dev/null && echo "Built $IMAGE"

docker run --rm \
  --entrypoint python \
  "$IMAGE" \
  -c 'import nemo_agents_plugin, nemo_fabric; print("Fabric runtime imports passed")'

uv run nemo agents create \
  --name "$AGENT_NAME" \
  --agent-config "$AGENT_CONFIG"

uv run nemo agents deploy \
  --agent "$AGENT_NAME" \
  --name "$DEPLOYMENT_NAME" \
  --mode docker \
  --image "$IMAGE" \
  --timeout 300

uv run nemo agents deployments get "$DEPLOYMENT_NAME"

uv run nemo agents invoke \
  --agent-deployment "$DEPLOYMENT_NAME" \
  --input "Reply with exactly: Fabric Docker deployment works" \
  --timeout 300
```

Optional cleanup:

```bash
uv run nemo agents deployments delete "$DEPLOYMENT_NAME" --yes
uv run nemo agents delete "$AGENT_NAME" --yes
```

## Summary by CodeRabbit

* **New Features**
  * Added support for packaging and building Fabric and NAT agent configurations.
  * Added automatic configuration-format detection and format-specific validation, rendering, and image creation.
  * Added Fabric package validation, skill-path and build-context checks, Dockerfile rendering, and image metadata.
  * Added secure runtime credentials for Platform inference gateway deployments.
  * Added separate configuration planning and diagnostic validation flows.
  * Added support for `NEMO_AGENTS_*` packaging environment variables.

* **Bug Fixes**
  * Improved cleanup of generated Dockerfiles and Docker ignore files after builds.
  * Prevented platform credentials from modifying source model configurations or affecting unrelated endpoints.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
@@ -393,7 +393,7 @@ def _register_package_command(app: typer.Typer) -> None:
             True, "--ignore/--no-ignore", help="Generate a .dockerignore file alongside the Dockerfile."
         ),
         skip_validation: bool = typer.Option(
-            False, "--skip-validation", help="Bypass validate_agent_config before build."
+            False, "--skip-validation", help="Bypass agent config validation before build."
         ),
         agent_version: Optional[str] = typer.Option(None, "--agent-version", help="Override agent version OCI label."),
         agent_author: Optional[str] = typer.Option(None, "--agent-author", help="Override agent author OCI label."),
--- a/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
@@ -634,27 +634,42 @@ def _package_render_only(
     # before we get here; assert for the developer who deletes that guard.
     assert format == "docker", f"unreachable: format={format!r}"
 
-    if config_format == NEMO_AGENTS_SPEC_CONFIG_FORMAT:
-        typer.echo("Error: Fabric agent packaging is not implemented yet.", err=True)
-        raise typer.Exit(code=1)
-
-    from nemo_agents_plugin.container.template import render_dockerignore, render_nat_dockerfile
+    from nemo_agents_plugin.container.template import (
+        render_dockerignore,
+        render_fabric_dockerfile,
+        render_nat_dockerfile,
+    )
 
     try:
-        content = render_nat_dockerfile(
-            agent_config,
-            pyproject,
-            base_image_url=base_image_url,
-            base_image_tag=base_image_tag,
-            python_version=python_version,
-            nat_version=nat_version,
-            uv_version=uv_version,
-            allow_root=allow_root,
-            sandbox_runtime=sandbox_runtime,
-            agent_version=agent_version,
-            agent_author=agent_author,
-            template_path=template,
-        )
+        if config_format == NEMO_AGENTS_SPEC_CONFIG_FORMAT:
+            content = render_fabric_dockerfile(
+                agent_config,
+                pyproject,
+                base_image_url=base_image_url,
+                base_image_tag=base_image_tag,
+                python_version=python_version,
+                uv_version=uv_version,
+                allow_root=allow_root,
+                sandbox_runtime=sandbox_runtime,
+                agent_version=agent_version,
+                agent_author=agent_author,
+                template_path=template,
+            )
+        else:
+            content = render_nat_dockerfile(
+                agent_config,
+                pyproject,
+                base_image_url=base_image_url,
+                base_image_tag=base_image_tag,
+                python_version=python_version,
+                nat_version=nat_version,
+                uv_version=uv_version,
+                allow_root=allow_root,
+                sandbox_runtime=sandbox_runtime,
+                agent_version=agent_version,
+                agent_author=agent_author,
+                template_path=template,
+            )
     except ValueError as exc:
         typer.echo(f"Error: {exc}", err=True)
         raise typer.Exit(code=1)
--- a/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
@@ -364,20 +364,25 @@ def _register_package_command(app: typer.Typer) -> None:
             "Defaults to 'Dockerfile' next to --pyproject when given (project root, "
             "so COPY statements resolve), otherwise next to the agent config.",
         ),
-        base_image_url: Optional[str] = typer.Option(None, "--base-image-url", envvar="NAT_BASE_IMAGE_URL"),
-        base_image_tag: Optional[str] = typer.Option(None, "--base-image-tag", envvar="NAT_BASE_IMAGE_TAG"),
-        python_version: Optional[str] = typer.Option(None, "--python-version", envvar="NAT_PYTHON_VERSION"),
+        base_image_url: Optional[str] = typer.Option(
+            None, "--base-image-url", envvar="NEMO_AGENTS_BASE_IMAGE_URL"
+        ),
+        base_image_tag: Optional[str] = typer.Option(
+            None, "--base-image-tag", envvar="NEMO_AGENTS_BASE_IMAGE_TAG"
+        ),
+        python_version: Optional[str] = typer.Option(
+            None, "--python-version", envvar="NEMO_AGENTS_PYTHON_VERSION"
+        ),
         nat_version: Optional[str] = typer.Option(
             None,
             "--nat-version",
-            envvar="NAT_VERSION",
             help=(
                 "NAT release to install (e.g. '1.7.0').  Strongly recommended: "
                 "pin explicitly so image tags/labels/deps are reproducible.  "
                 "When omitted, a baked-in default is used and a warning is printed."
             ),
         ),
-        uv_version: Optional[str] = typer.Option(None, "--uv-version", envvar="NAT_UV_VERSION"),
+        uv_version: Optional[str] = typer.Option(None, "--uv-version", envvar="NEMO_AGENTS_UV_VERSION"),
         allow_root: bool = typer.Option(
             False, "--allow-root", help="Disable non-root USER hardening in the rendered Dockerfile."
         ),
@@ -633,10 +638,10 @@ def _package_render_only(
         typer.echo("Error: Fabric agent packaging is not implemented yet.", err=True)
         raise typer.Exit(code=1)
 
-    from nemo_agents_plugin.container.template import render_dockerfile, render_dockerignore
+    from nemo_agents_plugin.container.template import render_dockerignore, render_nat_dockerfile
 
     try:
-        content = render_dockerfile(
+        content = render_nat_dockerfile(
             agent_config,
             pyproject,
             base_image_url=base_image_url,
--- a/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/cli.py
@@ -426,11 +426,29 @@ def _register_package_command(app: typer.Typer) -> None:
             template=template,
             platform=platform,
         )
-        _warn_if_nat_version_unpinned(nat_version)
+
+        from nemo_agents_plugin.container.builder import detect_agent_config_format
+
+        try:
+            config_format = detect_agent_config_format(agent)
+        except ValueError as exc:
+            typer.echo(f"Error: {exc}", err=True)
+            raise typer.Exit(code=1)
+
+        if config_format == NAT_WORKFLOW_CONFIG_FORMAT:
+            _warn_if_nat_version_unpinned(nat_version)
+        elif config_format == NEMO_AGENTS_SPEC_CONFIG_FORMAT and nat_version is not None:
+            typer.echo(
+                "Error: --nat-version is only valid for NAT workflow packaging; "
+                "Fabric agent packaging does not use NAT_VERSION.",
+                err=True,
+            )
+            raise typer.Exit(code=1)
 
         if no_build:
             _package_render_only(
                 agent_config=agent,
+                config_format=config_format,
                 pyproject=pyproject,
                 output=output,
                 format=format,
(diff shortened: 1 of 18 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/container/builder.py
```diff
@@ -1,10 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Docker image builder for NAT agents.
+"""Docker image builder for NeMo Platform agents.
 
 Builds a Docker image either from a pre-existing Dockerfile or by rendering
-one on-the-fly via :func:`~nemo_agents_plugin.container.template.render_dockerfile`.
+one on-the-fly via :func:`~nemo_agents_plugin.container.template.render_nat_dockerfile`.
```
mikeknep: Should this also mention `render_fabric_dockerfile`? They are "equal weight" relative to one another, right? (I'm guessing this diff occurred during just a function rename and so that's why only the one is mentioned currently.)

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mikeknep, line 1: Personal understanding check: we only expose `package` via the CLI, there's no SDK method for it?

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -448,28 +465,48 @@ def package(
...
+                    python_version=python_version,
+                    uv_version=uv_version,
+                    allow_root=allow_root,
+                    sandbox_runtime=sandbox_runtime,
+                    agent_version=agent_version,
+                    agent_author=agent_author,
+                    template_path=template,
+                    skip_validation=skip_validation,
+                    generate_ignore=generate_ignore,
+                    platforms=platform,
```
mikeknep, line 508: Gosh this is a long set of args 😅 and looks identical between the two functions except for the extra `nat_version`. I'm definitely open to pushback that it's too early, but: is there value in extracting a common `AgentPackagingMetadata` (naming?) dataclass that we create once and pass to either function?
```python
build_fabric_agent_image(agent, package_metadata)
build_nat_agent_image(agent, package_metadata, nat_version=nat_version)
```

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -590,23 +628,42 @@ def _package_render_only(
...
-            nat_version=nat_version,
-            uv_version=uv_version,
-            allow_root=allow_root,
-            sandbox_runtime=sandbox_runtime,
-            agent_version=agent_version,
-            agent_author=agent_author,
-            template_path=template,
-        )
+        if config_format == NEMO_AGENTS_SPEC_CONFIG_FORMAT:
+            content = render_fabric_dockerfile(
```
mikeknep, line 639: `<bernie>` I am once again asking if a protocol could be useful here `</bernie>`
```python
class AgentPackager(Protocol):
    def build_image(self, agent, render_params): ...
    def render_dockerfile(self, agent, render_params): ...

class FabricAgentPackager:
    ...

class NatAgentPackager:
    def __init__(self, nat_version):
        self._nat_version = nat_version
    ...
```
As with the other comment, not a blocker and maybe jumping the gun

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mikeknep: Update after reading more: most of this is already extracted to `SharedRenderParams` but that type isn't exposed beyond the `container.template` module.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mmogallapalli: Yeah thats correct, SDK handles only agent CRUD operations as well as deployment and invocation. Packaging is only via CLI

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mmogallapalli: I'm not opposed to either consolidation but soon (maybe next release 0.5) we will be tearing down NAT support from the Platform as NAT as a whole is being deprecated. So after that Fabric will be the only agent packaging option, so any abstraction may be short lived

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mmogallapalli: same as above

mikeknep (approved):

PR #1037, opened 2026-08-02
feat(iron-swarm): add the nemo-iron-swarm plugin (agent red-teaming and hardening)

koralchapnik:
## Summary

Adds the `nemo-iron-swarm` plugin: a security war-game that red-teams and hardens NAT agents. garak
attackers probe a sandboxed copy of the agent, defenders generate guardrails and OpenShell network
policy, and validators replay the attacks **plus a benign suite** to confirm the fix blocked the
attack without breaking ordinary behaviour. Before this, there was no way to measure whether a
hardening change actually helped; after it, `harden → apply → re-run` gives a comparable answer.

The plugin contributes a service, a job, a CLI, an SDK namespace, two entities and an agent skill.
It **never imports iron-swarm**: garak pulls `litellm → httpx>=0.28` and `torch`, which conflict with
`nvidia-nat`'s `httpx~=0.27`, so iron-swarm is provisioned into its own venv by
`nemo iron-swarm setup` and driven by subprocess. Nothing breaks when it isn't installed, and the
Studio UI ships **inside the plugin bundle** — installing the plugin is what puts it in the side
nav, so there is no feature flag to set.

## Changes

**The feature**

- **Two ways to pick a target.** `init --agent <name>` for any registered agent (it need not be
  deployed), or `init --project-dir <path>` for a local NAT project with nothing registered. Both
  save the same `IronSwarmManifest`, so `--manifest-id` is the single handle afterwards. CLI and
  Studio go through the same `POST /manifests`.
- **`run` is a pure consumer of the benign suite.** It never synthesises; `synth-benign` is a
  separate command and job, and the reviewed suite is cached on the manifest. A missing suite fails
  fast rather than doing something surprising.
- **A manifest is a frozen target.** `init` resolves once and stores the resulting scaffold as a
  fileset; runs download it instead of re-resolving, so two runs are comparable — which is what a
  "did the hardening help?" answer depends on. Agent edits land via an explicit
  `POST /manifests/{name}/refresh`, which `apply-mitigation` calls automatically.
- **Uploads never carry credentials.** Project bundles are selected with
  `git ls-files --exclude-standard` and drop `.env*` / `*.pem` / `*.key`, reusing `nemo-agents`'
  `DOCKERIGNORE_TEMPLATE`. Victim secrets travel by name: the manifest stores only the names, and
  values are resolved at run time from the operator's provisioned credential.
- **Studio** shows the target (source, port, egress, secrets, env, and the manifest itself), with
  refresh and an editor for non-secret env vars.

**Integration with `main`**

The branch was 251 commits behind, so this now includes a merge of `main` plus the fixes that
required:

- Studio wiring conflicts resolved where `main`'s anonymizer landed in the same slots (routes,
  route params, gate helpers, nav, route groups). `intakeEnabled` keeps `main`'s new `true` default,
  and the dead `traceId` route param is dropped — `main` reworked Intake to be session-based.
- Migrated 9 iron-swarm files from `react-router-dom` to `react-router`; `main` moved Studio to
  `react-router@8.3.0` and the old specifier is in no lockfile.
- The Studio UI image now copies the iron-swarm OpenAPI spec, without which
  `pnpm install --frozen-lockfile` aborted type generation and the image build failed.
- `nemo-iron-swarm-plugin` is installed in the `dev` dependency group. CI runs `uv run --frozen`,
  which syncs only default groups, so workspace membership alone left the package unimportable —
  the plugin's tests could not import it and the OpenAPI generator could not build its app.
- `plugins/nemo-iron-swarm/tests/unit` registered on the pytest `pythonpath`, so the sibling
  `_doubles` helper resolves when pytest runs from the repo root.
- Regenerated the iron-swarm OpenAPI spec against `main` (`db_version` for optimistic locking,
  `output_location`, and the `WarGameSpecOutput` → `WarGameSpec` rename).

**Review fixes**

- Studio's swarm event feed no longer freezes: an event with no numeric id used to fall back to
  `Date.now()`, which then advanced the poll cursor far above any real id so the server returned
  nothing for the rest of the run.
- `secrets_file` is constrained to the uploaded project directory. It is caller-supplied and read by
  `init` on the platform host, so it could previously name any readable file.
- User-controlled values are routed through `sanitize_for_log` before logging (7 CodeQL log-injection
  alerts, now cleared).
- `setup` redacts `index_url`; Artifactory embeds an access token in that URL.
- `delete_manifest` no longer deletes `project_fileset`. That fileset is supplied by the caller and
  two manifests can name the same bundle, so deleting either used to destroy the other's target.
  `agent_fileset` is service-created and is still cleaned up.
- The Iron Swarm nav entry and run list are gated on `usePluginInstalled('iron-swarm')`, matching the
  `nemo-agents` pattern, and fail open so an unresolved manifest never hides the feature.
- Pinned the OpenShell installer to `v0.0.92` instead of fetching `install.sh` from `main`.
- Corrected three stale statements about the manifest contract (README, `init` output, and a
  docstring naming the wrong exception).

## Type of Change

- [ ] Documentation only

## Quality Gates

281 unit tests in `plugins/nemo-iron-swarm/tests`. Two were added or corrected for the review fixes:
a regression test asserting `delete_manifest` leaves a caller-supplied `project_fileset` alone
(verified to fail against the previous behaviour), and the events-fallback test now names the log
after the run instead of `events.jsonl`, so it exercises the real `_events_path` contract rather than
passing by coincidence.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `uv run pre-commit run -a` | all 12 hooks pass, no files modified |
| `uv run --frozen pytest plugins/nemo-iron-swarm/tests` | 281 passed |
| `bash tools/lint/lint-python-style.sh` (`ruff check` + `ruff format --check`) | pass, 3097 files formatted |
| `bash tools/lint/lint-python-types.sh` (`ty`) | exit 0 |
| `pnpm --filter nemo-studio-ui typecheck` | pass |
| `pnpm --filter nemo-studio-ui lint` | pass at `--max-warnings 0` |
| `pnpm --filter nemo-studio-ui test` | 313 files, 2831 tests passed |
| `pnpm run format` (Prettier) | pass |
| `cd web && pnpm gen:check` | generated files up to date |
| `uv lock --check` | 580 packages, in sync |

Not verified locally, called out rather than claimed:

- The `nmp-studio-ui` Docker build could not be run here (`buildx` unavailable, and the legacy
  builder rejects `$BUILDPLATFORM`). The fix was validated by checking that every plugin spec orval
  reads has a matching `COPY`, and CI's **Build CPU smoke images** subsequently passed.
- Several CI runs failed inside a GitHub Actions incident (`Service Unavailable` / `Bad Gateway` /
  `Failed to resolve action download info` during *Set up job*). One `lint-web-sdk` failure occurred
  in that window and could not be reproduced locally under Node 26, Node 22.23.1, or a fresh
  `--frozen-lockfile` install with forced regeneration. It needs a clean run to confirm.

## Security notes for review

- The manifest is **read-only** in Studio by design. `overrides.defenders[].implementation` is a
  Python import path that iron-swarm loads *in the job process* (it logs `importing user-configured
  module … ensure the source is trusted`), so it executes on the platform host with the job's
  environment and Docker access — not inside the victim sandbox. A free-text YAML editor would turn
  manifest-write permission into code execution.
- `env` on a manifest is stored in plaintext and documented as non-secret; credentials use `secrets`,
  which stores only names and resolves values at run time.
- The war-game runs as a **subprocess executor** job and launches a Docker sandbox on the host.
  Correct for local/dev, explicitly not a production deployment — containerising the orchestrator
  and decoupling the sandbox from a host Docker daemon is the Phase-2 item.
- `apply-mitigation` writes another plugin's entity (`Agent.config`) while holding only
  `iron-swarm.runs.apply`, because the platform fail-closes on permission ids outside a service's own
  namespace. Treat that permission as an agent-write grant when assigning it.

## Known issues, logged rather than hidden

- A run that fails *after* the sandbox is up leaks the sandbox and its port forward; the next attempt
  then fails on a name collision, so one fault presents as a different fault each time. Cancellation
  tears down correctly — the failure paths do not.
- A synth producing zero rows reports `completed` while persisting nothing, and a cached
  `benign_interview` is not cleared when a new suite lands without one.
- `sanity_check` submits a job and returns immediately, so a failed validation cannot yet be gated on
  in CI. A `--wait` flag that polls to a terminal status is the follow-up.
- The plugin's tests carry no `unit` marker, so `make test-unit`'s `-m unit` filter deselects them.
  This is repo-wide (`nemo-anonymizer` and `nemo-agents` are the same) and is left alone here.

## Manual verification

Real war-games against the bundled `react-agent` and the `research` agent from `agents-lab` uploaded
as a project. The second is the interesting result — **attacks 11/11 blocked, benign 8/12 passed**
(3 refused, 1 error): the defenses stopped everything but broke a third of normal functionality,
which is exactly what the benign suite exists to catch.

Studio flows were driven end-to-end with Playwright: manifest creation, the target view, refresh,
the env editor, and a full benign-suite generation including the HITL interview.

## Not in this PR

Routing attack and detection through the NeMo Auditor (NAIS-356/357) is prototyped on a separate
branch and deliberately excluded here.

## Summary by CodeRabbit

- **New Features**
  - Introduced Iron Swarm for war games, benign-suite generation, mitigation review, defense composition, and hardened-agent validation.
  - Added Studio pages for manifests, runs, live swarm activity, suite editing, model configuration, and hardening workflows.
  - Added CLI, SDK, REST API, event replay, artifact handling, and agent/project setup capabilities.
  - Added model connectivity checks, secure credential handling, preflight diagnostics, and feature-flagged navigation.

- **Documentation**
  - Added comprehensive setup, workflow, configuration, troubleshooting, and CLI guidance.

- **Tests**
  - Added broad coverage for APIs, CLI workflows, execution, security, filesets, models, events, and Studio behavior.


plugins/nemo-iron-swarm/src/nemo_iron_swarm_plugin/cli/main.py
```diff
@@ -0,0 +1,487 @@
...
+
+            result = ctx.sdk.iron_swarm.sanity_check(
+                manifest_id=manifest_id,
+                mitigations=mitigations,
+                selected_defense_ids=selected,
+                replay_hitlog_fileset=replay_hitlog,
+                env_file=env_file,
+                workspace=ctx.workspace,
+            )
+            typer.echo(json.dumps(result, indent=2, default=str))
```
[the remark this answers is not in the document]
koralchapnik: Not applying this one — the suggested patch would break the command.

`sanity_check` is fire-and-forget: it returns `NemoJobScheduler().submit_remote(...)`, which is the job-*submission* response (`scheduler.py` documents it as "The decoded JSON response from the plugin service"). That payload has no `status` key at all — it carries the job name. So `result.get("status") == "completed"` is `None == "completed"` → always `False`, and **every successful submission would exit 1**.

It would also fail the existing test, which stubs the return as `{"name": "job-x"}` and asserts success:

```python
# plugins/nemo-iron-swarm/tests/unit/test_sanity_check_cli.py:119
assert result.exit_code == 0, result.output
```

The underlying concern is fair though: there's currently no way for CI to gate on a sanity-check verdict, because the verdict only exists once the job finishes (`nemo iron-swarm status`, where `RunStatus = Literal["running", "completed", "failed"]`). The right fix is a `--wait` flag that polls to a terminal status and *then* exits nonzero on `failed` — a feature, not a one-line change. Tracking that separately rather than in this PR.

plugins/nemo-iron-swarm/README.md
```diff
@@ -0,0 +1,447 @@
...
+</details>
+
+**Docker + OpenShell**, which run the victim sandbox:
+
+```bash
+brew install colima docker && colima start          # or Docker Desktop
+
+# Use the native installer — `uv tool install openshell` gives you the CLI
+# without the gateway service, and the war-game needs the gateway.
+curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
```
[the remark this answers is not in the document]
koralchapnik: Fixed in e3a32b28f — pinned to the `v0.0.92` release tag (verified it resolves) instead of `main`, matching the `openshell>=0.0.92` the deployments plugin already requires and the way the root README and CONTRIBUTING pin the uv installer. Resolving; the thread went outdated rather than auto-resolving because the line moved.

parkanzky (approved):
Plugin looks idiomatic enough to me. When tests pass I think we should just merge this to get it in and allow people to interact with it. However, I am not familiar enough with typescript to review the studio pieces. Finding someone to review that now.

marcusds (commented):
Rather than including changes directly into studio I think this would benefit from using the new support studio has for plugins.

This PR has details: https://github.com/NVIDIA-NeMo/nemo-platform/pull/594

Reach out if would like support on this. There some agent instructions that should be helpful.

Edit: since this would be the first real consumer of the studio plugin architecture I took a look at what moving it over would look like. For the most part it'll be straightforward but there are some changes I am going to make on our end to make it better.  Edit: This is in https://github.com/NVIDIA-NeMo/nemo-platform/pull/1180 and https://github.com/NVIDIA-NeMo/nemo-platform/pull/1192

marcusds (changes requested):

koralchapnik:
Working on this — moving the whole Studio UI into the plugin, as suggested. Thanks for #1180 and #1192! they're what made it possible.

The port is done locally and verified against a running platform: the UI now ships in the plugin wheel as a Studio bundle (`Root` + `navItems` via the `nemo.studio` entry point), its API client is generated inside the plugin (fetch-based, so no second axios), and `web/packages/studio` is back to matching `main` — no feature flag, no routes, no side-nav entry.

It sits on top of #1192, so I'll push once that merges and I can rebase .. pushing before then would pull its file moves into this PR's diff. Two plugin-API gaps came up while testing in the browser - commented on #1192.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
```diff
@@ -0,0 +1,798 @@
...
+  SelectTrigger,
+  Stack,
+  Text,
+  TextInput,
+} from '@nvidia/foundations-react-core';
+import { useQueryClient } from '@tanstack/react-query';
+import { FC, useCallback, useEffect, useRef, useState } from 'react';
+import { useParams } from 'react-router';
+
+type AttackIntensity = 'light' | 'standard' | 'thorough';
```
marcusds: Since this file is so large, I would split the types into a types.ts and the consts in a consts.ts

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
```diff
@@ -0,0 +1,798 @@
...
+      queryClient.invalidateQueries({
+        queryKey: getIronSwarmGetManifestQueryKey(workspace, ironSwarmManifestName),
+      });
+      return true;
+    } catch {
+      return false;
+    }
+  };
+
+  const downloadCsv = () => {
```
marcusds, line 173: Similarly to the consts and types, I would move some of these functions out to a utils.ts file to make this component more maintainable. Would require a little work on to move some items to arguments, but an agent should be able to do it really easily for you.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
```diff
@@ -0,0 +1,798 @@
...
+                    : 'Generate benign suite'}
+              </Button>
+              <Button color="brand" disabled={runWarGame.isPending} onClick={openRunDialog}>
+                Run war-game
+              </Button>
+            </Flex>
+          }
+        />
+
+        {gen.active ? (
```
marcusds, line 254: I would personally look into how you could split this component up into sub components (in a dir like ./components), its better for performance and maintability.

plugins/nemo-iron-swarm/web/src/routes/NewIronSwarmManifestRoute/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+} from '@nvidia/foundations-react-core';
+import { useQueryClient } from '@tanstack/react-query';
+import { FC, useEffect, useMemo, useState } from 'react';
+import { useForm } from 'react-hook-form';
+import { Link, useNavigate } from 'react-router';
+import { z } from 'zod';
+
+const NAME_PATTERN = /^[a-z0-9][a-z0-9-]*$/;
+
+const schema = z.object({
```
marcusds: We usually by practice put schema's in their own files.

marcusds (approved):

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
```diff
@@ -0,0 +1,798 @@
...
+        submitButtonText="Start"
+        loading={runWarGame.isPending}
+        submitDisabled={
+          (mode === 'replay' &&
+            (replaySource === 'upload'
+              ? !hitlogFileset || uploadHitlog.isPending
+              : !lastHitlogFileset)) ||
+          (benignSource === 'upload' && (!benignFileset || uploadBenign.isPending))
+        }
+        onSubmit={() => start()}
```
marcusds: Should have a e.preventDefault() here

```
onSubmit={(e) => {
    e.preventDefault();
    start();
}
```

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
marcusds: Though in this specific example we actually have a shared triggerDownload you should be able to use.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
```diff
@@ -0,0 +1,798 @@
...
+        label: row.label,
+        persona: row.persona,
+        rationale: row.rationale,
+      }))
+    );
+  }, [manifest]);
+
+  // Re-seed the launch dialog's config from the manifest default each time it opens, so per-run tweaks
+  // are ephemeral (they never persist unless the user hits "Save as default").
+  useEffect(() => {
```
marcusds: Any manifest refetch while the dialog is could wipe user's entries. I'd move this logic into openRunDialog

plugins/nemo-iron-swarm/web/src/components/ProjectManifestWizard.tsx
```diff
@@ -0,0 +1,259 @@
...
+              workspace={workspace}
+              defaults={modelDefaults}
+            />
+          </AccordionSection>
+        </AccordionRoot>
+        <Flex gap="density-md">
+          <Button color="brand" type="submit" disabled={isCreating}>
+            {isCreating ? 'Creating…' : 'Create Manifest'}
+          </Button>
+          <Button kind="tertiary" onClick={onReset}>
```
marcusds: add a type="button" so we don't submit the form.

plugins/nemo-iron-swarm/web/src/components/InterviewPanel.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+// Sentinel radio value for the "Other" escape hatch, so a question with options still accepts free text.
+const OTHER = '__other__';
+
+// The synth interview, rendered inline (manifest generate flow, or a run's fallback tab): one card per
+// question, options as radio cards with an "Other" free-text escape hatch. Returns the operator's answers
+// so the job can resume benign-suite synthesis.
+export const InterviewPanel: FC<InterviewPanelProps> = ({ prompt, loading, onSubmit }) => {
+  const [answers, setAnswers] = useState<Record<string, string>>(() =>
+    Object.fromEntries(prompt.questions.map((q) => [q.gap, defaultAnswer(q.options)]))
```
marcusds, line 35: Could this cause state to get reused in additional rounds?

plugins/nemo-iron-swarm/web/src/components/useSanityCheck.ts
```diff
@@ -0,0 +1,176 @@
...
+  const query = useQuery({
+    queryKey: ['iron-swarm-validation', workspace, jobName],
+    enabled: hasReport && Boolean(jobName),
+    queryFn: async (): Promise<ValidationReport> => {
+      const blob = await ironSwarmDownloadJobResult(workspace, jobName ?? '', VALIDATION_RESULT);
+      return JSON.parse(await blob.text()) as ValidationReport;
+    },
+  });
+
+  return { report: query.data, isLoading: Boolean(jobName) && !hasReport, hasReport };
```
marcusds: If a job fails, will it have a report? If so, will this always be stuck loading if failed?

plugins/nemo-iron-swarm/web/src/components/useMitigations.ts
```diff
@@ -0,0 +1,230 @@
...
+};
+
+// Fetches the run's mitigations artifact via the job results API (not the event bus): poll the result list
+// until it appears, then download + parse it once. Recommendations are derived client-side from the diffs.
+export const useMitigations = (workspace: string, jobName: string): UseMitigationsResult => {
+  const { data: results } = useIronSwarmListJobResults(workspace, jobName, {
+    query: {
+      enabled: Boolean(jobName),
+      refetchInterval: (query) =>
+        query.state.data?.data?.some((result) => result.name === MITIGATIONS_RESULT)
```
marcusds: Perhaps add a job status related escape hatch?

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
koralchapnik: Done in 0af3aa6e5 — split into `types.ts` (`AttackIntensity`, `ReplaySource`, `BenignSource`, `DefenderSelection`) and `constants.ts` (`INTENSITY_LABEL`, `REPLAY_SOURCE_LABEL`, `BENIGN_SOURCE_LABEL`, `CSV_FIELDS`).

The file is still 776 lines though, so your other two comments about `utils.ts` and sub-components are the ones that actually shrink it — replying to those separately.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
koralchapnik: Both done in 0af3aa6e5. The CSV serialisation moved to `utils.ts` (`toRequestsCsv` + `escapeCsv`), and `downloadCsv` now uses the shared `triggerDownload` from `@nemo/common` — thanks for the pointer, I'd missed it existed.

```ts
const downloadCsv = () => {
  triggerDownload(
    new Blob([toRequestsCsv(suite)], { type: 'text/csv' }),
    `${ironSwarmManifestName}-requests.csv`
  );
};
```

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
koralchapnik: Done in c039e9dfe — the run dialog moved to `components/RunWarGameDialog.tsx`; the route goes 776 → 421 lines.

It owns its form state instead of taking ~20 props, which also hardens the seeding rule from your other comment: it copies from the manifest once per opening, so a refetch can't wipe what was typed.

Verified in a running Studio, not just by inspection.

421 is still more than I'd like — the benign-suite section (~100 lines) and env dialog (~55) are the next seams. Here or follow-up, your call.

plugins/nemo-iron-swarm/web/src/routes/NewIronSwarmManifestRoute/index.tsx
koralchapnik: Done in 820cd65ec — moved to `NewIronSwarmManifestRoute/schema.ts`, following the `routes/*/schema.ts` convention Studio uses.

`ProjectManifestWizard.tsx` has a small inline `reviewSchema` too — say if you want that one moved as well.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
koralchapnik: Present on both modals — the env one at `index.tsx:380`, and the run dialog's, which moved to `RunWarGameDialog.tsx:246` in c039e9dfe.

plugins/nemo-iron-swarm/web/src/routes/IronSwarmManifestDetailRoute/index.tsx
koralchapnik: Fixed as you suggested, and hardened further in c039e9dfe: the dialog now owns its state and seeds from the manifest once per opening, guarded by a ref, so a refetch mid-edit can't reach it.

Verified in a running Studio — set rounds to 3, unchecked a defender, hit "Save as default" (which invalidates the query), and the form kept both.

plugins/nemo-iron-swarm/web/src/components/ProjectManifestWizard.tsx
koralchapnik: Added in fe1ba8cbd — the reset button is `type="button"` and the submit is `type="submit"`. "Detect Project" sits outside the `<form>`, so it can't submit either.

plugins/nemo-iron-swarm/web/src/components/InterviewPanel.tsx
koralchapnik: Good catch — it would have. Both call sites now remount per round with `key={interview.round}` (`IronSwarmRunDetailsRoute:254`, `IronSwarmManifestDetailRoute:276`), since the panel seeds its answers on mount.

plugins/nemo-iron-swarm/web/src/components/useSanityCheck.ts
koralchapnik: A failed job doesn't write `validation.json`, so this was a real risk. `useJobArtifact` takes the job status: it stops polling once the job has been terminal for a 20s grace period (results land slightly after the status flips) and then reports `missing`, which the caller renders as an error rather than a spinner.

plugins/nemo-iron-swarm/web/src/components/useMitigations.ts
koralchapnik: Added — this goes through the same `useJobArtifact` as the sanity-check report, so it takes the job status, stops polling after the terminal grace period, and surfaces `missing` instead of spinning.

marcusds (approved):

gabwow (approved):

crookedstorm (approved):
LGTM

PR #1039, opened 2026-08-03
fix(intake): bound trace hydration by page timestamps

BrianNewsom:
## Summary

- carry each selected trace root timestamp through the page reference
- bound second-phase trace_index hydration by the page timestamp range before FINAL
- preserve the flat trace-id bloom predicate and source-format identity predicate

## Production profile

For the reported 50-row page:

- deployed hydration: 7.6s, 13.09 GiB read, 28.65 GiB memory, MEMORY_LIMIT_EXCEEDED
- same hydration with the timestamp bound: 112ms, 62.89 MiB read, 59.77 MiB memory, all 50 rows returned

Count and page selection were already fast at 53ms and 90ms respectively.

## Validation

- 12 focused trace repository tests passed
- 2 ClickHouse trace-read integration tests passed
- Ruff check and format check passed
- ty check passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trace retrieval accuracy by incorporating precise start-time ranges when loading trace details.
  * Ensured trace timestamps are handled consistently across timezone-aware and timezone-naive values.
  * Updated preview results to preserve microsecond-level timestamp precision.


services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -434,12 +441,22 @@ def _unique_string_attribute_aggregate(source_alias: str, *, parameter: str, ali
...
+    started_at_us = [ref.started_at_us for ref in refs]
     return {
         "page_trace_ids": [ref.trace_id for ref in refs],
         "page_trace_keys": [ref.trace_key for ref in refs],
+        "page_started_at_min_us": min(started_at_us),
+        "page_started_at_max_us": max(started_at_us),
     }
 
 
+def _datetime_to_unix_microseconds(value: datetime) -> int:
```
asutermo: nit: is this something that exists or can/should be shared broadly in a utils type location? not a blocker or anything.

asutermo (approved):

services/intake/src/nmp/intake/repository/clickhouse/trace.py
BrianNewsom: Good push - there's a CH function that can handle this we use elsewhere. Going to standardize on that. Thanks!

PR #1044, opened 2026-08-03
feat(nemo-agents): update skills to use Fabric backed agents as preferred and NAT as legacy

mmogallapalli:
## Summary

This PR updates the NeMo Platform agent-building skills for AIRCORE-956 so new
agents use the Platform-owned `nemo-agents-spec-v1` configuration path by
default, while preserving existing NAT workflow YAML as an explicit
compatibility path.

It introduces a focused `nemo-agent-config` skill and aligns the surrounding
design, specification, model-selection, build, and invocation skills around the
same authoring flow:

```text
Agent idea
  -> nemo-explore
  -> AGENT-SPEC.md through nemo-spec
  -> Platform-owned agent.yaml through nemo-agent-config
  -> Agent registration and deployment through nemo-build-agent
  -> Invocation through nemo-try-agent
```

Users author Platform concepts such as harness selection, models,
instructions, skills, tools, environment, and telemetry. The skills do not ask
users to author raw Fabric SDK objects or treat Fabric configuration as the
Platform persistence contract.

## Changes

- Added the `nemo-agent-config` skill for creating, editing, validating, and
  migrating Platform-owned `agent.yaml` files.
- Added a canonical `nemo-agents-spec-v1` `agent.yaml` template covering:
  - Supported harness selection.
  - Default and per-harness model configuration.
  - System instructions.
  - Skills, MCP servers, and blocked tools.
  - Workspace and artifact paths.
  - Relay telemetry configuration.
- Added routing coverage for explicit and implicit `nemo-agent-config`
  requests, including legacy NAT migration requests.
- Updated `nemo-build-agent` to:
  - Use `agent.yaml` as the default implementation path.
  - Delegate config authoring to `nemo-agent-config`.
  - Preserve legacy NAT workflows instead of rewriting them automatically.
  - Check for existing Agent entities and deployments before replacement.
  - Avoid silently deleting an existing Agent during the build flow.
- Updated `nemo-skill-selection` to route focused Platform agent-config work to
  `nemo-agent-config` and keep full builds routed to `nemo-build-agent`.
- Updated `nemo-explore` to gather harness, model, tool, skill, environment,
  telemetry, and compatibility requirements needed by the new config path.
- Updated `nemo-spec` and its template so `AGENT-SPEC.md` captures the inputs
  needed to create a Platform-owned `agent.yaml` while remaining useful for NAT
  compatibility workflows.
- Updated `nemo-model-selection` to describe shared Platform model defaults,
  harness-specific overrides, and Inference Gateway behavior without exposing
  Fabric SDK configuration as the user contract.
- Updated `nemo-try-agent` to support both Platform-owned `agent.yaml` and
  legacy NAT YAML for local one-shot invocation, while distinguishing local
  provider requirements from deployed Inference Gateway normalization.
- Updated skill metadata and routing tests across the affected skills.
- Updated vendoring configuration to include YAML assets and regenerated the
  SDK copies of all changed skills and templates.

### `agent.yaml` is the preferred machine-readable contract

New agent workflows now produce a Platform-owned `agent.yaml` with
`config_format: nemo-agents-spec-v1`. The resulting config is stored in the
Platform Agent entity and translated behind the Platform execution boundary.

The skills deliberately use Platform-facing terms instead of asking users to
construct `FabricConfig` or other SDK-specific objects.

### `AGENT-SPEC.md` and `agent.yaml` have separate roles

`AGENT-SPEC.md` remains the human-readable design and acceptance contract.
`agent.yaml` is the machine-readable execution configuration derived from that
spec. Both live under the canonical agent spec directory:

```text
agents/<agent-name>-spec/
  AGENT-SPEC.md
  agent.yaml
```

This preserves the existing spec-first workflow while adding the new Platform
agent format.

### NAT remains a compatibility path

Existing NAT workflow YAML continues to be accepted. The updated skills do not
automatically rewrite or overwrite it. Users can deploy it unchanged or choose
a best-effort migration through `nemo-agent-config`.

If a NAT workflow, custom Python component, or tool has no equivalent supported
harness contract, the skills keep the NAT path or identify the need for a
custom adapter instead of claiming a mechanical conversion.

### Model defaults may be overridden by a harness

The template supports one shared `models.default` and an optional model block
on each harness. A harness-local model takes precedence when provided; otherwise
the shared default is used.

The default deployed path may rely on Platform Inference Gateway normalization.
Local one-shot invocation requires the selected model to already have a usable
provider endpoint and credentials because it does not apply that deployment
normalization.

### Existing resources are not replaced silently

The build workflow checks for existing Agent entities and deployments and asks
the user whether to reuse, update, or replace them. The default command sequence
assumes a new Agent and no longer performs an unconditional best-effort delete.

### YAML templates are vendored assets

The package vendoring configuration now includes `*.yaml` so the canonical
`agent.yaml` template is available from the SDK's bundled skill copy alongside
the Markdown skill content.

## Compatibility

- New agents default to `nemo-agents-spec-v1`.
- Existing configs without `config_format` continue to use the existing
  `nat-workflow-v1` default.
- Existing NAT workflow YAML can still be registered, deployed, and invoked.
- NAT-to-Platform migration is explicit and best effort.
- `nemo-try-agent` continues to support both local and deployed invocation
  targets.

## Validation

Focused skill registry and CLI tests:

```text
58 passed
```

Additional validation completed:

- Validated the example `agent.yaml` with the Platform `AgentConfig` model.
- Translated the example into a typed Fabric config and successfully ran
  Fabric planning.
- Confirmed `nemo skills list` includes `nemo-agent-config`.
- Confirmed canonical and vendored skill/template files match after
  `make vendor`.
- Ran repository Python style checks successfully.
- Ran `git diff --check` successfully.

## Summary by CodeRabbit

* **New Features**
  * Added guidance for authoring and validating NeMo Platform agent configurations.
  * Added an example configuration covering supported harnesses, models, workspace paths, and telemetry.
  * Added support for building, deploying, and invoking agents from local configurations or named deployments.

* **Improvements**
  * Updated model selection, exploration, specification, and skill routing for Platform agents.
  * Retained compatibility guidance for legacy NAT workflows.
  * Expanded validation, migration, troubleshooting, evaluation, and deployment workflows.
  * Improved packaged reference assets and skill coverage tests.


packages/nemo_platform_ext/pyproject.toml
```diff
@@ -91,7 +91,7 @@ package_root = "packages/nemo_platform_ext"
 # extension globs here as new companion file types are needed (`.json`,
 # `.yaml`, `.txt`, ...); we avoid `**/*` because dev environments leave
 # `__pycache__/*.pyc` in source modules, which the vendor tool doesn't filter.
-included_paths = ["**/*.py", "**/*.md"]
+included_paths = ["**/*.py", "**/*.md", "**/*.yaml"]
```
tylersbray: > **agent** (`pr-review`): P1 — vendoring still drops NAT/eval templates

`included_paths` now includes `**/*.yaml` (good for `agent.yaml`), but `nemo-build-agent` still references `references/templates/agent.yml` (NAT path) and `eval-job.json` (Step 4). After this PR's `make vendor`, the SDK skill tree still only has `data-designer-config.py` under that templates dir — those two assets remain missing for SDK-installed `nemo`.

Please add `**/*.yml` and `**/*.json` (the comment above already anticipates `.json`), re-run `make vendor`, and ideally add an asset-parity check so this does not regress.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/references/templates/agent.yaml
```diff
@@ -0,0 +1,63 @@
...
+      api_key_env: ANTHROPIC_API_KEY
+    settings:
+      permission_mode: dontAsk
+models:
+  default:
+    provider: replace-with-verified-provider
+    model: replace-with-verified-model-name
+
+skills:
+  paths: []
```
tylersbray, line 47: > **agent** (`pr-review`): P1 — relative `skills.paths` validate at create, fail at deploy

`nemo agents create` Fabric-validates `skills.paths` against the authoring directory (`base_dir=agent_config.parent`), then persists only the parsed config dict. Deploy materializes YAML into a fresh base dir / container mount (`_write_fabric_config` / `config_files`) and never copies skill trees.

So a non-empty relative path like `skills/review` can pass `create` and fail on `deploy`/runtime when the adapter looks for `SKILL.md` next to the rewritten `agent.yaml`.

Until packaging/upload exists, please have this skill (and build) refuse non-empty relative `skills.paths` for deployable agents, require an explicit packaging step, or document that only empty paths (or same-host absolute paths) work today.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/references/templates/agent.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+mcp:
+  servers: {}
```
tylersbray, line 50: > **agent** (`pr-review`): P2 — MCP servers need `transport` + `url`

Empty `servers: {}` is valid, but the skill says to use only template fields when adding MCP. `McpServerConfig` requires `transport` and `url` (`agent_config.py`). Please show a commented example server entry so agents do not invent invalid shapes.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+
+The canonical remote config location is derivable from workspace and agent name:
+`<workspace>/<agent-name>-spec#agent.yaml`. Do not invent a separate ref field.
+
+## What you do
+
+1. Confirm the agent name and config path. Default to
+   `agents/<agent-name>-spec/agent.yaml`.
+2. Start from `references/templates/agent.yaml` unless the user is editing an
+   existing file. Replace every model placeholder before validation.
```
tylersbray: > **agent** (`pr-review`): P2 — step order conflicts with model-selection

Step 2 says replace every model placeholder before validation; Step 4 says configure `models.default` only after `nemo-model-selection` returns a verified provider/model. Agents following the numbered list will fill placeholders first and defeat Step 4.

Please reorder so harness selection → `nemo-model-selection` → write verified model fields → prune unused harness blocks → validate. Also note create validates only `default_harness`, so leftover placeholder harness-local models can survive `create` and fail later on harness switch.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+| Tracing or telemetry settings | `telemetry` |
+
+For a NAT `codex_agent` workflow, apply this mapping instead of copying the
+workflow fields into `harnesses.codex.settings`:
+
+| NAT Codex field | Platform `agent.yaml` target |
+|---|---|
+| `working_directory` | `environment.workspace` |
+| `sandbox_mode` | `harnesses.codex.settings.sandbox` |
+| `approval_policy: never` | `harnesses.codex.settings.approval_mode: deny_all` |
```
tylersbray: > **agent** (`pr-review`): P2 — `approval_mode: deny_all` is unverified

This mapping is not present in repo examples (`sandbox` / `reasoning_effort` only) or Fabric translator tests. The skill also says adapter `settings_schema` is authoritative — please omit this key (or cite the schema property) rather than inventing `deny_all`.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+```bash
+.venv/bin/nemo agents deploy \
+  --agent "$AGENT_NAME" \
+  --name "$AGENT_NAME-deployment"
+```
+
+After deployment begins, wait for it and invoke it without another confirmation:
+
+```bash
+.venv/bin/nemo agents deployments wait \
```
tylersbray: > **agent** (`pr-review`): P2 — wait on `--agent` can diverge from invoke target

`deployments wait --agent` resolves the **latest active** deployment for that agent (`cli.py`), while the next `invoke` hardcodes `"$AGENT_NAME-deployment"`. Prefer `deployments wait "$AGENT_NAME-deployment"` (as `nemo-build-agent` does). Also, `deploy` already defaults to `--wait`, so this second wait is usually redundant.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
```diff
@@ -26,206 +30,337 @@ allowed-tools: [Bash, Read, Write, Edit]
...
-  sleep 10
-done
-.venv/bin/nemo evaluation benchmark-jobs results aggregate-scores download $AGENT_NAME-eval
+.venv/bin/nemo evaluation benchmarks list
+.venv/bin/nemo evaluation benchmark-jobs create "$AGENT_NAME-eval" \
+  --input-file "agents/$AGENT_NAME.eval-job.json"
 ```
 
-Verification: confirm the job reached `completed`, not `failed`. Display the score table. Stop. Ask if scores meet the bar from the spec.
+Poll until the job reaches `completed` or `failed`, then download aggregate
```
tylersbray, line 346: > **agent** (`pr-review`): P2 — restore concrete eval poll/download CLI

Main had a `get-status` poll loop and `aggregate-scores download`. This prose leaves agents without prescribed commands after `benchmark-jobs create`. Please restore something like:

```bash
for i in $(seq 1 24); do
  status=$(.venv/bin/nemo evaluation benchmark-jobs get-status "$AGENT_NAME-eval" 2>/dev/null)
  echo "$status"
  echo "$status" | grep -qE "completed|failed" && break
  sleep 10
done
.venv/bin/nemo evaluation benchmark-jobs results aggregate-scores download "$AGENT_NAME-eval"
```

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-build-agent/SKILL.md
```diff
@@ -26,206 +30,337 @@ allowed-tools: [Bash, Read, Write, Edit]
...
+scores. Show the score table and compare it with the success bar in
+`AGENT-SPEC.md`.
 
 ## Step 5: Guardrails (optional)
 
-If the spec lists constraints, add a content-safety intercept to the YAML (see `references/templates/agent-with-guardrails.yml` if present, or write inline) and redeploy:
+If the spec defines safety or policy constraints, invoke `nemo-guardrails` for
```
tylersbray: > **agent** (`pr-review`): P2 — no documented `nemo-agents-spec-v1` guardrails path

This step tells agents to invoke `nemo-guardrails` for the Platform path and apply integration without unsupported `agent.yaml` fields. `nemo-guardrails` only documents VirtualModel / IGW middleware; `AgentConfig` has no guardrail wiring (only `tools.blocked`).

Main previously used NAT `intercepts`. Please be honest here: keep NAT `intercepts` for the NAT path, and for Platform configs either document a real composition (e.g. IGW VirtualModel in front of the harness model) or say Platform guardrails are not skill-supported yet — do not invent YAML fields.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-try-agent/SKILL.md
```diff
@@ -36,42 +51,48 @@ curl -sS --connect-timeout 2 --max-time 5 http://localhost:8080/health/ready -o
...
-5. **Loop for follow-ups.** After the response, ask: "Another question, or done?" Keep the same target until the user changes it.
+2. **Offer another invocation.** Keep the same target unless the user changes it. Do not claim that separate CLI invocations preserve a conversation session.
 
 ## Verification
 
 A "successful" invocation requires both: (a) the CLI returns exit code 0, and (b) the response body is non-empty. An empty body on a question the spec says the agent should handle is a quality signal, not a success.
 
 ```bash
-RESP=$(.venv/bin/nemo agents invoke --agent <name> --input "<query>")
+RESP=$(.venv/bin/nemo agents invoke --agent-deployment <deployment-name> --input "<query>")
```
tylersbray: > **agent** (`pr-review`): P2 — verification always uses `--agent-deployment`

Local mode invokes with `--agent-config`, but mandatory verification reruns with `--agent-deployment`. A successful local-only invocation therefore fails verification. Please branch this snippet by invocation mode (`--agent-config` vs `--agent-deployment`).

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-try-agent/SKILL.md
```diff
@@ -36,42 +51,48 @@ curl -sS --connect-timeout 2 --max-time 5 http://localhost:8080/health/ready -o
 .venv/bin/nemo agents deployments list 2>/dev/null
 ```
 
-Do not use `nemo services status` for this check — it reports stale "running" from held locks after the process has died.
+Do not use `nemo services status` for this check; it reports stale "running" from held locks after the process has died.
 
-If `PLATFORM_DOWN`: route to `nemo-setup` and stop. If `PLATFORM_WEDGED`: route to `nemo-status` to surface the underlying error and stop.
+For a deployed invocation, if `PLATFORM_DOWN`, route to `nemo-setup` and stop. If `PLATFORM_WEDGED`, route to `nemo-status` and stop. Do not require these checks for a local one-shot invocation.
```
tylersbray: > **agent** (`pr-review`): P2 — NAT local invoke can still need Platform

Skipping readiness checks is fine for Platform-owned Fabric local invoke (`_local_fabric_invoke`). Legacy NAT local invoke goes through `temp_injected_config` → `inject_gateway_url`, which `setdefault`s the Platform IGW URL for `openai`/`nim` LLMs lacking `base_url`. Skipping Platform checks there can produce opaque connection failures.

Please distinguish Platform-owned local configs from NAT configs that still require a reachable gateway.

tylersbray (changes requested):
> **agent** (`pr-review`): Request changes — consolidated verified findings

Thanks for the clear `nemo-agents-spec-v1` preferred / NAT compatibility pivot, removing silent `agents delete`, and the confirmation gates. CI looks good.

**Request changes** on two blockers (inline below):

1. **Relative `skills.paths`** — create validates against the authoring dir; deploy only materializes YAML. Non-empty relative skill paths will fail after a successful create.
2. **Vendoring** — `*.yaml` was added, but build still depends on `agent.yml` + `eval-job.json`, which remain absent from the SDK skill tree.

Additional P2s are filed inline (eval poll CLI, try-agent verify branch, deploy-wait selector, guardrails honesty, model step order, `approval_mode`, MCP fields, NAT local gateway).

**Not actionable (replying on threads):** CodeRabbit’s `telemetry.atof` removal and `data[].id` advice — both are false positives against `TelemetryConfig` / `nemo models list` (`ModelEntity.name`).

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-agent-config/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+| NAT Codex field | Platform `agent.yaml` target |
+|---|---|
+| `working_directory` | `environment.workspace` |
+| `sandbox_mode` | `harnesses.codex.settings.sandbox` |
+| `approval_policy: never` | `harnesses.codex.settings.approval_mode: deny_all` |
+| `relay_atof_output_dir` | `environment.artifacts`, `telemetry.output_dir`, and `telemetry.atof` |
+| `skip_git_repo_check` | Omit; removed CLI-only setting |
+| `timeout_seconds` | Omit unless the installed Codex adapter settings schema declares it |
+| `max_output_chars` | Omit; no current Codex adapter setting |
+| `prefer_chatgpt_auth` | Omit; authentication is selected by the Codex adapter and model configuration |
```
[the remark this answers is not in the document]
tylersbray: > **agent** (`pr-review`): False positive — `telemetry.atof` is a real schema field

`TelemetryConfig` defines `atof: dict[str, Any] | None = None` in `plugins/nemo-agents/src/nemo_agents_plugin/agent_config.py`, and `fabric/translator.py` reads `telemetry.atof` (`_relay_atof_config`). Unit tests in `test_agent_config.py` / `test_fabric_translator.py` exercise it.

The migration mapping is therefore valid under `extra="forbid"`. What *is* missing is documenting the dict shape (`enabled`, `filename`, `mode`, sinks, …) in the skill/template — optional docs improvement, not removal of the mapping.

Please do **not** apply the suggested diff that drops `telemetry.atof` from the NAT Codex mapping.

packages/nemo_platform_ext/src/nemo_platform_ext/skills/nemo-model-selection/SKILL.md
```diff
@@ -41,20 +45,30 @@ python scripts/refresh-benchmark-cache.py
...
+- `namespace_to_type[]` — namespace-prefix → NAT `_type` mapping used only for legacy NAT workflow output.
 - `name_decomposition_rules[]` — pattern→hint rules for synthesizing `intent_hints` when an unknown model name lands.
 
-### 2. Fetch the live model list from the running platform
+### 2. Fetch the live model list for a Platform-routed harness
 
 ```bash
-curl -fsS http://localhost:8080/v1/models 2>/dev/null | python3 -c "import sys,json; d=json.load(sys.stdin); names=[m.get('id') for m in d.get('data',[])]; print('\n'.join(n for n in names if n))" 2>/dev/null || echo "PLATFORM_UNREACHABLE"
+nemo models list --all-pages --output-format json 2>/dev/null | python3 -c "import json,sys; d=json.load(sys.stdin); print('\n'.join(m['name'] for m in d.get('data', []) if m.get('name')))" 2>/dev/null || echo "PLATFORM_UNREACHABLE"
 ```
```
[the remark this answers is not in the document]
tylersbray: > **agent** (`pr-review`): False positive — wrong `nemo models` command tree

This finding analyzes `cli/commands/api/inference/models.py` (OpenAI `OpenAIModelResp`: `id` / `owned_by` only). That backs **`nemo inference models list`**, not the skill’s command.

**`nemo models list`** is the Models entity API (`cli/commands/api/models` → `client.models.list` → `GET /apis/models/v2/.../models`). JSON items are `ModelEntity` with **required** `id` and `name` (`openapi.yaml` `ModelEntity`). Table defaults are `name`, `description`.

- Use `data[].name` as the inference-facing identifier for `models.default.model`.
- Do **not** write autogenerated `ModelEntity.id` into `agent.yaml` (the skill’s note is correct).
- Filtering `m.get('name')` is correct for this command; switching to `data[].id` would put the wrong identifier in the config.

The author’s revision should stay as-is.

tylersbray (approved):
Giving an approval so you are not blocked.  Hopefully it is easy to have an agent crank through those recommended changes.

PR #1045, opened 2026-08-03
fix: use streaming NAT target for Studio agent evals

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Added support for NVIDIA Agent Toolkit evaluation targets.
  * Evaluations now use streaming `/generate/full` requests with configured request, response, and aggregation handling.
  * Agent evaluation targets now support filtered steps and encoded agent URLs.

* **Tests**
  * Updated evaluation coverage to verify the new streaming configuration.

### DIFF
--- a/web/packages/studio/src/routes/agents/AgentEvaluationsRoute/components/submitEvaluationSpec.test.ts
+++ b/web/packages/studio/src/routes/agents/AgentEvaluationsRoute/components/submitEvaluationSpec.test.ts
@@ -59,15 +59,23 @@ describe('bareName', () => {
 });
 
 describe('buildAgentTarget', () => {
-  it('targets the non-streaming /generate endpoint of the agent', () => {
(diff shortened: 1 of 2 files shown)


nv-odrulea:
@mckornfield We have actually pointed Studio to nemo-evaluator, and are not using the NAT eval endpoint anymore. 
I think your change is probably correct to fix these tests, but I'm not sure these tests are even needed any longer if we're not pointing there anymore?

nv-odrulea (changes requested):
Our implementation uses generic + stream: false for both agent eval paths (dataset-driven and task-driven evals). 

This PR's tests assert `nemo_agent_toolkit` format, so those tests would fail against our current code. The tests don't match our implementation.

mckornfield:
These change the impl based on a bug. Let me dm as well

PR #1064, opened 2026-08-04
feat(plugin): add EntityUpdateClientProtocol and parent-aware get

SandyChapman:
## Problem

The entity-client protocols in `nemo_platform_plugin` describe a narrower surface than `EntityClient` actually has, so a plugin needing more must declare a private protocol or type against the concrete class. Two capabilities are missing:

- **`update`** — the read-modify-write against the `db_version` optimistic lock.
- **`get(..., parent=...)`** — needed to address a *child* entity, which is unique within `(workspace, entity_type, parent, name)` rather than by name alone.

## Approach

**`update` becomes its own protocol**, composed with the existing ones rather than folded into them:

```python
class Store(EntityClientProtocol[T], EntityUpdateClientProtocol[T], Protocol[T]):
    ...
```

Folding it into `EntityClientProtocol` would be wrong: protocols are structural, so a new member silently invalidates every existing implementer — including every test double — even for services that never call it. `MetricService` only creates and reads; it shouldn't be typed against `update`, and its fakes shouldn't grow methods to satisfy a type.

**`parent` is added to `EntityGetterProtocol.get`** rather than given its own protocol, because a second protocol declaring a conflicting `get` couldn't compose with the first. It's optional, so fetching a root entity is unchanged.

That does require the existing test doubles to accept it — 10 fakes, one argument each. That's a correctness fix rather than churn: each stands in for a client that already accepts `parent`, so their signature was simply inaccurate.

## Drift guard

The tests include a static conformance assertion:

```python
def _static_conformance(client: EntityClient) -> _ReadWriteStore[_Entity]:
    return client
```

`ty` fails here if the client and protocols ever diverge. Verified it actually catches drift — renaming `update` in the protocol produces:

```
info: type `EntityClient` is not assignable to protocol `_ReadWriteStore[_Entity]`
info: └── protocol member `update_renamed` is not defined on type `EntityClient`
```

`_ReadWriteStore` is itself the composition shown above, so the test doubles as proof that the pieces compose.

## Verification

- 1575 tests pass (`nemo_platform_plugin` + `nemo-evaluator`)
- `tools/lint/lint-all.sh` — 13/13
- `ty` across both packages: **130 diagnostics before, 128 after** — two fewer, none new

## Context

Prompted by review feedback on #1023, which had to declare a private `EntityStoreProtocol` to publish entity revisions. That PR will drop its private copy and compose these instead. No evaluator-specific behavior is included here — the only evaluator changes are the test-double signatures the protocol change requires.

## Summary by CodeRabbit

* **New Features**
  * Entity retrieval and deletion now support optional parent context for nested or child entities.
  * Added dedicated entity update support, including optimistic-locking safeguards.
  * Expanded the public API with additional entity client capabilities.

* **Tests**
  * Added coverage for parent-aware entity operations, update compatibility, and child-entity workflows.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/__init__.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/__init__.py
@@ -69,6 +69,9 @@ from nemo_platform_plugin.entities.base import (
     EntityTypeLike as EntityTypeLike,
 )
 from nemo_platform_plugin.entities.base import (
+    EntityUpdateClientProtocol as EntityUpdateClientProtocol,
+)
+from nemo_platform_plugin.entities.base import (
     EntityValidationError as EntityValidationError,
 )
 from nemo_platform_plugin.entities.base import (
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
@@ -231,14 +231,29 @@ EntityTypeLike = Type[EntityT] | EntityToken
 
 
 class EntityGetterProtocol(Protocol[EntityT]):
-    """Protocol for entity clients that can fetch entities by workspace/name."""
+    """Protocol for entity clients that can fetch entities by workspace/name.
 
-    async def get(self, entity_type: Type[EntityT], *, name: str, workspace: str) -> EntityT: ...
+    ``parent`` addresses a **child** entity, which is unique within
+    ``(workspace, entity_type, parent, name)`` rather than by name alone. It is optional, so
+    fetching a root entity is unchanged.
+    """
+
+    async def get(
+        self,
+        entity_type: Type[EntityT],
+        *,
+        name: str,
+        workspace: str,
+        parent: Optional[str] = None,
+    ) -> EntityT: ...
 
 
 class EntityDeleteClientProtocol(EntityGetterProtocol[EntityT], Protocol[EntityT]):
     """Protocol for entity clients that can list and delete entities."""
 
+    # ``filter_str`` and ``filter_obj`` exist on the client but are deliberately absent here:
+    # ``filter_operation`` is the sanctioned structured form, and the other two are a JSON-string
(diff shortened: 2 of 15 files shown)


maxdubrinsky (approved):

ironcommit (approved):
We should probably follow up with further checks.  delete also appears to be missing the parent field.

SandyChapman:
Confirmed — and the audit turned up one more. Comparing each protocol method against `EntityClient` parameter by parameter:

```
get:    complete
create: complete
delete: protocol is missing ['parent']
list:   protocol is missing ['filter_obj', 'filter_str']
```

Folded the fixes into this PR:

- **`delete` now takes `parent`.**
- **`list` deliberately keeps `filter_str`/`filter_obj` off**, with a comment saying why: `filter_operation` is the sanctioned structured form and those two are a JSON-string variant and an exact-match shorthand kept for older callers.
- **Added the further check you asked for.** The existing conformance test could not have caught this: structural conformance is one-directional, so a class satisfies a protocol by providing *at least* what it declares and extra client parameters pass silently — `ty` reported success with all three parameters missing. The new test compares signatures parameter by parameter and fails on anything absent, with an explicit allowlist for the two intentional omissions. Reverting the `delete` fix makes it fail with `protocol is missing client parameters: {'delete': {'parent'}}`.

Full suite green (1576 passed); no new `ty` diagnostics in the touched files.

SandyChapman:
One follow-on from reviewing this: **`update` does not need `parent`**, and adding it would be wrong.

The split is by whether the caller holds an entity. `create` and `update` read `entity._parent` (base.py:634, :742); `get` and `delete` address by name with nothing in hand, so they have to take it explicitly — which is why only those two had the gap. Giving `update` a `parent` argument would create two sources of truth and a silent conflict when they disagree.

That correctness argument rests on `get` populating `_parent` on the way out, and nothing was pinning it, so I added two tests to `test_entity_client.py`: a get→mutate→update round trip asserting the parent query param survives both hops, and a direct check that `delete` forwards it. Both mutation-tested — breaking either half of the round trip fails the test.

PR #1013, opened 2026-07-31
feat(evaluator)!: typed run metadata and a required runner identity contract

SandyChapman:
Closes AALGO-451.

## Problem

A finished run couldn't answer *"what produced this, with what settings, and how long did it take?"* — no timing, no target identity, no SDK version.

Callers improvised provenance inside the untyped `benchmark: dict[str, Any]` bag with no convention — `{"name": ...}` in one example, `{"benchmark": ...}` in another, plus `mode`/`backend`/`task`/`score_source`. The auto-derived value also shape-shifted between `str` and `list[str]` depending on how many benchmarks the tasks declared.

## Change

- **`AgentEvalResult.benchmark` → `metadata: RunMetadata`** — typed `labels`, `target`, `started_at`/`finished_at`/`duration_sec`, `sdk_version`.
- **`runner_info() -> RunnerInfo` is now part of `AgentTaskRunner`.** Identity is universal, not an optional capability, so every shipped runner implements it with a curated name (`gym`, `harbor`, `docker_sandbox`, `callable`, `codex_cli`, `codex_docker_cli`, `fabric`, `fabric_container`) plus the settings that shape its results. Credentials are deliberately excluded — `FabricContainerRuntime` reports its provider but never `_secrets`.
- **`benchmark` dropped entirely** (field, config parameter, and task-metadata derivation). Nothing computed on it, it duplicated what `labels` already expresses, and it invited misuse as a run name — one in-tree example was using it for exactly that.
- **Bundle artifact `benchmark.json` → `metadata.json`.**

Example output from a live Gym run:

```json
{
  "target": { "name": "gym", "kind": "runner",
              "config": { "resources_server": "mcqa", "agent": "simple_agent",
                          "model_type": "inference_provider", "num_repeats": 2 } },
  "started_at": "2026-07-31T11:46:40Z", "finished_at": "2026-07-31T14:14:08Z",
  "duration_sec": 8848.3, "sdk_version": "0.0.0", "labels": {}
}
```

## Breaking change

`AgentTaskRunner` now requires `runner_info()`. The protocol is `@runtime_checkable` and isinstance-dispatched, so a runner lacking it no longer matches and raises `NotImplementedError: unsupported agent-eval target type`. All in-tree runners and test doubles are updated.

## Verification

- Full SDK suite: **25 failures, identical to clean `main`** (23 ragas + 1 git_patch + 1 fabric live-integration, all pre-existing); 1287 passing vs main's 1283 — the +4 are new provenance tests.
[...]

PR #1065, opened 2026-08-04
feat(evaluator): aggregate agent-eval results natively and import Gym's own

SandyChapman:
Closes AALGO-434.

## Why

Agent-eval reported a mean and little else, and a runner's own aggregations were discarded entirely — so a Gym run's headline numbers (pass@k, per-environment metrics) were visible in Gym's output files but nowhere in the eval result.

## Native aggregation

- **pass@k** per task via the unbiased estimator (Chen et al. 2021), for every score-like metric output. An attempt passes on full credit — deliberately not configurable, since it's a reporting-time interpretation and a tunable version would produce numbers that look comparable across runs but aren't.
- **Percentiles** on every aggregate, reusing the deterministic-metric helper (now public as `compute_percentiles`) so both paths report distributions the same way.
- **`median`** added to `AggregateScoreBase`, populated natively from `p50`, so the field means the same thing whether a score was computed here or imported from a backend that reports a median without a full distribution.
- **Both standard-deviation conventions, named explicitly.** `std_dev`/`variance` remain population (÷n); `sample_std_dev`/`sample_variance` are new (÷n−1, `None` when n<2). Gym computes its spread with pandas (ddof=1), so naming both lets the two sets of numbers coexist without either silently changing meaning. No existing value changes.

## Importing Gym's own numbers

`RunAggregationsProvider` is a single method — `run_aggregate_scores()` — returning typed aggregates that merge into `summary.scores`. There is no parallel opaque passthrough: Gym already writes its raw `rollouts_aggregate_metrics.json` inside the run's work dir, so a second copy on the result would duplicate a file already in the bundle *and* re-introduce the untyped `dict[str, Any]` bag that #1013 removes.

Names are **`runner.gym.<metric>`** — namespaced by *runner*, not agent, since each run uses a single agent and what a reader needs to know is which backend produced the number. Falls back to `runner.gym.<agent>.<metric>` only if a run produced several. Gym's `reward` is skipped: the SDK already scores it natively as `gym_reward.reward` from the same rollouts.

### The part most worth reviewing

Imports read **`agent_metrics`**, not `key_metrics`. This matters: `key_metrics` is a *subset* of `agent_metrics` chosen by the resources-server, and Gym's default `get_key_metrics` keeps **only the `mean/*` entries** — so max/min/median/std never appear there, and sourcing from it would degrade every distribution into a lone `runner.gym.mean/<name>` scalar. `test_imports_read_agent_metrics_not_the_key_metrics_subset` pins this.

Gym flattens each distribution into `<stat>/<metric>` keys. Its `describe_dataframe` emits `mean/max/min/median/std/histogram` together, and `prepare_for_serialization` strips `histogram` before writing — so the family in the file is exactly those five, and re-assembly requires **all five**. That guard matters because **36 of Gym's ~97 resources-servers override `compute_metrics`**, emitting keys in their own shapes (`arena_elo/score`, `easy/pass@1/accuracy`), and a server is free to define a metric literally named `mean`. A partial match stays as standalone scalars rather than being renamed into a statistic of a distribution that never existed.

`test_nothing_numeric_is_dropped_from_a_custom_environment_payload` pins the invariant: importing is a *renaming*, never a filter. Only `reward` (redundant by construction) and non-numeric values may disappear.

## New `AggregateScalarScore`

A single pre-computed value with no underlying sample. Distinct from `AggregateRangeScore` so a reader can tell "this is the whole story" from "this summarizes `count` samples", instead of seeing a range score with a suspicious `count` of 1.

`count` is now `int | None`, and imported aggregates carry **`None`** — Gym reports statistics without the n behind them, and `0` would assert that nothing was evaluated (besides being a division hazard). The dashboard and example render `None` as an em dash while still showing a genuine `0`.

## Also

- Dashboard gains Value / Median / Std dev columns and renders scalars by value.
- New `examples/gym/inspect_results.py` — companion to `run_gym_eval.py`, showing how to reach each kind of result, with lift-and-paste accessors.

## Known scope boundary

Gym's `group_level_metrics` (its **per-task** aggregation) is not imported. `summary.scores` is run-level, so it has no home there yet — the SDK computes per-task groupings internally for pass@k but doesn't surface that granularity. Native per-task rollups plus importing Gym's into them is tracked separately; the raw data remains in Gym's own file inside the bundle.

## Verification

- Full run: **1883 passed / 25 failed**, all 25 the pre-existing ragas + live-Fabric baseline (verified: zero failures outside it).
- Ruff clean; `ty` at exactly the `main` baseline (198); `make vendor` run with no drift.

## Summary by CodeRabbit

* **New Features**
  * Evaluation summaries and dashboards now include median, variance, standard deviation, percentiles, and improved count handling.
  * Gym results can include runner-provided aggregate metrics.
  * Added pass@k reporting for eligible task-level metrics.
  * Added a command-line example for inspecting saved results and per-task outcomes.

* **Bug Fixes**
  * Improved scalar metric display and handling of incomplete or malformed Gym aggregate data.
  * Corrected pass@k calculations across supported metric types.

* **Documentation**
  * Updated Gym example instructions for running evaluations and inspecting saved results.

### DIFF
--- a/packages/nemo_evaluator_sdk/examples/gym/README.md
+++ b/packages/nemo_evaluator_sdk/examples/gym/README.md
@@ -31,7 +31,7 @@ YAML
 From the nemo-platform repo root (any Python with `nemo_evaluator_sdk` importable — the runner shells out to Gym's own venv):
 
 ```bash
-python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval --gym-root /path/to/Gym
+uv run python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval --gym-root /path/to/Gym
 ```
 
 Useful flags: `--resources-server`, `--agent`, `--model-type` (`inference_provider` for OpenAI-compatible **chat** endpoints; `openai_model` uses the OpenAI **Responses API** and 500s against chat-only endpoints), `--num-repeats`, `--output-dir`.
@@ -51,6 +51,20 @@ aggregate scores:
 
 Run bundle (run.json, trials.jsonl, scores.jsonl, report.html): /var/folders/.../gym-eval-ab12cd34
 ```
+
+## Read the results
+
+`inspect_results.py` is the companion to the above: it reads a bundle and shows how to reach each kind of result — headline aggregates, `pass@k`, per-task outcomes, and the runner's own imported numbers. Its accessors (`aggregate`, `per_task_outcomes`) are written to be lifted into your own code, and everything it shows also works on the in-memory `AgentEvalResult` that `AgentEvaluator().run(...)` returns — reading a bundle just makes it runnable without a live run.
+
+No bundle is checked in; the run above produces one. Give it a stable `--output-dir` and point the reader at the same path:
+
+```bash
+uv run python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval \
+    --gym-root /path/to/Gym --output-dir /tmp/gym-eval
+uv run python -m packages.nemo_evaluator_sdk.examples.gym.inspect_results --bundle /tmp/gym-eval
+```
+
+Aggregates named `runner.gym.*` are Gym's own figures, imported into `summary.scores` so they are addressable exactly like the SDK's — the prefix is what tells you which side computed them. The script cross-checks Gym's `pass@1` against the SDK's natively-computed one (Gym reports accuracy on a 0-100 scale where the SDK uses 0-1). It works on any agent-eval bundle: pass `--metric-type`/`--output-name` for a run scored with a different metric.
 
 ## How it runs Gym
 
--- a/packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
+++ b/packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
@@ -1,0 +1,222 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Read and display the results of an agent-eval run.
+
+Companion to ``run_gym_eval.py``: that script *produces* a run bundle, this one *reads* it and shows
+how to get at each kind of result — headline aggregates, ``pass@k``, per-task outcomes, and the
+runner's own aggregations.
+
+The helpers below (:func:`aggregate`, :func:`per_task_outcomes`) are written to be lifted directly
+into your own code. Everything shown here also works on the in-memory ``AgentEvalResult`` returned by
+``AgentEvaluator().run(...)`` — reading from a bundle just makes the example runnable without a live
+run.
+
+There is no bundle checked into the repo — ``run_gym_eval.py`` makes one. It writes to a fresh
+temporary directory by default, so give it an explicit ``--output-dir`` and point this script at the
+same path. From the repository root::
+
+    uv run python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval \\
+        --gym-root /path/to/Gym --output-dir /tmp/gym-eval
+    uv run python -m packages.nemo_evaluator_sdk.examples.gym.inspect_results --bundle /tmp/gym-eval
+
+Any agent-eval bundle works, not just a Gym one: pass ``--metric-type``/``--output-name`` for the
+metric it was scored with. Only the ``runner.gym.`` section is Gym-specific, and it is skipped when a
+run has no imported aggregations.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+from collections.abc import Sequence
+from pathlib import Path
+
+from nemo_evaluator_sdk.agent_eval.results import AgentEvalSummary
+from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEvalTaskScore, is_trial_failure
+from nemo_evaluator_sdk.values.results import AggregateScalarScore, AggregateScore
+
+#: Value at which an attempt counts as a pass, matching the SDK's pass@k definition (full credit).
+PASS_VALUE = 1.0
+
+#: Namespace the Gym runner's own aggregations are imported under, so they never collide with ours.
+RUNNER_PREFIX = "runner.gym."
+
+# --------------------------------------------------------------------------------------------------
+# Accessors — lift these into your own code.
+# --------------------------------------------------------------------------------------------------
+
+
+def aggregate(summary: AgentEvalSummary, name: str) -> AggregateScore:
+    """Look up one aggregate by name, e.g. ``"gym_reward.reward.pass@2"``.
+
+    Aggregates are a flat list, so this is a scan. Raises with the available names on a miss, which is
+    the failure you actually want when a metric or output was renamed.
+    """
+    for score in summary.scores.scores:
+        if score.name == name:
+            return score
+    available = ", ".join(sorted(score.name for score in summary.scores.scores))
+    raise KeyError(f"no aggregate named {name!r}; available: {available}")
+
+
+def per_task_outcomes(
+    scores: Sequence[AgentEvalTaskScore],
+    *,
+    metric_type: str,
+    output_name: str,
+) -> dict[str, list[float | None]]:
+    """Group per-trial score values by task: ``task_id -> [value per attempt]``, ``None`` if it died.
+
+    A run with ``num_repeats=R`` produces R trials per task, and the scores are a flat
+    task x trial x metric list — so answering "which tasks failed?" means grouping them yourself.
+
+    Failed trials are kept, as ``None``. Dropping them would show a task that passed once and crashed
+    once as solved, and disagrees with how the SDK computes pass@k (a dead rollout is an attempt that
+    did not pass). A failed *metric* is dropped instead: it leaves the attempt unmeasured rather than
+    unsuccessful, so counting it against the agent would turn a judge timeout into a failure.
+    """
+    by_task: dict[str, list[float | None]] = {}
+    for score in scores:
+        if score.metric_type != metric_type:
+            continue
+        if is_trial_failure(score):
+            by_task.setdefault(score.task_id, []).append(None)
+            continue
+        if score.status == AgentEvalScoreStatus.FAILED:
+            continue
+        for output in score.outputs:
+            if output.name == output_name and isinstance(output.value, int | float):
+                by_task.setdefault(score.task_id, []).append(float(output.value))
+    return by_task
+
+
+# --------------------------------------------------------------------------------------------------
+# Bundle loading (see the run.json manifest for the full artifact list).
+# --------------------------------------------------------------------------------------------------
+
+
+def load_bundle(bundle: Path) -> tuple[AgentEvalSummary, list[AgentEvalTaskScore]]:
+    """Hydrate the pieces of a persisted run bundle used below.
+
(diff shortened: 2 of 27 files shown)


packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -0,0 +1,202 @@
...
+runner's own aggregations.
+
+The helpers below (:func:`aggregate`, :func:`per_task_outcomes`) are written to be lifted directly
+into your own code. Everything shown here also works on the in-memory ``AgentEvalResult`` returned by
+``AgentEvaluator().run(...)`` — reading from a bundle just makes the example runnable without a live
+run.
+
+Run from the repository root::
+
+    uv run python -m packages.nemo_evaluator_sdk.examples.gym.inspect_results --bundle /tmp/gym-eval
```
ngoncharenko, line 21: should we commit example of a bundle to run this command?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -143,9 +163,87 @@ def _aggregate_scores(
...
+        attempts_and_passes: dict[str, list[int]] = {}  # task_id -> [n_attempts, n_passes]
+        for score in scores:
+            if score.metric_type != metric_type:
+                continue
+            if score.status not in (AgentEvalScoreStatus.COMPLETED, AgentEvalScoreStatus.PARTIAL):
+                continue
+            output = _score_output(score, output_name)
+            value = _semantic_value(output) if output is not None else None
+            if value is None:
+                continue
```
ngoncharenko, line 248: [P1] **Count failed attempts in pass@k.** These guards discard `FAILED` scores and missing outputs before incrementing `n`. With one full-credit attempt and one Gym failure-sidecar attempt, `pass@1` is reported as `1.0` instead of `0.5`, and `pass@2` disappears. Count every matching attempt in the denominator, while incrementing passes only for a usable full-credit value.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -732,6 +735,17 @@ def _describe_target(
...
 
 
+def _collect_runner_aggregate_scores(target: object) -> list[AggregateScore]:
+    """The typed subset of a runner's own aggregations, for merging into ``summary.scores``.
+
+    A runner that maps its numbers onto aggregate scores namespaces them under ``runner.<name>.``, so
+    they sit alongside the SDK's own without being mistaken for them.
+    """
+    if isinstance(target, RunAggregationsProvider):
+        return list(target.run_aggregate_scores())
```
ngoncharenko: [P2] **Enforce runner score namespaces.** `RunAggregationsProvider` requires names under `runner.<runner_name>.*`, but this accepts any name and merges it directly into `summary.scores`. A provider returning `reward.reward` can therefore collide with an SDK-computed aggregate. Validate the required runner namespace here and reject invalid entries; add a negative contract test.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/dashboard.py
```diff
@@ -83,18 +84,41 @@ def _metric_rollups(result: AgentEvalResult) -> str:
...
+
+    A scalar score has no mean — rendering the column straight off ``score.mean`` would leave every
+    runner-imported figure blank in the table where it is the only thing worth reading.
+    """
+    return score.value if isinstance(score, AggregateScalarScore) else score.mean
+
+
+def _median(score: AggregateScore) -> float | None:
+    percentiles = getattr(score, "percentiles", None)
+    return percentiles.p50 if percentiles is not None else None
```
ngoncharenko, line 119: [P2] **Render the standalone median.** Imported range scores may report `median` without a percentile distribution, which the new model explicitly supports. This helper returns `None` for such scores, so the dashboard renders `n/a` despite having a median. Read `score.median` directly, optionally falling back to `percentiles.p50` for older data.

ngoncharenko (commented):
Three inline findings from the merge-base review.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
SandyChapman: Fixed: failed trials now count toward `n` (never `c`); metric failures stay out, since a judge timeout says nothing about whether the agent passed — and pass@k's `nan_count` was structurally always 0, so unusable tasks are now reported too.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
SandyChapman: Agreed — `_collect_runner_aggregate_scores` now validates `runner.<name>.` and drops violators with a warning (not a raise: this runs after `run_tasks`), plus the negative test.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/dashboard.py
SandyChapman: Confirmed — `_median` read only `percentiles.p50`, which blanked the column for every Gym import; now reads `score.median` first with a `p50` fallback.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
SandyChapman: It chains off `run_gym_eval.py --output-dir`, which nothing actually said — documented the pairing in the docstring and README rather than committing a bundle that would pin Gym's non-contractual output schema into the repo.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/scores.py
```diff
@@ -20,6 +20,14 @@ class AgentEvalScoreStatus(str, Enum):
     PARTIAL = "partial"
 
 
+#: Diagnostic detail key stamped when a score is ``FAILED`` because the *trial* failed — the agent
+#: produced nothing to score — rather than because the metric itself raised. Both are reported as
+#: ``FAILED``, but they mean different things to a reader: a failed trial is a failed *attempt*, a
+#: failed metric is a failed *measurement*. Consumers that must tell them apart read this key via
+#: :func:`is_trial_failure`.
+TRIAL_STATUS_DETAIL = "trial_status"
```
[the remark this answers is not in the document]
SandyChapman: Now matches the detail's value (`failed`) rather than the key's presence, which also rejects a stray `trial_status: completed`; namespacing would make it the only such key in the SDK. Test added.

packages/nemo_evaluator_sdk/examples/gym/README.md
```diff
@@ -52,6 +52,20 @@ aggregate scores:
...
+
+No bundle is checked in; the run above produces one. Give it a stable `--output-dir` and point the reader at the same path:
+
+```bash
+uv run python -m packages.nemo_evaluator_sdk.examples.gym.run_gym_eval \
+    --gym-root /path/to/Gym --output-dir /tmp/gym-eval
+uv run python -m packages.nemo_evaluator_sdk.examples.gym.inspect_results --bundle /tmp/gym-eval
+```
+
+Aggregates named `runner.gym.*` are Gym's own figures, imported into `summary.scores` so they are addressable exactly like the SDK's — the prefix is what tells you which side computed them. The script cross-checks Gym's `pass@1` against the SDK's natively-computed one (Gym reports accuracy on a 0-100 scale where the SDK uses 0-1). It works on any agent-eval bundle: pass `--metric-type`/`--output-name` for a run scored with a different metric.
```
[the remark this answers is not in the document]
SandyChapman: Skipping: this is an examples README, not Sphinx docs — MyST tab-sets don't render on GitHub, and both commands already are the Python SDK.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -0,0 +1,210 @@
...
+) -> dict[str, list[float]]:
+    """Group per-trial score values by task: ``task_id -> [value per attempt]``.
+
+    A run with ``num_repeats=R`` produces R trials per task, and the scores are a flat
+    task x trial x metric list — so answering "which tasks failed?" means grouping them yourself.
+    """
+    by_task: dict[str, list[float]] = {}
+    for score in scores:
+        if score.metric_type != metric_type or score.status == AgentEvalScoreStatus.FAILED:
+            continue
```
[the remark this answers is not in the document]
SandyChapman: Good catch, and worse than described: the example rendered a task that passed once and crashed once as *solved*, contradicting the pass@k rules in the same PR. Now keeps failed trials as `None` attempts and drops only metric failures.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -729,6 +743,138 @@ def _failures_path_for(rollouts_path: Path) -> Path:
...
+    isn't contractual, so the SDK does not type it.
+    """
+    path = _aggregate_metrics_path_for(rollouts_path)
+    if not path.exists():
+        return None
+    try:
+        parsed = json.loads(path.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError):
+        logger.warning("Could not parse Gym aggregate metrics at %s; skipping run aggregations.", path)
+        return None
```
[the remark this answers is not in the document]
SandyChapman: Valid, `UnicodeDecodeError` is a `ValueError` so it slipped the handler; added it, with a test writing a truncated-codepoint sidecar.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -729,6 +743,138 @@ def _failures_path_for(rollouts_path: Path) -> Path:
...
+                median=stats["median"],
+                # Gym computes this with pandas (ddof=1), so it is the sample standard deviation.
+                sample_std_dev=stats["std"],
+                sample_variance=stats["std"] ** 2,
+            )
+        )
+    scores.extend(
+        AggregateScalarScore(name=f"{prefix}{key}", count=None, nan_count=0, value=value)
+        for key, value in sorted(scalars.items())
+        if key not in _GYM_REDUNDANT_METRICS
```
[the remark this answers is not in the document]
SandyChapman: Valid; moved the redundancy check ahead of the incomplete-family fallback so a partial `reward` family can't survive as `mean/reward` scalars. Test added.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/scores.py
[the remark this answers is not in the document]
SandyChapman: Holding: that requires a score whose own diagnostic asserts the trial failed while meaning a metric failure, which no SDK path produces and which is self-contradictory data rather than a collision — and the key would be the only namespaced one in the SDK (`exception_type` is equally generic and bare).

PR #1071, opened 2026-08-04
feat(evaluator): make a stored task runner-polymorphic via kind

SandyChapman:
> **#1023 has merged.** This branch is rebased onto `main` and is now a single commit; the earlier stacked-review note no longer applies.

## Why

A task is an evaluation unit; *how* it runs is a property of the task, not a different kind of record. The target side already models this — `AgentRunnerTarget` is a `kind`-discriminated union of `codex | fabric | harbor`, with a comment saying "widen with more members as runners land."

This makes the stored side match, so a user manages every evaluation unit in one place regardless of which runner executes it. It's the foundation for publishing Harbor tasks into NeMo.

## What

Task content moves under a discriminated `spec`:

| kind | content |
|---|---|
| `evaluator` | `intent`, `inputs`, `reference`, `metrics`, `views` |
| `harbor` | `archive_ref` (fileset), `archive_digest`, `instruction`, opaque Harbor `config` |

**Nested rather than flattened with nullable per-kind fields.** Flat would make every field optional with nothing enforcing a coherent set; nesting keeps each variant's required fields required, and the revision digest then covers the spec as a unit — so two kinds with coincidentally similar metadata can't collide on content. There's a test for that.

**Named `…Definition`, not `…Spec`.** `AgentEvalTaskSpec` already names the *runtime* task DTO in `jobs.agent_spec`; two models sharing a class name across modules break OpenAPI generation with a schema-name collision. (Found by regenerating the spec — see below.)

**One fileset per Harbor task**, so a task shared by several tasksets is stored once. `archive_ref` is shape-validated, so a malformed reference is rejected at publish rather than surfacing as a download failure mid-run — matching how a metric reference is checked when a task is stored.

## Grader-only `reference` on stored tasks

`EvaluatorTaskDefinition` gains `reference` — held-out ground truth, surfaced to metrics but never seeded into the agent's workspace or shown to the agent. It has existed on the inline `AgentEvalTaskInput` since #566, where persisting it was deferred because the stored schemas then lived in the root OpenAPI/SDK; they're plugin-owned now, so that reason has lapsed.

Until this, a taskset-driven run expanded to an *empty* reference, so any task needing ground truth the agent cannot edit had to give up stored tasks and tasksets entirely — which is the reward-hacking hole #566 was opened to close.

Held out from the *agent*, not from the API: anyone who can read the task can read it.

## What the revision digest covers

The rule: **the digest covers anything that affects a task's execution output or how it is graded.** A field may be excluded only if it's a derived view of content the digest already covers by another route.

`reference` is therefore **in** — it decides what a metric grades against, so two revisions that score differently must not share a digest, or publish-time dedup would collapse them and a pin would stop fixing the grading.

Harbor's `config` is the one **exclusion**, and it doesn't contradict that rule: it's a projection of `task.toml`, Harbor reads the real file out of the materialized archive at run time, and `archive_digest` is authoritative over every file in that directory. So a config change that genuinely alters execution or grading already moves the digest, while hashing the projection would make revision history sensitive to Harbor's serialization — a release that reordered keys would cut a revision for byte-identical files.

That makes `archive_digest` load-bearing, and there's a test guarding it: a Harbor field ever read from the stored record rather than from the archive would have to be digested.

## Scope: storage, not execution

This PR makes the **storage** model runner-polymorphic. It touches no execution-path source file.

|  | `evaluator` kind | `harbor` kind |
|---|---|---|
| Stored-task management (CRUD, revisions, tasksets) | ✅ | ✅ |
| Execution **from a stored task** | ✅ today | ❌ follow-up |

`resolve_taskset_ref` previously projected every member onto an agent-eval DTO. A Harbor task's content is a *directory of files*, not fields, so that projection would silently produce a task with no intent and no metrics — an evaluation that runs and scores nothing. It now raises `UnsupportedTaskKindError`, surfaced as a **422** before the run starts.

Harbor execution itself already works on `main` (`HarborRunnerTarget`, driven by `harbor_dataset_path` metadata) and is unaffected by this PR — it just isn't fed by stored task entities yet.

**Follow-up: [AALGO-481](https://linear.app/nvidia/issue/AALGO-481/bridge-stored-harbor-tasks-to-the-harbor-runner)** — bridge stored harbor tasks to the harbor runner, and encode runner ↔ task-kind compatibility declaratively instead of via an `isinstance` check in the expansion path. That ticket also covers fixing the current rejection message, which unhelpfully suggests submitting against "a target whose runner executes 'harbor' tasks" when no such target exists yet.

## Breaking change

No migration. Rows stored in the previous flat shape fail validation on read, which surfaces as a 500 when listing tasks. Clear them before upgrading.

## Verification

- 733 unit tests
- `tools/lint/lint-python-types.sh` — 0 errors
- `ruff check` / `ruff format --check` — clean
- Plugin OpenAPI spec regenerated

Docs updated: examples, the `TaskInput` field table, the task-kinds section, and the evaluator skill reference.

## Summary by CodeRabbit

- **New Features**
  - Added support for evaluator and Harbor task types with runner-specific task details.
  - Tasks now use a unified specification for inputs, metrics, references, and Harbor configuration.
  - Added grader-only references and pinned taskset grading content.
  - Changes to grading references or Harbor archives now create appropriately tracked revisions.
  - Added validation for task definitions, metric references, metadata, and taskset membership.

- **Documentation**
  - Updated task creation, retrieval, stored-task, taskset, and held-out-reference examples for the new format.

### DIFF
--- a/docs/evaluator/manage-tasks-tasksets.mdx
+++ b/docs/evaluator/manage-tasks-tasksets.mdx
@@ -73,34 +73,63 @@ Reference the stored metric with a `MetricRef` (`workspace/name`, or a bare `nam
 the task's workspace). The service returns the stored `Task`.
 
 ```python
-from nemo_evaluator.api.schemas import MetadataItem, MetricRef, TaskInput, TaskInputs
+from nemo_evaluator.api.schemas import EvaluatorTaskDefinition, MetadataItem, MetricRef, TaskInput, TaskInputs
 
 task = TaskInput(
-    intent="Answer the user's geography question with the capital city.",
-    inputs=TaskInputs(instruction="What is the capital of France?"),
-    metrics=[MetricRef("default/answer-exact-match")],
+    spec=EvaluatorTaskDefinition(
+        intent="Answer the user's geography question with the capital city.",
+        inputs=TaskInputs(instruction="What is the capital of France?"),
+        metrics=[MetricRef("default/answer-exact-match")],
+    ),
     metadata=[MetadataItem(key="suite", value="geography")],
 )
 
 stored = tasks.create("capital-of-france", task=task)
-print(stored.id, stored.metrics)
+print(stored.id, stored.spec.metrics)
 ```
 
 ### `TaskInput` fields
 
 | Field | Type | Required | Description |
 |-------|------|----------|-------------|
-| `intent` | `str` | Yes | Human-readable description of the desired agent behavior. |
-| `inputs` | `TaskInputs` | No | The task's recognized input fields. `instruction` is the agent's prompt; it falls back to `intent` when unset. |
-| `metrics` | `list[MetricRefOrInline]` | No | The metrics that score the task, as `MetricRef` references (`workspace/name`) to stored metrics. Pre-built inline metric bundles (`MetricInline`) are also accepted and are normalized to stored metrics on create. |
-| `views` | `dict[str, SemanticView]` | No | Optional reporting views mapping metric outputs into named semantic scores. |
+| `spec` | `TaskDefinition` | Yes | The task's content, discriminated by `kind` — see below. |
 | `metadata` | `list[MetadataItem]` | No | Key/value annotations. Keys must be unique. |
 | `tags` | `list[str]` | No | Tags to point at the revision this request publishes. `latest` is always applied server-side. |
+
+### Task kinds
+
+A task is an evaluation unit; its `kind` says which runner executes it. Both kinds are stored as the
+same record type, so a taskset can group them and you manage every evaluation unit in one place.
+
+`EvaluatorTaskDefinition` (`kind="evaluator"`) — scored by platform metrics:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `intent` | `str` | Yes | Human-readable description of the desired agent behavior. |
+| `inputs` | `TaskInputs` | No | The task's recognized input fields. `instruction` is the agent's prompt; it falls back to `intent` when unset. |
+| `reference` | `dict[str, Any]` | No | Grader-only ground truth (held-out tests, expected outputs, rubric data). Surfaced to metrics but never seeded into the agent's workspace or shown to the agent. Held out from the *agent*, not from the API. |
+| `metrics` | `list[MetricRefOrInline]` | No | The metrics that score the task, as `MetricRef` references (`workspace/name`) to stored metrics. Pre-built inline metric bundles (`MetricInline`) are also accepted and are normalized to stored metrics on create. |
+| `views` | `dict[str, SemanticView]` | No | Optional reporting views mapping metric outputs into named semantic scores. |
+
+`HarborTaskDefinition` (`kind="harbor"`) — a Harbor task, scored by Harbor's own reward:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `archive_ref` | `str` | Yes | Files reference to the task's packaged directory (`workspace/fileset#path`). One fileset per task, so a task shared by several tasksets is stored once. |
+| `archive_digest` | `str` | Yes | Content hash Harbor computed over the task directory. |
+| `instruction` | `str` | No | The task's instruction text, when it has one. |
+| `config` | `dict` | No | Harbor's own task configuration (verifier, agent, environment, steps), stored as published. |
+
+<Note>
+Storing a Harbor task is supported; **running one from storage is not yet**. A taskset may group both
+kinds, but expanding a `harbor` member is rejected with `422` before the run starts, whatever target
+you submit against. Harbor evaluations continue to run through the existing dataset-driven path.
+</Note>
 
 <Note>
 A stored task holds **metric references only**. Any inline metric bundle you pass on create is stored
 as a content-addressed *derived* metric, and the task record is normalized to reference it. This is
-why `stored.metrics` always comes back as a list of `MetricRef` references.
+why `stored.spec.metrics` always comes back as a list of `MetricRef` references.
 </Note>
 
 ### Retrieve, list, and delete
@@ -108,12 +137,12 @@ why `stored.metrics` always comes back as a list of `MetricRef` references.
 ```python
 # Retrieve one task by name (its current content)
 task = tasks.retrieve("capital-of-france")
-print(task.revision, task.tags)  # e.g. 1 {'latest': 1}
+print(task.spec.kind, task.revision, task.tags)  # e.g. evaluator 1 {'latest': 1}
 
 # List tasks in the workspace (paginated)
 page = tasks.list(page=1, page_size=100, sort="-created_at")
 for item in page.data:
-    print(item.name, item.intent)
+    print(item.name, item.spec.kind)
 
 # Delete a task (this also removes all of its revisions)
 tasks.delete("capital-of-france")
@@ -361,9 +390,10 @@ A fragment that no longer resolves fails the evaluation rather than falling back
 revision.
 
 <Note>
-Stored tasks carry no grader-only `reference` (held-out ground truth): that field lives only on inline
-`AgentEvalTaskInput`. Taskset-driven tasks therefore run with an empty `reference`, so use a taskset
-when your metrics score the agent's output directly rather than against per-task held-out data.
+A member's grader-only `reference` (held-out ground truth) is loaded from the pinned revision along
+with the rest of its content, so a taskset-driven run grades against the ground truth that revision
+fixed. Because `reference` is covered by the revision digest, changing it publishes a new revision —
+a pin fixes the grading, not just the prompt.
 </Note>
 
 ## Async usage
--- a/docs/evaluator/manage-tasks-tasksets.mdx
+++ b/docs/evaluator/manage-tasks-tasksets.mdx
@@ -77,6 +77,7 @@ from nemo_evaluator.api.schemas import EvaluatorTaskDefinition, MetadataItem, Me
 
 task = TaskInput(
     spec=EvaluatorTaskDefinition(
+        kind="evaluator",
         intent="Answer the user's geography question with the capital city.",
         inputs=TaskInputs(instruction="What is the capital of France?"),
         metrics=[MetricRef("default/answer-exact-match")],
--- a/docs/evaluator/manage-tasks-tasksets.mdx
+++ b/docs/evaluator/manage-tasks-tasksets.mdx
@@ -99,10 +99,15 @@ print(stored.id, stored.spec.metrics)
 
 ### Task kinds
 
-A task is an evaluation unit; its `kind` says which runner executes it. Both kinds are stored as the
-same record type, so a taskset can group them and you manage every evaluation unit in one place.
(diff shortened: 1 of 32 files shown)


docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -73,47 +73,77 @@ Reference the stored metric with a `MetricRef` (`workspace/name`, or a bare `nam
...
 
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `spec` | `TaskDefinition` | Yes | The task's content, discriminated by `kind` — see below. |
+| `metadata` | `list[MetadataItem]` | No | Key/value annotations. Keys must be unique. |
+| `tags` | `list[str]` | No | Tags to point at the revision this request publishes. `latest` is always applied server-side. |
+
+### Task kinds
+
+A task is an evaluation unit; its `kind` says which runner executes it. Both kinds are stored as the
```
ngoncharenko: we mention "both kinds" before providing info about what are they

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -73,47 +73,77 @@ Reference the stored metric with a `MetricRef` (`workspace/name`, or a bare `nam
...
 | `metrics` | `list[MetricRefOrInline]` | No | The metrics that score the task, as `MetricRef` references (`workspace/name`) to stored metrics. Pre-built inline metric bundles (`MetricInline`) are also accepted and are normalized to stored metrics on create. |
 | `views` | `dict[str, SemanticView]` | No | Optional reporting views mapping metric outputs into named semantic scores. |
-| `metadata` | `list[MetadataItem]` | No | Key/value annotations. Keys must be unique. |
-| `tags` | `list[str]` | No | Tags to point at the revision this request publishes. `latest` is always applied server-side. |
+
+`HarborTaskDefinition` (`kind="harbor"`) — a Harbor task, scored by Harbor's own reward:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `archive_ref` | `str` | Yes | Files reference to the task's packaged directory (`workspace/fileset#path`). One fileset per task, so a task shared by several tasksets is stored once. |
```
ngoncharenko, line 124: nit: I think of archive as something that is stored and **not** used often. Not sure if that's what we mean here. Why not `task_ref`?

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
```diff
@@ -0,0 +1,308 @@
...
+    secrets: dict[str, SecretRef] = Field(
+        default_factory=dict, description="Secret references required to execute the metric."
+    )
+    payload: MetricPayload = Field(description="Format-specific serialized metric.")
+
+
+# An entity reference is ``name`` or ``workspace/name``, each segment using the platform name charset.
+# Shared by every ``workspace/name`` reference type (metrics, tasks). Enforced on the field so
+# empty/malformed refs are rejected at validation rather than during parsing.
+_ENTITY_REF_PATTERN = r"^[\w\-.]+(/[\w\-.]+)?$"
```
ngoncharenko: should we already have this in the code base somewhere? The same question for `FILESET_REF_PATTERN` below

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
```diff
@@ -0,0 +1,308 @@
...
+#: The tag applied to every publish and used when a ref carries no fragment.
+LATEST_TAG = "latest"
+# A *fileset* reference: ``workspace/fileset#path/inside.ext``. The fragment is a file path, so
+# unlike a revision fragment it admits ``/`` and ``.``. Validated on the field so a malformed
+# reference is rejected at publish rather than surfacing as a download failure mid-run — the same
+# reason a metric reference is checked when a task is stored.
+FILESET_REF_PATTERN = r"^[\w\-.]+/[\w\-.]+#[\w\-./]+$"
+
+
+def parse_entity_ref(root: str, default_workspace: str) -> tuple[str, str]:
```
ngoncharenko: should have this function already

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
```diff
@@ -0,0 +1,308 @@
...
+    )
+
+
+class MetadataItem(BaseModel):
+    """A single key/value annotation on a task."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    key: str = Field(description="Annotation key.")
+    value: str = Field(description="Annotation value.")
```
ngoncharenko, line 239: what is the value of this compared to dict?

ngoncharenko (approved):

plugins/nemo-evaluator/src/nemo_evaluator/entities.py
```diff
@@ -209,27 +240,23 @@ class _RevisionedCommon(BaseModel):
...
-    metrics: list[MetricRef] = Field(
-        default_factory=list,
-        description="References to the metrics that score this task. Inline metrics submitted with the "
-        "task are normalized to (derived) stored metrics, so a persisted task only ever holds refs.",
-    )
-    views: dict[str, SemanticView] = Field(
-        default_factory=dict,
-        description="Optional reporting views mapping this task's metric outputs into named semantic scores.",
-    )
+    spec: TaskDefinition = Field(description="The task's content, discriminated by which runner executes it.")
```
ngoncharenko, line 259: **[P1] Handle existing flattened task records**

Every task persisted by `origin/main` has top-level `intent`, `inputs`, `metrics`, and `views`, with no `spec`. The entity client validates stored data against this model, so after upgrading, existing task reads and lists fail because `spec` is required; historical revisions and taskset expansion fail similarly. Add a migration or legacy deserialization path, accounting for existing revision hashes and taskset digest pins, before requiring this shape.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
```diff
@@ -5,23 +5,73 @@
...
+)
+from nemo_evaluator.api.fields import (
+    REF_FRAGMENT_SEPARATOR as REF_FRAGMENT_SEPARATOR,
+)
+from nemo_evaluator.api.fields import (
+    MetadataItem as MetadataItem,
+)
+from nemo_evaluator.api.fields import (
+    MetricInline as MetricInline,
+)
```
ngoncharenko, line 32: **[P2] Restore the moved metric-payload exports**

Before this refactor, `CloudpickleMetricPayload`, `InlineMetricPayload`, and `MetricPayload` were importable from `nemo_evaluator.api.schemas`. They moved to `fields.py`, whose module documentation promises compatibility re-exports, but this import block omits them, so existing callers now get `ImportError`. Re-export all three here alongside `MetricInline`.

docs/evaluator/manage-tasks-tasksets.mdx
```diff
@@ -73,47 +73,77 @@ Reference the stored metric with a `MetricRef` (`workspace/name`, or a bare `nam
...
 stored = tasks.create("capital-of-france", task=task)
-print(stored.id, stored.metrics)
+print(stored.id, stored.spec.metrics)
 ```
 
 ### `TaskInput` fields
 
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `spec` | `TaskDefinition` | Yes | The task's content, discriminated by `kind` — see below. |
```
ngoncharenko, line 96: **[P2] Update the revision examples to the new spec shape**

This snippet still passes removed top-level arguments to `TaskInput`; `make docs-check-python-snippets DOCS_PATH=docs/evaluator/manage-tasks-tasksets.mdx` fails with a missing `spec` and unknown `intent`, `inputs`, and `metrics` arguments. Wrap these fields in `spec=EvaluatorTaskDefinition(kind='evaluator', ...)`, and update the later `original.inputs`/`current.inputs` accesses to `.spec.inputs`. The repository requires this snippet check in [`docs/AGENTS.md`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/fdd6c679ecfe729e0452e00fd39266e4385882e5/docs/AGENTS.md#L14-L16).

ngoncharenko (commented):
Manual review against origin/main: three actionable issues.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Reordered — both kinds are now named up front, before the sentence that refers to them collectively.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Keeping `archive_ref` — it points at the packaged directory and pairs with `archive_digest`, whereas `TaskRef` already means a `workspace/name` reference to a stored task, so `task_ref` here would name the wrong thing.

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
SandyChapman: Good catch — both now live in `nemo_platform_plugin.refs` next to `parse_entity_ref`/`FilesetRef`, and `_SUBENTITY_REF_PATTERN` is spliced from the shared constant so the two shapes can't drift.

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
SandyChapman: You're right, it was a straight duplicate — deleted, and `parse_subentity_ref` now delegates to `nemo_platform_plugin.refs.parse_entity_ref` (which ~10 services already use) and just adds the revision fragment.

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
SandyChapman: A dict can't reject duplicate keys as a validation error, can't carry per-key docs into OpenAPI, and reorders freely — which would perturb the revision digest — so keeping the model.

plugins/nemo-evaluator/src/nemo_evaluator/entities.py
SandyChapman: Keeping the breaking change as documented, but you're right it needs a plan — filed AALGO-496, where the hard part is digest continuity: nesting under `spec` changes the hash, so migrated revisions would re-derive new digests and dangle every existing taskset pin.

plugins/nemo-evaluator/src/nemo_evaluator/api/schemas.py
SandyChapman: Fixed — all three re-exported from `api.schemas` again, honoring what `fields.py` promises in its module docstring.

docs/evaluator/manage-tasks-tasksets.mdx
SandyChapman: Fixed, and the reason it slipped through: the snippet linter was passing ty a rule name that was renamed upstream, so it failed for *every* doc regardless of content — corrected that too, and added an integration test that runs the doc walkthrough against a real platform.

PR #1081, opened 2026-08-04
fix(auth): role binding lookup across all workspaces

ironcommit:
## Summary by CodeRabbit

- **New Features**
  - Added reliable single-entity lookup using names or filters.
  - IAM role bindings can now be retrieved and revoked across workspaces.
  - Added clear responses when no matching role binding exists or duplicate names are found.

- **Bug Fixes**
  - Improved role-binding operations in non-default workspaces.
  - Ensured revoked role bindings remain accessible for verification.
  - Clarified lookup behavior when multiple matching entities are available.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
@@ -312,9 +312,10 @@ class EntityNotFoundError(EntityStoreError):
 class EntityConflictError(EntityStoreError):
     """Entity conflict error.
 
-    Raised in two scenarios:
+    Raised when an entity operation conflicts with the current store state:
     - Entity already exists (conflict on create)
     - Entity version mismatch (optimistic locking conflict on update)
+    - Query expected one result but matched multiple entities
     """
 
     pass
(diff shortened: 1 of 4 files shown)


anastasia-nesterenko (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
```diff
@@ -585,6 +585,49 @@ async def count_by(
...
+        """Find exactly one entity matching a query.
+
+        This is intentionally query-based, not identity-based. It is useful for
+        cross-workspace lookups where the caller expects a field such as ``name``
+        to be globally unique by convention even though the entity store only
+        enforces workspace-scoped uniqueness.
+
+        Raises:
+            EntityNotFoundError: No entity matches the query
+            EntityConflictError: More than one entity matches the query
```
tylersbray, line 610: > **agent** (`pr-review`): EntityConflictError docstring drift

Nit: `find_one` introduces a third `EntityConflictError` scenario (ambiguous query), but the class docstring at ~L315 still documents only two:

- Entity already exists (conflict on create)
- Entity version mismatch (optimistic locking conflict on update)

Worth a one-line update on the class docstring so callers don't assume 409 always means create/update conflict.

packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py
```diff
@@ -585,6 +585,49 @@ async def count_by(
...
+        filter_str: Optional[str] = None,
+        filter_obj: Optional[Dict[str, Any]] = None,
+        sort: Optional[str] = None,
+    ) -> EntityT:
+        """Find exactly one entity matching a query.
+
+        This is intentionally query-based, not identity-based. It is useful for
+        cross-workspace lookups where the caller expects a field such as ``name``
+        to be globally unique by convention even though the entity store only
+        enforces workspace-scoped uniqueness.
```
tylersbray, line 604: > **agent** (`pr-review`): Overlap with get_by_field

Nit / optional follow-up: `get_by_field` (below in this file) still does `page_size=1` and returns the first match with no ambiguity check. `find_one` is the safer uniqueness-by-convention API.

Not blocking for this PR, but it would help future callers if we briefly document when to prefer `find_one` vs `get_by_field`, or eventually have `get_by_field` delegate to `find_one` when uniqueness is expected.

services/core/auth/tests/integration/test_iam_role_bindings.py
```diff
@@ -86,6 +86,52 @@ def test_role_binding_crud_lifecycle(self, http_client: TestClient):
         revoked = response.json()
         assert revoked["revoked_at"] is not None
 
+    def test_role_binding_crud_lifecycle_non_default_workspace(self, http_client: TestClient):
+        """Test role binding CRUD by name outside the default workspace."""
+        headers = {"X-NMP-Principal-Id": SERVICE_PRINCIPAL}
+        test_principal = f"system-user-{uuid.uuid4().hex[:8]}@example.com"
+        test_workspace = "system"
```
tylersbray: > **agent** (`pr-review`): Consider covering a custom workspace too

Nice regression coverage for `system` (the important `workspace: null` → `system` case). Optional polish: parametrize this over an additional arbitrary workspace (e.g. `"custom-ws"`) so get/revoke-by-name is exercised outside both `default` and `system`. Same code path, so not required for confidence.

tylersbray (approved):
Dredged up three non-blocking issues.  Also watch for the DCO.

PR #1090, opened 2026-08-05
feat(anonymizer): advertise the preview num_records maximum [ASTD-356]

marcusds:
Puts the anonymizer preview's `num_records` ceiling into the API schema, so clients stop having to guess it.

Closes [ASTD-356](https://linear.app/nvidia/issue/ASTD-356/express-the-anonymizer-preview-num-records-bound-in-the-api-schema).

## The problem

The preview endpoint has always capped `num_records`, but only imperatively:

```python
# functions/preview.py:168-178
if requested_num_records > config.preview_num_records.max:
    raise HTTPException(status_code=422, detail=f"Max num records for preview requests is {...}")
```

The field itself declared just a lower bound (`Field(default=10, ge=1)`), so the maximum never reached `openapi.yaml`, the generated `PreviewRequest.ts`, or the zod validator. A client that trusted the schema sent `num_records: 500` and got a 422 it had no way to anticipate. ASTD-332 worked around this by hardcoding `MAX_PREVIEW_ROWS = 10` in Studio — duplicating a server-owned value.

## Why `json_schema_extra` and not `le=`

`le=10` would be simpler and self-enforcing, and it's wrong here. The effective ceiling is `preview_num_records.max`, which an operator can raise via `NEMO_ANONYMIZER_PREVIEW_NUM_RECORDS_MAX` or `platformConfig.anonymizer`. Pydantic validates before the handler runs, so a static `le=` would reject 11 on a server configured to accept 50 — quietly capping the knob at its default and turning the config into a lie.

Building the constraint from `get_config()` at import time was the other option and is worse: `make refresh-openapi` output would depend on local env, so anyone with that variable set would commit spec drift.

So the annotation states the *default* bound — what a client can know without asking the server — and runtime enforcement stays where it belongs. The accepted cost is that a deployment which raises the max advertises a stale `maximum: 10`. Strictly better than advertising nothing.

## What lands

`DEFAULT_MAX_PREVIEW_NUM_RECORDS` is now the single source for both the `PreviewNumRecords` default and the field annotation, so they can't drift.

The spec regenerated to exactly one added line (`maximum: 10.0`) — no incidental churn. `web/packages/sdk/generated/` is gitignored, so there's nothing to commit for it, but `pnpm gen` confirms it picks the bound up: `PreviewRequest.ts` gains `@maximum 10` and the zod schema emits `anonymizerPreviewfunctionRouteBodyNumRecordsMax = 10` with `.max(...)`. Studio inherits it on the next regen.

Two tests: one asserts the schema reports `maximum`, the other constructs a `PreviewRequest` with 11 and asserts it *succeeds* — that one is the real guard, since it fails the moment someone "tidies" this into `le=`.

## Verification

`ruff check` and `ruff format --check` clean. Plugin unit tests pass (12/12 across `test_task_config.py`, `test_config.py`, `test_preview_function.py`). `ty` reports 7 diagnostics in the plugin — identical before and after this change (verified by stashing), all pre-existing on `main`.

## Summary by CodeRabbit

* **New Features**
  * Preview requests now advertise a maximum of 10 records in the API schema.
  * The default preview record limit is consistently set to 10.

* **Bug Fixes**
  * Preview request validation now correctly supports higher record counts when deployment-specific limits allow them.
  * Preview requests continue to enforce a minimum of 1 record.

### DIFF
--- a/plugins/nemo-anonymizer/openapi/openapi.yaml
+++ b/plugins/nemo-anonymizer/openapi/openapi.yaml
@@ -1301,6 +1301,7 @@ components:
           $ref: '#/components/schemas/SelectedModelsOverrides'
         num_records:
           type: integer
+          maximum: 10.0
           minimum: 1.0
           title: Num Records
           default: 10
--- a/plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/app/task_config.py
+++ b/plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/app/task_config.py
@@ -12,6 +12,7 @@ from anonymizer.config.anonymizer_config import AnonymizerConfig
 from anonymizer.config.replace_strategies import Annotate, Hash, Redact, ReplaceMethodBase, Substitute
 from nemo_anonymizer_plugin.app.input import AnonymizerInputSpec
 from nemo_anonymizer_plugin.app.model_configs import SelectedModelsOverrides
+from nemo_anonymizer_plugin.config import DEFAULT_MAX_PREVIEW_NUM_RECORDS
 from pydantic import BaseModel, Field, field_serializer
 
 _REPLACE_METHOD_KINDS: dict[type[ReplaceMethodBase], str] = {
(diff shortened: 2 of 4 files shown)


marcusds:
Note: you need access to a gliner model to run this properly, build.nvidia.com recently deprecated it, so it needs to be self hosted.

lipikaramaswamy (approved):
Thanks for sharing. I checked this from the standalone Anonymizer side and don’t see any changes needed there; the change is contained within the NeMo Platform plugin. I’ll defer the OpenAPI/configuration design review to the platform maintainers.

mckornfield (approved):

PR #1091, opened 2026-08-05
chore: remove unused perl packages and bump cryptography to 50.x

ironcommit:
## Summary

Reduces CVE surface in container images by removing unused Perl packages, upgrading `cryptography` to 50.x, and trimming unused Go dependencies in the jobs launcher.

### Perl removal
- **`Dockerfile.auditor-tasks`**: Inline purge of `perl-base`, `perl`, `perl-modules-*`, and `libperl*` with build-time assertions that `perl` is no longer on `PATH` and no CVE-tracked Perl packages remain installed.
- **`docker/scripts/cve-cleanup.sh`**: Added reusable Perl CVE cleanup block that removes Perl packages, verifies removal, and fails the build if any remain. Intended for images that source this script during build.

### `cryptography` 48 → 50
- Bumped `cryptography` constraint from `>=48.0.1,<49` to `>=50.0.0,<51` across:
  - `pyproject.toml` (constraint-dependencies + override-dependencies)
  - `Dockerfile.auditor-tasks`
  - `Dockerfile.safe-synthesizer-tasks`
  - `Dockerfile.nmp-automodel-base`
  - `plugins/nemo-safe-synthesizer/constraints.txt`
- Addresses GHSA-g6cj-pr64-35w5.
- `uv.lock` updated accordingly.

### Jobs launcher dependency cleanup
- Bumped Go toolchain from 1.25.8 to 1.25.12.
- Removed `go.opentelemetry.io/contrib/exporters/autoexport` and all transitive Prometheus/gRPC exporter dependencies that were no longer needed.
- Upgraded `google.golang.org/grpc` from 1.81.1 to 1.83.0 and `golang.org/x/text` from 0.38.0 to 0.39.0.
- `go.sum` updated accordingly.


docker/scripts/cve-cleanup.sh
```diff
@@ -27,6 +27,33 @@ for venv in /app/.venv /opt/venv; do
     fi
 done
 
+# Remove only the Perl packages currently tracked with CVEs. Apt will also
```
mckornfield: "tracked with CVEs" sounds weird

will any of our python libs break because because they rely on perl?

mckornfield (approved):

docker/scripts/cve-cleanup.sh
ironcommit: updated

PR #1098, opened 2026-08-05
fix(auditor): isolate garak cache per pytest-xdist worker

ironcommit:
When running tests in parallel with pytest-xdist, multiple workers share the same XDG_CACHE_HOME, causing garak to read and write cached data concurrently and producing flaky test failures.

Set a worker-specific XDG_CACHE_HOME (scoped by test run UID and worker ID) before any imports so that each xdist worker gets its own isolated cache directory. A no-op env var (NMP_PYTEST_XDIST_CACHE_HOME_ISOLATED) prevents nested conftest files from re-applying the override. Adds unit tests for the path-derivation logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when running tests in parallel by isolating cache data for each worker.
  * Prevented cache-related conflicts during test collection.
  * Preserved existing cache behavior for standard, non-parallel test runs.

* **Tests**
  * Added coverage for worker-specific cache directory paths.
  * Added coverage confirming isolation is disabled when parallel execution is inactive or already configured.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -12,14 +12,21 @@ files for more specific fixtures.
 import logging
 import os
 import tempfile
+from collections.abc import Generator
 from pathlib import Path
-from typing import Generator
 
-import pytest
-from nmp.testing.pytest_outcomes import pytest_skip as skip_test
+from tests.xdist_cache import xdist_worker_xdg_cache_home
 
-from tests.auth_idp.xdist import append_xdist_group_suffix
-from tests.discovery_exclusions import TEST_DISCOVERY_EXCLUSIONS
+_xdist_cache_home = xdist_worker_xdg_cache_home(os.environ)
+if _xdist_cache_home is not None:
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -12,14 +12,41 @@
...
+from collections.abc import Generator, Mapping
 from pathlib import Path
-from typing import Generator
 
-import pytest
-from nmp.testing.pytest_outcomes import pytest_skip as skip_test
 
-from tests.auth_idp.xdist import append_xdist_group_suffix
-from tests.discovery_exclusions import TEST_DISCOVERY_EXCLUSIONS
+def _xdist_worker_xdg_cache_home(environ: Mapping[str, str]) -> str | None:
```
mckornfield: do we want this just for Garak tests or everything? i guess in case similar issues come up

mckornfield (approved):

PR #1099, opened 2026-08-05
docs(experimentalist): point the nooa pin at the file that actually has it

callingmedic911:
the nooa framework skill tells you to use the revision "pinned in `pyproject.toml`," but `plugins/nemo-experimentalist/pyproject.toml` declares a bare `nooa` with no revision.

The pin is in the **workspace root** `pyproject.toml` under `[tool.uv.sources]`, which is where uv resolves workspace members from, and `uv.lock` pins the commit accordingly. So this is a docs pointer bug, not a missing pin. 

## Summary by CodeRabbit

- **Documentation**
  - Updated NOOA dependency guidance to reference the workspace-root pinned revision.
  - Documented support for callable `@strategy(llm=...)` usage with newer NOOA revisions.
  - Clarified which related files must be updated together and where to review compatibility details.
  - Removed outdated references to the previous version and transport-timeout fix.


gabwow (approved):
LGTM

PR #1100, opened 2026-08-05
perf(studio): code-split the entry chunk (1,641 kB -> 326 kB)

marcusds:
## Summary

Studio's entry chunk had grown to **1,641 kB (gzip 535 kB)**. This splits it down to **326 kB (gzip 76 kB)** — an 80% reduction in raw bytes, 86% gzipped.

| chunk | before | after |
| --- | --- | --- |
| `index-*.js` | 1,641.04 kB / gzip 535.38 kB | **326.38 kB / gzip 75.84 kB** |

## Why it was big

Two static imports in the app shell — which renders on every route — were pulling the entire NeMo Copilot chat surface into the entry chunk:

```
GlobalNav
  -> ClaudeCodeTopBarChat
    -> ClaudeCodeChatThread
      -> @assistant-ui/*, remark-gfm -> micromark stack
      -> ClaudeCodeToolCallPart -> Chat/MessageContent -> MarkdownDataViewTable
           -> DataView/internal -> @tanstack/table-core, @dnd-kit/core, date-fns
      -> BlockingInputComposer -> AgentBlockingInput
           -> DatasetFileSelect -> FileContentPreview -> CodeEditor
                -> @codemirror/*, @lezer/{common,javascript,python}, yaml, papaparse
```

`ClaudeCodeChatRoute` was already `lazy()`, but that made no difference — everything it needed had already been hoisted into the entry chunk by the nav's static import. The route chunk was 2.2 kB.

Separately, `main.tsx` imported `@studio/telemetry/telemetry` as a top-level side effect, putting the OpenTelemetry SDK + `zone.js` (~162 kB) on the critical path.

## Changes

**`ClaudeCodeTopBarChat.tsx`** — `lazy()` the chat thread, gated on a `hasOpened` latch so the chunk fetches on first pop-out open and stays mounted after (matching the previous always-mounted behaviour). The trigger button and its thinking/unread badges stay synchronous. Warmed on hover so the first open is instant.

**`FileContentPreview`** — `lazy()` the `CodeEditor` behind a `Suspense` spinner. Moves 602 kB off the boot path.

**`CodeEditor/constants.ts`** — `import type { BasicSetupOptions }`. It was a value import, so importing `ContentType` alone still dragged in `@uiw/react-codemirror`; without this the split above wouldn't hold.

**`CodeEditor/linters/yaml.ts`** — `await import('yaml')` inside the lint source (`linter` accepts an async `LintSource`). Splits 103 kB out of the editor chunk into its own.

**`main.tsx`** — telemetry moved from a top-level side-effect import to a fired-not-awaited `import()`, awaited in `Promise.all` alongside `waitForThemeStylesheet()` before `root.render`. OTel still patches `fetch`/XHR before the first app request; the 140 kB now downloads in parallel instead of inflating the entry chunk.

## Resulting chunk layout

```
index-*.js                326.38 kB │ gzip:  75.84 kB   (was 1,641.04 / 535.38)
CodeEditor-*.js           602.50 kB │ gzip: 209.00 kB   lazy, on file preview
telemetry-*.js            140.22 kB │ gzip:  43.18 kB   parallel, pre-render
browser-*.js (yaml)       103.05 kB │ gzip:  31.54 kB   lazy, on first YAML lint
ClaudeCodeChatThread-*.js  32.45 kB │ gzip:   9.83 kB   lazy, on first chat open
```

## Measured load impact

Built `main` at the branch point and this PR head, served each `dist` from a local static server, loaded cold-cache in headless Chromium. Median of 9 runs after a discarded warm-up; backend requests aborted so both variants see identical (zero) API latency.

**Desktop, no throttling**

| | before | after | Δ |
| --- | --- | --- | --- |
| FCP | 124 ms | **96 ms** | −23% |
| LCP | 124 ms | **96 ms** | −23% |
| DOMContentLoaded | 105 ms | **64 ms** | −39% |
| load | 106 ms | **65 ms** | −39% |
| ScriptDuration | 87 ms | 81 ms | −6% |
| JS files before load | 48 | **23** | −52% |
| JS bytes before load | 4,457 kB | **3,008 kB** | **−1,449 kB** |

**Throttled — 4× CPU, 1.6 Mbps / 150 ms RTT**

| | before | after | Δ |
| --- | --- | --- | --- |
| FCP | 25,628 ms | **18,956 ms** | **−6.7 s (−26%)** |
| LCP | 25,628 ms | **18,956 ms** | −26% |
| DOMContentLoaded | 25,568 ms | **17,978 ms** | −30% |
| load | 25,570 ms | **17,980 ms** | −30% |
| ScriptDuration | 289 ms | 273 ms | −6% |
| Long-task time | 182 ms | **109 ms** | −40% |
| JS bytes before load | 4,457 kB | **3,008 kB** | −1,449 kB |

The 1,449 kB drop in critical-path JS is the whole story. Script execution barely moves, which is expected — the deferred code was never *executing* at boot, it was being downloaded and parsed.

**Caveat:** the absolute throttled numbers are inflated — the measurement server is HTTP/1.1, so every request pays the full 150 ms RTT. The relative delta is the trustworthy part. Note that the *uncompressed* bytes may not be an artifact: see next steps, I could not find compression on the Studio static mount either.

Netting out the shared vendor bundles (2,379 kB, byte-identical in both builds), Studio's own critical-path JS goes **1,973 kB → 558 kB (−72%)**.

## Recommended next steps

After this PR the critical path is **24 resources / 3,488 kB uncompressed**, broken down as:

| resource | bytes | share |
| --- | --- | --- |
| `vendor/foundations.js` | 1,996 kB | 57% |
| `index.css` | 480 kB | 14% |
| `index.js` (entry) | 327 kB | 9% |
| `vendor/react-router.js` | 200 kB | 6% |
| `vendor/react-dom.js` | 185 kB | 5% |
| `oidc-client-ts.js` | 108 kB | 3% |
| zod / react-query / axios | 151 kB | 4% |
| 15 smaller chunks | ~41 kB | 1% |

Ordered by measured impact:

**1. Serve Studio's assets compressed.** I could not find any compression on the static mount — no `GZipMiddleware` (or equivalent) in `services/studio/src/nmp/studio/service.py`, which mounts `SPAStaticFiles` (a plain Starlette `StaticFiles` subclass), and no gzip/brotli configuration in `k8s/helm`. If that's accurate for deployed environments too, the whole 3,488 kB critical path is going over the wire uncompressed. The chunks gzip at roughly 4:1 (the entry chunk is 326 kB → 76 kB), so this is on the order of a **~2.5 MB reduction for a few lines of middleware** — a bigger win than this entire PR, and much cheaper. Worth confirming against whatever ingress fronts a real deployment before acting.

**2. `vendor/foundations.js` — 1,996 kB, 57% of what's left.** The vendor shim in `vite.config.ts` is a blanket `export * from '@nvidia/foundations-react-core'` built with `codeSplitting: false`, so the entire design system ships regardless of how much of it Studio actually renders, and nothing tree-shakes. The blanket re-export exists for a good reason — the bundle has to satisfy the union of what Studio *and* every runtime-loaded plugin imports, and there must be exactly one instance. But that union could be generated from the components actually imported across `packages/studio`, `packages/common` and `plugins/*/web` rather than assumed to be everything. Biggest single lever left.

**3. `index.css` — 480 kB, render-blocking.** Emitted by the Tailwind build. Worth checking whether the `content` globs are over-broad (the postcss config points at all of `packages/`) and whether KUI's full stylesheet is being inlined alongside Studio's own utilities.

**4. The generated SDK barrel — 142 kB, 44% of the remaining entry chunk.** `packages/sdk/generated/platform/api.ts` is a single orval-generated barrel pulled in eagerly. Per-tag entry points would let each route import only the operations it uses.

**5. `oidc-client-ts` — 108 kB, `modulepreload`ed in `index.html`.** Eager at boot. Worth checking whether it's genuinely needed before first paint or only on the auth/silent-renew path.

**6. `ClaudeCodeChatProvider` — ~90 kB, mostly `@assistant-ui/core`.** Still eager in `PageLayout`; see the note below on why it wasn't done here.

**7. Add a size budget to CI.** `build.chunkSizeWarningLimit` only warns and is easy to ignore. An assertion on the entry chunk's gzip size would stop this regressing back to 535 kB the next time something is imported into the app shell — which is exactly how it got there.

## Notes

- The second commit is a pass with the Vercel React best-practices rules. Worth calling out one finding: KUI's `PopoverTrigger` spreads `...props` **after** its own `onPointerEnter`, so passing `onPointerEnter` (or `onFocus`) to a `Popover` trigger silently replaces the trigger's handler and breaks opening the pop-out. The hover preload uses `onMouseEnter` for that reason — there are unit tests covering it.
- Not addressed: `ClaudeCodeChatProvider` is still eager in `PageLayout` (~90 kB, mostly `@assistant-ui/core`). Every no-remount way to defer it requires making `ClaudeCodeChatContextValue` nullable — `useClaudeCodeChatContext` currently throws on null. That's a design change, not a perf tweak, so it's left out.
- Everything else that turned up while measuring is in **Recommended next steps** above; none of it is in scope here.

## Testing

- `tsc --noEmit` clean
- `@nemo/common`: 1379/1379 pass (four `FileContentPreview` assertions moved to `findByTestId` for the now-suspended editor)
- `nemo-studio-ui`: full suite green except `SafeSynthesizerNewRoute`, which times out at 10 s under full-suite contention and passes on its own — unrelated to these files
- eslint + prettier clean
- Headless load of the production build: app renders, no module errors, `telemetry-*.js` loads, and neither `CodeEditor-*.js` nor `ClaudeCodeChatThread-*.js` is requested at boot

## Summary by CodeRabbit

* **Performance**
  * Code editors and compact chat now load on demand, improving initial application load times.
  * Chat resources begin loading when hovering over the chat control for a faster opening experience.

* **Bug Fixes**
  * Added clearer loading states for file previews and chat.
  * Copilot chat errors are isolated with a retry option instead of disrupting the surrounding interface.
  * Improved handling of YAML linting and application startup initialization.


steramae-nvidia (approved):

web/packages/studio/src/main.tsx
```diff
@@ -34,7 +34,7 @@ function waitForThemeStylesheet(): Promise<void> {
 
 const rootElement = document.getElementById('app')!;
 if (!rootElement.innerHTML) {
-  waitForThemeStylesheet().then(() => {
+  Promise.all([waitForThemeStylesheet(), telemetryReady]).then(() => {
```
aray12, line 40: I understand this reduces the initial bundle size but doesn't it just create a synchronous roundtrip that is required before first render?

aray12 (approved):

web/packages/studio/src/main.tsx
marcusds: Yeah but it's happening when we are already waiting for stylesheets so I don't think there would be much or any negative impact.

PR #1101, opened 2026-08-05
fix: set envoy upstream idle timeout

ironcommit:
## Summary

This centralizes the connection keep-alive fix in this PR. Before this change, Envoy could keep pooled upstream HTTP/1 connections around longer than the Uvicorn API kept idle sockets open, which allowed intermittent reuse of a closed backend connection. This PR makes the API keep-alive timeout explicit, sets Envoy's upstream idle timeout below it, and validates the rendered Envoy configs in chart-aware CI instead of general unit tests.

## Changes

- Adds a `keep_alive_timeout_seconds` platform runner setting with validation and propagates it through direct, socket-bound, reload, embedded, and background Uvicorn startup paths.
- Adds `--keep-alive-timeout-seconds` to `nemo services run`, `start`, and `restart`, preserving the previous value on restart unless explicitly overridden.
- Vendors the local service keep-alive changes into the generated Python SDK package and updates the CLI reference docs.
- Adds Helm values for API keep-alive and Envoy upstream idle timeout, with a Helm guard that requires Envoy's upstream idle timeout to be a positive whole-second value lower than the API keep-alive timeout.
- Updates the base Helm Envoy config and Authentik reference Envoy configs to use explicit HTTP/1 upstream options and a `4s` upstream idle timeout against the `5s` API keep-alive default.
- Moves Envoy config validation out of the unit suite and into Helm/auth-idp CI paths, using the deployment-specific Envoy images for Helm/Auth (`v1.37.0`) and the static Compose gateway (`v1.36.2`).

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

- `uv run --frozen pytest tests/auth_idp/static -v` — 101 passed
- `uv run --frozen pytest tests/unit/test_helm_clickhouse.py -v` — 6 passed
- `uv run --frozen pytest tests/auth_idp/static/test_envoy_config_validation.py tests/auth_idp/static/test_authentik_kubernetes_demo.py::test_authentik_static_ci_prepares_envoy_validation_inputs -v` — 3 passed
- `tools/lint/lint-helm.sh` — passed, including rendered Envoy validation
- `uv run --frozen ruff check tests/auth_idp/static/test_envoy_config_validation.py tests/auth_idp/static/test_authentik_kubernetes_demo.py` — passed
- `uv run --frozen ruff format --check tests/auth_idp/static/test_envoy_config_validation.py tests/auth_idp/static/test_authentik_kubernetes_demo.py` — passed
- `git diff --check` — passed
- `uv run pre-commit run -a` — passed

CI spot-check after the latest push:

- `Python auth-idp static tests` — passed on `d065da0626fd950b2ed627c86a6e9b7f8421a0ac`
- `Helm lint` — passed on `d065da0626fd950b2ed627c86a6e9b7f8421a0ac`
- `Helm chart verifier` — passed on `d065da0626fd950b2ed627c86a6e9b7f8421a0ac`

### DIFF
--- a/contrib/auth/authentik/gateway/envoy.yaml
+++ b/contrib/auth/authentik/gateway/envoy.yaml
@@ -251,6 +251,16 @@ static_resources:
     - name: nemo
       connect_timeout: 5s
       type: LOGICAL_DNS
+      # Envoy's default upstream HTTP idle timeout is 1h. Keep this below the
+      # API keep-alive timeout so stale pooled API connections are retired
+      # before the backend closes them.
+      typed_extension_protocol_options:
+        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+          "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+          common_http_protocol_options:
+            idle_timeout: "4s"
+          explicit_http_config:
+            http_protocol_options: {}
       load_assignment:
         cluster_name: nemo
         endpoints:
--- a/k8s/helm/values.yaml
+++ b/k8s/helm/values.yaml
@@ -647,6 +647,10 @@ api:
 
   # -- Number of replicas for the API service.
   replicaCount: 1
+  # -- Platform API server settings.
+  server:
+    # -- Seconds Uvicorn keeps idle HTTP connections open. Must be greater than envoyProxy.timeouts.upstreamIdle when Envoy is enabled.
+    keepAliveTimeoutSeconds: 5
   # -- Additional arguments to pass to the Platform API service
   extraArgs: []
   # -- Additional volume mounts to add to the Platform API container.
@@ -1053,6 +1057,8 @@ envoyProxy:
     route: "0s"
     # -- Cluster connect timeout (time to establish connection to backend).
     connect: "30s"
+    # -- Positive whole-second upstream connection idle timeout. Must be less than api.server.keepAliveTimeoutSeconds when Envoy is enabled.
+    upstreamIdle: "4s"
 
   # -- Kubernetes deployment resources configuration for the Envoy service. Utilization-based autoscaling requires a matching resource request.
   resources: {}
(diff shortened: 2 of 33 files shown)


contrib/auth/authentik/gateway/envoy.yaml
```diff
@@ -251,6 +251,16 @@ static_resources:
...
       connect_timeout: 5s
       type: LOGICAL_DNS
+      # Envoy's default upstream HTTP idle timeout is 1h. The chart
+      # default is 20s so stale pooled API connections are retired quickly
+      # if the backend closes keep-alive sockets earlier than Envoy would.
+      typed_extension_protocol_options:
+        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+          "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+          common_http_protocol_options:
+            idle_timeout: "20s"
```
mckornfield: lol this rings a bell...

k8s/helm/values.yaml
```diff
@@ -1047,6 +1047,8 @@ envoyProxy:
     route: "0s"
     # -- Cluster connect timeout (time to establish connection to backend).
     connect: "30s"
+    # -- Upstream connection-pool idle timeout. Envoy defaults this to 1h, which can outlive backend keep-alive sockets.
+    upstreamIdle: "20s"
```
mckornfield: is the other way to fix this to set a lease time or something for these connections? maybe that's exactly what this is doing? lol

mckornfield (approved):

k8s/helm/values.yaml
ironcommit: So the idea is that the envoy connection pool timeout needs to be less than the keepalive on the fastapi side.  That prevents the random 500s when the connection pool tries to reuse a connection that has been cleaned up by fastapi.  I tried to made that clear in the helm docs.

svvarom (approved):

PR #1120, opened 2026-08-05
fix: improve setup recovery diagnostics

mckornfield:
## Summary
- Report setup port conflicts with the port and `EADDRINUSE`, including when service startup exits early after a bind failure.
- Keep `nemo setup --resume` as a retry marker, but clarify it uses the normal idempotent setup path rather than a special recovery flow.
- Remove the `nemo setup --resume` example from command help so the loaded help matches the manual manifest entry.
- Vendor the setup/process changes into the SDK copy.

## NVBugs
- 6556551
- 6556559

## Validation
- `uv run --frozen pytest packages/nemo_platform_ext/tests/cli/test_app.py::test_manifest_help_matches_loaded_manual_entry packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestNonTtyEarlyExit::test_resume_runs_idempotent_setup_path -q`: 13 passed
- `uv run --frozen ruff check packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py packages/nemo_platform_ext/tests/cli/commands/test_setup.py sdk/python/nemo-platform/src/nemo_platform/cli/commands/setup.py`: passed
- `make vendor-nemo-platform-ext`
- `make generate-cli-reference-docs`
- Earlier focused setup lifecycle tests: 248 passed against source and vendored SDK copies
- Earlier live source and SDK port-conflict smoke with a temporary listener

## Review follow-up
- `0b355c100` tones down `--resume` help/banner text and fixes the manifest help mismatch that failed CI.
- `eeb5037c7` resolves the merge conflict against current main with a signed merge commit.

## Summary by CodeRabbit

* **New Features**
  * Added a `--resume` option to retry interrupted setup through the normal setup process.
  * Added clearer guidance when required ports are already in use.

* **Bug Fixes**
  * Port conflicts now show specific address-in-use diagnostics instead of generic startup failures.
  * Error messages include the `EADDRINUSE` code, affected-port details, and troubleshooting guidance.

* **Documentation**
  * Documented the new `--resume` setup option in the CLI reference.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
@@ -49,6 +49,7 @@ from nemo_platform_ext.client.tls import client_verify_from_env
 from nemo_platform_ext.config.config import Config
 from nemo_platform_ext.config.models import DEFAULT_BASE_URL, ConfigFile, ConfigParams, LocalServicesConfig, NoAuthUser
 from nemo_platform_ext.local.process import (
+    PortConflict,
     check_port_available_for_start,
     compute_scope,
     format_port_conflict,
@@ -997,6 +998,14 @@ _DOCKER_FAILURE_LOG_MARKERS = (
     "Error while fetching server API version",
 )
 
+_PORT_CONFLICT_LOG_MARKERS = (
+    "EADDRINUSE",
+    "address already in use",
+    "address is already in use",
+    "Errno 98",
+    "Only one usage of each socket address",
+)
+
 
 def _services_log_suggests_docker_failure(log_path: Path | None) -> bool:
     """Return True when services.log contains known Docker skip / daemon errors."""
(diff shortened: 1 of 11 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -941,6 +950,36 @@ def _services_log_suggests_docker_failure(log_path: Path | None) -> bool:
...
+        return False
+    try:
+        text = log_path.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return False
+    lower_text = text.lower()
+    return any(marker.lower() in lower_text for marker in _PORT_CONFLICT_LOG_MARKERS)
+
+
+def _detect_startup_port_conflict(base_url: str, log_path: Path | None) -> PortConflict | None:
```
mckornfield, line 1033: wonder how much of an edge case this is...

mckornfield (commented):
one thing

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
mckornfield: I kept this behavior as-is. It covers the startup race after preflight where the subprocess exits before readiness probing can succeed; port collision is a common cause, and surfacing the log tail makes the failure actionable without changing the happy path.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -2032,6 +2085,8 @@ def setup_command(
     base_url = cli_context.get_base_url() or DEFAULT_BASE_URL
 
     console.print("\n[bold cyan]NeMo Platform Setup[/bold cyan]\n")
+    if resume:
```
tylersbray, line 2175: > **agent** (`pr-review`): `--resume` is a no-op relative to its help text

This is the only use of `resume` — it prints a banner and then falls through the normal setup path. The `--resume` help ("reusing saved config and existing platform resources") implies special recovery behavior that isn't implemented here.

If setup is already idempotent, please either drop the flag (and document plain `nemo setup` / `--auto` as the resume path) or tone the help/examples so they don't claim special reuse logic. As written, the flag oversells what it does.

tylersbray (approved):
Here is your non-blocking approval.  Agentic review found one issue around --resume flag.  Fix CI please and thank you.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
mckornfield: Fixed in 0b355c100. I kept the flag for retrying interrupted setup, but changed the option help and banner to say it runs the normal idempotent setup path, and removed the `nemo setup --resume` example that implied a separate recovery flow.

PR #1125, opened 2026-08-06
fix(client): FileSet upload retries and timeout propagation

albcui:
## Summary

It was observed that a FileSet upload that hit a transient failure failed with `h11._util.LocalProtocolError: Too little data for declared Content-Length` error. However, this is a bit of a red herring, because it hides the underlying issue, which is a bit gnarly, so let's dig a bit deeper.

## Root cause

The core crux of the issue is in `filesystem.py`'s `FilesetFileSystem._put_file`, which implements `fsspec`. 

```python
file_size =  (await anyio.Path(lpath).stat()).st_size

async def stream_file() -> AsyncIterator[bytes]:
    async with await anyio.open_file(lpath, "rb") as f:
        while chunk := await f.read(self.blocksize):
            callback.relative_update(len(chunk))
            yield chunk

await self._client.with_headers({"Content-Length": str(file_size)}).upload_file(
    ...
    content=stream_file(),
)
```

The `stream_file()` is a generator object that gets called and stored once in `PreparedRequest.content` and reused by every pass of `NemoClient`'s retry loop. If any transient failures occur, then the same `Content-Length` followed by zero bytes (the generator is already exhausted) is sent, which breaks the HTTP/1.1 protocol, and we get a `httpx.LocalProtocolError`, which subclasses `TransportError`, which gets retried until attempts ran out. This is bad because we actually lose the information for the original error.

**So what could've caused the original failure?**

Let's start with entrypoint from Customizer's `file_io` taks:

```python
UPLOAD_TIMEOUT = httpx.Timeout(30.0, write=10 * 60, read=5 * 60)
self.sdk.with_options(timeout=UPLOAD_TIMEOUT).files.upload(...)
```

The expected behaviour is that `UPLOAD_TIMEOUT` would propagate down to `files.upload`, which should extend the write timeout to a generous 600s. However, this is not the case. What actually happened was:
- `with_options(...)` created a new `NeMoPlatform` client with the new timeout based on `UPLOAD_TIMEOUT`, but keeps the _original_ `httpx.Client`, which has its own timeout (60s). So now there are two timeouts:
  - `NeMoPlatform.timeout = Timeout(30, write=600, read=300)`
  - `NeMoPlatform._client.timeout = Timeout(60, ...)` 
- then `client_from_platform` shares the `NeMoPlatform._client`. The sharing is not wrong by itself, it's the fact that we didn't propagate the `NeMoPlatform.timeout` in the first place.
- the issue is made worse by the fact that downstream,`FilesetFileSystem._ensure_async` built a brand new `https.AsyncClient` with no timeout at all, which defaults to `5s`.

Essentially, `FilesetFileSystem._put_file`'s httpx client had a timeout of 5s (on both writes and reads). This could error for both scenarios:
1. Streaming all the chunks of a large file took longer than 5s, times out on the write phase
2. Even if we streamed all the chunks, the Files service might not respond within 5s because it needs to commit all the chunks to storage, triggering a timeout

The fix involved changes in multiple layers of the code:

- `client/client.py` -- don't retry requests with generator bodies, at least not at this layer of the code
- `tasks/file_io/run.py` -- retry here instead, since it can rebuild the generator
- `client/adapter.py` + `filesystem/filesystem.py` -- propagate the timeout properly
  - Adapter applies `client.with_options(timeout=platform.timeout)`
  - `ensure_async` propagates the timeout onto both the new `https.AsyncClient` as well as the `AsyncFilesClient`

## Bug Fixes

* **Bug Fixes**
  * Client timeout settings are now preserved across synchronous, asynchronous, and filesystem operations, including unlimited-timeout configurations.
  * Retry behavior now avoids replaying partially consumed streaming uploads.
  * File operations retry more reliably for temporary connection, rate-limit, and server errors.
  * Upload failures now surface consistently after retry attempts are exhausted.
  * Fileset metadata is validated consistently during creation and updates.
  * File listings now provide consistent file details.

* **Documentation**
  * Clarified retry behavior for one-time request bodies and streaming uploads.


ironcommit (approved):
LGTM.  This works around our very messy transport system.  But it will do.

packages/filesets/src/filesets/filesystem/filesystem.py
```diff
@@ -358,20 +358,27 @@ def _ensure_async(client: FilesClient | AsyncFilesClient) -> AsyncFilesClient:
 
         import httpx
 
+        # Carry the sync client's timeout onto the new transport. Left unset,
+        # httpx applies its 5s default to every phase, which a multi-GB upload
+        # blows through waiting for the server to commit the body to storage.
+        timeout = client._timeout if client._timeout is not None else client._http.timeout
```
matthewgrossman: this is a rough part, having both the high-level client have a `_timeout` as well as the `client._http` having a timeout. We should really only be using one field as the source of truth

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/adapter.py
```diff
@@ -72,21 +72,35 @@ def client_from_platform(
...
+    # ``platform.with_options(timeout=...)`` would otherwise be silently dropped
+    # on the way to the typed client — the httpx client it copies is the *same*
+    # object, with the *original* timeout still on it.
+    timeout = platform.timeout
+    if timeout is None:
+        # ``None`` on the platform means "no timeout at all", but ``with_options``
+        # reads None as "leave unchanged". Say the same thing in the form httpx
+        # itself uses, so the override survives.
+        timeout = httpx.Timeout(None)
+    return client.with_options(timeout=timeout)
```
matthewgrossman: this feels like a smell that we have to mutate the client we _just created_. Why can't we just properly thread in the timeout field into the client_cls constructors above?

matthewgrossman (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/adapter.py
albcui: yup that's cleaner for sure

packages/filesets/src/filesets/filesystem/filesystem.py
albcui: Yea... I just assumed that the higher `_timeout` is the override to the inner client's timeout. However, I'm looking forward to @ironcommit's refactor to clean this up.

PR #1127, opened 2026-08-06
feat(experimentalist): support ATIF agent traces

schuellc-nvidia:
## Summary

The Experimentalist read agent traces as OTLP only, so an agent emitting ATIF produced trials with no trace at all — silently. Discovery found no `*.jsonl`, the trial was skipped, and nothing said why. This teaches the trace path to carry its format from discovery through analysis, so an ATIF-emitting agent can be optimized. OTLP remains the default and its behavior is unchanged.

## Changes

**ATIF trace support** (`feat` commit)

- **Discovery** — `_trial_resources` recognises `*.atif.json` alongside `*.jsonl` and stamps `trace_format` on the `ResourceRef`, so the format is decided once and carried forward rather than re-derived downstream. New `HarborEvaluatorConfig.trace_format` selects which artifact becomes the trial's trace; both stay exposed in `resources`.
- **Upload** — new `experimentalist/atif.py` reads the trajectory and builds the ingest payload; `_persist_trial` dispatches on the stamped format and posts to Intake's ATIF ingest with `evaluation_context` carrying experiment and test-case identity. Trial identity rides on the trajectory's own `session_id`, which Intake adopts as the trace id.
- **Read-back** — Intake promotes known attributes into typed columns and strips those keys from `raw_attributes`, so a reader that knows only the OpenInference names finds nothing. Token counts and error fields were missing on **every** span read back from storage, OTLP included. `from_intake` now parses the columns directly; `_load_trace_models` keeps a per-record shape check because it deliberately accepts Intake-shaped payloads from disk. An ATIF branch keyed on the `source` column covers fields that have no column at all.
- **Three silent failures become diagnosable** — discovery warns when the configured format matches nothing but the other format's artifacts are present; `persist_evaluation` logs the trial it drops; `from_ref` rejects a local `.atif.json` with a message naming the real cause instead of reporting an empty trace.
- **Docs** — a README section covering the `trace_format` setting, the artifact contract, the Intake requirement, and the two diagnostics.

**Independent of ATIF**, kept as separate commits so they can be reviewed or cherry-picked on their own:

- `fix(tau3-nooa-agent)` — when the agent dies before writing traces, `/app/traces` never exists and the environment refuses to copy it, raising `RuntimeError`. That escaped the metrics guard and replaced the agent's own error with a Docker copy failure. Catching it lets the run reach the existing check that raises with the agent's exit code and stderr.
- `chore(experimentalist)` — ignore `artifacts/experimentalist-benchmarks/`. It is generated output, and because the optimizer writes candidate agents there, a candidate whose generated code does not compile was failing `ruff check plugins/nemo-experimentalist/`.

**Review fixes** (`fix(experimentalist)` commit, from CodeRabbit)

- `tool_call_id` has no Intake column, so `tool_call.id` / `tool.id` / `tool_call_id` survive in `raw_attributes`. The new Intake parser hardcoded an empty value for non-ATIF rows, so OTLP traces read back from Intake lost their tool call ids — a regression against the shared parser it replaced. Restores the attribute lookup, with ATIF's preserved payload as the fallback only for ATIF rows, and adds a regression test.
- The README documented `trace_format` under an `optimizer` key. That is the benchmark config's shape; an agent profile roots the same model at `experiment_config`, where `evaluator` is a top-level field. Corrected.

## Type of Change

- [ ] Documentation only

## Quality Gates

31 tests added across four files: `test_atif.py` (payload construction), `test_evaluator_harbor.py` (format discovery and selection), `test_experimentalist_backend.py` (ATIF upload), `test_trace_explorer_from_ref.py` (Intake-row parsing, both backends). The `fix` commit's 3 tests were confirmed to fail without the fix and pass with it.

## Verification

Targeted validation:

```
uv run pre-commit run -a
  → all hooks Passed (ruff, ruff format, ty typechecks, config reference,
    helm docs, uv lock, uv lock drift, copyright headers, UI lint-staged,
    plugin import boundary, merge conflicts)

uv run --frozen pytest <the 5 test files this PR touches> -q
  → 142 passed

uv run --frozen pytest plugins/nemo-experimentalist/tests/ -q
  → 631 passed

uv run ruff check plugins/nemo-experimentalist/
  → All checks passed

uv run ruff format --check
  → 3030 files already formatted

git diff --check origin/main...HEAD
  → clean
```

**End-to-end verification against a live platform.** Beyond the unit tests, the ATIF path was exercised with a real agent and a running Intake: trajectories were accepted by Intake's validators, 7 traces / 86 spans landed with correct `evaluation_id` and `test_case_id` grouping, and spans read back with token counts and tool call ids populated. The agent-side scaffolding used for that run was reverted before this PR; only the feature remains.

**Known limitation.** Per-step durations are absent on ATIF-derived spans: Intake takes a step's end time only from a producer-supplied invocation block, and infers nothing from adjacent steps. Trajectory-level duration and the Studio rollup are unaffected. This is documented in the README.

### DIFF
--- a/plugins/nemo-experimentalist/artifacts/.gitignore
+++ b/plugins/nemo-experimentalist/artifacts/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore
--- a/plugins/nemo-experimentalist/artifacts/.gitignore
+++ b/plugins/nemo-experimentalist/artifacts/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/evaluator/harbor.py
+++ b/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/evaluator/harbor.py
@@ -20,7 +20,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path
 from types import ModuleType, TracebackType
-from typing import Any, TypeAlias, TypedDict
+from typing import Any, Literal, TypeAlias, TypedDict
 from uuid import uuid4
 
 from harbor.constants import MAIN_SERVICE_NAME
@@ -112,6 +112,7 @@ _TASK_TREE_RESOURCES: tuple[TreeResourceSpec, ...] = (
 )
 _TRACE_ARTIFACT_SOURCE = "/app/traces"
 _TRACE_ARTIFACT_DESTINATION = "traces"
+_ATIF_TRACE_SUFFIX = ".atif.json"
 _SHELL_SYNTAX_TIMEOUT_SEC = 10.0
 _AGENT_IMPORT_ROOT = "_nemo_experimentalist_eval_agents"
 _IDENTIFIER_RE = re.compile(r"\W+")
@@ -188,6 +189,13 @@ class HarborEvaluatorConfig(EvaluatorConfig):
     retry: RetryConfig = Field(default=RetryConfig(exclude_exceptions=set()))
     import_path: str = Field(default="harbor_wrapper:WrappedAgent")
     trace_dir: str = Field(default=_TRACE_ARTIFACT_SOURCE)
+    trace_format: Literal["otlp", "atif"] = Field(
+        default="otlp",
+        description=(
+            "Which trace artifact becomes the trial's trace. Both are still collected and "
+            "exposed in resources; this only selects the one that gets uploaded and analysed."
+        ),
+    )
 
 
 class HarborDependencyRuntime(DependencyRuntime):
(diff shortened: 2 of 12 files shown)


plugins/nemo-experimentalist/artifacts/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
gaiadilorenzo: chore(experimentalist) — ignore artifacts/experimentalist-benchmarks/. It is generated output, and because the optimizer writes candidate agents there, a candidate whose generated code does not compile was failing ruff check plugins/nemo-experimentalist/.

-> Shouldn't be the case?

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/evaluator/harbor.py
```diff
@@ -680,8 +691,15 @@ def _trial_resources(trial_dir: Path) -> tuple[dict[str, ResourceRef], ResourceR
         elif path.suffix == ".jsonl" and "traces" in Path(relative_path).parts[:-1]:
```
gaiadilorenzo, line 691: We should probably switch to the same "otlp.jsonl" suffix pattern.

gaiadilorenzo (approved):

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/evaluator/harbor.py
schuellc-nvidia: Yes, but let's postpone this

plugins/nemo-experimentalist/artifacts/.gitignore
schuellc-nvidia: Addressed by last commit

PR #1129, opened 2026-08-06
feat(evaluator): let evaluations pin a taskset revision

SandyChapman:
## What

A `TasksetRef` could only ever name a taskset's **head**, so replacing the taskset changed what a re-submitted spec evaluated. Member content was already digest-pinned within a revision, which made an evaluation *look* reproducible while its membership could still move underneath it.

This lets a reference pin the taskset revision too: `default/geography-suite#<tag-or-digest>`.

Fast-follow to #1023, raised in review there ([thread on `api/schemas.py`](https://github.com/NVIDIA-NeMo/nemo-platform/pull/1023#discussion_r3726233918)). **Stacked on #1023** — please merge that first; this targets its branch.

## The two pins compose

They cover different things:

| Reference | Member content | Which members |
|---|---|---|
| `TasksetRef("suite")` | pinned | current revision |
| `TasksetRef("suite#<digest>")` | pinned | pinned |

A bare ref already grades identical task *content* across re-runs. Pinning the taskset additionally holds *membership* still across a `replace` that adds or drops a task — what you want when a benchmark number has to stay comparable.

## How

`TasksetRef` moves onto `_SUBENTITY_REF_PATTERN`, so it carries `#<tag-or-digest>` exactly as `TaskRef` does, with an absent fragment meaning `latest`. `MetricRef` stays on the narrower pattern until metrics gain revisions — the comment above that constant already anticipated this split and is updated.

`resolve_taskset_ref` resolves the fragment and expands the members of the named revision. **Both the pinned and the bare paths go through `get_revision`** rather than reading the head's own `tasks`: a head and its `latest` revision are guaranteed to agree, and resolving one way for pinned refs and another for bare ones would let the two drift apart on the next bug. It also buys content verification for the bare case.

A fragment that no longer resolves fails the evaluation rather than falling back to the current revision.

## Reviewing this

Small: 6 files, +249/−51, and 160 of the added lines are tests.

1. `task_refs.py` — the resolution change; everything else follows from it.
2. `api/schemas.py` — the pattern move and why `MetricRef` stays behind.
3. Tests and docs are mechanical once the above is settled.

**The test fixtures needed fixing, and that is worth a look.** They were building tasksets the services cannot produce: inserted with no published revision, and with bare (unpinned) member refs. Routing resolution through `get_revision` surfaced that immediately, and `TasksetRevisionEntity` then rejected the unpinned members outright. `_create_published` / `_pin_members` now publish and digest-pin the way `_resolved_content` does, so every fixture describes a state reachable through the real API.

`test_sibling_ref_types_still_reject_fragments` was parameterized over `MetricRef` **and** `TasksetRef`; it is split, since half of its premise no longer holds.

## Testing

- 693 evaluator unit tests pass (685 before).
- The three pinning tests were verified to **fail against the old head-reading behavior** and pass with the fix, rather than assumed to be meaningful.
- Covered: bare ref follows the tip; digest-pinned ignores members added later; tag-pinned resolves through the tag; a pinned ref survives a `replace` with identical expansion; unresolvable digest and unresolvable tag both raise; field-level accept/reject of fragment shapes.
- `tools/lint/lint-all.sh`: 12/13. `lint-web-sdk` fails locally only — this box has no `web/node_modules` and runs Node 22.19 against a `>=22.23.2` engine pin. Its output is gitignored and no web code is touched here, so CI should pass it; I could not verify that locally.
- `openapi.yaml` regenerated with `make refresh-openapi` (+11/−10: the `TasksetRef` pattern and description only).

## Notes

- **No back-compat concern:** widening a validation pattern only admits input that was previously rejected. Every existing bare ref keeps its current meaning.
- The docs Note added in #1023 recording this as a known limitation is replaced by a "Pin the taskset itself" section, and the edge-case table gains the unresolvable-fragment row.

## Summary by CodeRabbit

- **New Features**
  - Taskset references can be pinned to a specific revision, tag, or content digest.
  - Supports tracking the latest taskset membership or freezing membership to a selected revision.
  - Pinned taskset members remain stable when tasksets are replaced.
  - Taskset reference validation now supports valid revision fragments.

- **Bug Fixes**
  - Evaluation now fails with a clear error when a referenced taskset revision cannot be resolved.

- **Documentation**
  - Added guidance on taskset revision pinning, membership behavior, and related API edge cases.

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/task_refs.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/task_refs.py
@@ -18,8 +18,8 @@ from __future__ import annotations
 
 from typing import cast
 
-from nemo_evaluator.api.schemas import TasksetRef, parse_entity_ref, parse_subentity_ref
-from nemo_evaluator.entities import TaskEntity, TaskRevisionEntity, TasksetEntity
+from nemo_evaluator.api.schemas import TasksetRef, parse_subentity_ref
+from nemo_evaluator.entities import TaskEntity, TaskRevisionEntity, TasksetEntity, TasksetRevisionEntity
 from nemo_evaluator.jobs.agent_spec import AgentEvalTaskInput
 from nemo_evaluator.revisions import RevisionNotFoundError, get_revision
 from nemo_platform_plugin.entities import EntityClientProtocol
@@ -49,10 +49,11 @@ def _entity_to_task_input(entity: TaskEntity, revision: TaskRevisionEntity) -> A
     )
 
 
-#: Expanding a taskset reads three entity types through one client — the taskset head, each member
-#: task's head, and the pinned revision of each member. Python has no intersection types, so the
-#: parameter is annotated at one of them and the other two are taken as typed views of the same
-#: object; the concrete client's methods are generic over the entity type and satisfy all three.
+#: Expanding a taskset reads four entity types through one client — the taskset head, its pinned
+#: revision, each member task's head, and the pinned revision of each member. Python has no
+#: intersection types, so the parameter is annotated at one of them and the rest are taken as typed
+#: views of the same object; the concrete client's methods are generic over the entity type and
+#: satisfy all four.
 TasksetStoreProtocol = EntityClientProtocol[TasksetEntity]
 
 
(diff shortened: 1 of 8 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/task_refs.py
```diff
@@ -85,12 +93,21 @@ async def resolve_taskset_ref(
             "or pass an inline task list instead."
         ) from exc
 
-    if not taskset.tasks:
+    # Expand the taskset revision the ref names. Members are digest-pinned inside a revision, so a
+    # member republishing on its own never moves this. A bare ref still follows the taskset's own
+    # revisions, and a ``replace`` re-resolves members on write — so it can change both which members
+    # are named and what they resolve to. Only a pinned ref holds both steady.
+    try:
+        taskset_revision = await get_revision(taskset_revision_store, TasksetRevisionEntity, taskset, taskset_fragment)
```
arpitsardhana, line 101: If not revision specified, does it fetch the latest or some default?

arpitsardhana (approved):

plugins/nemo-evaluator/src/nemo_evaluator/task_refs.py
SandyChapman: Yes, exactly. It uses the `latest` tag to get the current revision.

PR #1130, opened 2026-08-06
fix(deps): raise the Pydantic floor to 2.12.0

SandyChapman:
## Summary

`packages/nemo_platform_plugin/src/nemo_platform_plugin/entities/base.py:192` and `plugins/nemo-insights/src/nemo_insights_plugin/service.py:53` call `model_dump(exclude_computed_fields=True)`. That keyword arrived in Pydantic **2.12.0**, but the declared floors sat at `>=2.10.3` / `>=2.10.6`. Before this change, installing the published wheels without the lockfile could resolve Pydantic 2.10.x and fail at runtime with `TypeError: model_dump() got an unexpected keyword argument`; after it, the dependency metadata matches what the code actually requires.

`uv.lock` already resolved 2.13.4, so lockfile-based installs were unaffected — which is exactly why this went unnoticed.

## Changes

- Raise `pydantic` to `>=2.12.0` in the three packages that declare it directly and use the keyword (transitively or in tests): `packages/nemo_platform_plugin`, `plugins/nemo-evaluator`, `plugins/nemo-insights`, plus the root workspace `pyproject.toml`.
- Propagate the floor into the generated `[tool.bundle-package]` blocks via `make vendor`: `packages/nemo_platform`, `plugins/nemo-data-designer`, `plugins/nemo-safe-synthesizer`. These are generated — not hand-edited. `tools/lint/lint-sdk-vendored.sh` fails until they are committed.
- Update `uv.lock` with the eight matching `requires-dist` specifier bumps (see below for why this was done by hand rather than by a full `uv lock`).

### Note on the lock diff

`uv.lock` carries only the eight matching `requires-dist` specifier bumps — one per changed declaration. Package-block count (582), wheel-URL count (2231), and every existing package are byte-identical to `main`.

A full `uv lock` run produces far more churn on uv 0.9.30: it strips 361 platform wheel URLs (armv7l, ppc64le, s390x, riscv64, musllinux), drops three orphaned packages (`pydantic-graph`, `logfire-api`, `pydantic-monty`), and adds `extra == 'all' / 'plugins' / 'services'` marker entries this lock does not carry. **That churn is unrelated to this change**: `uv lock --check` reports the same drift against an unmodified `main`, verified by swapping `main`'s own lock in and re-running. So the lock here was updated by hand to the intended eight lines rather than letting an unrelated toolchain disagreement ride along.

Consequence worth flagging for review: with this lock, local `uv lock --check` and the `uv-lock-check` pre-commit hook report drift — but they do so on unmodified `main` too, on the same machine. CI is the arbiter, and this lock matches `main`'s format.

Possible follow-up: the repo allows `uv>=0.9.14,<0.10.0` while CI uses `setup-uv@v7.6.0`, so a contributor's local uv can disagree with the committed lock format. Pinning a single uv version would remove this trap.

### Follow-up not included here

Root `pyproject.toml:300` still pins `pydantic-monty==0.0.18` with a comment citing `pydantic-ai-harness==0.3.0` CodeMode, which no longer exists in the tree. The pin is harmless (a constraint does not pull a package in) but vestigial. Left for separate cleanup to keep this diff to the version floor.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Tests not applicable — justification: this changes declared dependency metadata only, with no source change. The runtime behavior it protects is already exercised by the existing entity-projection tests (`plugins/nemo-evaluator/tests/test_{task,taskset,metric,result}_entity.py`), which call `model_dump(exclude_computed_fields=True)` and would fail on any Pydantic below the new floor. A test asserting the floor itself would only restate the `pyproject.toml` line.
- [x] Documentation not applicable — justification: no user-visible surface changes; this is an install-time constraint.

## Verification

Targeted validation:

| Command | Result |
|---|---|
| `uv lock --check` | reports drift — but identically against unmodified `main` on this machine (see lock note) |
| `uv sync --frozen --all-packages --python 3.12` | pass — resolves pydantic 2.13.4 |
| `make test-unit` | 12201 passed, 57 skipped, 2 xfailed, **1 failed** (see below) |
| `tools/lint/lint-all.sh` | 12/13 pass — `lint-web-sdk` blocked (see below) |
| `uv run pre-commit run -a` | all hooks pass except `studio-lint-staged` (blocked, see below) and `uv-lock-check` (same pre-existing uv drift, see lock note) |

Runtime check that the keyword behaves as required under the new floor:

```
$ uv run python -c "...M().model_dump()"                            -> {'a': 1, 'b': 2}
$ uv run python -c "...M().model_dump(exclude_computed_fields=True)" -> {'a': 1}
```

**Not passed — `make test-unit`, 1 failure.** `packages/nemo_evaluator_sdk/tests/agent_eval/test_sandbox_compose_cli.py::test_streaming_timeout_retains_partial_output` failed with `assert '' == 'started\n'`. It passes 3/3 when run in isolation and asserts on subprocess-streaming timing, so I believe it is flaky under `pytest-xdist` parallel load rather than caused by this change. I did not reproduce it on unmodified `main`, so this is reasoned, not proven — worth a second opinion if CI shows the same failure.

**Blocked — Node-based checks.** `lint-web-sdk` and the `studio-lint-staged` pre-commit hook could not run locally: this machine has Node v22.19.0 against the `>=22.23.2 <23` engine pin in `web/package.json`, and `web/node_modules` is absent. Not marked as passing. This change touches no `web/` code, and the web SDK's generated output lives in the gitignored `web/packages/sdk/generated/`, so no tracked generated file derives from it. Relying on CI to confirm.

## Summary by CodeRabbit

* **Chores**
  * Updated the minimum supported Pydantic version to 2.12.0 across the platform, evaluator, insights, data designer, and safe synthesizer packages.
  * Ensured all related package configurations use consistent version requirements for improved compatibility and more predictable installations.


plugins/nemo-evaluator/pyproject.toml
```diff
@@ -8,7 +8,9 @@ dependencies = [
     "nemo-evaluator-sdk",
     "nemo-platform-plugin",
     "nemo-platform-sdk",
-    "pydantic>=2.10.6",
+    # 2.12.0 introduced `model_dump(exclude_computed_fields=...)`, which this
```
mckornfield: do wonder how necessary comments like these are 🤷

mckornfield (approved):

plugins/nemo-evaluator/pyproject.toml
SandyChapman: I chop 'em out.

ironcommit (approved):
LGTM

PR #1145, opened 2026-08-06
fix(deployments): deliver config_files on the openshell backend (AIRCORE-999)

maxdubrinsky:
## Summary

`DeploymentConfig.config_files` is a first-class API field (accepted, validated, given a default `mode`, echoed back), but only the k8s backend implemented it. The openshell backend accepted the field, reached `READY`, and never wrote the files: a silent drop, so a caller had every reason to believe its config was delivered. This makes openshell deliver the files, or fail loudly when it cannot.

## Related Issue

AIRCORE-999.

## Changes

- Deliver each `config_files` entry into the sandbox before serve launch via `ExecSandbox`, with the content on the RPC's first-class `stdin` bytes field (`sh -c 'set -e; mkdir -p <dir>; cat > <path>; chmod <mode> <path>'`). The content is streamed, so it never touches the argv nor the single-line, size-capped sandbox environment. `_exec_detached` gained a `stdin` kwarg.
- Delivery runs inside the once-only launch guard in `_advance_provisioning`, immediately before launch, so files are written exactly once per deployment. Delivery runs as the non-root sandbox user (ExecSandbox has no user field; the policy pins `run_as_user`), so a target that user cannot write (e.g. `/workspace`, which the packaged agent image chowns to `agent` even though the sandbox policy lists it read-write) fails with a terminal `FAILED` naming the path and the shell's error, and the sandbox is torn down, rather than reaching `READY` file-less.
- Delivery is self-attesting: the script prints a completion marker only after `set -e` clears mkdir/cat/chmod, and a write is accepted only when that marker is present in the output. This closes the residual silent-drop path where an `ExecSandbox` stream ends without an exit event, without changing the shared exec probes' "only positive evidence flaps a deployment" semantics.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-visible API or CLI surface change. A doc note on the openshell writable-path constraint can ride the docs work already tracked under AIRCORE-981.

## Verification

Targeted validation:

- `pre-commit run --files <changed files>`: ruff, ruff format, ty, copyright headers, plugin-import guard, and merge-conflict hooks all pass; the remaining hooks skip as not-applicable. Run scoped to the changed files rather than `-a` because a full workspace sync could not be built in this worktree (a transitive dependency needs `Python.h`); the applicable hooks are the ones exercised above.
- `pytest plugins/nemo-deployments/tests/unit/backends/openshell`: 93 passed. Adversarial cases include stdin-streaming with delivery-before-launch, multi-file, octal mode, shell-metacharacter content that must not be interpreted, spaced-path quoting, unwritable `/workspace` failing loudly and deleting the sandbox, rpc-error failing loudly, unconfirmed-delivery (no completion marker) failing loudly, multi-file partial failure, empty-list no-op, and write-exactly-once.
- Full `plugins/nemo-deployments/tests/unit`: passes except `test_ports.py::test_find_available_port_excludes_pending_assignments`, a pre-existing real-socket flake that reproduces identically on unmodified `main` (host port 9000 is occupied on the dev box) and is unrelated to this change, which touches only the openshell backend.

## Notes

Docker delivery and deleting the nemo-agents `NAT_CONFIG_YAML` env-var workaround are deferred to a follow-up ticket, along with a possible fileset-id delivery path for large artifacts. This PR scopes to the openshell backend, whose delivery mechanism is proven.

## Summary by CodeRabbit

* **New Features**
  * Deployment configuration files are now delivered to the sandbox before the service starts.
  * Files are written to their configured paths with the correct permissions.
  * Multiple configuration files are supported and delivered in the correct order.

* **Bug Fixes**
  * Deployment now stops safely when configuration delivery fails, is incomplete, or cannot be verified.
  * Failed deployments clean up the sandbox and do not launch the service.

### DIFF
--- a/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
+++ b/plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
@@ -89,6 +89,23 @@ _SERVE_PIDFILE = "/tmp/nemo-serve.pid"
 # otherwise cannot prove the write happened is a failed, not a silent, delivery.
 _CONFIG_DELIVERED_MARKER = "__nmp_config_delivered__"
 
+
+def _delivery_script(path: str, mode: int) -> str:
+    """Shell one-liner that writes a config file from stdin and self-attests the write.
+
+    ``set -e`` aborts before ``cat`` if the parent directory cannot be made; the content
+    arrives on stdin so it stays opaque bytes to the shell (no escaping, no argv/env size
+    limit, no metacharacter interpretation). The trailing ``printf`` runs only if every
+    step succeeded, so its marker on stdout is positive proof the file was written.
+    """
+    directory = posixpath.dirname(path) or "/"
+    return (
+        f"set -e; mkdir -p {shlex.quote(directory)}; "
+        f"cat > {shlex.quote(path)}; chmod {mode:o} {shlex.quote(path)}; "
+        f"printf %s {_CONFIG_DELIVERED_MARKER}"
+    )
+
+
 # OpenShell rejects a longer sandbox or service name with INVALID_ARGUMENT: three
 # segments plus two "--" delimiters must fit a 63-char DNS label.
 _MAX_ROUTABLE_NAME_LEN = 19
@@ -675,32 +692,21 @@ class OpenShellDeploymentBackend(DeploymentBackend):
         """Write each declared config file into the sandbox before serve launch.
 
         Returns None on success, or a terminal FAILED update naming the file when a write
-        does not succeed. A missing config file is a deterministic config error, so it is
-        reported loudly rather than silently dropped (the historical defect) and the
-        provisioning caller tears the sandbox down.
+        does not succeed. A file that cannot be written fails loudly and the provisioning
+        caller tears the sandbox down; it never reaches READY with the file absent.
 
-        Delivery runs through ``ExecSandbox`` as the sandbox user (the RPC has no user
-        field; the policy pins ``run_as_user`` to the non-root sandbox user). It can
-        therefore only write paths that user owns -- ``/home/sandbox``, ``/sandbox``,
-        ``/tmp`` -- and NOT ``/workspace``, which the packaged agent image chowns to its
-        own ``agent`` user even though the sandbox policy lists it read-write. A target
-        the sandbox user cannot write fails here with the path and the shell's own error,
-        instead of reaching READY with the file absent. When OpenShell grows a native
-        file-transfer RPC, swap the ``cat`` for it; callers only ever emit a first-class
-        ``ConfigFile(path, content)`` and never learn how the bytes arrived.
+        Delivery runs through ``ExecSandbox`` as the sandbox user (the RPC pins
+        ``run_as_user`` to the non-root sandbox user), so it can only write paths that
+        user owns -- ``/home/sandbox``, ``/sandbox``, ``/tmp`` -- and not ``/workspace``,
+        which the packaged agent image chowns to its own ``agent`` user even though the
+        sandbox policy lists it read-write. A path that user cannot write to fails here
+        with the shell's own error. When OpenShell grows a native file-transfer RPC, swap
+        the ``cat`` for it; callers only ever emit a first-class ``ConfigFile(path,
+        content)`` and never learn how the bytes arrived.
         """
         for config_file in config_files:
             path = config_file.path
-            directory = posixpath.dirname(path) or "/"
-            # set -e aborts before cat if the directory cannot be made; content arrives on
-            # stdin so it is opaque bytes to the shell (no escaping, no argv/env limit). The
-            # trailing printf runs only if every step succeeded, so its marker on stdout is
-            # positive proof the file was written -- see the marker check below.
-            script = (
-                f"set -e; mkdir -p {shlex.quote(directory)}; "
-                f"cat > {shlex.quote(path)}; chmod {config_file.mode:o} {shlex.quote(path)}; "
-                f"printf %s {_CONFIG_DELIVERED_MARKER}"
-            )
+            script = _delivery_script(path, config_file.mode)
             try:
                 exit_code, output = await self._exec_detached(
                     sandbox_id, ["/bin/sh", "-c", script], stdin=config_file.content.encode("utf-8")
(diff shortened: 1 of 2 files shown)


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -497,8 +504,14 @@ async def _advance_provisioning(self, sandbox: Any, sandbox_nm: str, workspace:
...
         # Launch the serve command once; the launcher writes a marker so a later poll
-        # does not relaunch it (read_status keeps no state across calls).
+        # does not relaunch it (read_status keeps no state across calls). Config files are
+        # delivered inside this once-only guard, immediately before launch, so the serve
+        # process finds them on disk and they are written exactly once per deployment.
         if not await self._serve_launched(sandbox_id):
+            failure = await self._deliver_config_files(sandbox_id, config.config_files)
+            if failure is not None:
+                await self._delete_sandbox_best_effort(sandbox_nm)
+                return failure
```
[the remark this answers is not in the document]
maxdubrinsky: Declining this one: reconciliation is serial, so there is no concurrent `read_status` for the same sandbox. The controller drives it through an awaited `for deployment in deployments:` loop (`controller.py:135-146`) with no `gather`/`create_task`, and `read_status` is only called from there (`deployment_reconciler.py:165`). Delivery also sits inside the pre-existing once-only `if not _serve_launched(...)` guard, so it adds no race the launch did not already have; it is idempotent (re-writes identical content) and status writes are optimistically locked (`NemoEntityConflictError` retry). The real sequential-poll pattern is covered by `test_config_files_not_redelivered_once_serve_is_launched`. A sandbox-wide lock would add complexity for a race this execution model does not have.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -646,6 +669,65 @@ def _drain() -> tuple[int | None, str]:
...
+                detail = output.strip()
+                message = f"Failed to write config file {path} (exit {exit_code})"
+                if detail:
+                    message = f"{message}: {detail}"
+                return BackendStatusUpdate(status="FAILED", status_message=message)
+            # Self-attesting delivery: require the completion marker rather than trusting a
+            # zero/None exit. This closes the residual silent-drop hole where a stream ends
+            # without an exit event (exit_code None) yet the write never happened -- an
+            # unconfirmed delivery fails loudly instead of advancing to READY file-less.
+            if _CONFIG_DELIVERED_MARKER not in output:
```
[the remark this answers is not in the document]
maxdubrinsky: Declining, and pinning it with a test instead: the completion marker is already positive proof of the write, so requiring `exit_code == 0` would regress. `ExecSandbox` is server-streaming and the file content is sent atomically in the request's `stdin` bytes field (not the response stream), so a truncated response cannot produce a partial file, and `printf <marker>` runs only after `set -e` clears `mkdir`+`cat`+`chmod`. Marker-present therefore means the file is fully written regardless of whether the exit event survived, and rejecting that case would false-fail a good delivery and delete the sandbox, violating this backend's "only positive evidence flaps a deployment" rule. Added `test_config_delivery_with_marker_but_no_exit_event_succeeds` (the `_stream_without_exit(stdout=_CONFIG_DELIVERED_MARKER)` case you named) to document it, alongside the existing no-marker->FAILED coverage.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -646,6 +669,65 @@ def _drain() -> tuple[int | None, str]:
...
+        Returns None on success, or a terminal FAILED update naming the file when a write
+        does not succeed. A missing config file is a deterministic config error, so it is
+        reported loudly rather than silently dropped (the historical defect) and the
+        provisioning caller tears the sandbox down.
+
+        Delivery runs through ``ExecSandbox`` as the sandbox user (the RPC has no user
+        field; the policy pins ``run_as_user`` to the non-root sandbox user). It can
+        therefore only write paths that user owns -- ``/home/sandbox``, ``/sandbox``,
+        ``/tmp`` -- and NOT ``/workspace``, which the packaged agent image chowns to its
+        own ``agent`` user even though the sandbox policy lists it read-write. A target
```
mckornfield: comment seems weirdly cut off here though it's just phrasing "A target the sandbox user cannot write"

Wonder if this comment can be stated more succintly

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -646,6 +669,65 @@ def _drain() -> tuple[int | None, str]:
...
+            # A definitively non-zero exit is a failure (e.g. a /workspace write denied by the
+            # sandbox user's permissions); surface the shell's own error with the path.
+            if exit_code not in (None, 0):
+                detail = output.strip()
+                message = f"Failed to write config file {path} (exit {exit_code})"
+                if detail:
+                    message = f"{message}: {detail}"
+                return BackendStatusUpdate(status="FAILED", status_message=message)
+            # Self-attesting delivery: require the completion marker rather than trusting a
+            # zero/None exit. This closes the residual silent-drop hole where a stream ends
```
mckornfield: silent-drop hole

plugins/nemo-deployments/tests/unit/backends/openshell/test_openshell_backend_mocked.py
```diff
@@ -608,3 +652,254 @@ def test_liveness_probe_is_pending_when_the_marker_is_unusable(tmp_path: Path, m
...
+    mock_stub.GetSandbox.return_value = _sandbox(pb.SANDBOX_PHASE_READY)
+    mock_stub.ExecSandbox.side_effect = [_exec_events(1), _delivery_ok(), _delivery_ok(), _exec_events(0)]
+
+    update = await openshell_backend.read_status(workspace="default", name="srv")
+
+    assert update.status == "STARTING"
+    deliveries = _delivery_requests(mock_stub)
+    assert len(deliveries) == 2
+    assert deliveries[0].stdin == b"a: 1\n"
+    assert "cat > /home/sandbox/agent.yaml" in deliveries[0].command[-1]
```
mckornfield, line 714: some of these asserts 🤷

mckornfield (approved):
wonder if tests could be DRY a bit but 🤷 maybe it will look worse

ironcommit (approved):
LGTM

plugins/nemo-deployments/tests/unit/backends/openshell/test_openshell_backend_mocked.py
maxdubrinsky: yeah I had the robot write a better test, this one is uh, not super useful 🫠

PR #1148, opened 2026-08-06
feat(evaluator): agent-evaluate wire publish_to_intake() to API

nv-odrulea:
## Summary

Agent-evaluation jobs can now publish their results to Intake, so eval runs show up under
Experiments and are associated with a specific agent. Previously `publish_to_intake` existed but
nothing called it except tests — there was no API to request publication, so Studio couldn't
trigger it.

## Related Issue

https://linear.app/nvidia/issue/ASTD-384/agent-experiments-evaluator-publish-to-intake-from-api

## Changes

Adds an opt-in `publication.intake` block to the agent-evaluate job spec:

```json
{
  "publication": {
    "intake": {
      "evaluation_id": "email-security-baseline",
      "agent_name": "email-security-analyst",
      "required": true
    }
    "intake": {
      "evaluation_id": "email-security-baseline",
      "agent_name": "email-security-analyst",
      "required": true
    }
  }
}
```

Absent → no publish, zero Intake calls. Present → the worker publishes after the run, under an
Evaluation that must already exist (it

## Also fixes: publishing was not idemp

`trial_to_atif_ingest` emitted a step with no timestamp, so Intake fell back to a per-request
ingest clock — and `spans` is a `ReplacingMergeTree` keyed on `start_time`. Re-publishing wrote
rows that never collapse. Measured agaise, one re-published trial:

| | spans (FINAL, and via the public API) |
|---|---|
| before | **4** (should be 2) |
| after | **2** |

`traces.list` returned `1` either way, which is why this would not have been noticed. Fix is one
field: stamp `result.metadata.started_at` on the step. `publish_to_intake` now refuses to publish
without it rather than writing uncollap

This matters because the recovery storye-publish" — without idempotency
that instruction corrupts the data inst

## Notable decisions

- **`required` defaults to true.** A run that asked to publish shouldn't report success with
  nothing in Experiments. The bundle isn runs, so a failed job still
  leaves results intact to re-publish. `required: false` opts out.
- **Publication runs last**, after the bundle artifact and queryable result record are durable.
  It's the only step that can fail the job.
- **Agent identity is derived only whergentTarget`, `HarborRunnerTarget`).
  Otherwise `agent_name` is required and rejected at submit with a 422 — not after a 20-minute eval.
- **Job output uses `PublicationOutcome`**, typed with the shared `PlatformJobStatus`
  (`COMPLETED` / `ERROR`), not the publisher's internal `PublishReport` — which names the
  Evaluation `experiment_id`, contradiche API accepts.

## Testing

- 30 new unit tests; 748 total green in
- 3 integration tests pass against live Intake + ClickHouse, including a new
  double-publish idempotency test.
- Also fixes a pre-existing limitation in that integration fixture: it spawned `nemo services run`
  without `--port`, so it always bound the 8080 default and couldn't run alongside a local dev
  platform. Now derives the port from `NMP_BASE_URL`.

## Follow-ups

- `nmp-v7j` — recovery endpoint to re-pUntil it ships, recovery after a
  failed publication means re-running t
- `nmp-046` — Studio submit wiring. Splg with in-flight modal changes
  (`nmp-bjm`). Note `web/packages/sdk/generated/` is gitignored, so run
  `pnpm --filter @nemo/sdk gen:evaluato

## Not in this PR

Trajectories published today are single; real `steps[]` from
`trial.evidence` is deferred to D2 / AAscores plus a stub trace.

This reverses AALGO-290's stance that py* optional (no feature flag).
The explicit opt-in spec field preserve = no publish.

## Summary by CodeRabbit

* **New Features**
  * Added optional publication of agent-evaluation results to Intake, including evaluation ID, agent metadata, and configurable failure behavior.
  * Publication outcomes are returned when enabled.
  * Added scalar scores and expanded aggregate statistics, including median, sample variance, and standard deviation.
  * Aggregate score counts are now optional when sample counts are unknown.
  * Studio now supports displaying scalar aggregate scores.

* **Bug Fixes**
  * Improved republishing safety with timestamps and stable identifiers to prevent duplicate records.
  * Added validation and clearer handling for missing timestamps, evaluations, clients, or publication failures.

### DIFF
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -2132,6 +2132,11 @@ components:
           title: Labels
           description: Caller-supplied tags recorded on the run's metadata (e.g. benchmark,
             mode, backend).
+        publication:
+          allOf:
+          - $ref: '#/components/schemas/PublicationSpec'
+          description: Where the completed run publishes its results, beyond its own
+            result bundle. Omit to publish nowhere.
         tasks:
           anyOf:
           - $ref: '#/components/schemas/TasksetRef'
@@ -2298,6 +2303,11 @@ components:
           title: Labels
           description: Caller-supplied tags recorded on the run's metadata (e.g. benchmark,
             mode, backend).
+        publication:
+          allOf:
+          - $ref: '#/components/schemas/PublicationSpec'
+          description: Where the completed run publishes its results, beyond its own
+            result bundle. Omit to publish nowhere.
         tasks:
           items:
             $ref: '#/components/schemas/AgentEvalTaskSpec'
@@ -3736,6 +3746,48 @@ components:
         so no code is shipped or executed on load. Used for platform-recognized
 
         built-in metric types.'
+    IntakePublicationSpec:
+      properties:
+        evaluation_id:
+          type: string
+          minLength: 1
+          title: Evaluation Id
+          description: Name of the existing Evaluation to publish under. Must already
+            exist; the job does not create it.
+        agent_name:
+          title: Agent Name
+          description: Agent name recorded on each published trajectory. Derived from
+            the target when it names one; required otherwise.
+          type: string
+        agent_version:
+          type: string
+          title: Agent Version
+          description: Agent version recorded on each published trajectory. Neither
+            a Model nor an Agent carries a version, so this defaults to 'unknown'
+            unless the submitter supplies one.
+          default: unknown
+        required:
+          type: boolean
+          title: Required
+          description: Fail the job when publication fails. Defaults to True so a
+            run that asked to publish does not report success with nothing in Experiments.
+            The result bundle is saved before publication runs, so a failed job still
+            leaves the results intact to re-publish. Set False to keep the job successful
+            and report the failure in its output instead.
+          default: true
+      additionalProperties: false
+      type: object
+      required:
+      - evaluation_id
+      title: IntakePublicationSpec
+      description: "Publish this run's trials and scores to Intake, under an Evaluation\
+        \ that already exists.\n\n``evaluation_id`` is the *name* of a ``client.evaluations``\
+        \ record. Intake stores that record as\nits ``Experiment`` entity and the\
+        \ SDK's ``publish_to_intake`` calls the argument\n``experiment_id``, but the\
+        \ value is the same one either way \u2014 the parent ``client.experiments``\n\
+        group is a different resource and is not what goes here. The job never creates\
+        \ the Evaluation: a\nmissing one is an error, because nothing in an eval spec\
+        \ can supply the dataset identity\n``evaluations.create`` requires."
     JsonValue:
       title: JsonValue
     MetadataItem:
@@ -4483,6 +4535,17 @@ components:
       - created_at
       - updated_at
       title: PlatformJobTaskStatusResponse
+    PublicationSpec:
+      properties:
+        intake:
+          allOf:
+          - $ref: '#/components/schemas/IntakePublicationSpec'
+          description: Publish trials and scores to Intake. Omit to publish nowhere.
+      additionalProperties: false
+      type: object
+      title: PublicationSpec
+      description: Where a completed run publishes its results, beyond its own result
+        bundle.
     ReasoningParams:
       properties:
         end_token:
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -3899,8 +3899,10 @@ components:
           description: Agent name recorded on each published trajectory. Derived from
             the target when it names one; required otherwise.
           type: string
+          minLength: 1
         agent_version:
           type: string
+          minLength: 1
           title: Agent Version
           description: Agent version recorded on each published trajectory. Neither
             a Model nor an Agent carries a version, so this defaults to 'unknown'
--- a/plugins/nemo-evaluator/openapi/openapi.yaml
+++ b/plugins/nemo-evaluator/openapi/openapi.yaml
@@ -2675,9 +2675,12 @@ components:
           title: Name
           description: Name of the score.
         count:
-          type: integer
           title: Count
-          description: Number of samples evaluated (excluding NaN).
+          description: "Number of samples evaluated (excluding NaN). None when the\
+            \ sample size is unknown \u2014 e.g. a figure imported from a backend\
+            \ that reports statistics without the n behind them. Distinct from 0,\
+            \ which asserts that nothing was evaluated."
+          type: integer
         nan_count:
           type: integer
           title: Nan Count
@@ -2698,13 +2701,33 @@ components:
           title: Max
           description: Maximum score value.
           type: number
+        median:
+          title: Median
+          description: Median score value. Equal to percentiles.p50 when a percentile
+            distribution is also present; carried separately because a backend may
+            report a median without one.
+          type: number
         std_dev:
           title: Std Dev
-          description: Standard deviation of the scores.
+          description: Population standard deviation of the scores (divides by n).
+            Describes the spread of the values actually evaluated. See sample_std_dev
+            to estimate the spread of the wider process.
           type: number
         variance:
           title: Variance
-          description: Variance of the scores.
+          description: Population variance of the scores (divides by n). See sample_variance.
+          type: number
+        sample_std_dev:
+          title: Sample Std Dev
+          description: "Sample standard deviation of the scores (Bessel-corrected,\
+            \ divides by n-1). Estimates the spread of the process the values were\
+            \ drawn from \u2014 the right choice when repeated trials sample a stochastic\
+            \ system. None when fewer than two values (undefined, not zero)."
+          type: number
+        sample_variance:
+          title: Sample Variance
+          description: Sample variance of the scores (Bessel-corrected, divides by
+            n-1). None when fewer than two values.
           type: number
         score_type:
           type: string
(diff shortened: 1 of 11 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/jobs/publication.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    except PublishError as error:
+        return fail(str(error), error.report)
+    except NotFoundError:
+        return fail(
+            f"Evaluation {spec.evaluation_id!r} does not exist in workspace {workspace!r}. "
+            "Create it before submitting the job; the evaluation does not create it."
+        )
+    except NeMoPlatformError as error:
+        return fail(f"{type(error).__name__}: {error}")
```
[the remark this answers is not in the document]
nv-odrulea: done

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -2665,9 +2675,12 @@ components:
...
           description: Name of the score.
         count:
-          type: integer
           title: Count
-          description: Number of samples evaluated (excluding NaN).
+          description: "Number of samples evaluated (excluding NaN). None when the\
+            \ sample size is unknown \u2014 e.g. a figure imported from a backend\
+            \ that reports statistics without the n behind them. Distinct from 0,\
+            \ which asserts that nothing was evaluated."
+          type: integer
```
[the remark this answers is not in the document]
nv-odrulea: done

plugins/nemo-evaluator/tests/jobs/test_publication.py
```diff
@@ -0,0 +1,462 @@
...
+def test_job_publishes_through_the_real_sync_bridge(tmp_path: Path, mocker: MockerFixture) -> None:
+    # `run` is synchronous and has already driven one event loop to completion via
+    # `evaluator.run_sync` by the time publication starts. Publication then drives another through
+    # `run_sync` on the *same* injected async SDK. Nothing is patched out here, so a regression to
+    # `asyncio.run` or to a loop-bound client surfaces as "Event loop is closed" (cf. nmp-1hr.2).
+    mocker.patch.object(AgentEvalJob, "_build_evaluator", return_value=_FakeEvaluator())
+    client = _FakeClient()
+    ctx = _job_context(tmp_path)
+
+    result = AgentEvalJob().run(_job_spec().model_dump(), ctx=ctx, async_sdk=cast(AsyncNeMoPlatform, client))
```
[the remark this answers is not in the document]
nv-odrulea: done

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -3736,6 +3886,48 @@ components:
...
+          type: string
+          minLength: 1
+          title: Evaluation Id
+          description: Name of the existing Evaluation to publish under. Must already
+            exist; the job does not create it.
+        agent_name:
+          title: Agent Name
+          description: Agent name recorded on each published trajectory. Derived from
+            the target when it names one; required otherwise.
+          type: string
```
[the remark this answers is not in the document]
nv-odrulea: done

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -2777,18 +2832,113 @@ components:
       type: object
       required:
       - name
-      - count
       - nan_count
       - rubric_distribution
       title: AggregateRubricScore
       description: Aggregated statistics for a rubric-type score with category distribution.
+    AggregateScalarScore:
```
SandyChapman, line 2839: There's a bit of churn that somehow happened where these aren't edited in this PR but are being regenerated in this spec. I have a PR that's adding these as well, so if you see a conflict in this file, we probably just need a rebase.

plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
```diff
@@ -95,12 +97,24 @@ def trial_to_atif_ingest(
     a minimal single-step trajectory carrying the trial's final output text, so
     the session/score path works end to end. Real ``steps[]`` reconstructed from
     ``trial.evidence`` arrive with D2.
+
+    ``started_at`` (the run's start time) is stamped on the step because it is what
+    makes re-ingest idempotent. Intake's ``spans`` table is a ``ReplacingMergeTree``
+    keyed on ``(workspace, session_id, start_time, id)``, and a step with no timestamp
+    falls back to the server's per-request ingest clock — so the same trajectory sent
+    twice lands as two rows that never collapse. An explicit timestamp makes the root
+    span's ``start_time`` a function of the run, not of when it was published.
```
SandyChapman, line 106: Thanks for adding this. I don't think this field existed when this was first implemented.

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
```diff
@@ -127,6 +134,17 @@ async def publish_to_intake(
...
+    # Required, not defaulted to "now": a fallback would silently reintroduce the
+    # publish-time clock that makes re-ingest duplicate rows (see the ingest note on
+    # ``mapping.trial_to_atif_ingest``). A real run always sets it; a hand-built result
+    # must say when it ran.
+    started_at = result.metadata.started_at
+    if started_at is None:
+        raise PublishError(
+            f"Cannot publish run {result.run_id!r}: metadata.started_at is unset, and publishing "
+            "without it would write trajectories that duplicate on re-publish."
+        )
```
SandyChapman, line 146: I think we should just change started_at to be non-optional in the `packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py` file.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -152,6 +153,76 @@ class HarborRunnerTarget(BaseModel):
...
+    Only targets that carry a real name yield one — nothing here invents an identity, because a
+    made-up agent name is worse than an explicit one the submitter had to supply. A ``ModelTarget``
+    has a model but no agent; the runners other than Harbor name a harness, not an agent. Those
+    cases return ``None`` and the spec must carry ``publication.intake.agent_name``.
+
+    Distinct from ``result_persistence._agent_target_fields``, which flattens the same targets to
+    ``(kind, name, url)`` filter traits and folds runner *models* into its ``name`` slot.
+    """
+    if isinstance(target, AgentTarget):
+        return getattr(target.agent, "name", None), None
```
SandyChapman: Do we need the getattr here? I think this should correspond to https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/agents.py#L68 which has a required `name` property.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -418,4 +420,20 @@ def run(
...
 
-        return {"status": "completed", "artifact": artifact.model_dump()}
+        output = {"status": "completed", "artifact": artifact.model_dump()}
+
+        # Publication runs last, after the bundle and the queryable record are both durable, so a
+        # failed publish costs a re-publish rather than a re-run. It is also the only step here that
+        # can fail the job (when `required`), which is why nothing depends on its result.
+        intake = spec.publication.intake if spec.publication is not None else None
+        if intake is not None:
+            outcome = publish_agent_eval_result(
```
SandyChapman, line 430: Maybe add a logger statement here for intake publication.

SandyChapman (approved):
thanks for adding this! I left a couple comments, but overall this lgtm!

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
nv-odrulea: Type level guarantee beats runtime check, I agree with this direction.
On the other hand, the blast radius is a bit bigger than I'm comfortable with for this PR.

I searched and found this:

Agreed, that's the better home for it. Filed separately since it cascades.

`AgentEvalResult.metadata` uses `default_factory=RunMetadata`, so making `started_at` required means `metadata` becomes required too, touching ~12 test constructions across `nemo_evaluator_sdk`, `nemo-evaluator` and `nemo-optimization`. The only production constructor (`AgentEvaluator.run`) already sets it, so no behavior change. Keeping the runtime guard here as the safety net until then; it gets deleted when the type makes it unreachable.
 
I think adding 12 files to this PR, plus opening the can of worms on the `finished_at` & `duration_sec`, merits a follow up PR? I've filed a ticket and will put the changes in a new PR:

https://linear.app/nvidia/issue/ASTD-389/eval-make-started-at-non-optional-replacing-runtime-check

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
nv-odrulea: confirmed.  bad habit of my claude, for some reason. added a note to stop doing this in my AGENTS.local.md

nv-odrulea: fixed

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
nv-odrulea: done.

PR #1158, opened 2026-08-07
fix(cli): persist Anthropic auth header format during setup

BrianNewsom:
## Summary

`nemo setup` currently stores Anthropic's `X-Api-Key` template in `required_extra_headers`, where it remains the literal string `{{ auth_secret }}` and Anthropic rejects every request. This change persists the existing provider definition's `auth_header_format` field on both create and update, so the inference gateway renders the secret at request time; default Bearer-auth providers are unchanged.

## Changes

- Pass the provider definition's `auth_header_format` through setup provider create and update calls.
- Repair existing Anthropic providers by explicitly clearing legacy `required_extra_headers` when interactive or automatic setup is rerun.
- Add create, interactive-update, auto-update, and raw-key non-disclosure regression coverage.
- Regenerate the vendored NeMo Platform SDK source and tests.
- Keep the fix scoped to setup: no Models or inference-gateway implementation changes.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
- [x] Documentation not applicable — justification: the supported Anthropic provider command and `auth_header_format` behavior are already documented; this fixes `nemo setup` to follow that contract.

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [x] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

Targeted validation on current head (`242b66df17`):

```bash
uv run --frozen pytest \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py -v
# 217 passed

uv run --frozen pytest \
  sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/cli/commands/test_setup.py -v
# 217 passed

uv run ruff check \
  packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py
# passed

uv run ruff format --check \
[...]

PR #1159, opened 2026-08-07
feat(agents): allow optimization agents to configure models

BrianNewsom:
## Summary

This PR makes the Analyst, Eval Author, and Experimentalist use the models selected through `nemo setup`. It replaces plugin-specific endpoints, credentials, hardcoded model names, and Experimentalist's three-tier model configuration with one Platform-native pair: `default_model` for quality-oriented work and `fast_model` for latency-oriented work, with fast falling back to default.

Before this change, each optimization plugin configured inference independently. Setup-selected models and credentials in Platform Secrets were not reused, so OpenAI-compatible providers required manual plugin configuration and Anthropic could not be selected in setup and consumed through its native Messages contract.

### Platform-native model selection

`nemo setup` now stores two workspace-qualified Model Entity references in the active CLI context:

```yaml
default_model: default/gpt-5
fast_model: default/gpt-5-mini
```

Shared logic in `nemo_platform_plugin` reads the pair, resolves each entity through the existing Platform SDK, uses its actual served model name/provider route/backend format/Platform headers, constructs the corresponding Nooa/LiteLLM client, and activates those clients for the agent run.

Backend selection comes from the Model Entity rather than model-name heuristics:

- `OPENAI_CHAT` uses the OpenAI-compatible Chat Completions contract.
- `ANTHROPIC_MESSAGES` uses Anthropic's native Messages contract.

The plugins do not translate request or response bodies. Provider credentials stay in Platform Secrets and are never passed into the optimization agents.

### Consistent agent roles

- `default_model` is the existing quality-oriented default and replaces the previous smart role.
- `fast_model` is used only for deliberately latency-sensitive or high-volume work, such as summarization and control steps.
- Experimentalist's former mid-tier uses are assigned to default or fast based on their quality requirements; setup does not try to infer a third provider-independent capability tier.
- Individual plugins only request the active default or fast client; Model Entity parsing, served-name lookup, routing headers, backend selection, and client lifetime are centralized.

Experimentalist's checked-in `models.yaml` remains separate: it controls model mutations for the agent under test, not the models running the optimizer itself. Application and benchmark credentials such as `AUT_MODEL_NAME`, `OPENAI_API_KEY`, and `OPENAI_BASE_URL` may therefore still be required by the evaluated workload.

### Setup and compatibility

Interactive setup asks for a default and fast model. The fast prompt defaults to the chosen default, so users who want one model can press Enter twice without understanding the internal role mapping.

`nemo setup --auto` preserves main's behavior:

- Use `NEMO_DEFAULT_MODEL` when set; otherwise select the alphabetically first discovered Model Entity.
- Use `NEMO_FAST_MODEL` when set; otherwise reuse the effective default.

Existing contexts that only contain `default_model` remain compatible because fast automatically falls back to default.

For CI, containers, and other environments that cannot read the host CLI context, the pair can be overridden with Model Entity references:

```bash
export NEMO_DEFAULT_MODEL=default/gpt-5
export NEMO_FAST_MODEL=default/gpt-5-mini
```

The referenced entities and providers must already exist on the target Platform.

### Removed optimizer-specific configuration

The optimization agents no longer consume:

- `AUTHOR_API_BASE`, `AUTHOR_API_KEY`, or `AUTHOR_*_MODEL_NAME`
- `NEMO_EXPERIMENTALIST_API_BASE` or `NEMO_EXPERIMENTALIST_API_KEY`
- `NEMO_EXPERIMENTALIST_MODELS_SMART`, `NEMO_EXPERIMENTALIST_MODELS_MID`, or `NEMO_EXPERIMENTALIST_MODELS_FAST`
- `INFERENCE_API_KEY` as an optimization-agent credential

### Scheduled Analyst runs

`nemo insights analysis enable` now persists the effective pair in the server-side analysis config. The controller does not need access to the operator's local `~/.config/nmp/config.yaml`. Re-run `analysis enable` after changing the pair; previously enabled records must also be re-enabled once to capture it.

Anthropic setup's `auth_header_format` fix is intentionally isolated in #1158.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

- `uv run --frozen pytest packages/nemo_platform_ext/tests/cli/commands/test_setup.py -q` — 229 passed.
- `uv run --frozen pytest sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/cli/commands/test_setup.py -q` — 229 passed.
- Shared adapter + Analyst + Eval Author + Experimentalist scope before the latest main sync — 2,386 passed, 3 intentional live-test skips.
- Latest main conflict-resolution scope — Experimentalist: 618 passed; Eval Author: 43 passed, 3 intentional live-test skips; focused cross-plugin regression set: 25 passed; targeted `ty`: passed.
- Latest main sync (`ff047d2a57`) — preserved the new SDK/bootstrap/mise documentation and the model-pair setup text; relevant file-level pre-commit hooks passed.
- `uv run pre-commit run -a` after the latest main sync — Ruff, formatting, `ty`, config docs, Helm docs, copyright, UI, import-boundary, and merge-conflict checks passed. `uv-lock-check` is blocked because `uv lock` adds newly published wheel metadata despite no dependency-input change; that generated drift is excluded, and the committed lock resolves successfully with `uv run --frozen`.
- `make docs-check` — passed; 216 MDX files parsed cleanly and no published-to-gated links found.
- `make docs-broken-links` — command exited successfully but reported five existing broken links in unchanged pages (`telemetry-and-privacy.mdx`, `using-authentication.mdx`, and `studio/index.mdx`). None are in this pull request's changed documentation.

Live provider validation:

- Analyst completed against an internal NVIDIA inference model (`default/openai-openai-gpt-5-5`) and created two Insights from 19 observed sessions. Separate Analyst runs completed with native Anthropic models and with an OpenAI default + Anthropic fast pair.
- Eval Author's live repair canary completed with distinct configured models and invoked the direct OpenAI `default/gpt-5-mini` fast path.
- Experimentalist completed a full one-round Tau3 optimization with the internal GPT-5.5 default and direct OpenAI GPT-5 Mini fast model. Platform logs recorded 39 successful default calls and one successful fast call; the candidate won held-out validation with reward 0.5.

## Summary by CodeRabbit

* **New Features**
  * Added separate default and fast model selection through setup, including discovery, fallback, and validation.
  * Analysis configurations now persist and use both selected models.
  * Eval Author, Experimentalist, and Analyst now use centrally configured Platform models.

* **Documentation**
  * Updated setup guides, examples, troubleshooting, and configuration references.
  * Replaced legacy provider-specific model and credential instructions.

* **Bug Fixes**
  * Improved model routing, resource cleanup, preflight checks, and scheduled-analysis model propagation.


plugins/nemo-experimentalist/benchmarks/run.py
```diff
@@ -350,7 +342,10 @@ def summarize_experimentalist_jobs(results_dir: Path) -> dict[str, Any]:
             "cache_tokens",
             "output_tokens",
         ):
-            totals[key] += job[key]
+            value = job[key]
+            if not isinstance(value, int):
+                raise TypeError(f"Expected integer job summary field {key!r}, got {type(value).__name__}")
+            totals[key] += value
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in 64add97586. Raw Harbor integer fields are validated before aggregation; booleans, floats, and numeric strings are rejected with the output field name. Parameterized tests cover all three invalid shapes.

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
```diff
@@ -0,0 +1,207 @@
...
+
+    default: CompletionClient
+    fast: CompletionClient
+    refs: ConfiguredModelRefs | None = None
+
+    async def aclose(self) -> None:
+        """Close each distinct client owned by this pair."""
+        await self.default.aclose()
+        if self.fast is not self.default:
+            await self.fast.aclose()
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in 64add97586. Pair shutdown attempts every distinct client even after a close failure, and partial model resolution closes every client already constructed before re-raising the original error. Both failure paths have tests.

.github/workflows/insights-testbed.yml
```diff
@@ -114,6 +114,11 @@ jobs:
       # Same gateway key; litellm/tau2 read it under the OpenAI-conventional name.
       OPENAI_API_KEY: ${{ secrets.NVIDIA_INFERENCE_KEY }}
       OPENAI_API_BASE: ${{ secrets.NVIDIA_INFERENCE_URL }}
+      # nemo setup registers this provider and persists its credential in Platform Secrets.
+      NEMO_DEFAULT_INFERENCE_KEY: ${{ secrets.NVIDIA_INFERENCE_KEY }}
+      NEMO_DEFAULT_INFERENCE_BASE_URL: ${{ secrets.NVIDIA_INFERENCE_URL }}
+      NEMO_DEFAULT_MODEL: ${{ vars.INSIGHTS_TESTBED_MODEL || 'default/openai-openai-gpt-5-5' }}
+      NEMO_FAST_MODEL: ${{ vars.INSIGHTS_TESTBED_MODEL || 'default/openai-openai-gpt-5-5' }}
```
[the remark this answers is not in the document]
BrianNewsom: Validated against the current reconciler and a live Platform run. This provider actually discovered and routed default/openai-openai-gpt-5-5; the gateway returned 200 on that exact Model Entity route. I retained the proven entity ID and removed the redundant NEMO_FAST_MODEL entries in b6763b9737 so auto mode exercises the fast-to-default fallback.

plugins/nemo-insights/examples/research-agent/tests/test_analyst_e2e.py
```diff
@@ -147,11 +149,17 @@ def _cli_cmd(name: str, *args: str) -> list[str]:
 
 def _require_keys() -> None:
     env = _subprocess_env()
-    missing = [key for key in ("NVIDIA_API_KEY", "TAVILY_API_KEY", "INFERENCE_API_KEY") if not env.get(key)]
+    missing = [key for key in ("NVIDIA_API_KEY", "TAVILY_API_KEY") if not env.get(key)]
     if missing:
         pytest.skip(f"missing required API keys: {', '.join(missing)}")
```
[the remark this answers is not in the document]
BrianNewsom: Not applied as written. NVIDIA_API_KEY is required by the research agent under test itself; NEMO_DEFAULT_INFERENCE_KEY configures the separate Analyst optimizer provider and cannot replace the AUT credential. b6763b9737 clarifies that contract in the test documentation so the existing preflight accurately reports the real requirements.

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/run.py
```diff
@@ -101,37 +103,41 @@ async def run_experimentalist(
...
+                framework_skills_dirs=framework_skills_dirs,
+            )
+            result = await experimentalist.run(deps)
+        winner = result.winner
+        reporter.run_finished(
+            winner=winner.label if winner is not None else None,
+            scores=dict(winner.reward("validation").metrics) if winner and winner.reward("validation").metrics else {},
+            report_path=(experiment_dir / "eval-and-optimize" / "OPTIMIZATION.md") if winner is not None else None,
+        )
+        return result.summary
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in 64add97586. The benchmark resolves the effective optimizer pair once, passes those exact references into run_experimentalist, and records default and fast under optimizer_models in summary.json.

docs/get-started/example-agent.mdx
```diff
@@ -156,11 +165,8 @@ sbx exec --workdir "$repo" \
   --env INFERENCE_API_BASE \
   --env OPENAI_API_KEY \
   --env OPENAI_BASE_URL \
-  --env NEMO_EXPERIMENTALIST_API_KEY \
-  --env NEMO_EXPERIMENTALIST_API_BASE \
-  --env NEMO_EXPERIMENTALIST_MODELS_SMART \
-  --env NEMO_EXPERIMENTALIST_MODELS_MID \
-  --env NEMO_EXPERIMENTALIST_MODELS_FAST \
+  --env NEMO_DEFAULT_MODEL \
+  --env NEMO_FAST_MODEL \
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in 64add97586. The Git-backed clone-mode flow now carries the confidentiality warning before the command, while the existing local-only warning remains.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -378,6 +387,31 @@ def _configure_remote_connection(cli_context: CLIContext, base_url: str, workspa
...
+    Config.write(
+        params,
+        context_name=_LOCAL_CONTEXT_NAME,
+        set_current_on_create=True,
+    )
+    cli_context.overrides["base_url"] = DEFAULT_BASE_URL
+    cli_context.overrides["current_context"] = _LOCAL_CONTEXT_NAME
+    cli_context.reset_sdk_context()
+
+
```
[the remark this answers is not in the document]
BrianNewsom: Fixed in b6763b9737. Setup now reuses only an existing no-auth context for the same local endpoint; otherwise it chooses the first unused local-N name and leaves an existing local context untouched. Source and vendored SDK tests cover the collision.

docs/cli/reference.mdx
```diff
@@ -47,8 +47,8 @@ Set up NeMo Platform: connect or start services, configure a provider, install s
 
 Uses an already-running platform, starts local services, or connects the
 CLI to an existing remote deployment. Then selects and registers an
-inference provider, picks a default model, installs coding agent skills,
-and optionally deploys a demo agent.
+inference provider, picks default and fast agent models, installs coding
+agent skills, and optionally deploys a demo agent.
```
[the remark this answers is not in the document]
BrianNewsom: This is already generated from the command source. The source text is present in `packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py` and mirrored in the vendored SDK/manifest registry; commit d0f1aed was produced by `make generate-cli-reference-docs`. I also ran `bash tools/lint/lint-sdk-vendored.sh`, which verifies this generated page is synchronized, and it passes. No additional source change is needed.

SETUP.md
```diff
@@ -139,18 +139,20 @@ Without one of those, a system or nvm-managed Node.js takes precedence and may n
 
 If `nemo setup` is too high-level for the task (e.g. debugging startup, custom service set, custom plugin install after bootstrap), use the manual sections below.
 
-### Default model selection (under `--auto`)
+### Default and fast model selection
```
gaiadilorenzo, line 152: A section on suggested pairs might be helpful in either the setup doc or the inference skill

script/generate_config_docs.py
```diff
@@ -39,7 +39,6 @@
 from typing import Any, get_args, get_origin
 
 import yaml
-from nemo_experimentalist_plugin.settings import ExperimentalistConfig
```
gaiadilorenzo, line 42: Why?

plugins/nemo-experimentalist/examples/tau3-nooa-agent/.env.example
```diff
@@ -7,14 +7,14 @@ export INFERENCE_API_KEY=sk-...
...
 export OPENAI_API_KEY="$INFERENCE_API_KEY"
 export OPENAI_BASE_URL="$INFERENCE_API_BASE"
-export NEMO_EXPERIMENTALIST_API_KEY="$INFERENCE_API_KEY"
-export NEMO_EXPERIMENTALIST_API_BASE="$INFERENCE_API_BASE"
+
+# `nemo setup` stores the optimizer's default and fast Platform Model Entities.
+# A Docker Sandbox cannot read the host's CLI config, so uncomment these with the
+# workspace-qualified IDs shown by setup before following the sandbox example.
+# export NEMO_DEFAULT_MODEL=default/gpt-4-1
+# export NEMO_FAST_MODEL=default/gpt-4-1-mini
```
gaiadilorenzo, line 15: Why the change in default models ids?

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
```diff
@@ -0,0 +1,240 @@
...
+) -> CompletionClient:
+    """Build a Nooa client that routes through the Model Entity's Platform URL."""
+    api_base = client.models.get_model_entity_route_openai_url(model_entity)
+    extra_headers = dict(client.models.get_client_default_headers())
+    # Inference Gateway's direct passthrough session preserves compressed
+    # response bytes. Nooa needs decoded JSON/SSE, so make that requirement
+    # explicit at this adapter boundary for every configured agent client.
+    extra_headers[_ACCEPT_ENCODING_HEADER] = _IDENTITY_ENCODING
+    if model_entity.backend_format == _OPENAI_FORMAT:
+        litellm_model = f"openai/{served_model_name}"
```
sklinglernv, line 119: is this always correct or this looks like our internal inference API convention?

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
```diff
@@ -0,0 +1,240 @@
...
+            extra_headers=extra_headers,
+            drop_params=True,
+            # The configured backend contract is explicitly OPENAI_CHAT. Newer
+            # LiteLLM versions otherwise bridge GPT-5.4+ tool calls with any
+            # reasoning_effort value (including "none") to /responses.
+            _skip_responses_api_bridge=True,
+            # Chat Completions tool calls on current OpenAI frontier models
+            # require reasoning to be disabled. Use LiteLLM's capability map to
+            # omit that value only when the served model explicitly rejects it.
+            reasoning_effort=_openai_chat_reasoning_effort(litellm_model),
```
sklinglernv, line 137: we can keep this for now: but rather than guessing we should allow users to configure LLM parameters, e.g. where would I set max output tokens?

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+    elif model_entity.backend_format == _ANTHROPIC_FORMAT:
+        api_base = api_base.removesuffix("/v1")
+    else:
```
sklinglernv, line 141: are we on purpose only supporting OPENAI and anthropic models? NOOA should be able to handle everything that is supported by LiteLLM.

gaiadilorenzo (commented):
Looks good to me but I'm missing a lot of nemo platform context

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def _active_model_clients() -> ConfiguredModelClients:
+    model_clients = _ACTIVE_MODEL_CLIENTS.get()
+    if model_clients is None:
+        raise RuntimeError("Configured model clients were not activated for this agent run")
+    return model_clients
+
+
+def get_default_model() -> CompletionClient:
```
sklinglernv: true for all return types: I am not sure if we makes sense to commit to CompletionClient as the interface. UnifiedLLM is the base class of both CompletionClient and ResponsesClient and would be more general

SETUP.md
BrianNewsom: That could be useful, but I kept setup provider-neutral because available Model Entity IDs depend on the registered provider and workspace. The inference skill already covers discovering and selecting models, and fast defaults to default. I’d prefer provider-specific recommendations as a follow-up rather than maintaining a static list here.

script/generate_config_docs.py
BrianNewsom: `ExperimentalistConfig` only contained the plugin-specific endpoint/key and smart/mid/fast model settings removed by this PR. Keeping it in the generator would continue documenting settings that are no longer read. The optimizer model pair now comes from the shared Platform CLI context.

plugins/nemo-experimentalist/examples/tau3-nooa-agent/.env.example
BrianNewsom: These are not new provider defaults—the new lines are commented examples. The old values were upstream LiteLLM model strings used by Experimentalist’s private configuration. `NEMO_DEFAULT_MODEL` and `NEMO_FAST_MODEL` now contain workspace-qualified Platform Model Entity references created by `nemo setup`.

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
BrianNewsom: This is LiteLLM’s adapter selection syntax, not an internal inference API naming convention. `OPENAI_CHAT` says the Platform route exposes an OpenAI-compatible Chat Completions contract; `api_base` still points at the Model Entity route, and the served name comes from provider registration. I added a source comment clarifying that distinction in `64451c7170`.

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
BrianNewsom: Agreed that parameters such as maximum output tokens should eventually be configurable rather than inferred here. I’d keep that as a follow-up because this PR is focused on model/provider selection and preserving existing agent behavior.

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
BrianNewsom: This is not a provider allowlist. `OPENAI_CHAT` and `ANTHROPIC_MESSAGES` are the backend wire formats currently exposed by Platform Model Entities. NVIDIA, Mistral, and other OpenAI-compatible providers use the `OPENAI_CHAT` branch; native Anthropic Messages uses the other branch. I added a source comment to make that explicit in `64451c7170`.

packages/nemo_platform_plugin/src/nemo_platform_plugin/nooa_model_client.py
BrianNewsom: Agreed. I widened the stored clients and public getters to `UnifiedLLM` in `64451c7170`. The current factory still constructs `CompletionClient` because both currently supported paths use chat-shaped Nooa clients, but consumers are no longer committed to that implementation and a future `ResponsesClient` will not require a shim.

gaiadilorenzo (approved):

.github/workflows/insights-testbed.yml
```diff
@@ -114,6 +114,10 @@ jobs:
       # Same gateway key; litellm/tau2 read it under the OpenAI-conventional name.
       OPENAI_API_KEY: ${{ secrets.NVIDIA_INFERENCE_KEY }}
       OPENAI_API_BASE: ${{ secrets.NVIDIA_INFERENCE_URL }}
+      # nemo setup registers this provider and persists its credential in Platform Secrets.
+      NEMO_DEFAULT_INFERENCE_KEY: ${{ secrets.NVIDIA_INFERENCE_KEY }}
+      NEMO_DEFAULT_INFERENCE_BASE_URL: ${{ secrets.NVIDIA_INFERENCE_URL }}
+      NEMO_DEFAULT_MODEL: ${{ vars.INSIGHTS_TESTBED_MODEL || 'default/openai-openai-gpt-5-5' }}
```
svvarom, line 120: Is `default/openai-openai-gpt-5-5` a valid default? Should it be something like `azure/openai/gpt-5.5`?

BrianNewsom: This is confusingly the platform identifier, which is scoped by workspace (default) and then a model string. I /think/ this is correct.

svvarom (approved):

.github/workflows/insights-testbed.yml
svvarom: Ah got it

PR #1163, opened 2026-08-07
fix(experimentalist): stop discarding usable work in the optimization loop

schuellc-nvidia:
## Summary

Two places where the optimization loop threw away work it had already paid for. A
proposal round died if any single improvement was imperfect, and trajectory
scoring killed the entire run when it met a trace it could not cite. Both now
degrade to less rather than to nothing.

Found while building a deterministic smoke fixture (#1089), but neither is
specific to it.

## Changes

**Proposer — salvage imperfect proposals.** Three conditions discarded a whole
round rather than the offending improvement. The strictest required every
improvement in a round to carry a distinct `optimization_type`; nothing justifies
that, and it also meant a type used once could not be used again later in the
run. Two real runs died on it after round 1, having already produced usable
candidates. Surplus improvements, disallowed types, and duplicate optimization
text now drop the individual improvement and log why. Only an empty result is
fatal. Adds the first tests for this validation, which had none.

**GroupLeafScorer — survive a trace with no turns.** Every score must cite span
IDs, and the only sanctioned lookups (`get_span_id`, `get_turn_data`) are indexed
by turn. An agent that makes no LLM calls produces a trace with zero turns, so
both return `None` however often they are called. The scorer could not satisfy
its own contract, retried to the CodeAct ceiling, and raised `GenerationError` —
and because scoring runs inside `asyncio.gather` with no handler above it, that
ended the run, discarding candidates that had already evaluated successfully.

The evidence was never missing: a turn-less trace still carries a call graph of
which methods ran, in what order, and their status. The contract now names the
observable condition (`Turns: 0`), says to leave `span_ids` empty, and points at
the call graph instead. Guessed or abbreviated ids are still refused, and traces
that do have turns are unaffected.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: internal component contracts; no user-visible surface changes.

## Verification

Targeted validation:

- `uv run --frozen pytest plugins/nemo-experimentalist/tests/experimentalist/` — **320 passed**
- `uv run ruff check` / `ruff format --check` on the changed files — **clean**
- `uv run pre-commit run -a` — every hook passes **except `Check for uv.lock drift`**, which also fails on a pristine `origin/main` checkout with no changes applied. It is `uv 0.9.30` on this machine re-locking to a narrower platform set (dropping armv7l, ppc64le, s390x, riscv64 wheels); committing that would break other architectures, so `uv.lock` is deliberately left untouched by this PR.

**Behavioural verification.** The proposer fix was confirmed by a multi-round run
that previously died at round 2 and now completes. The scorer fix was confirmed
by a run against a deterministic agent: the scorer completed and separated two
candidates that were tied on reward, citing method names and statuses from the
call graph.

## Summary by CodeRabbit

* **Improvements**
  * Proposal handling now removes unsupported or duplicate-text suggestions, retains repeated optimization types, and limits results to the configured candidate count.
  * Clear errors are reported when no usable proposals remain.
  * Turn-less trace scoring now supports scores without span IDs and provides guidance for grounding explanations in available trace evidence.

* **Tests**
  * Added coverage for proposal filtering, logging, truncation, and empty-result handling.
  * Added validation for turn-less trace scoring and span ID behavior.

### DIFF
--- a/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
+++ b/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import logging
 from pathlib import Path
 from typing import Any, Literal, get_args
 
@@ -20,6 +21,8 @@ from .cards import Optimize
 from .model_config import get_fast_model, get_smart_model
 from .tools import WorkspaceTool
 from .util import load_framework_skills
+
+logger = logging.getLogger(__name__)
 
 
 class Improvement(BaseModel):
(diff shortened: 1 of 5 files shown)


plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
```diff
@@ -157,41 +160,75 @@ async def run(
...
+        there is then nothing to build.
+
+        Repeated ``optimization_type`` is fine. Two candidates may need the same
+        kind of edit in different places, and the collision grows *more* likely as
+        optimization succeeds -- each problem closed removes a distinct kind of
+        remaining work, so what is left tends to need the same treatment.
+
+        Drops are logged rather than silent: a Proposer that consistently emits
+        unusable output would otherwise look healthy while doing a fraction of the
+        work.
```
gaiadilorenzo: :nitpick: there's a lot of history of the code in this docstring, I would keep it more coincise and stick to what to expect from this method

plugins/nemo-experimentalist/tests/experimentalist/test_proposer_validation.py
```diff
@@ -0,0 +1,127 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pin what `Proposer._filter_improvements` keeps, drops, and refuses.
+
+This used to be `_validate_improvements`, and every problem was fatal: a
```
gaiadilorenzo: :nitpick: another historic comment

gaiadilorenzo (approved):

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
```diff
@@ -157,41 +160,75 @@ async def run(
...
+
         if len(improvements) > max_candidates:
-            raise ValueError(f"Proposer returned {len(improvements)} improvements; maximum is {max_candidates}")
-        seen_types: set[str] = set()
+            logger.warning(
+                "Proposer returned %d improvements; keeping the first %d",
+                len(improvements),
+                max_candidates,
+            )
+            improvements = improvements[:max_candidates]
```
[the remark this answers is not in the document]
schuellc-nvidia: Confirmed and fixed in b4959e3640. Reproduced it first: `max_candidates=2` over `[add_subagent, add_subagent, edit_config]` truncated to the two disallowed entries, dropped both, and raised — while a usable improvement sat one slot past the cut. That is the exact failure this PR set out to remove.

Truncation now runs after the filter loop, on `kept`. `test_surplus_is_counted_after_the_unusable_are_dropped` pins it, and I verified it fails against the pre-fix proposer with the reported symptom.

plugins/nemo-experimentalist/tests/experimentalist/test_trace_scorer_contract.py
```diff
@@ -0,0 +1,65 @@
...
+    contract = _run_contract()
+    assert "turns: 0" in contract, "the contract must name the condition the scorer can observe"
+    assert "no turns" in contract
+    assert "leave `span_ids` empty" in contract, "the contract must say what to return instead"
+
+
+def test_contract_offers_an_alternative_grounding() -> None:
+    """Permission to return nothing is not enough; the reason must still be evidence-backed."""
+    contract = _run_contract()
+    assert "call graph" in contract, "a turn-less trace still carries a call graph to cite"
```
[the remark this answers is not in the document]
schuellc-nvidia: Fixed in b4959e3640. The three constraints are now asserted individually, so dropping any one of them fails the test.

One thing the suggestion as written would not have caught: the prompt wraps mid-phrase, so `"order they ran in"` spans a newline and the literal substring is absent. `_run_contract()` now collapses whitespace before matching, which also stops the pre-existing assertions in this file from breaking the next time the docstring re-wraps.

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
schuellc-nvidia: Trimmed in b4959e3640. `_filter_improvements` now documents what it drops and what it raises, in six lines instead of sixteen; the rationale for the change lives in the commit message and PR description where it belongs.

plugins/nemo-experimentalist/tests/experimentalist/test_proposer_validation.py
schuellc-nvidia: Trimmed in b4959e3640 — the module docstring now states the behaviour the tests pin, with no reference to what the method used to be called. I applied the same trim to `test_trace_scorer_contract.py`, which had the same problem in the same PR.

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/proposer.py
```diff
@@ -157,41 +160,70 @@ async def run(
...
-        self._validate_improvements(
+        # allowed_types is every type, not just the untried ones. available_types
+        # still reaches the prompt, so novelty stays a *preference*: a type that
+        # worked in an earlier round can be used again when it is the right tool.
+        return self._filter_improvements(
             improvements=improvements,
             max_candidates=max_candidates,
-            allowed_types=set(available_types) or all_types,
+            allowed_types=all_types,
         )
```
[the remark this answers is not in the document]
schuellc-nvidia: Confirmed and fixed in aa5c0abd85. You are right that the two halves disagreed, and the comment I left at the `allowed_types=all_types` call site was describing the validator rather than the instruction the model actually follows — while any type was untried, the prompt told it never to propose a reused one, so the allowance underneath was unreachable.

The pick is legal from either set and preferred from the untried one now. Two further places restricted it the same way and would have contradicted the MANDATORY block, so they moved too: per-improvement step 3 (`pick ONE optimization_type from available_types`) and the exploitation-phase instruction to `pick the best untried direction`.

Added `test_proposer_contract.py` to pin the prompt half, since a validator that permits reuse is not observable if the prompt never emits it. All three assertions fail against the pre-fix docstring. Suite: 324 passed.

PR #1146, opened 2026-08-06
build: allow newer uv for source checkouts

ironcommit:
## Summary

Relaxes the `uv` version upper bound from `>=0.9.14,<0.10.0` to `>=0.9.14`, allowing developers and CI to use newer uv releases (0.10.x, 0.11.x, etc.) for source checkouts. Lockfile updates are still pinned to uv 0.9.14 for reproducibility.

## Changes

- **`pyproject.toml`**: Removed `<0.10.0` upper bound from `required-version` and `constraint-dependencies`
- **`script/uv-lock.sh`** (new): Wrapper script that enforces uv 0.9.14 for `uv lock` operations, ensuring lockfile consistency across contributors
- **`.pre-commit-config.yaml`**: Updated the `uv-lock` hook to use the new wrapper script
- **CI (`.github/workflows/ci.yaml`)**: Added two new jobs:
  - `uv-lock`: Validates lockfile with pinned uv 0.9.14
  - `uv-latest-compatibility`: Runs `uv sync --frozen --all-packages` with latest uv to verify forward compatibility
- **Documentation**: Updated `README.md`, `AGENTS.md`, `docs/get-started/setup.mdx`, `docs/requirements.mdx`, `docs/support-matrix.mdx`, and `plugins/nemo-experimentalist/README.md` to reflect the relaxed requirement
- **`uv.lock`**: Re-resolved with uv 0.9.14 (removes transient packages no longer pulled in)

## Type of Change

- [x] Code change with documentation updates
- [x] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Documentation updated for user-visible behavior

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] No secrets, API keys, or credentials are included

## Summary by CodeRabbit

- **Documentation**
  - Updated setup and support guidance to allow `uv` version 0.9.14 or later.
  - Clarified that `pip` remains supported for published packages.

- **CI**
  - Added checks for lockfile consistency and compatibility with both pinned and latest `uv` versions.

- **Developer Experience**
  - Added validation and guidance for using the required `uv` version when updating lockfiles.

PR #1088, opened 2026-08-05
feat(skills): add Experimentalist workflow

gaiadilorenzo:
## Summary
- add the platform-bundled `nemo-experimentalist` skill
- document agent, spec, dataset, run, and review inputs
- add routing fixtures and platform skill-loader coverage

## Validation
- `uv run --frozen --python <arm64-python> pytest packages/nemo_platform_ext/tests/cli/commands/skills/test_skill_content.py -q` (12 passed)

## Summary by CodeRabbit

* **New Features**
  * Added guidance for improving existing agent source code or evaluation harnesses using Insights, benchmark datasets, and Harbor suites.
  * Clarified routing between experimental improvement workflows and deployed-agent optimization.
  * Documented setup, configuration, evaluation modes, verification, and recovery procedures.

* **Tests**
  * Added coverage for experimentalist requests involving Insights, benchmark datasets, Harbor datasets, and train/validation splits.
  * Added negative cases for unrelated tuning, new agent design, and trace uploads.

PR #1165, opened 2026-08-07
docs(experimentalist): rewrite the plugin README for developers

schuellc-nvidia:
## Summary

The `nemo-experimentalist` README had grown into a 246-line operator runbook that
duplicated the Experimentalist skill and the docs site while never explaining what
the Experimentalist actually does — the evolutionary loop, the levers it changes,
and what the train and validation splits are each for. This rewrites it for human
developers: function first, links out for procedure, and a record of what the
plugin ships in the box.

## Changes

- **Rebased onto current `main`** and reconciled with #1159, which rewrote the
  model-configuration section while this branch was open. The
  `objective_function` / `regression_metrics` reference from that PR is carried
  forward under a task-oriented heading; the relaxed `uv >=0.9.14` floor (#1146)
  and the `[tool.uv.sources]` pointer (#1099) are carried forward too.
- **Lead with the walkthrough.** The example-agent guide is now the first thing
  under "Run it", so readers start from a worked run rather than a cold invocation.
- **Explain the loop.** New "How it works" covers the six optimization levers
  (LLM call, deterministic code, tools, skills, structure, model/config), the
  one-targeted-change-per-candidate rule, and the round order taken from
  `EvolutionaryOptimizer` — including the train evaluation the old text omitted
  entirely. States plainly that train diagnoses and validation selects.
- **Add a parameter table** built from the `run` command's Typer options, plus
  the `doctor` flags.
- **Add missing operational facts:** run cost and duration (use `tmux`),
  smoke-first guidance, resume-on-rerun via the same `--experiment-dir`, that
  `doctor` checks neither datasets nor task template without a loaded profile,
  and that the `.git@` marker is required to pin a ref.
- **Document what ships:** the `nooa` and `langchain-framework` framework skills,
  the two example agents, and the canonical benchmark suites.
- **Fix two errors:** `nemo experimentalist doctor` -> `nemo agents
  experimentalist doctor` (the plugin registers under `nemo.cli.agents`), and a
  stale note describing the pinned NOOA revision as "one past v0.0.6" when the
  root `pyproject.toml` pins past v0.0.8 for a different reason.
- Trim content duplicated by the skill and `docs/agents/insight-driven-optimization.mdx`.

The `## Recommended laptop isolation` heading is retained deliberately: the
Experimentalist skill added in #1088 links to that anchor.

## Type of Change

- [x] Documentation only

## Quality Gates

- [x] Tests not applicable — justification: documentation-only change to a single
      Markdown file; no code, config, or generated output is touched.

## Verification

Targeted validation:

- `uv run pre-commit run --files plugins/nemo-experimentalist/README.md` — passed
  (merge-conflict check ran; all language-specific hooks correctly skipped).
- `uv run pre-commit run -a` — passed. Note: the `uv-lock` hook re-resolved
  `uv.lock` in my local environment, dropping ppc64le/s390x wheel entries. That is
  unrelated environment drift, so it was reverted and is **not** part of this PR;
  the only file changed is `plugins/nemo-experimentalist/README.md`.
- Every factual claim added to the README was checked against the plugin source
  rather than against the skill or the docs. Statements verified this way include
  the CLI option list (`cli.py`), the round order and resume behavior (`loop.py`),
  the six optimization cards (`cards.py`), the one-change rule (`proposer.py`),
  the `storage.publish_winner: true` default (`config.py`), the profile-gated
  `doctor` checks (`cli.py`), the `.git@` ref parsing (`repository.py`), and the
  89-task count for Terminal-Bench 2.1 (`benchmarks/suites/terminal-bench-2.1.yaml`).
- All relative links in the file were confirmed to resolve.
A second verification pass against the source found three claims that did not hold,
all inherited from the previous README, now corrected:

- The agent under test does not read a fixed forwarded set. What reaches the
  evaluation container is declared per task in `task.toml`; `AUT_MODEL_NAME` is a
  convention of the example agents and `benchmarks/run.py`, not a plugin contract.
- NeMo Experiments receives a structure-only mirror — `experiment_mirror.py` states
  that rewards and trials are not copied into it.
- The model-tier table read as exhaustive. `get_smart_model` also drives the
  optimizer loop, and `get_fast_model` drives sub-steps inside most components.

- No test suite covers README prose, so none was run.

### DIFF
--- a/plugins/nemo-experimentalist/README.md
+++ b/plugins/nemo-experimentalist/README.md
@@ -3,108 +3,335 @@
 
 # NeMo Experimentalist
 
-NeMo Experimentalist is the agent-improvement plugin for
-[NeMo Platform](https://github.com/NVIDIA-NeMo/nemo-platform). It consumes
-Insights produced by the Platform-owned Insights plugin and improves a local
-or Git-backed agent against Harbor-compatible train and validation datasets.
+The Experimentalist improves an agent (source code). Give it a baseline
+agent and a pair of Harbor-compatible train and validation datasets, and it
+runs an evolutionary loop: LLM agents read the failures, propose code changes,
+implement them as candidate agents, and score each one. Only candidates that
+improve the held-out validation split survive. The winner can be published as a
+draft PR.
 
-## Install and develop
+Use it when the fix belongs in the agent's harness — its workflow, tool use,
+state handling, or prompts-in-code. To tune a *deployed* agent's model routing
+or cost without changing its implementation, use `nemo agents optimize`
+instead.
+
+## How it works
+
+This is an agent-system optimizer, not a prompt tuner. It reads your agent as a
+graph of deterministic code, LLM calls, and nested subagents, and it can change
+any of those optimization levers:
+
+| Levers | Example change |
+| --- | --- |
+| LLM prompts | Rewrite a prompt |
+| Deterministic code | Move repeatable parsing or validation out of the LLM and into code |
+| Tools | Add, remove, or constrain a tool |
+| Skills and context | Add or correct domain guidance |
+| Structure | Split a method, add a subagent, change routing |
+| Model and config | Swap the model, or raise a limit the agent keeps hitting |
+
+Which one it reaches for is decided by the diagnosed root cause. Every candidate
+makes **exactly one** targeted change, so a score movement can be attributed to
+it.
+
+The loop builds a baseline (`agent-0`) and scores it on validation, which is
+what the first selection has to work from. Then it repeats each round:
+
+1. **Select** survivors from the previous round's validation scores.
+2. **Evaluate on train** — the diagnostic split.
+3. **Analyze** the failures, comparing each trace against a generated reference
+   solution path and separating mechanical breakage from bad reasoning.
+4. **Propose** one-change improvements, alternating between exploring new
+   directions and refining the current best.
+5. **Implement** each in its own copy of the agent: a bounded sub-problem
+   optimization loop refines the change until the behavior it targets actually
+   moves, then an integration check gates the result; candidates that fail the
+   integration check are dropped.
+6. **Evaluate on validation** — the selection split.
+
+The run stops on the round budget (`max_rounds`) or when a convergence check
+decides the rounds have stopped producing improvement.
+
+**The two splits do different jobs, and validation is hidden while candidates
+are generated.** Train diagnoses; validation selects. At run start the
+validation data is moved to a held-out directory and the bash tool used by the
+coding and analysis agents blocks reads of that path, restoring it only while
+validation scoring runs — so candidates cannot be tuned against the data that
+picks them. A third held-out test split, if you keep one, is yours to run
+manually; the profile and CLI take only train and validation.
+
+Candidates are ranked on two signals. The **outcome reward** comes from the
+Harbor verifier: metrics normalized to `[0.0, 1.0]`, averaged across trials,
+with failed trials counted as zero. The optional **trajectory score** rates the
+path the agent took against a goal tree — a weighted rubric of the capabilities
+a task needs. Outcome is the primary signal; trajectory breaks ties and keeps
+candidates whose process is better than their score, which matters because
+long-horizon agent failures are sparse and a directionally right change can
+still fail the final answer. Survivors are kept as a Pareto front rather than a
+single leader, so complementary strengths stay alive across rounds.
+
+## Where it fits
+
+- [NeMo Insights](../nemo-insights/README.md) produces the Insight — a
+  failure pattern inferred from traces — that the Experimentalist optimizes
+  against. Run `nemo agents analyst run` first. The Experimentalist does not
+  analyze traces or host an Insight API.
+- [NeMo Eval Author](../nemo-eval-author/README.md) builds the
+  Insight-specific evaluation suite, invoked automatically in Insight mode.
+- **Harbor** runs the task containers that score every candidate.
+- **NeMo Experiments** mirrors each run and its candidates as an experiment
+  group, so the lineage is visible in Studio. Structure only — rewards and
+  trials are not copied there.
+
+Both the Analyst and the Experimentalist read the same `optimizer.yaml`
+profile, discovered by walking up from the current directory. See
+[Insight-driven optimization](../../docs/agents/insight-driven-optimization.mdx)
+for the full concept guide.
+
+## Run it
+
+**Start with
+[Get started with an example agent](../../docs/get-started/example-agent.mdx).**
+It walks a checked-in example agent end to end — platform setup, dataset
+preparation, `.env` contents, the run itself, and how to read the result. It is
+the only complete worked path, and it is far easier to point this at your own
+agent once you have watched a run finish. The rest of this section is reference
+for when you get there.
 
 This plugin lives in the `nemo-platform` monorepo and shares the root `.venv`.
 From the root of the checkout:
 
 ```bash
 uv sync
-export NEMO="$PWD/.venv/bin/nemo"
 ```
 
-Requires `uv >=0.9.14`.
+Budget for it before you start: a run is an LLM-driven loop that builds and
+scores agents in Harbor containers, so even a deliberately shortened run takes
+about an hour and a real one takes several. Start it in a persistent session
+such as `tmux`, and redirect output to a log if you need to disconnect.
 
-**To verify an end-to-end run, follow
-[Get started with an example agent](../../docs/get-started/example-agent.mdx).**
-
-The source dependencies are pinned to tagged or immutable revisions in the
-workspace root `pyproject.toml` under `[tool.uv.sources]`. NVIDIA-labs OO Agents
-(NOOA) is pinned to a public GitHub commit past `v0.0.8` that carries the
-callable `@strategy(llm=...)` support this plugin depends on.
-
-## Insight-to-experiment flow
-
-The supported handoff is:
-
-```text
-nemo agents analyst run → .nemo-optimizer/insights.yaml or Platform Insight ID
-                       → nemo agents experimentalist doctor
-                       → nemo agents experimentalist run
-```
-
-Run `nemo agents analyst run` using the Platform Insights plugin and its
-documented trace, workspace, and output options. It always persists an Insight
-on Platform and reports its ID; `--insights-file-output` additionally mirrors
-those rows into a local file, which this plugin can consume in place of an ID.
-The Experimentalist does not analyze traces, schedule analysis, or host an
-Insight API.
-
-From an agent directory with an `optimizer.yaml` profile, validate the
-effective inputs:
+Validate the effective inputs before spending anything. Run `doctor` from the
+agent directory containing `optimizer.yaml`, or pass `--profile` — without a
+loaded profile it cannot check the datasets or task template, and reports
+success having verified neither:
 
 ```bash
-$NEMO agents experimentalist doctor
+uv run nemo agents experimentalist doctor
 ```
+
+### Smoke first
+
+Prove that source checkout, Harbor, credentials, and artifact collection work
+together before committing to a real run. Use
+[`experimentalist-smoke.yaml`](examples/tau3-nooa-agent/experimentalist-smoke.yaml)
+with small copied subsets of your train and validation splits — one round, one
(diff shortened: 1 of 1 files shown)


plugins/nemo-experimentalist/README.md
```diff
@@ -3,107 +3,301 @@
...
+The loop builds a baseline (`agent-0`), then repeats each round:
+
+1. **Select** survivors from the previous round's validation scores.
+2. **Evaluate on train** — the diagnostic split.
+3. **Analyze** the failures, comparing each trace against a generated reference
+   solution path and separating mechanical breakage from bad reasoning.
+4. **Propose** one-change improvements, alternating between exploring new
+   directions and refining the current best.
+5. **Implement** each in its own copy of the agent, behind an integration check
+   with a bounded repair loop; candidates that cannot be repaired are dropped.
```
gaiadilorenzo:
```suggestion
   with a bounded sub-problem optimization loop; candidates that fail the integration check are dropped.
```

plugins/nemo-experimentalist/README.md
```diff
@@ -3,107 +3,301 @@
...
-those rows into a local file, which this plugin can consume in place of an ID.
-The Experimentalist does not analyze traces, schedule analysis, or host an
-Insight API.
+### Two modes
 
-From an agent directory with an `optimizer.yaml` profile, validate the
-effective inputs:
+**Insight-driven** — optimize against a diagnosed failure pattern. Uses the
+local `.nemo-optimizer/insights.yaml` beside the profile by default; `--insight`
+names another local file or a platform Insight ID. Requires a `--task-template`
```
gaiadilorenzo, line 139: We should add a task template under the tau agent example

plugins/nemo-experimentalist/README.md
```diff
@@ -3,107 +3,301 @@
...
+`nemo agents experimentalist doctor` reports what is unset.
 
-3. Run against a platform, which ATIF requires:
+| Variable | Config key | Default | Used by |
+|---|---|---|---|
+| `NEMO_EXPERIMENTALIST_API_BASE` | `api_base` | `https://inference-api.nvidia.com/v1` | all optimizer agents |
+| `NEMO_EXPERIMENTALIST_API_KEY` | `api_key` | — (required) | all optimizer agents |
+| `NEMO_EXPERIMENTALIST_MODELS_SMART` | `models.smart` | — (required) | Top-level reasoning in the optimizer loop, Coder, Analyzer, Proposer, Rationalizer, and TraceAnalyzer |
+| `NEMO_EXPERIMENTALIST_MODELS_MID` | `models.mid` | — (required) | Trajectory scorer and the architecture doc |
+| `NEMO_EXPERIMENTALIST_MODELS_FAST` | `models.fast` | — (required) | Terminator and goal tree, plus high-volume sub-steps and summarizers inside most other components |
```
gaiadilorenzo: Let's make sure to update the README's in Brian's PR

plugins/nemo-experimentalist/README.md
```diff
@@ -121,126 +315,63 @@ sbx exec --workdir "$repo" \
...
-```
-
-Use `--insight` to name another local file. Local files can contain one
-Insight object or an `insights` list. A list with multiple entries requires
-`--insight-id`, which accepts an exact ID, exact title, or zero-based index:
+`sbx stop nemo-experimentalist` preserves the VM, output, packages, and private
+Docker cache; `sbx rm nemo-experimentalist` deletes them. Copy artifacts you
+want to keep to the host with `sbx cp` — the
+[example walkthrough](../../docs/get-started/example-agent.mdx#5-optimize-performance-with-the-experimentalist)
+has concrete inspection and copy commands.
```
gaiadilorenzo, line 368: As a future ticket, we should have a local copy of the artifacts dir in the sandbox while running since it is a long running process

plugins/nemo-experimentalist/README.md
```diff
@@ -3,107 +3,292 @@
 
 # NeMo Experimentalist
 
-NeMo Experimentalist is the agent-improvement plugin for
-[NeMo Platform](https://github.com/NVIDIA-NeMo/nemo-platform). It consumes
-Insights produced by the Platform-owned Insights plugin and improves a local
-or Git-backed agent against Harbor-compatible train and validation datasets.
-
-## Install and develop
+The Experimentalist improves an agent (source code). Give it a baseline
```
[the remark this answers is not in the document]
schuellc-nvidia: Skipping this one — the cited guideline is not scoped to plugin READMEs, and applying it literally would break rendering.

`.cursor/rules/writing-docs.mdc` carries `alwaysApply: false` and declares no `globs:`, so it is an agent-requested rule rather than a path-scoped one. `.coderabbit.yaml` scopes its own doc instructions to `documentation/docs/**/*` and `web/**`; neither pattern matches this file.

It is also written for the MyST/Sphinx docs site, not for GitHub-rendered Markdown. Its dropdown guidance is `:::{dropdown}`, which appears nowhere outside `packages/nemo_nb/` and renders as literal text on a GitHub README — the `<details>` element is the equivalent here, and this page already uses none.

Empirically the convention does not exist for these files: of 18 plugin READMEs, 15 have neither section and only `nemo-safe-synthesizer` has both.

On the substance, the content is mostly present, just not under those headings — checkout, `uv sync`, the time budget and the `doctor` precheck are in `Run it`; credentials and models are in `Configuration`; the onward links are `Examples and benchmarks` plus the `Get started with an example agent` pointer in the second paragraph, which is deliberately the first thing a reader hits rather than the last.

Happy to revisit if the intent is to adopt the rule for plugin READMEs generally — that is a repo-wide change to 15 files and a `globs:` entry on the rule, not a change to this PR.

plugins/nemo-experimentalist/README.md
schuellc-nvidia: Applied in 6fe1d9207c, with one deviation I want to flag rather than bury.

You are right on both halves: there is a bounded sub-problem optimization loop and the walkthrough omitted it, and "candidates that fail the integration check are dropped" is exactly what `IntegrationCheckFailed` means. But taking the line verbatim would put the sub-problem loop *inside* the integration check, and `Coder.implement` runs them in sequence — `optimize_subproblem` at coder.py:683, then `integration_check` at :685 — so I reordered it and dropped the second mention of the check:

> 5. **Implement** each in its own copy of the agent: a bounded sub-problem optimization loop refines the change until the behavior it targets actually moves, then an integration check gates the result; candidates that fail the integration check are dropped.

Tell me if you would rather have the literal wording and I will switch it.

plugins/nemo-experimentalist/README.md
schuellc-nvidia: Agreed this is a real gap — the page says insight-driven mode requires `--task-template` and then points at an example that has none.

Leaving it out of this PR, because it is content work rather than a docs edit: there is no checked-in task template anywhere in the repo to copy from. The only hit is `plugins/nemo-insights/examples/tau2-nemo-oo-agent/dataset/template/task_template`, which is untracked local scratch and contains nothing but empty `environment/` and `tests/` directories. Writing a real one for tau3 airline means authoring the Harbor task itself.

Want to take it as a follow-up, or is there a template you already have that should be checked in here?

plugins/nemo-experimentalist/README.md
schuellc-nvidia: Filed an issue for it

plugins/nemo-experimentalist/README.md
schuellc-nvidia: Correcting myself: I said there was no checked-in task template to copy from. That is true of the current tree, but not of history — there is a complete, documented one at commit `709a3ce87a` ("feat(experimentalist): add Harbor examples and guides"), under `examples/hello-harbor-agent/dataset/task-template/`:

```
README.md
instruction.md
task.toml
environment/Dockerfile
tests/expected.txt
tests/test.sh
```

That commit is not an ancestor of `main`, so the files are reachable only through `git show 709a3ce87a:<path>`. Its README also records a constraint worth keeping: `HarborDataset._find_task_dirs` skips any directory literally named `task_template`, so the template has to sit outside `train/` and `validation/` or be named `task-template`.

So this is less open-ended than I made it sound — the shape exists and the tau3 version is an adaptation rather than a from-scratch authoring job. Still proposing it as a follow-up rather than folding it into a docs-only PR, but say the word if you want it here.

schuellc-nvidia: I created an issue for it

plugins/nemo-experimentalist/README.md
schuellc-nvidia: I adapted it to Brian's PR

plugins/nemo-experimentalist/README.md
```diff
@@ -3,244 +3,376 @@
...
+| Tools | Add, remove, or constrain a tool |
+| Skills and context | Add or correct domain guidance |
+| Structure | Split a method, add a subagent, change routing |
+| Model and config | Swap the model, or raise a limit the agent keeps hitting |
+
+Which one it reaches for is decided by the diagnosed root cause. Every candidate
+makes **exactly one** targeted change, so a score movement can be attributed to
+it.
+
+The loop builds a baseline (`agent-0`), then repeats each round:
```
gaiadilorenzo: This could benefit from a mermaid graph

schuellc-nvidia: Added in 3ce94523c7 — a flowchart above the list, not replacing it. The list holds detail that does not fit in a node; the diagram carries what the prose hid, namely that this is a cycle at all: the edge back to Select, the drop path out of Implement, and where each of the two splits is consumed.

```mermaid
flowchart TD
    Base["Build the baseline<br/>(agent-0)"]

    subgraph Round [One round]
        direction TB
        Select["1. Select survivors<br/>Pareto front of the last validation scores"]
        Train["2. Evaluate on train<br/>the diagnostic split"]
        Analyze["3. Analyze the failures<br/>against a generated reference solution path"]
        Propose["4. Propose improvements<br/>exactly one change each"]
        Implement["5. Implement in its own copy of the agent<br/>sub-problem loop, then an integration check"]
        Validate["6. Evaluate on validation<br/>the selection split"]
        Dropped["Candidate dropped"]

        Select --> Train
        Train --> Analyze
        Analyze --> Propose
        Propose --> Implement
        Implement -->|passes the check| Validate
        Implement -->|fails the check| Dropped
    end

    Stop{"Round budget spent,<br/>or converged?"}
    Winner["Winner<br/>optionally published as a draft PR"]

    Base --> Select
    Validate --> Stop
    Stop -->|no| Select
    Stop -->|yes| Winner
```

Syntax verified rather than eyeballed: I extracted the fenced block from the committed README and ran it through mermaid`s own parser, with a deliberately malformed diagram as a negative control so the check was not passing vacuously. The constructs used are the ones already rendering under `docs/`.

Happy to add the held-out-validation boundary as a second diagram if useful — I left it out here to keep this one to the round itself.

plugins/nemo-experimentalist/README.md
```diff
@@ -3,244 +3,407 @@
...
+
+```mermaid
+flowchart TD
+    Base["Build the baseline<br/>(agent-0)"]
+
+    subgraph Round [One round]
+        direction TB
+        Select["1. Select survivors<br/>Pareto front of the last validation scores"]
```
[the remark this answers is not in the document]
schuellc-nvidia: Confirmed and fixed in 881b5602a0. Verified against the loop rather than taken on trust: `_run` has an explicit `# ---- Baseline validation evaluation (round 0)` step at loop.py:558 that evaluates `agent-0` on validation before the cycle starts, and `EvolutionaryOptimizer`'s own docstring records the order as `baseline -> [convergence-check -> select -> train-eval -> ...]`. Running `Base` straight into `Select` left the first selection with nothing to select on.

The diagram now has a `BaseEval` node between them, and I updated the sentence above it to match so the prose and the graph do not drift apart.

One thing I checked and did *not* change: the convergence check sits after `Validate` in the diagram, looping back to `Select`. That reads as check-at-bottom, but the traversal `Validate -> Stop -> (no) -> Select` is the same order the docstring gives as `convergence-check -> select`, so it is already consistent.

Mermaid re-parsed after the edit.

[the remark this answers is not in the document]
schuellc-nvidia: no longer relevant

BrianNewsom (approved):

PR #1178, opened 2026-08-07
feat(auth): implement RFC 8693 workload on-behalf-of token exchange

ironcommit:
## Summary
Implements RFC 8693-style on-behalf-of (OBO) token exchange so that job workloads (Docker containers and Kubernetes pods) obtain access tokens scoped to the job submitter's identity instead of relying on static platform credentials baked into the workload. Adds the core delegation primitives, wires them through the auth service's token exchange endpoint and both jobs backends, and updates the SDK/CLI clients and docs to match.

## Related Issue
No linked issue.

## Changes
- **Auth primitives (`nmp_common`)**: add `WorkloadDelegationEntity`/`WorkloadDelegationStore` (an auth-owned entity binding a workload to the submitting user's identity, with optimistic-concurrency updates, expiry, and revocation), an opaque HMAC-verifiable "Docker proof token" format for workloads without a verifiable IdP identity, and RFC 8693 `act` (actor) claim parsing so resolved bearer tokens carry both the delegated principal and the acting workload.
- **Auth service**: `/apis/auth/token` resolves a delegation record by deterministic name (job/attempt/step for Docker, verified Pod UID for Kubernetes) and issues an access token whose `sub` is the job submitter and `act.sub` is the workload, capped by delegation expiry; the exchange strictly enforces the RFC 8693 profile (rejects `resource`, `actor_token`, `actor_token_type`, multiple `audience`) and retries briefly to absorb Kubernetes pod-startup races. `authenticate.py` now fails closed if the external IdP's issuer/JWKS overlaps with NeMo's own workload-token issuer/JWKS.
- **Jobs controllers**: the Docker backend replaces the old background OAuth password-grant subject-token refresher with per-step delegation registration (proof token issued before container start, revoked on completion/failure); the Kubernetes backend registers a delegation per pod keyed by the verified `authentication.kubernetes.io/pod-uid`, reconciled on status sync and revoked on deletion/terminal status.
- **Go job launcher**: selects the correct `subject_token_type` (opaque Docker proof token vs. JWT) automatically based on the token it receives.
- **Clients**: SDK/plugin clients pick the right `subject_token_type` automatically and can bootstrap workload auth implicitly from `NMP_WORKLOAD_IDENTITY_TOKEN_FILE`.
- **Docs**: document the new flow end-to-end, including the gateway-forwarded `x-nmp-principal-on-behalf-of[-email|-groups]` headers, and update Kubernetes RBAC guidance (only `tokenreviews` needed, no Pod/Job read).
- OpenAPI spec and generated Python SDK regenerated for the updated token exchange contract.
- Extensive unit and integration test coverage for delegation lifecycle (create/lookup/update/revoke, optimistic locking), actor-claim parsing, proof-token validation, issuer-boundary misconfiguration, and Docker/Kubernetes controller behavior.

## Breaking Changes
- Removes the Docker `workload_identity` OAuth password-grant config (`token_endpoint`, `username`, `password_env_var`, etc.). Deployments using the old Docker workload identity config must migrate to the new delegation-based flow, which needs no per-deployment OAuth client config but does require `auth.oidc.workload_token_exchange_enabled`.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

- [x] Every commit includes an appropriate Signed-off-by trailer
- [x] uv run pre-commit run -a passes, or any blocked checks are identified below

Targeted validation:

- uv run --frozen pytest packages/nmp_common/tests/auth/test_jwt.py::TestJWTValidator::test_validate_token_ignores_actor_with_whitespace_subject packages/nmp_common/tests/auth/test_workload_delegations.py -v — 30 passed.
- uv run --frozen pytest packages/nmp_common/tests/auth/test_jwt.py packages/nmp_common/tests/auth/test_token_resolver.py packages/nmp_common/tests/auth/test_middleware.py packages/nmp_common/tests/auth/test_workload_delegations.py services/core/auth/tests/test_middleware.py -v — 114 passed.
- uv run pre-commit run -a — passed after removing plans from the PR.
- DCO audit over origin/main..HEAD — passed for ba3ad00cab350b849be2756d95d7b7ff69711738, 5f0abe73666184e4ae04a8536060d57368f55a14, and 9d68debe8841b2bb92e593f07e331dd934fd48cb.

CI note:

- Pushing 9d68debe8841b2bb92e593f07e331dd934fd48cb started a fresh CI run for the code-only PR diff.


ironcommit:
@coderabbitai review

ironcommit:
@coderabbitai review

ironcommit:
@coderabbitai review

ironcommit:
@coderabbitai review

ironcommit:
@coderabbitai review

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/oidc.py
```diff
@@ -32,6 +32,13 @@
 import httpx
 from nemo_platform_plugin.client.constants import WORKLOAD_IDENTITY_TOKEN_FILE_ENVVAR
 from nemo_platform_plugin.client.tls import client_verify_from_env
+from nmp.common.auth import (
```
ironcommit: **R1 - P1 blocking:** This public client path now imports `nmp.common.auth` at module import time.

`nemo-platform-plugin` and the generated `nemo-platform-sdk` do not declare `nmp-common`, and the plugin auth shim documents `nmp_common` as platform-image-only and lazy. A standalone plugin or SDK install can fail on import before workload token exchange runs. Consider keeping these token-type constants and subject-token detection in the public package/SDK layer instead of importing `nmp.common`.

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -1753,9 +1912,6 @@ def docker_state_debug_fields(self, container: Container) -> dict[str, Any]:
         }
 
     def create_step_update(self, step: PlatformJobStepWithContext, container: Container) -> JobUpdate:
-        if step.status in (PlatformJobStatus.ACTIVE, PlatformJobStatus.PENDING):
-            self._restore_workload_identity_refresher_for_container(container)
-
         status, status_details, error_stack = self.map_docker_container_status_to_platform_status(step, container)
```
ironcommit, line 1842: **R3 - P1 blocking:** Docker terminal reconciliation does not revoke the workload delegation.

Docker registers an opaque proof-token delegation before container start with expiry `ttl_seconds_active + 300`, but terminal `create_step_update` only updates task state and returns. Cleanup also only stops the refresher and removes the container/volumes. If a workload copied the proof token before exit, it can continue exchanging OBO access tokens until expiry. Kubernetes revokes the same delegation on terminal status.

contrib/auth/authentik/compose/implementation-details.md
```diff
@@ -115,23 +118,25 @@ trusts the local gateway certificate.
...
-service validates Authentik OIDC tokens, NeMo workload-exchange access tokens,
-and NeMo Scoped Access Keys, then returns trusted `X-NMP-Principal-*` and
-`X-NMP-Scopes` headers for Envoy to forward upstream.
+protected `/apis/` requests, Envoy calls NeMo Platform's `/apis/auth/ext-authz`
+endpoint with the presented bearer token. The auth service validates Authentik
+OIDC tokens, NeMo Platform workload-exchange access tokens, and NeMo Platform
+Scoped Access Keys, then returns trusted `X-NMP-Principal-*` and `X-NMP-Scopes`
+headers for Envoy to forward upstream. The public `/apis/auth/authenticate`
+endpoint remains a JSON diagnostic/API endpoint for direct callers and is not
+the configured Envoy callout.
```
mckornfield, line 135: envoy callout?

contrib/auth/authentik/compose/implementation-details.md
```diff
@@ -158,9 +163,21 @@ NMP_WORKLOAD_IDENTITY_TOKEN_FILE=/var/run/secrets/nemo-platform/workload/token
...
 The SDK reads that file and sends an RFC 8693 token exchange request to the
-NeMo auth service through the gateway. The NeMo auth service validates the
-Authentik subject token, mints a NeMo-signed access token, and returns it to the
-workload. The workload uses that exchanged token for normal NeMo API calls.
+NeMo Platform auth service through the gateway. The Docker backend registered
+an internal workload delegation row before the container started. The
+NeMo Platform auth service validates the proof token, checks the matching row,
+mints a NeMo Platform-signed delegated access token, and returns it to the
+workload. The access token uses the captured job submitter as the top-level
+subject and the Docker workload as the RFC 8693 `act.sub` actor.
```
mckornfield, line 171: "Docker workload" is specific to local right? should this just say "container" here and below? since in kube it won't be docker

contrib/auth/authentik/README.md
```diff
@@ -88,8 +89,10 @@ The 2-minute CLI access-token lifetime is a local demo/testing setting so token
 refresh is easy to observe. Do not use it as a production default; use a longer
 value such as `hours=1` outside the refresh demonstration.
 
-In the Docker Compose runtime, Authentik issues the demo workload subject token,
-but it does not accept the RFC 8693 token exchange grant directly. The Docker
-backend refreshes the Authentik subject token file, the SDK posts that token to
-the NeMo auth service, and the gateway trusts the NeMo auth service JWKS for
-exchanged workload access tokens.
+In the Docker Compose runtime, Authentik authenticates users and controller
```
mckornfield: hehe `Authentik authenticates` sounds awkward

contrib/auth/authentik/README.md
```diff
@@ -88,8 +89,10 @@ The 2-minute CLI access-token lifetime is a local demo/testing setting so token
...
+service principals, but managed Docker job OBO uses a NeMo Platform-owned
+opaque workload proof token. The Docker backend writes that proof token into the
```
mckornfield, line 94: maybe for my docker comment above, this is only for the authentik flow, and not for the "regular" flow?

docs/set-up/config-reference.mdx
```diff
@@ -301,26 +301,8 @@ jobs:
...
-        # Controller environment variable that contains the Docker demo issuer password grant shared secret. | default: 'AUTHENTIK_WORKLOAD_IDENTITY_PASSWORD'
-        password_env_var: AUTHENTIK_WORKLOAD_IDENTITY_PASSWORD
-        # OAuth scope for the Docker demo issuer.
-        scope:
-        # Fallback subject-token lifetime when the Docker demo issuer response omits expires_in.
-        subject_token_ttl_seconds: 600
-        # Seconds before subject-token expiry when the Docker refresher issues a replacement token. | default: 60
-        refresh_margin_seconds: 60
+      # Docker workload identity configuration.
+      workload_identity: {}
```
mckornfield: is there a reason we're not "strongly typing" this config anymore?

packages/nmp_common/src/nmp/common/auth/jwt.py
```diff
@@ -36,6 +44,7 @@ class TokenClaims:
     groups: list[str]
     scopes: list[str]
     raw_claims: dict
+    actor: Optional[ActorClaims] = None
```
mckornfield: Probably a nit, but should we prefer
```suggestion
    actor: ActorClaims | None = None
```

mckornfield: also below when we use optional

packages/nmp_common/src/nmp/common/auth/jwt.py
```diff
@@ -134,6 +152,39 @@ def _extract_token_claims(self, claims: dict) -> Optional[TokenClaims]:
...
+        """Extract normalized groups from configured or provider-specific claims."""
+        for claim_name in [self.config.oidc.groups_claim, "cognito:groups"]:
+            if claim_name not in claims:
+                continue
+            value = claims[claim_name]
+            if isinstance(value, str):
+                return [group.strip() for group in value.split(",") if group.strip()]
+            if isinstance(value, list):
+                return [str(group).strip() for group in value if str(group).strip()]
+            return []
```
mckornfield: do we need this return? saves us one loop I guess?

services/core/auth/src/nmp/core/auth/api/v2/authenticate.py
```diff
@@ -88,23 +97,106 @@ def _scopes_from_claims(claims: dict[str, object]) -> list[str]:
...
 
 
-def _response_from_claims(
-    claims: TokenClaims,
-    token_kind: ResolvedTokenKind,
-) -> AuthenticateResponse:
-    jti = claims.raw_claims.get("jti")
+def _normalize_jwk_public_material(jwk: dict[str, object]) -> tuple[tuple[str, str], ...] | None:
+    kty = jwk.get("kty")
+    if kty == "RSA":
```
mckornfield: match/case?

services/core/auth/src/nmp/core/auth/api/v2/authenticate.py
```diff
@@ -88,23 +97,106 @@ def _scopes_from_claims(claims: dict[str, object]) -> list[str]:
...
+        if isinstance(jwk, dict) and _normalize_jwk_public_material(jwk) == nemo_material:
+            raise TokenIssuerBoundaryMisconfigurationError(
+                "OIDC JWKS must not contain the NeMo workload token signing key"
+            )
+
+
+def _actor_from_claims(claims: dict[str, object]) -> ActorClaims | None:
+    actor_claims = claims.get("act")
+    if not isinstance(actor_claims, dict):
+        return None
```
mckornfield, line 176: for any of these (i.e., chains where we return None), is it worth having debug logging or nah?

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
```diff
@@ -345,6 +418,37 @@ async def decode_subject_token(self, config: AuthConfig, subject_token: str, aud
...
+        store: WorkloadDelegationStore,
+        name: str,
+        *,
+        retry: bool,
+    ) -> WorkloadDelegationEntity | None:
+        """Fetch a delegation row by name with a bounded retry for Kubernetes startup races."""
+        if not retry:
+            return await store.get(name)
+
+        retry_attempts = (
```
mckornfield, line 438: lol quite the calculation

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
```diff
@@ -427,6 +551,14 @@ def _allowed_subject_issuers(config: AuthConfig) -> set[str]:
     return {issuer for issuer in config.oidc.workload_subject_issuers if issuer}
 
 
+def _strip_private_bound_reference_claims(claims: dict[str, Any]) -> dict[str, Any]:
+    sanitized = dict(claims)
+    sanitized.pop(_BOUND_REFERENCE_NAME_CLAIM, None)
```
mckornfield: is there a prefix we can strip by as well or no? just curious if we'll accidentally have missed one

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -308,62 +256,212 @@ def init(self) -> None:
...
+                        extra={"delegation_name": delegation_name},
+                    )
+            raise
+        return delegation_name
+
+    def _revoke_workload_delegation_after_failed_start(self, delegation_name: str | None) -> None:
+        if not delegation_name:
+            return
+        try:
+            self._revoke_workload_delegation(delegation_name)
```
mckornfield: should this try/except logging be reusable? I guess maybe the log message is it?

mckornfield (commented):
nothing major I saw, can also pepper with the triple review

contrib/auth/authentik/compose/implementation-details.md
ironcommit: Envoy does a callout to check the key, to make sure it hasn't expired.

contrib/auth/authentik/compose/implementation-details.md
ironcommit: Docker and K8s are handled separately, and they have different requirements.  K8s is the more important one, but Docker may be required in the future.  The docs here are specific to the compose example which implements something similar to what quickstart used to be, just using docker-compose.  No k8s here.

contrib/auth/authentik/README.md
ironcommit: This doc was intended to cover both the compose and k8s examples.

services/core/auth/src/nmp/core/auth/api/v2/authenticate.py
ironcommit: I don't think so, it's a very common case.  Anything that isn't OBO basically.

docs/set-up/config-reference.mdx
ironcommit: removed

services/core/auth/src/nmp/core/auth/api/v2/authenticate.py
ironcommit: updated

packages/nmp_common/src/nmp/common/auth/jwt.py
ironcommit: updated

packages/nmp_common/src/nmp/common/auth/jwt.py
```diff
@@ -134,9 +159,43 @@ def _extract_token_claims(self, claims: dict) -> Optional[TokenClaims]:
...
+            actor=actor,
+        )
+
+    def _extract_groups_from_claims(self, claims: JsonObject) -> list[str]:
+        """Extract normalized groups from configured or provider-specific claims."""
+        for claim_name in [self.config.oidc.groups_claim, "cognito:groups"]:
+            if claim_name not in claims:
+                continue
+            value = claims[claim_name]
+            if isinstance(value, str):
```
anastasia-nesterenko: can re-use `groups_from_claim`?

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
```diff
@@ -308,62 +248,206 @@ def init(self) -> None:
...
+    def _revoke_workload_delegation(self, delegation_name: str) -> None:
+        store = create_authenticated_workload_delegation_store(
+            self._nmp_sdk,
+            missing_headers_message="Docker workload delegation requires authenticated SDK headers",
+        )
+        store.revoke(delegation_name)
+
+    def _provision_docker_workload_proof_token(self, delegation_name: str) -> DockerWorkloadProofToken:
+        token, token_hash = create_opaque_docker_proof_token(delegation_name)
+        expires_at = workload_delegation_expires_at(
```
anastasia-nesterenko, line 315: If I understand correctly, the token might be active during the whole job run that might take hours. Do we have/need any check or refresh if the proof token is misused before it expires/job fails?

services/core/auth/src/nmp/core/auth/api/v2/workload_token_exchange.py
ironcommit: I have removed this and replaced with something less hacky.

packages/nmp_common/src/nmp/common/auth/jwt.py
ironcommit: yup

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
ironcommit: cleaned up

services/core/jobs/src/nmp/core/jobs/controllers/backends/docker.py
ironcommit: Yes, I thought about this, and there are a few options with tradeoffs.
- follow k8s convention and have the token rotated on a regular basis.  This is the previous implememtation for docker, but I have removed it because I don't think the complexity is justified for the docker usecase.  This required that the controller have extra priviledges that I want to remove.
- Mint the opaque proof token valid for the life of the container, revoked on terminal state.  This is the replacement. It's simpler, but has the drawback of. widening the window of abuse for an exfiltrated token.

There is room for improvement here for sure.  But given that docker isn't actually a fully supported solution yet, I think this is good enough for now.  K8s is the more security sensitive deployment.  Future improvements might involve adding back the rotated tokens, and scoping the tokens to the container id.

But really, if the job is able to exfiltrate the token, then non of this will help at all, a compromised job is compromised for the entire duration of the job anyways.  So I think in this case simple is best.

contrib/auth/authentik/README.md
ironcommit: updated

mckornfield (approved):

packages/nmp_common/src/nmp/common/auth/jwt.py
ironcommit: updated

anastasia-nesterenko (approved):
👏

PR #1107, opened 2026-08-05
refactor(eval-author): simplify dataset augmentation

aleckhoury:
## Summary

Refactors Eval Author’s portable metric handoff and moves Experimentalist to metric-contract-driven optimization. Regression metrics are selector context rather than a hard eligibility gate, while candidates must still report every Pareto objective.

## Changes

- Stage and hand off authored Eval Author datasets and metric contracts.
- Use objective-aware Pareto ranking, direction-aware peer analysis, and LLM-guided soft regression constraints.
- Safely replace Harbor task directories only after validating the staged copy.
- Update metric configuration documentation and focused regression coverage.

## Type of Change

- [x] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Documentation updated for user-visible behavior

## Verification

- [x] Pull request title follows the repository’s Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [x] Targeted tests pass
- [x] No secrets, API keys, or credentials are included
- [ ] `uv run pre-commit run -a` passes — blocked by 38 existing type-check diagnostics from rebased `main` files outside this PR.

Targeted validation:

- `.venv/bin/pytest plugins/nemo-experimentalist/tests/test_metric_contract.py plugins/nemo-experimentalist/tests/experimentalist/test_dataset_staging.py plugins/nemo-experimentalist/tests/test_experimentalist_analyzer.py plugins/nemo-experimentalist/tests/experimentalist/test_terminator.py -v` — 32 passed.
- `.venv/bin/pytest plugins/nemo-experimentalist/tests/test_experimentalist_backend.py -v` — 19 passed.
- Focused Ruff lint and formatting checks — passed.
- `uv run generate-config-docs && git diff --exit-code docs/set-up/config-reference.mdx` — passed.

PR #1195, opened 2026-08-10
fix(experimentalist): retain the baseline when no candidate beats it

schuellc-nvidia:
## Summary

Finalization picked the run's winner with `select_diverse_survivors(..., k=1)`. That agent is prompted to always include a candidate created in the current round, so at `k=1` the rule consumes the only slot and the round-0 baseline can never be returned. Before this change, every run shipped a diff — including runs where no candidate measured better than doing nothing. After it, finalization is deterministic: the Pareto front over the configured objectives, with the oldest candidate winning ties, so a run that finds no improvement keeps the baseline.

Winner selection and survivor selection are different questions. Picking a set to carry into the next round wants diversity and fresh blood; picking a winner wants the best measured result and no judgement call. Mid-loop survivor selection is untouched and still uses the diversity selector.

## Changes

- Add `select_winner_node()` in `experimentalist/components/loop.py`: `pareto_front` over `pareto_objectives`, then `min(front, key=round)`. Used by `_finalize` in place of the `select_diverse_survivors(..., 1, ...)` call.
- Preserve the existing objective-metric contract: `has_metric_dimensions` eligibility filtering and `pareto_objectives` ranking are unchanged.
- Leave the mid-loop `select_diverse_survivors` call site unchanged.
- Add `plugins/nemo-experimentalist/tests/experimentalist/test_winner_selection.py` (10 tests). This path had no coverage, which is how the regression shipped.

Preferring the lowest round only decides ties: a candidate that genuinely improves *dominates* its ancestor and removes it from the front before the tie-break is consulted, so it can never block a real gain. The previous behaviour was correct only by accident — `pareto_front` preserves input order and `agent-0` is inserted first, so ties happened to fall to the baseline. The rule is now stated explicitly rather than relying on dict insertion order.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal selection behaviour with no documented user-facing surface; the rationale is captured in the function docstring and the tests.

## Verification

Targeted validation:

```
uv run --frozen pytest plugins/nemo-experimentalist/tests/experimentalist/test_winner_selection.py -q
  → 10 passed

uv run --frozen pytest plugins/nemo-experimentalist/tests/ -q
  → 628 passed

uv run ruff check <loop.py> <test_winner_selection.py>          → All checks passed!
uv run ruff format --check <loop.py> <test_winner_selection.py> → 2 files already formatted
uv run --frozen ty check <loop.py>                              → All checks passed!

uv run pre-commit run -a
  → all hooks passed EXCEPT "Run uv lock with platform uv", which failed with:
       "uv.lock must be checked or updated with uv 0.9.14 ... Current uv: uv 0.9.30"
```

**The one failing hook is environmental and pre-existing, not caused by this change.** It was reproduced on pristine `main` with these changes stashed, and fails identically. This PR modifies no `pyproject.toml` and no `uv.lock`; the separate "Check for uv.lock drift" hook passed. It needs uv 0.9.14 locally, which this machine does not have. The `Verification` checkbox above is left unchecked rather than claiming a gate that did not pass.

End-to-end behaviour was also confirmed against a fixture whose expected outcome is known in advance: a single-round generalization scenario where all three agents tie at 1/3 on validation now finalizes with `winner=agent-0` (the baseline). Under the previous code the baseline was structurally ineligible, so a candidate that improved nothing would have been published as the winner.

## Provenance

The finalization behaviour corrected here was introduced in #1107, which also added `selector.py`. Before that change, finalization ranked with `pareto_front` and did not call an agent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved final candidate selection using objective comparisons across multiple metrics.
  * Ensured genuine improvements win over regressions while respecting maximized, minimized, and protected metrics.
  * Restored the original baseline for reliable regression comparisons, even if it ended early.
  * Prevented incomplete or incomparable candidates from displacing the baseline.
  * Added deterministic tie-breaking for consistent results.
  * Finalization now handles empty, single-candidate, and baseline-free results consistently.

### DIFF
--- a/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
+++ b/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
@@ -52,8 +52,10 @@ from nemo_experimentalist_plugin.experimentalist.components.holdout_utils import
     restore_heldout_splits,
 )
 from nemo_experimentalist_plugin.experimentalist.components.models import (
+    EvolutionNode,
     EvolutionTree,
     OptimizationType,
+    pareto_front,
     pareto_sort,
 )
 from nemo_experimentalist_plugin.experimentalist.components.proposer import Improvement, Proposer
@@ -127,6 +129,25 @@ def _coerce_optimization_type(optimization_type: str | None) -> OptimizationType
     if optimization_type in get_args(OptimizationType):
         return cast(OptimizationType, optimization_type)
     return None
+
+
+def select_winner_node(
+    eligible: list[EvolutionNode],
+    objective_function: list[MetricTarget],
+) -> EvolutionNode | None:
+    """Return the run's winner: non-dominated, oldest candidate wins ties.
+
+    Deliberately *not* ``select_diverse_survivors``. Picking a winner and picking a
+    set to carry into the next round are different questions. The selector is told to
+    always include a candidate created this round, so at ``k=1`` that rule consumes the
+    only slot and the round-0 baseline can never be returned -- meaning a run always
+    ships a diff, even one that measured no better than doing nothing.
+
+    Preferring the lowest round only decides ties: a candidate that genuinely improves
+    dominates its ancestor and removes it from the front before this is consulted.
+    """
+    front = pareto_front(eligible, lambda node: pareto_objectives(node.val_reward, objective_function))
+    return min(front, key=lambda node: node.round) if front else None
 
 
 def _with_insight_objective(
@@ -1786,17 +1807,8 @@ class EvolutionaryOptimizer(Agent):
         # Only survivors that actually have a validation reward are eligible winners.
         scored = [n for n in evolution_tree.nodes.values() if n.is_survivor and n.val_reward]
         eligible = [node for node in scored if has_metric_dimensions(node.val_reward, self.config.objective_function)]
-        ranked_nodes = pareto_sort(
-            eligible,
-            lambda node: pareto_objectives(node.val_reward, self.config.objective_function),
-        )
-        finalists = await self.select_diverse_survivors(
-            [node.candidate for node in ranked_nodes],
-            1,
-            self.config.objective_function,
-            self.config.regression_metrics,
-        )
-        best_id = finalists[0].label if finalists else None
+        best = select_winner_node(eligible, self.config.objective_function)
+        best_id = best.label if best else None
 
         restore_heldout_splits(self.working_dir)
 
--- a/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
+++ b/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
@@ -131,23 +131,60 @@ def _coerce_optimization_type(optimization_type: str | None) -> OptimizationType
     return None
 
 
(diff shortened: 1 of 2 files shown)


plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
```diff
@@ -129,6 +131,25 @@ def _coerce_optimization_type(optimization_type: str | None) -> OptimizationType
...
+    eligible: list[EvolutionNode],
+    objective_function: list[MetricTarget],
+) -> EvolutionNode | None:
+    """Return the run's winner: non-dominated, oldest candidate wins ties.
+
+    Deliberately *not* ``select_diverse_survivors``. Picking a winner and picking a
+    set to carry into the next round are different questions. The selector is told to
+    always include a candidate created this round, so at ``k=1`` that rule consumes the
+    only slot and the round-0 baseline can never be returned -- meaning a run always
+    ships a diff, even one that measured no better than doing nothing.
```
gaiadilorenzo: :nitpick: similar comment as usual, I would keep documentation short and stick to interface contracts rather than historic comments

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
```diff
@@ -129,6 +131,25 @@ def _coerce_optimization_type(optimization_type: str | None) -> OptimizationType
...
+
+    Preferring the lowest round only decides ties: a candidate that genuinely improves
+    dominates its ancestor and removes it from the front before this is consulted.
+    """
+    front = pareto_front(eligible, lambda node: pareto_objectives(node.val_reward, objective_function))
```
gaiadilorenzo: This chooses the best one even if regression metrics have worsened. I'd still keep it llm based or improve the selection with a hard constrain on regression metrics

plugins/nemo-experimentalist/tests/experimentalist/test_winner_selection.py
```diff
@@ -0,0 +1,142 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
gaiadilorenzo, line 1: this needs a test with multiple metrics (both objective and regression metrics)

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
schuellc-nvidia: ok, addressed

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/loop.py
schuellc-nvidia: Ok, addressed it with a hard constraint

plugins/nemo-experimentalist/tests/experimentalist/test_winner_selection.py
schuellc-nvidia: ok, addressed

gaiadilorenzo (approved):

PR #1198, opened 2026-08-10
feat(evaluator): add by-name aggregate lookup to AgentEvalSummary

SandyChapman:
## Summary

Reading a single aggregate from an eval run required a hand-written scan over a doubly-nested attribute (`next(s for s in result.summary.scores.scores if s.name == "...")`). This adds `score(name)` and a `scores_by_name` mapping view so the same read is one call, and converts the seven sites that were scanning by hand.

Linear: [AALGO-445](https://linear.app/nvidia/issue/AALGO-445/add-score-lookup-accessor-to-agentevalsummary-no-way-to-fetch-one)

## Changes

- `AggregatedMetricResult.score(name)` — returns the aggregate, raises `KeyError` on a miss. Placed here rather than only on the agent-eval summary so the dataset path (`EvaluationResult.aggregate_scores`) gains the same surface.
- `AggregatedMetricResult.scores_by_name` — mapping view for `in`, `.get()`, and iteration, for when a score's absence is a legitimate outcome. First-wins on a duplicate name, matching the `next(...)` scans it replaces.
- One-hop delegates on `AgentEvalSummary`.
- Converted 7 hand-rolled scans: the private `_score` helper plus 4 call sites in `test_evaluator.py`, one each in `test_gym_aggregate_scores.py`, `test_hermes_example.py`, two in `test_pass_at_k.py`, and the user-facing `examples/hermes/example.py`.

### Design notes

**A miss raises instead of returning `None`.** An unknown name is nearly always a typo or a metric that did not run. Both are bugs worth surfacing at the lookup, where the name is in hand, rather than as an `AttributeError` on `.mean` further downstream. `scores_by_name.get()` covers legitimate absence.

**The miss message leads with close matches.** A run with several metrics times pass@k carries dozens of names, so enumerating them all buries the answer exactly when it is most needed. It reports how many others exist so a wrong suggestion is not a dead end, and truncates the fallback listing at 10.

**Left alone deliberately.** `harbor_runtime.py` and `dashboard.py` iterate rather than look up by name; `plugin_examples.py` filters on two candidate names. Converting any of them would change behaviour, not just style.

**No schema impact.** `score` is a method and `scores_by_name` a property, so neither enters the JSON schema — verified against `AggregatedMetricResult`, `AgentEvalSummary`, and `EvaluationResult`. The committed OpenAPI spec is unchanged.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the new surface is documented in its own docstrings; there is no SDK reference page enumerating result accessors to update.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pytest packages/nemo_evaluator_sdk/tests/agent_eval packages/nemo_evaluator_sdk/tests/values` | 598 passed |
| `pytest .../test_summary_accessors.py` | 11 passed (new) |
| `ruff check` + `ruff format --check` on changed files | clean |
| `tools/lint/lint-python-types.sh` | exit 0, no diagnostics in changed files |
| JSON-schema drift check on the 3 affected models | no drift |
| `uv run pre-commit run -a` | **2 hooks failed for local-environment reasons — see below** |

**`uv run pre-commit run -a` — two failures, neither caused by this change:**

1. `uv-lock` — refuses to run because the local uv is 0.9.30 and the repo pins `>=0.9.14` for lockfile writes. This is a toolchain mismatch, not lock drift: the separate `Check for uv.lock drift` hook **passed**, and this branch touches no `pyproject.toml` or `uv.lock`.
2. `studio-lint-staged` — `mise` has no `pnpm` shim configured in this worktree. This branch touches no web files.

Every other hook passed, including `ruff`, `ty`, copyright headers, config-reference doc, and Helm docs. Flagging both rather than marking the gate green; happy to re-run under a corrected local toolchain if a reviewer wants the clean sweep.

## Summary by CodeRabbit

* **New Features**
  * Added convenient score lookup by name for evaluation summaries and aggregated results.
  * Added score mappings, membership checks, and optional lookups.
  * Improved missing-score errors with typo suggestions and available-name guidance.

* **Tests**
  * Expanded coverage for successful lookups, duplicate names, empty results, and long score lists.
  * Updated examples and tests to use streamlined score accessors.
  * Added validation for available-name reporting and first-match behavior.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
@@ -419,9 +419,11 @@ class AggregateScalarScore(AggregateScoreBase):
 AggregateScore = AggregateRangeScore | AggregateRubricScore | AggregateScalarScore
 
 
-#: Names listed in a lookup-miss message when nothing resembles what was asked for. Enough to spot a
-#: naming-convention mistake, few enough to read; a run with several metrics times pass@k can carry
-#: dozens, and a wall of them buries the answer rather than giving it.
+#: How many names a lookup-miss message lists when nothing resembles what was asked for. A judgement
+#: call rather than a measured optimum -- enough to show the naming convention, few enough to stay
+#: readable, since a run with several metrics times pass@k can carry dozens. Only this fallback is
+#: truncated; a near-miss is surfaced by similarity, so finding the name you meant never depends on
+#: where it happens to fall alphabetically.
 _MISS_NAME_LIMIT = 10
 
 
@@ -462,15 +464,22 @@ class AggregatedMetricResult(BaseModel):
         raise KeyError(self._unknown_score_message(name))
 
     def _unknown_score_message(self, name: str) -> str:
-        """Explain a lookup miss, leading with near-misses when the name looks like a typo."""
+        """Explain a lookup miss, leading with near-misses when the name looks like a typo.
+
+        "Close" is :func:`difflib.get_close_matches`: SequenceMatcher (Ratcliff/Obershelp) similarity
+        of at least 0.6, best three first. That is a subsequence-overlap ratio, not an edit distance.
+
+        Listing every name alphabetically would be simpler, and is the better answer if the list is
+        left whole. Truncating one is what breaks it -- the name a caller meant is not reliably in
+        the first :data:`_MISS_NAME_LIMIT`, since a typo'd ``view.solved`` sits behind a page of
+        ``gym_reward.*`` in a run carrying pass@1..8 for two metrics.
+        """
         # Deduplicated: names are expected unique, but runner-contributed extras are appended as-is,
         # and a repeat would otherwise be suggested twice, listed twice, and counted twice in the
         # "N other aggregates" tally -- making a collision look like two distinct near-misses.
         available = sorted({score.name for score in self.scores})
         if not available:
             return f"no aggregate score named {name!r}: this result has no aggregates at all"
-        # Suggestions beat enumeration for the common case (a typo, or the wrong pass@k), and stay
-        # useful when a run carries dozens of names.
         close = get_close_matches(name, available, n=3)
         if close:
             suggestions = ", ".join(repr(match) for match in close)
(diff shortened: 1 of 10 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
```diff
@@ -417,12 +419,75 @@ def _serialize(self, handler):
 AggregateScore = AggregateRangeScore | AggregateRubricScore | AggregateScalarScore
 
 
+#: Names listed in a lookup-miss message when nothing resembles what was asked for. Enough to spot a
```
arpitsardhana: this is bit confusing description with added ":" Also why 10 is sweet spot for that?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
```diff
@@ -417,12 +419,75 @@ def _serialize(self, handler):
...
+        """Explain a lookup miss, leading with near-misses when the name looks like a typo."""
+        # Deduplicated: names are expected unique, but runner-contributed extras are appended as-is,
+        # and a repeat would otherwise be suggested twice, listed twice, and counted twice in the
+        # "N other aggregates" tally -- making a collision look like two distinct near-misses.
+        available = sorted({score.name for score in self.scores})
+        if not available:
+            return f"no aggregate score named {name!r}: this result has no aggregates at all"
+        # Suggestions beat enumeration for the common case (a typo, or the wrong pass@k), and stay
+        # useful when a run carries dozens of names.
+        close = get_close_matches(name, available, n=3)
```
arpitsardhana, line 483: what constitutes close match? Is it calcuting lavenshein distance? 

I am wondering rather than lookup, we can just offer the list of metrics sorted by name?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
SandyChapman: The `#:` is the Sphinx attribute-doc marker already used across this package (60 uses, e.g. `_PASS_VALUE` in `agent_eval/results.py`) — happy to drop it here if you'd rather, though it'd be the odd one out; 10 is a judgement call, not a measured optimum, and the comment now says so.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
SandyChapman: It's `difflib.get_close_matches` — SequenceMatcher (Ratcliff/Obershelp) overlap ≥ 0.6, not an edit distance. Updated the docs here.

arpitsardhana (approved):

PR #1201, opened 2026-08-10
feat(evaluator): give AgentEvalResult the dataset path's display surface

SandyChapman:
## Summary

`EvaluationResult` ships `format_summary()`, `print_summary()`, `to_records()`, `to_table()`, `to_pandas()`, and `__str__`. `AgentEvalResult` had none of them, so anyone inspecting an agent-eval run wrote their own formatting — two result types in one SDK with unrelated ergonomics. This adds the same six methods, with the same signatures and the same `view` values.

Linear: [AALGO-447](https://linear.app/nvidia/issue/AALGO-447/agentevalresult-has-no-display-surface-no-print-summaryformat)

## Changes

- `to_records(view="rows"|"aggregate")` on `AgentEvalResult`, plus `to_table`, `to_pandas`, `format_summary`, `print_summary`, `__str__`
- Row projection, error-text and diagnostics columns, header, and failed-score section as module-level helpers alongside the existing ones

### Design: no mixin

The ticket proposed extracting a shared mixin. `BenchmarkEvaluationResult` (`values/multi_metric_results.py`) already faced this and answered it differently — import the module-level helpers, implement the methods directly. The genuinely shared logic (`format_table`, `summary_aggregate_record`, `serialize_value`, `flatten_dict`) is already free functions, so a mixin would share method *names* while every body still needed overriding, and would add a third pattern to unify two that already agree. This follows the established one.

Reversible, but reversing it means refactoring all three result types rather than just this one.

### What a "row" is here

One record per metric score. The fan-out is preserved rather than collapsed: `task_id` and `trial_id` are columns, so a consumer can still group by task — which is what pass@k depends on. The aggregate view matches the dataset path exactly (percentiles flattened, histograms as JSON strings).

The error section is agent-eval's own, because it has a distinction the dataset path cannot make: a failed **trial** is an attempt the agent is answerable for; a failed **metric** is a measurement that never happened. Both arrive as `FAILED`.

### One deliberate divergence — `to_table` column union

`pa.Table.from_pylist` takes its schema from the **first record alone**. Here `error` and `diagnostics.*` appear only on failed scores, so a run whose first score passed exported a table missing exactly the columns you needed — while `to_pandas`, which unions keys, included them. Verified: 5 columns vs 7 on the same run.

Columns are now unioned before the table is built. **Both siblings have the same latent behaviour.** Matching a known-lossy export seemed worse than diverging and saying so — but that makes this the odd one out until `EvaluationResult` and `BenchmarkEvaluationResult` get the same fix. Happy to file that follow-up, or to drop this and match the siblings if reviewers would rather keep the three identical.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the new surface is documented in its own docstrings and mirrors an existing documented API; there is no SDK reference page enumerating result methods to update.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pytest packages/nemo_evaluator_sdk/tests` | 1464 passed |
| `pytest .../test_result_display.py` | 16 passed (new) |
| `ruff check` + `ruff format --check` on changed files | clean |
| `tools/lint/lint-python-types.sh` | exit 0, no diagnostics in changed files |
| Mutation check (14 mutants against the new tests) | 14/14 caught |
| `uv run pre-commit run -a` | **2 hooks failed for local-environment reasons — see below** |

Three adversarial review rounds were run before this was opened. Round 1 surfaced three tests that passed against a broken implementation — most notably, the failed-trial/failed-metric assertion still passed when the two labels were **swapped** — plus an unreachable `if count` guard, now removed. Round 2 left one survivor (`max_error_rows` defaulting to `max_rows` was documented but untested). Round 3: 14/14 caught, zero survivors.

**`uv run pre-commit run -a` — two failures, neither caused by this change:**

1. `uv-lock` — refuses to run because the local uv is 0.9.30 while the repo pins `>=0.9.14` for lockfile writes. A toolchain mismatch, not lock drift: the separate `Check for uv.lock drift` hook **passed**, and this branch touches no `pyproject.toml` or `uv.lock`.
2. `studio-lint-staged` — `mise` has no `pnpm` shim in this worktree. This branch touches no web files.

Every other hook passed, including `ruff`, `ruff format`, `ty`, copyright headers, config-reference doc, Helm docs, and the plugin import-boundary check.

## Summary by CodeRabbit

- **New Features**
  - Added options to export evaluation results as row-based or aggregate records, tables, and pandas dataframes.
  - Added formatted summaries, console printing, and compact string representations.
  - Improved visibility into score outputs, diagnostic severity, metadata, errors, and failed trials or metrics.
  - Added controls for limiting displayed rows and error details.

- **Bug Fixes**
  - Improved consistency of exported data and diagnostic information across supported formats.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
@@ -5,19 +5,35 @@
 
 from __future__ import annotations
 
+import json
 import math
 from collections.abc import Sequence
 from datetime import datetime
 from pathlib import Path
+from typing import Any
 
-from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEvalTaskScore, is_trial_failure
+from nemo_evaluator_sdk.agent_eval.scores import (
+    AgentEvalDiagnosticSeverity,
+    AgentEvalScoreStatus,
+    AgentEvalTaskScore,
+    is_trial_failure,
+)
 from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalTask, SemanticReducer, ViewSignal
 from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial, RunnerInfo
 from nemo_evaluator_sdk.metrics.aggregation import compute_percentiles
 from nemo_evaluator_sdk.metrics.protocol import MetricOutput
 from nemo_evaluator_sdk.metrics.utils import metric_type_name
 from nemo_evaluator_sdk.values.protocol import BooleanValue, ContinuousScore
-from nemo_evaluator_sdk.values.results import AggregatedMetricResult, AggregateRangeScore, AggregateScore
+from nemo_evaluator_sdk.values.results import (
+    AggregatedMetricResult,
+    AggregateRangeScore,
+    AggregateScore,
+    ResultView,
+    flatten_dict,
+    format_table,
+    serialize_value,
+    summary_aggregate_record,
+)
 from pydantic import BaseModel, ConfigDict, Field
 
 #: Metric-output value schemas eligible for pass@k (a per-attempt "did it pass?" signal). Labels,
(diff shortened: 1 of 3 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -179,6 +195,232 @@ def persist(self, output_dir: str | Path | None = None, *, write_dashboard: bool
...
+    Both arrive as ``FAILED``, but they mean different things to a reader: a failed trial is an
+    attempt the agent is answerable for, a failed metric is a measurement that never happened. The
+    dataset path has no equivalent distinction to make, so this section is agent-eval's own rather
+    than a reuse of :func:`format_error_details`.
+    """
+    failed = [score for score in scores if score.status is AgentEvalScoreStatus.FAILED]
+    if not failed:
+        return []
+
+    shown_limit = len(failed) if max_error_rows is None else max(0, max_error_rows)
```
arpitsardhana, line 429: why max(0, max_error_rows) ? is it expected to be negative number or were you trying to compare len(failed) and max_error_rows

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -179,6 +195,232 @@ def persist(self, output_dir: str | Path | None = None, *, write_dashboard: bool
...
+            max_error_rows: Maximum number of failed scores included in the error-details section.
+                Defaults to ``max_rows``.
+
+        Returns:
+            Multi-line summary string suitable for terminal/notebook display.
+        """
+        if max_error_rows is None:
+            max_error_rows = max_rows
+        aggregate_records = [summary_aggregate_record(score) for score in self.summary.scores.scores]
+        preview = [_score_preview_record(score) for score in self.scores[:max_rows]]
```
arpitsardhana, line 290: This has potential for exception with max_rows is larger than size of self.scores. max_rows need to default to min(10, len(self.scores)

arpitsardhana (approved):
minor nits comments

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -179,6 +195,232 @@ def persist(self, output_dir: str | Path | None = None, *, write_dashboard: bool
...
+    return record
+
+
+def _score_record(score: AgentEvalTaskScore) -> dict[str, Any]:
+    """Full export record for one score: preview columns plus error text and diagnostics."""
+    record = _score_preview_record(score)
+    if error_text := _score_error_text(score):
+        record["error"] = error_text
+    record.update(_score_diagnostics_columns(score))
+    return record
```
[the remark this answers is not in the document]
SandyChapman: Fixed — `id`, `run_id`, and flattened `metadata.*` now in the row export; the summary preview stays narrow on purpose, matching the dataset path's `to_records` vs `summary_row_base_record` split.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
SandyChapman: Guards a *negative* limit, which slicing would read as an offset from the end (`failed[:-2]` shows all but the last two rather than none) — over-large limits need no guard, and it mirrors `format_error_details` in the dataset path.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
SandyChapman: Slicing past the end returns the whole list rather than raising — `format_summary()`'s default of 10 already runs over 3-score results throughout the test suite — but I've added an explicit test at `max_rows=1000` so it's pinned.

PR #1202, opened 2026-08-10
refactor(evaluator)!: collapse the dataset entrypoints into evaluate_dataset

SandyChapman:
## Summary

The backend contract carried two dataset entrypoints that differed only in metric arity and return type — `evaluate` took one metric and returned an `EvaluationResult`, `evaluate_benchmark` took a list and returned a `BenchmarkEvaluationResult`. One metric is the degenerate case of several, so this collapses them into a single `evaluate_dataset(metrics=[...])`.

This is deliberately a subtraction: 210 insertions against 781 deletions. It also frees the name `evaluate`, which a follow-up will use for the taskset entrypoint — nothing claims it here, so `evaluate_dataset` is the only method on the contract after this lands.

## Related Issue

_None._

## Changes

- `EvaluationBackend` / `SyncEvaluationBackend`: `evaluate` and `evaluate_benchmark` replaced by `evaluate_dataset`. Same for `LocalBackend` and the sync adapter.
- `Evaluator.run` / `run_sync`: the three single-metric overloads on each are gone, leaving the three that discriminate on config and target. Return type narrows to `BenchmarkEvaluationResult`.
- `BenchmarkEvaluationResult.__str__`, matching `EvaluationResult`. Without it, collapsing to one result type would have quietly regressed `print(result)` to a pydantic dump for every caller.
- The plugin's evaluate job drops `metrics if len(metrics) > 1 else metrics[0]`, which existed only to reach the singular overload.
- A backend's flavour is now decided by one method, so the async/sync/mixed three-way collapses to two — a lone method is either a coroutine function or it is not. `_MixedBackend` and its test go with it.
- Call sites updated to `metrics=[...]` across tests, examples, skill references and the walkthrough notebook.

## Type of Change

- [ ] Documentation only

## Quality Gates

Existing tests were retargeted rather than duplicated: the backend fakes now carry one dataset result instead of two, and `test_api.py` asserts the multi-metric shape, whose row columns are metric-qualified (`output.exact-match.exact-match`) and whose preview is per-metric. Two tests were removed because their subjects no longer exist — the single-vs-sequence dispatch fallthrough, and the mixed sync/async client.

## Verification

| Command | Result |
| --- | --- |
| `uv run --frozen pytest packages/nemo_evaluator_sdk/tests plugins/nemo-evaluator/tests docs/evaluator -q` | 2225 passed, 24 skipped |
| `uv run --frozen pytest e2e -q` | collected clean; 238 skipped without a live platform |
| `uv run ruff check` / `ruff format --check` | passed |
| `bash tools/lint/run-ty-check.sh` | passed |
| `make vendor` | vendored copy matches source |
| `uv run pre-commit run -a` | all hooks pass except the two below |

Two hooks fail on this machine only, neither related to the diff:

- `Run uv lock with platform uv` requires uv 0.9.14 exactly and exits on the version check before reading a file. `Check for uv.lock drift` passes and no `pyproject.toml` changed.
- `Run UI lint-staged` fails with `mise ERROR No version is set for shim: pnpm` in a fresh worktree. This PR changes nothing under `web/`.

## Notes for reviewers

Two things surfaced only once this was split out of a larger branch:

- `is_metric` / `is_metric_sequence` are **not** dead yet. The plugin's `bundle_metrics_for_spec` still dispatches on single-vs-sequence, so they survive until the plugin's own API goes plural.
- Dropping `_is_sync_backend` cost the `TypeGuard` narrowing on the final branch, so it stays, with an unreachable `else` that exists so the checker can narrow.

## Summary by CodeRabbit

* **New Features**
  * Evaluation now consistently accepts one or more metrics and returns a unified benchmark result.
  * Added concise result summaries with previews of evaluated rows.
  * Exposed the benchmark result type through the SDK’s public API.
* **Documentation**
  * Updated SDK, plugin, walkthrough, and execution examples to use collection-based metrics.
* **Tests**
  * Expanded coverage for multi-metric evaluation, synchronous and asynchronous execution, result formatting, and backend validation.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/execution/evaluator.py
```diff
@@ -413,7 +269,7 @@ def run_sync(
 
     def run_sync(
         self,
-        metrics: Metric | Sequence[Metric],
+        metrics: Sequence[Metric],
```
SandyChapman, line 272: This PR gets rid of the single metric eval mode (in favour of the multi-metric call).

packages/nemo_evaluator_sdk/tests/execution/backends/local/test_backend.py
```diff
@@ -45,7 +45,7 @@ async def test_delegates_to_sdk_evaluate_benchmark_with_unique_metric_keys(self,
         )
         metrics = [DuplicateMetric(), DuplicateMetric()]
 
-        result = await backend.evaluate_benchmark(metrics=metrics, dataset=dataset, params=params)
+        result = await backend.evaluate_dataset(metrics=metrics, dataset=dataset, params=params)
```
SandyChapman, line 48: We rename `evaluate_benchmark` to `evaluate_dataset` to better align with actual difference.

ngoncharenko (approved):

PR #1196, opened 2026-08-10
refactor(evaluator)!: resolve the Gym CLI from PATH instead of a configured checkout

SandyChapman:
## Summary

`GymRuntimeConfig` no longer carries `gym_root` or a `gym_bin` path — the runner resolves `gym` from PATH, so NeMo Gym must be installed in the same environment as the SDK. Before, a caller had to name a Gym checkout and the runner derived the executable from `<gym_root>/.venv/bin/gym`; now there is no path configuration at all, and a missing CLI fails with an actionable message before the run starts instead of an `ENOENT` partway through.

Two reasons the old shape does not survive. These runner configs become **serialized job specs** when Gym runs as a governed platform job, and a path into somebody's checkout or venv means nothing on the other side of that boundary. And a checkout is **no longer required**: NeMo-Gym now ships its component trees in the wheel, so `pip install nemo-gym` resolves environments and their example data with nothing on disk.

## Related Issue

Tracked in Linear as AALGO-485 (validating the Gym runner across a representative sample of built-in environments); this is the configuration cleanup that came out of it. No GitHub issue.

## Changes

- **`gym_runtime.py`** — remove `gym_root` and `gym_bin` from `GymRuntimeConfig`. Add a module-level `_gym_executable()` that resolves `_GYM_CLI` via `shutil.which`, mirroring how `CodexCliAgentRuntime` already resolves its own CLI, and raises with install guidance when absent.
- **Subprocess cwd** — no longer set from config. The `gym env start` and `gym eval run` subprocesses inherit this process's working directory, which is where Gym looks for the gitignored `env.yaml` holding the collector's credentials before falling back to its install root. Running from a Gym checkout still makes its components take precedence, so that capability survives without a field.
- **`examples/gym/run_gym_eval.py`** — drop `--gym-root`; default the dataset to the packaged `resources_servers/<name>/data/example.jsonl` via `importlib.resources`, with an actionable error when Gym is not importable.
- **`examples/gym/README.md`** — lead with `pip install nemo-gym`. The previous text asserted "a working NeMo Gym checkout is required — Gym resolves its environments from the repo, not from a package install", which is no longer true.
- **`examples/gym/inspect_results.py`** — update the usage example.
- **`tests/agent_eval/test_run_metadata.py`** — drop `gym_root` from the two `GymRuntimeConfig` constructions. `runner_info()` output is unchanged, so persisted `metadata.json` provenance keeps its shape.
[...]

PR #1203, opened 2026-08-10
feat(evaluator): pre-flight Gym config with `gym env validate` and take overrides as a dict

SandyChapman:
## Summary

Makes the Gym runner's preconditions checkable before a run commits to anything. It now runs `gym env validate` before starting servers, takes `env_overrides` as a nested dict instead of pre-serialized Hydra strings, and stops Gym littering the caller's working directory.

Before, a config mistake surfaced only after a Ray cluster and several uvicorn servers were up, as a readiness timeout of up to `startup_timeout_s` (240 s) whose message said nothing about the cause. Now it fails in about a second with Gym's own diagnosis.

Follows #1196 (merged). Rebased onto `main`; the two commits here are the whole diff.

## Related Issue

Tracked in Linear as AALGO-485. No GitHub issue.

## Changes

- **Validate before starting.** `_run_two_step` runs `gym env validate` with the identical selection arguments it then passes to `gym env start`, raising with Gym's report on rejection. Gym does not publish what configuration an environment requires — the typed `*ResourcesServerConfig` classes cover behavioural knobs, while model wiring lives in per-environment YAML under names that vary — so validate is the only way to find out short of running. Bounded at 120 s; output captured to `gym_validate.log` in the run's work dir.
- **`env_overrides` is now `dict[str, Any]`**, flattened to Hydra's `++a.b=1` grammar at invocation. Hydra's grammar is typed, so serialization is not `str()`: unquoted, `true` parses as a bool, `null` as `None`, `1.5` as a float, `a,b` as a *ChoiceSweep*, and `A[B` does not parse at all. Strings are single-quoted (only `'` escaped — Hydra does not decode `\\` inside quotes), and a value ending in a backslash raises rather than emitting something unparseable. Interpolations survive quoting, so `${policy_base_url}` still resolves. Redaction moved with it and matches on the full dotted path, so nesting cannot hide a credential behind an innocuous leaf name.
- **`hydra.run.dir` redirected** under the run's work dir. Gym is a Hydra app and writes a timestamped run directory per invocation, defaulting to `outputs/<date>/<time>/` under the current directory; since the subprocesses inherit this process's cwd so Gym can find `env.yaml`, the default littered wherever the caller ran from. Applies to every Gym entry point, `gym list` included.
- **Corrected the missing-CLI message.** It told the user to install Gym into this environment, which cannot work: Gym imports Ray at module load, and nemo-platform excludes Ray by constraint over an unfixed CVE (root `pyproject.toml`, `"ray; sys_platform == 'never'"`). Verified — `pip install nemo-gym` here yields `ModuleNotFoundError: No module named 'ray'` on every `gym` invocation. It now says to install Gym in its own environment and put that `bin` on PATH.
- **`_selection_args` extracted** from `_run_two_step`, so "validate and start receive identical arguments" is a property a test can assert rather than a comment to trust.
- **New `test_gym_environment_coverage.py`** — drives the real CLI against five environments spanning Gym's dependency and wiring categories.
- **Nine unit tests** in `test_gym_runtime.py` for the pre-flight and override serialization.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the example and its README were updated in #1196; the runner's module docstring is updated in this commit.

## Verification

```
pytest packages/nemo_evaluator_sdk/tests/agent_eval/ -q  → 573 passed, 10 skipped
ruff check packages/nemo_evaluator_sdk/                  → All checks passed!
ruff format --check packages/nemo_evaluator_sdk/         → 230 files already formatted
ty check .../agent_eval/runtimes/gym_runtime.py          → All checks passed!
make vendor                                              → vendored copy in sync
```

**The new tests were checked by mutation, not by coverage percentage.** Deleting the `hydra.run.dir` line fails only `test_selection_redirects_hydra_output_under_the_run_work_dir`; dropping the `None → "null"` branch fails only `test_hydra_scalars_use_hydra_spellings_not_python_ones`. `_validate_config` is driven against a stub `gym` that records its argv and exits how the test wants, so it needs no Gym install and covers both the success path (argv plus log written) and the failure path (raises carrying Gym's report).

The coverage sweep against a real Gym install (Gym in its own venv, its `bin` on PATH):

```
  mcqa[baseline]              validate ✓
  gpqa_diamond[tiny]          validate ✓
  gdpval[heavy-cpu-judge]     validate ✓  (with the five overrides it requires)
  legal_agent_bench[docker]   validate ✓
  wmt_translation[gpu]        validate ✓  (validate needs no GPU — only its rollout does)
  all five rollout tests      skipped — NEMO_GYM_POLICY_BASE_URL not set
```

That sweep is what found the `gdpval` case: it registers its resources-server as `gdpval_resources_server`, not `gdpval`, so `bind_resources_server`'s assumption that the two names match does not hold. Deliberately **not** worked around in the runner — guessing would make it wrong for the environments that do follow the convention. The test records what a caller has to supply instead.

### Known limitation

**The rollout path has not been run end to end.** Every `gym env validate` invocation is exercised, and the runner code around it is unit-tested, but `validate → env start → eval run` as a sequence has never executed here: the rollout tests need a reachable model endpoint and skip without one. A stub endpoint that removes that dependency is the next piece of work, tracked under AALGO-485.

### Pre-commit

**10 of 11 hooks pass**, run with the repository-pinned toolchain (uv **0.9.14** installed to a temp dir, `mise trust` + `mise install` for pnpm **10.34.5**). That includes `Run uv lock with platform uv`, which I had previously reported as blocked — it was my toolchain, not the change.

The remaining failure is `Run UI lint-staged`: `Command "lint-staged" not found`, because this worktree has never run `pnpm install` in `web/`. No `web/` files in this diff, so the hook has nothing of mine to lint.

Worth noting the pin conflict this surfaces: that hook requires uv **0.9.14**, while NeMo-Gym's `pyproject.toml` sets `required-version = ">=0.9.30"`. One uv install cannot satisfy both, which matters because this runner shells out to Gym.

## Summary by CodeRabbit

* **New Features**
  * Added bounded pre-flight validation for Gym environments before execution.
  * Configuration overrides now support nested mappings with typed values.
  * Hydra run outputs are redirected into the active run workspace.
  * Credential-like values are automatically redacted from run metadata.
  * Added clearer installation guidance for Gym environments using Ray dependencies.

* **Bug Fixes**
  * Improved handling and reporting of invalid configurations and missing Gym executables.

* **Tests**
  * Expanded coverage across CPU, Docker, GPU, and end-to-end Gym environments.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
@@ -28,10 +28,14 @@ rollout record, giving a total, order-independent ``index → task`` map. This a
 means a caller can run a **subset** of tasks without Gym rolling out the rest.
 
 **Execution** is the two-step Gym flow (the one that reads a dataset directly
-without triggering Gym's split-driven data-prep): ``gym env start`` brings up the
-resources-server + agent + model servers, then ``gym eval run --no-serve --input
-<materialized dataset>`` collects rollouts against them. The runtime shells out
-to the ``gym`` CLI on PATH, so this SDK never imports ``nemo_gym``. Subprocess
+without triggering Gym's split-driven data-prep), preceded by a pre-flight:
+``gym env validate`` merges the composed config and reports unset ``???`` values,
+bad paths, and dangling cross-references without starting anything; then ``gym env
+start`` brings up the resources-server + agent + model servers, and ``gym eval run
+--no-serve --input <materialized dataset>`` collects rollouts against them. Both
+commands receive the identical selection arguments, so what is validated is what
+runs. The runtime shells out to the ``gym`` CLI on PATH, so this SDK never imports
+``nemo_gym``. Subprocess
 output is streamed to log files under the run's work dir *and* mirrored to this
 module's logger at ``DEBUG``, so callers choose terminal visibility through
 ordinary ``logging`` configuration.
@@ -95,6 +99,11 @@ DEFAULT_REWARD_KEY = "reward"
 #: once, and that path is what the subprocesses run — so a child whose PATH differs from ours cannot
 #: end up executing a different Gym.
 _GYM_CLI = "gym"
+#: Bound on `gym env validate`. It merges config without starting anything and returns in about a
+#: second; this only exists so a wedged invocation cannot stall the run before it begins.
+_VALIDATE_TIMEOUT_S = 120.0
+#: Where Gym's Hydra run directories are redirected, relative to the run's work dir.
+_HYDRA_SUBDIR = "gym_hydra"
 #: Gym's index fields on each rollout record. ``_ng_task_index`` is the only join back to the input
 #: rows that survives a round-trip: Gym mutates ``responses_create_params`` (even the prompt) and
 #: copies only a fixed allowlist of row keys onto the result, so no field we invent comes back. Gym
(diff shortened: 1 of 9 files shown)


packages/nemo_evaluator_sdk/tests/agent_eval/test_gym_environment_coverage.py
```diff
@@ -0,0 +1,341 @@
...
+from nemo_evaluator_sdk.agent_eval.runtimes.gym_runtime import (
+    GymAgentTaskRunner,
+    GymRuntimeConfig,
+    _flatten_overrides,
+    discover_gym_tasks,
+)
+from nemo_evaluator_sdk.agent_eval.tasks import AgentEvalRunConfig
+from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrialStatus
+
+pytestmark = [pytest.mark.integration, pytest.mark.slow]
```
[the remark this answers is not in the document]
SandyChapman: Re-ran with the pinned toolchain (uv 0.9.14, pnpm 10.34.5): 10 of 11 hooks pass, including `uv-lock`, which I'd wrongly reported as blocked. The last one needs `pnpm install` in `web/`; no `web/` files in this diff.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -107,34 +116,123 @@
...
+#: Substrings that mark an override as carrying a credential. Matched case-insensitively against the
+#: full dotted path, so nesting cannot hide one behind an innocuous leaf name.
 _SECRET_KEY_MARKERS = ("api_key", "apikey", "token", "secret", "password", "passwd", "credential")
 #: Stand-in written in place of a redacted override value.
 _REDACTED = "<redacted>"
 
 
-def _redact_env_overrides(overrides: Sequence[str]) -> list[str]:
-    """Redact credential-looking values from Hydra overrides before they are recorded as provenance.
+def _hydra_scalar(value: Any) -> str:
```
JashG, line 159: (I'm not super familiar with Hydra's grammar so feel free to ignore) Do we need to handle `dict`s here? i.e. is something like a list of `dict`s (each `dict` would be converted to a string today) a possible input shape?

JashG (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Good question. I'll confirm and update if needed.

SandyChapman: Confirmed — you're right. A dict inside a list falls through to `str()`, and Hydra rejects the quoted keys Python's repr emits (`++a=[{'b': 1}]` is a parse error); it wants `[{b:1}]`. Fixed in the latest push, along with an empty-dict case that was being dropped silently.

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -3778,12 +3778,13 @@ components:
...
-          type: array
+          additionalProperties: true
+          type: object
           title: Env Overrides
-          description: Extra Hydra '+key=value' overrides for `gym env start` (applied
-            after the auto-derived resources-server binding).
+          description: "Extra config overrides for `gym env start`, as nested data\
+            \ \u2014 {'model': {'temperature': 0.7}} rather than pre-serialized Hydra\
+            \ strings \u2014 so a spec survives being sent as JSON. Flattened to Hydra's\
+            \ grammar at invocation, after the auto-derived resources-server binding."
```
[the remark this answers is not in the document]
SandyChapman: `GymRunnerTarget` landed on main ~14h ago in #1257 and is in no release tag (0.3.0 predates it by two weeks), so no shipped client can send `env_overrides` in either shape. Accepting the list form would also preserve exactly the pre-serialized-Hydra-string footgun this PR removes. Declining.

PR #1212, opened 2026-08-10
feat(eval-author): add standalone Intake trace query tools

aleckhoury:
## Summary

Closes [ASE-686](https://linear.app/nvidia/issue/ASE-686). Eval Author can now find and read production traces on its own, without an Insight handed to it by the Experimentalist loop.

The tool surface is raw queries plus one composite. The agent writes its own filters, so a new question does not need a new method here.

| Tool | What it does |
| --- | --- |
| `query_spans` | Passes a filter to `spans.list`. `group_by="trace_id"` rolls any filter up into the traces that match it, newest first. |
| `query_traces` | Passes a filter to `traces.list`, including `{"id": {"$in": [...]}}`. Carries the server's exact `span_count` and `error_count`. |
| `find_agent_traces` | Recent traces of one agent, newest first. Built on the two above. |
| `read_trace` | Delegates to Experimentalist's `TraceExplorer`. |
| `analyze_trace` | Delegates to Experimentalist's `TraceAnalyzer`. |

## Spans choose, traces tell

`query_spans` returns individual spans, and a span read on its own hides the trajectory around it. A slow call may be waiting on the retry before it, and an error may be one the next span handles, so a span quoted as evidence can say the opposite of what the run actually did.

The division is now explicit in the tool docs: `query_spans` chooses which traces to open, `read_trace` reads what happened in them. Every span row and every `trace_id` group carries `trace_ref`, so the way back to the whole trace is one call rather than a string the caller assembles. Grouping by `session_id` spans many traces and gets no ref, because there is no single trace to open.

## Why `find_agent_traces` exists

No single endpoint answers "recent traces of this agent," because only spans carry `agent_name` and `traces.list` cannot scope to one. So it takes two calls: group spans by trace id sorted by time to get the recent traces, then read the summaries of exactly those ids in chunks of 50.

The second call is not redundant. A group counts only the spans that matched the filter, so an agent-scoped group of a 40-span trace with three errors reports one span and no errors. The server summary gives whole-trace totals. A trace whose root span was never ingested keeps `None` counts rather than zero, so an unknown never reads as healthy.

This used to page spans newest first and collect distinct ids, which @BrianNewsom flagged as wanting a grouped find. He was right, and the missing sort was a real gap in Intake rather than a fact of life: #1242 added it, so the scan is now one grouped call. That also fixed a correctness bug, not just cost. The scan could only see traces inside the window it happened to read, so a quiet agent behind a noisy one went missing no matter how large the budget. `span_budget` is gone with it, having never reached the agent's surface.

## Filter vocabulary is what Intake actually serves

The docstrings list the 15 span filters and 6 trace filters that a live Intake genuinely applies, not the larger set the schema published. Five span filters (`dataset_id`, `dataset_name`, `dataset_version`, `prompt_name`, `prompt_version`) answered HTTP 500; #1225 removed them at the source, and they now read as rejected fields.

## Replicate vs. consume

Replicated from Analyst: the thin read mechanics only — drain a paginator, dump rows, cap and report truncation.

Consumed from Experimentalist: `TraceExplorer` for full trace reads and `TraceAnalyzer` for diagnosis. Both already exist and neither is reimplemented.

Not built: clustering. Failure-mode discovery stays with the Analyst.

## Notes for review

- `analyze_trace` synthesizes the `Task` and `TrialResult` that `TraceAnalyzer.run` expects, because a production trace has no evaluator trial behind it. The synthetic task declares no dependencies, so `start_deps()` starts nothing.
- `TraceExplorer` and `TraceAnalyzer` imports are `_BORROWED_BEHAVIOUR` in `test_plugin_boundary.py`, which is tracked debt rather than a permanent dependency.
- Errors carry a corrective action: 401 names `nemo auth login`, 400 points at the filter vocabulary, 404 names the workspace, and a connection failure names `NMP_BASE_URL`. An empty result is not an error; it returns a note that the agent name is the usual cause.
- `tests/manual/` holds two scripts that run against a live platform. Nothing automated runs them, and pytest does not collect them. They exist because the unit tests fake every Intake call, so they prove the logic but never prove Intake answers the way the logic expects.

## Verified against a live Intake

Both merged Intake changes are in this branch, and the manual checks were run against an instance holding real agent traces. 28 checks pass, 0 fail. The parts worth naming:

```
PASS  groups come back newest first :: ['2026-08-06T21:03:05.958510', '2026-08-06T21:03:04.669397', ...]
PASS  the grouped ids are the traces find_agent_traces returned
PASS  groups also sort by size :: [64, 45, 45, 43, 41]
PASS  a filter Intake once answered with 500 :: 400 - "Unknown filter field 'dataset_name'"
```

The size sort returns different traces from the time sort, so the two orders are not accidentally the same. The vocabulary probe independently confirms the docstrings: 15 span filters, 6 trace filters, `$in` only on `traces.id`, grouping only by `trace_id` and `session_id`.

## Test plan

- [x] `uv run --frozen pytest plugins/nemo-eval-author/tests/` — 72 passed, 3 skipped
- [x] `uv run ruff check` and `ruff format --check` clean
- [x] `tools/lint/lint-python-types.sh`, the script CI runs — no errors
- [x] `test_plugin_boundary.py` passes, so no new cross-plugin import escaped the allowlist
- [x] Manual checks against a live Intake carrying #1225 and #1242 — 28 pass, 0 fail
- [x] Up to date with `main`, including both merged Intake fixes


plugins/nemo-eval-author/tests/manual/intake_vocabulary_probe.py
```diff
@@ -0,0 +1,207 @@
...
+
+Not a test. Pytest does not collect this file. No automation runs it. See README.md in
+this directory.
+
+The docstrings of ``query_spans`` and ``query_traces`` are the only guide the agent has
+when it builds a query, so a wrong docstring sends the agent down a dead end. Neither
+the Pydantic filter schemas nor the "Unknown filter field" error list can be trusted for
+this. A field can appear in both and still fail:
+
+* ``SpanFilter`` declares ``dataset_name``, and Intake answers HTTP 500.
```
BrianNewsom: We should fix this on the intake side, this feels like a workaround that shouldn't need to exist.

plugins/nemo-eval-author/src/nemo_eval_author_plugin/traces.py
```diff
@@ -0,0 +1,404 @@
...
+    workspace: str,
+    since: datetime | None = None,
+    limit: int = DEFAULT_TRACE_LIMIT,
+    span_budget: int = DEFAULT_SPAN_BUDGET,
+) -> dict[str, Any]:
+    """Find the most recent traces of one agent, newest first.
+
+    No single endpoint answers this. The scan walks spans newest first for the
+    distinct trace ids, because only spans carry ``agent_name`` and span groups do
+    not sort by time. It then reads the summary of exactly those ids, because counts
```
BrianNewsom: I would have expected to do the grouped find here - sounds like this is maybe missing a sort?

aleckhoury: You were right, and it was a missing sort. Span groups could only be sorted by `span_count`, and `SpanGroup` carried no timestamp at all, so there was no way to ask a grouped query for the most recent traces. That is the only reason this drains spans instead.

Fixed on the Intake side in #1242: groups now carry `started_at` and accept `started_at` / `-started_at`. Validated against a running instance with real trace data, where an unpatched Intake answers the same request with `422 Input should be 'span_count' or '-span_count'`.

Once that lands this helper collapses to a single grouped call, which also removes the window limitation the current scan has, where it only finds traces that fit inside the scanned span budget. I'll follow up with that change rather than fold it in here, so this PR does not depend on an unmerged service fix.

plugins/nemo-eval-author/tests/manual/intake_vocabulary_probe.py
aleckhoury: Agreed, and fixed on the Intake side in #1225 rather than worked around. Root cause was a three-way mismatch: `SpanFilter` published five fields (`dataset_id`, `dataset_name`, `dataset_version`, `prompt_name`, `prompt_version`) that the attribute catalog had no entry for, so `spec_for_field` raised and the request became a 500. That PR removes them, derives the filter list from the catalog so the two cannot drift again, and adds a contract test asserting every published filter reaches SQL.

The probe script itself is now gone from this branch, since it only runs against a local platform and CI cannot verify it.

plugins/nemo-eval-author/src/nemo_eval_author_plugin/traces.py
```diff
@@ -0,0 +1,428 @@
...
+              children of one span)
+            - agent: ``agent_name``, ``agent_id``
+            - call: ``kind`` (``LLM``, ``TOOL``, ``AGENT``, ``CHAIN``, ``EVALUATOR``),
+              ``status`` (``success``, ``error``, ``cancelled``, ``unknown``),
+              ``tool_name``, ``model``, ``provider``
+            - context: ``project``, ``source``, ``evaluation_id``, ``test_case_id``
+            - time: ``started_at`` as ``{"gte": ..., "lte": ...}``, either bound or both
+
+            No other field is filterable, and ``$in`` is not accepted here: only
+            ``id`` on ``query_traces`` takes a list. Note that ``dataset_id``,
```
BrianNewsom: remove this section

BrianNewsom (approved):
Let's merge in the intake changes, update the docstrings and then merge.

PR #1224, opened 2026-08-10
feat(evaluator): expose per-task attempts with trial identity

ngoncharenko:
## Summary

- **Why:** the run summary only published cross-task averages and pass@k. Answering "which tasks were flaky, and which attempt failed?" meant re-deriving it by hand from the flat task × trial × metric score list.
- **What:** adds `AgentEvalSummary.task_metric_attempts` — the ordered attempts per task and metric output — and persists it to `summary.json`.
- Each attempt names the trial that produced it, so it joins across metric outputs and out to `trials.jsonl`:

```json
"task-47": { "harbor_reward.reward": [
  { "trial_id": "task-47__7f3a9c", "value": 1.0, "value_type": "number" },
  { "trial_id": "task-47__1b8e42", "value": null, "value_type": "missing" }
]}
```

**Important notes**
- `value: null` = the trial died before scoring — an attempt that did not pass. A failed *metric* leaves no entry at all: unmeasured, not unsuccessful.
- pass@k now derives from that same mapping instead of a second walk over the scores, so the per-attempt view and the published pass@k cannot drift apart. **Aggregates are unchanged.**

## Related Issue

- AALGO-310 (P3.1) — this is the deliverable.
- Unblocks AALGO-441 (P3.4, Harbor `reward_stats`) and AALGO-428 (P3.2, exception rollup).

## How this change moves us closer to Harbor parity

**Target shape:** [`AgentDatasetStats`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L15-L25) — Harbor's per-dataset aggregate, the thing `summary.json` has to be able to reproduce. Per-trial source is [`VerifierResult.rewards`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/verifier/result.py#L4-L5); it is populated by [`JobStats.increment`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L149-L150) as `reward_stats.setdefault(value, []).append(trial_result.trial_name)`.

| Harbor field | Status | From |
| --- | --- | --- |
| [`pass_at_k`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L19) | ✅ | `summary.score("<metric>.<output>.pass@k").mean` — pre-existing |
| [`reward_stats`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L20-L22) | ✅ **new** | `task_metric_values` — trial ids + raw floats |
| [`n_trials`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L16) | ✅ | `summary.trial_count` |
| [`n_errors`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L17) | ✅ **newly possible** | distinct `trial_id` where `value is None` |
| [`exception_stats`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L23-L25) | ❌ | needs the exception type → AALGO-428 |

- **The unlock** — Harbor groups by `trial_name`; we grouped by `task_id`. `harbor_runtime` already stamps `trial_name` onto `AgentEvalTrial.id`, so carrying `trial_id` on each attempt makes the shapes *the same*, not merely similar:

```python
# Harbor's real reward_stats, from summary.json alone — no re-walk of result.scores
for attempt in attempts:
    stats.setdefault(attempt.value, []).append(attempt.trial_id)
# -> {"reward": {1.0: ["alpha__a", "beta__a", "beta__b"], 0.0: ["alpha__b", "gamma__a"]}}
```

- **Two divergences closed** — grouping key (`trial_name`, not `task_id`) and inner key type (raw `float | int`, not stringified). Pinned by `test_harbor_reward_stats_is_derivable_from_summary_task_metric_attempts`.
- **`n_errors` needs the trial id specifically** — one dead trial contributes a `null` to *every* metric key, so a naive count over-counts by the number of metrics. Deduping requires identity.
- **Caveats, so this doesn't overclaim** — `pass_at_k` matches in content but not shape (Harbor keys `{k: float}`, we emit named aggregates), and `n_errors` is derivable but not implemented.

**Why attempts carry an id but no ordinal**

Worth stating, since "shouldn't each attempt have an index?" is the obvious review question:

- **Harbor has no attempt ordinal to carry.** It expands `n_attempts` by discarding the loop index — [`job.py:416`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/job.py#L416): `for _ in range(self.config.n_attempts)` — and runs the repeats concurrently ([`job.py:129`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/job.py#L129)). Each repeat gets a fresh random `trial_name`. There is no first attempt.
- **That is a feature, not a gap.** Exchangeable repeats are exactly what the unbiased pass@k estimator assumes — `_pass_at_k(n, c, k)` is a function of counts and never of position. An ordinal would imply an ordering the runs do not have.
- **So identity does the work, not order.** For Harbor the list position is `sorted(glob("*/result.json"))` over a random suffix, i.e. arbitrary; `trial_id` is the only meaningful handle, which is why it is carried rather than parsed. For contrast, the experimentalist's [`_trial_attempt`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/454034a69a46c4eb5dc880dc51783221abd97fad/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/components/evaluator/harbor.py#L577-L581) recovers the ordinal by parsing the trial-name suffix, which `isdigit()`-fails against Harbor's ShortUUID — so it returns `None` on every real Harbor run (verified against a live `n_attempts=2` job: `hello-world__4c3VrKY`, `hello-world__NXRG3pE`).
- **The one case where an ordinal would be real: retries.** A retried trial is causally after the one it replaces, unlike a parallel repeat. Harbor counts retries only at job level ([`JobStats.n_retries`](https://github.com/harbor-framework/harbor/blob/v0.21.0/src/harbor/models/job/result.py#L34)) and its per-trial `TrialResult` carries no retry field, so "did the retry do better?" is a gap in Harbor's model — not something the evaluator can synthesize. Flagging it in case it matters later.

**Follow-ups**

- **AALGO-441 (P3.4)** — unblocked. `reward_payload_from_result` still walks `result.scores` and emits the legacy task-keyed, stringified shape; rewiring it onto the summary is now mechanical.
- **AALGO-428 (P3.2)** — `exception_stats` is the last gap. With `trial_id` present it's a join against `trials.jsonl` (already in the bundle), not a schema change. Recommend it first add a typed `error` to `AgentEvalTrial`, replacing the untyped `metadata["exception_type"]` convention.

## What to review

- **`_task_metric_attempts` in `agent_eval/results.py`** — the core. Docstring carries an `in → out` worked example; the two bullet lists map 1:1 onto the branches below them.
- **The asymmetry is deliberate** — a *failed trial* is `value: null` (counted in `n`); a *failed metric* is no entry (kept out of `n`, so a judge timeout is never charged to the agent). Most likely thing to get wrong later.
- **Why a list of records, not a dict keyed by `trial_id`** — `persistence.py` writes with `sort_keys=True` (would reorder attempts lexicographically), and nothing enforces trial-id uniqueness, so a dict would collapse two attempts into one and silently drop pass@k's `n`. Pinned by `test_duplicate_trial_ids_are_two_attempts_not_one`.
- **`AgentEvalAttemptValue` is frozen, and `value` has no default** — both guard the same class of silent corruption:
  - *Frozen*: the summary hands these out by reference, so a consumer rescaling in place (Gym reports reward on 0-100 where we use 0-1) would rewrite the run's own results, and a later `persist` would save the rewrite.
  - *Required `value`*: `None` means "the trial died" and pass@k counts it toward `n`, so a record that merely omits the key must not quietly become a failed attempt. Explicit `"value": null` still works.
- **`nan_count` semantics** — `test_a_task_that_produced_no_trial_is_unmeasured_and_counted_in_pass_at_k_nan`. Previously such a task was absent from the denominator; now reported as missing coverage. Means unaffected. Note this is a `from_scores` path only — a full run still fails loudly on a trial-less task, and this PR deliberately does not relax that guard.
- **Schema exclusions are keyed by task** — the flat set in the first commit let one task's declaration suppress another task's output. Worth a look if you think exclusions should instead be run-global.
- **Exception type deferred to AALGO-428** — the only source is untyped `trial.metadata["exception_type"]` that only Harbor stamps, and `exception_type` already means two opposite things in this codebase. With `trial_id` present it becomes a join against `trials.jsonl`.

## Changes

- **New field** — `task_metric_attempts: dict[task_id, dict["<metric_type>.<output>", list[AgentEvalAttemptValue]]]`, where `AgentEvalAttemptValue = {trial_id, value}`.
- **pass@k reads it through a one-line projection** — `values_by_task = [attempt_values(outputs[key]) ...]`. Everything below it (`measured`, `unmeasured`, `max_n`, `_pass_at_k`) is byte-identical.
- **Retention follows the declared schema** — continuous/discrete/boolean kept; labels and free models dropped even when the emitted value is numeric. With `tasks=None` there are no specs to filter on, so every numeric output observed is kept.
- **Coverage change** — a task that produced no trial now lands in `pass@k` `nan_count` instead of silently shrinking the denominator. Reachable only via `AgentEvalSummary.from_scores` called directly with a task list wider than the scores; a full run cannot get here, because `_score_trials` refuses to score when a task produced no trial (`test_run_rejects_tasks_without_trials`).
- **Harbor parity unlocked** — `trial_id` *is* Harbor's `trial_name` (`_trial_from_harbor_result` stamps it), so Harbor's own `reward_stats` is rebuildable from the summary alone. See the parity section above.
- **Schema exclusions are per task** — one task declaring an output under an unretained schema must not strip it from another task that never declared it. Tasks in one run need not agree on an output's schema.
- **NaN survives JSON** — this is the first summary field carrying a raw metric value rather than a filtered aggregate, so a NaN score would have been written as a bare `NaN` token and made `summary.json` unparseable by strict readers. Serialized as `"NaN"`, matching `MetricOutput`; round-trips back to a float.
- **Single pass over scores** — grouped by `(task_id, metric_type)` once, replacing a per-task/per-key rescan.
- **Example rewritten** — gym `inspect_results.py` reads the summary directly; `per_task_outcomes` keeps its bare-value shape, new `per_task_attempts` exposes the records. It now rejects a bundle predating the field rather than loading it and silently showing no per-task section.
- **Vendored SDK copy** regenerated via `make vendor`, pinned byte-exact by a test.

------
## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pytest packages/nemo_evaluator_sdk/tests` (live/docker deselected) | 1496 passed |
| `pytest packages/nemo_evaluator_sdk/tests/agent_eval` (live/docker deselected) | 588 passed |
| `pytest .../test_task_metric_attempts.py` | 18 passed |
| `pytest plugins/nemo-evaluator/tests plugins/nemo-optimization/tests` | 849 passed, 22 skipped |
| `uv run ruff check packages/nemo_evaluator_sdk` | passed |
| `uv run ruff format --check packages/nemo_evaluator_sdk` | passed (232 files) |
| `uv run --frozen ty check <changed files>` | 9 diagnostics, all pre-existing (12 at HEAD) |
| `make vendor` | mirror only; byte-exact test passes |
| end-to-end: real Harbor bundle → `persist_run` → `inspect_results` | records in `summary.json`; per-task output unchanged |

**pass@k invariance** — a golden `{name: (mean, count, nan_count)}` table was captured from the pre-change implementation *before* the producer was touched, over a fixture covering every branch (always-passes, dead trial, metric-raised, two never-measured). Post-change it compares identical, and is pinned by `test_pass_at_k_aggregates_are_unchanged_by_carrying_trial_ids`.

Not passing / not run:

- **`uv-lock` pre-commit hook**: needs uv `0.9.14` to match CI; local toolchain is `0.9.30`. Environment mismatch, not from this PR — no `pyproject.toml` is touched, and `uv-lock-check` passes.
- **Live/e2e suites**: `test_harbor_runtime_e2e.py`, `test_codex_runtime_live.py`, `test_sandbox_docker_provider_live.py`, `test_sandbox_compose_provider_live.py` need Docker and provider credentials.

## New Features
- Added typed trial results for numeric, boolean, label, missing, and failed-trial values.
- Added per-task outcome views, trial IDs, and numeric value access.
- Harbor examples now support configurable attempt counts and job names.
- Gym evaluations can limit the number of tasks processed.

## Bug Fixes
- Improved result validation, persistence, and failure handling.
- Updated pass@k calculations for unmeasured outputs and failed trials.
- Added clear handling for invalid or outdated result bundles.

## Documentation
- Expanded result-reading guidance with value formats and failure semantics.
- Updated the Gym results inspector to display task outcomes and aggregates.


ngoncharenko:
@coderabbitai continue the review

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -70,10 +99,132 @@ class AgentEvalSummary(BaseModel):
             "'<metric_type>.<output>', plus per-semantic-view rollups named 'view.<name>'. "
             "Failed or missing scores are surfaced as nan_count."
         ),
+        examples=[
```
arpitsardhana, line 264: nit: Will these examples show up in openapi.yaml? In service mode, i am wondering if extensive example can blow up the size of openapi spec

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -70,10 +99,132 @@ class AgentEvalSummary(BaseModel):
...
+            "Per task, the attempts each '<metric_type>.<output>' measured, in trial order. Each "
+            "attempt names the trial that made it, so attempts join across keys -- and out to "
+            "trials.jsonl and scores.jsonl -- by trial_id. A failed trial has value None: an attempt "
+            "that did not pass. An unmeasured attempt (metric failed, output absent) has no entry at "
+            "all, so each key's list is independent: align by trial_id, never by position. An empty "
+            "list means nothing was measured, including a task that produced no trial."
+        ),
+        examples=[
+            {
+                "contract-review-msa-indemnity": {
```
arpitsardhana, line 359: So this essentially builds the reverse index of results ie task : {runner: [{trials: (name, score)}]}

ngoncharenko: yes, to match harbor format

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
ngoncharenko: - On the service side, this is stored in [summary.json](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/plugins/nemo-evaluator/src/nemo_evaluator/entities.py#L191-L192) and not exposed in the API spec
- Added this on purpose because the results is not a straightforward data structure and having illustrative examples should help both agents and humans.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -61,54 +59,64 @@ def aggregate(summary: AgentEvalSummary, name: str) -> AggregateScore:
 
 
 def per_task_outcomes(
-    scores: Sequence[AgentEvalTaskScore],
+    summary: AgentEvalSummary,
     *,
     metric_type: str,
     output_name: str,
 ) -> dict[str, list[float | None]]:
```
SandyChapman: I wonder if we should provide a stronger type here. Instead of:
```
{
"task-47": { "harbor_reward.reward": [
  { "trial_id": "task-47__7f3a9c", "value": 1.0 },
  { "trial_id": "task-47__1b8e42", "value": null }
]}
}
```
maybe use a model like:
```
class TrialOutcome:
  trial_id: str
  value: float | int | bool | None

class PerTaskOutcome(BaseModel):
  metric_name: str
  trials: list[TrialOutcome]

class PerTaskOutcomes(BaseModel):
  task_id: str
  outcomes: list[PerTaskOutcome]
```

We actually have the type `AgentEvalAttemptValue` which seems to match this TrialOutcome.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -61,54 +59,64 @@ def aggregate(summary: AgentEvalSummary, name: str) -> AggregateScore:
...
-        for line in (bundle / "scores.jsonl").read_text(encoding="utf-8").splitlines()
-        if line.strip()
-    ]
-    return summary, scores
+    payload = json.loads((bundle / "summary.json").read_text(encoding="utf-8"))
+    if "task_metric_attempts" not in payload:
+        raise SystemExit(
+            f"{bundle / 'summary.json'} predates summary.task_metric_attempts, which this script reads "
+            "per-task outcomes from. Re-run the eval to produce a current bundle."
+        )
```
SandyChapman: `SystemExit` seems extreme for the exception, right? Maybe a custom exception would be appropriate here. Also, I assume the `AgentEvalSummary.model_validate` would have failed anyway? Why couldn't we just handle the model_validate failure instead?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
```diff
@@ -58,8 +61,48 @@ class AgentEvalMetricOutputCoverage(BaseModel):
...
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    trial_id: str = Field(
+        description=(
+            "Identifier of the trial that made this attempt. Joins to AgentEvalTrial.id "
+            "(trials.jsonl) and AgentEvalTaskScore.trial_id (scores.jsonl)."
+        )
+    )
+    value: float | None = Field(
```
SandyChapman: Are there more values than `float | None`? I think this would correspond to int | bool too, right? What about `str`? I know our native metrics can output labels too.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
ngoncharenko: - AgentEvalSummary.model_validate won't raise since default value is a dict
- Updated SystemExit part

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py
ngoncharenko: Great catch. Modeled it after pass@k which operates on floats only. Fixed this

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -61,54 +68,75 @@ def aggregate(summary: AgentEvalSummary, name: str) -> AggregateScore:
...
-        if score.status == AgentEvalScoreStatus.FAILED:
-            continue
-        for output in score.outputs:
-            if output.name == output_name and isinstance(output.value, int | float):
-                by_task.setdefault(score.task_id, []).append(float(output.value))
-    return by_task
+    key = f"{metric_type}.{output_name}"
+    return {
+        task_id: list(by_metric[key]) for task_id, by_metric in summary.task_metric_values.items() if key in by_metric
+    }
```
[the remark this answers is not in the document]
ngoncharenko: Declining this one — the premise doesn't hold. Both states you name *are* retained, and the example already reports them as `unmeasured`.

Keys are not discovered from emitted values alone. `_task_metric_values` seeds each task's key set from that task's **declared** metric specs before it reads a single score ([`results.py:969-976`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/ngoncharenko/aalgo-310-passatk-harbor-runner/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/results.py#L969-L976)), and `AgentEvaluator` always passes the task list ([`evaluator.py:195`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/ngoncharenko/aalgo-310-passatk-harbor-runner/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py#L195)). So a declared-but-unmeasured output always has a present, empty key — never an absent one.

I built the four cases and ran them through the example:

| task | key in `task_metric_values` | `show_per_task` reports |
|---|---|---|
| metric raised on every trial | present, `[]` | `unmeasured` |
| completed but omitted the output | present, `[]` | `unmeasured` |
| scored normally | present, `[1.0]` | `solved` |
| scored only by a **different** metric | absent | dropped |

```
? task-a-metric-fa…  []   unmeasured
? task-b-output-om…  []   unmeasured
+ task-c-scored-fi…  [1]  solved

  1 solved · 0 flaky · 0 failed · 2 unmeasured
```

The only task the `if key in by_metric` filter drops is the last row — one that was never scored by the requested metric at all. Applying `by_metric.get(key, ())` would list it as "unmeasured" for a metric it was never eligible for, which makes any mixed-metric run's report wrong. That filter is deliberate, and the rule is documented in `reading-results.mdx`: *"keys are per task, so a task scored by a different metric simply has none."*

On the coverage request: the declared-but-unscored case is already pinned in [`test_pass_at_k.py:124`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/ngoncharenko/aalgo-310-passatk-harbor-runner/packages/nemo_evaluator_sdk/tests/agent_eval/test_pass_at_k.py#L124) — `"verdict.category": []`, with the comment *"t2 declares the label metric but no verdict was scored for it"*.

packages/nemo_evaluator_sdk/tests/agent_eval/test_task_metric_values.py
```diff
@@ -0,0 +1,656 @@
...
+def test_gym_example_rejects_a_bundle_written_before_task_metric_values(tmp_path: Path) -> None:
+    # The field defaults to empty, so an older bundle would load cleanly and simply show no per-task
+    # section -- a reader would take that as "no per-task outcomes" rather than "this script cannot
+    # see them". Fail with a version message instead.
+    from packages.nemo_evaluator_sdk.examples.gym.inspect_results import BundleFormatError, load_bundle
+
+    (tmp_path / "summary.json").write_text(json.dumps({"task_count": 2}), encoding="utf-8")
+
+    with pytest.raises(BundleFormatError, match="predates summary.task_metric_values"):
+        load_bundle(tmp_path)
```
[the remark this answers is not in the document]
ngoncharenko: Declining this one.

The import is the documented invocation, not a test-only namespace trick:

```
uv run python -m packages.nemo_evaluator_sdk.examples.gym.inspect_results
```

These tests load that same module path on purpose. Collection already succeeds under `--import-mode=importlib` from repo root (the CI path). Switching to `spec_from_file_location` would keep the tests green even if the documented `-m packages...` path broke, which is the failure we actually want to catch.

The other example tests use file-path loading because those scripts are not advertised as importable modules. This one is.

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
ngoncharenko: Added this

packages/nemo_evaluator_sdk/tests/agent_eval/test_sandbox_compose_provider_live.py
```diff
@@ -17,6 +17,11 @@
     DockerComposeSandboxProvider,
 )
 
+# Keep-alive containers ignore SIGTERM; use a 1s stop grace so teardown does not wait
+# the provider default of 30s on every create/close cycle.
+_SHUTDOWN_TIMEOUT_SECONDS = 1.0
```
ngoncharenko, line 22: updated the test - was waiting for sandbox spin down for too long

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
SandyChapman: Return value is still `-> dict[str, list[float | None]]:` instead of a model though.

ngoncharenko: missed this point. Don't need these functions actually now. `summary.task_outcomes(f"{args.metric_type}.{args.output_name}")` returns `list[PerTaskOutcomes]`

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
```diff
@@ -60,55 +68,42 @@ def aggregate(summary: AgentEvalSummary, name: str) -> AggregateScore:
...
-                by_task.setdefault(score.task_id, []).append(float(output.value))
-    return by_task
-
+# Per-task values need no accessor here: `summary.task_outcomes("<metric_type>.<output>")` returns
+# `list[PerTaskOutcomes]` already sorted by task, each naming its own `task_id` and `metric_name`,
+# and each trial naming the `trial_id` that produced it — so a value can be joined back to
+# `trials.jsonl` (where a failed trial's error lives) or across to another metric's outcomes.
+# Records are ordered by trial, and `trial_id` rather than position is what lines two outputs up:
+# a trial whose metric failed is absent rather than null, so two lists for one task need not be the
+# same length.
```
SandyChapman: This comment probably isn't all that helpful. Just claude/codex saying "I had stuff here but removed it".

SandyChapman (approved):

packages/nemo_evaluator_sdk/examples/gym/inspect_results.py
ngoncharenko: Done

PR #1225, opened 2026-08-10
fix(intake): stop publishing span filters that answer HTTP 500

aleckhoury:
## Summary

Five span filters returned HTTP 500 with the body `An unexpected error occurred.`:

```
dataset_id  dataset_name  dataset_version  prompt_name  prompt_version
```

`ATTRIBUTE_EQ_FILTER_FIELDS` in `spans/api/spans.py` routed all five to the span attribute catalog. `SpanAttributeField` has no entry for any of them, so `spec_for_field` raised `ValueError` while the SQL was built, and nothing caught it. A bad request read as a server fault.

Reproduced against a running platform before the fix:

```
dataset_id: InternalServerError status=500 - {'detail': 'An unexpected error occurred.'}
dataset_name: InternalServerError status=500 - {'detail': 'An unexpected error occurred.'}
dataset_version: InternalServerError status=500 - {'detail': 'An unexpected error occurred.'}
prompt_name: InternalServerError status=500 - {'detail': 'An unexpected error occurred.'}
prompt_version: InternalServerError status=500 - {'detail': 'An unexpected error occurred.'}
```

This was reachable from the public API docs: the endpoint description named `prompt_name` and `prompt_version` explicitly.

## Why remove them instead of adding catalog entries

`SpanSemanticAttributes` has no dataset or prompt-name field, so Intake never ingests, stores, or returns this data. `SpanEvaluationContext` exposes only `evaluation_id` and `test_case_id`. A catalog entry would let the query run and return zero rows every time, and a silently empty result is harder to diagnose than a rejected field.

## Changes

- `spans_schemas.py`: drop the five fields from `SpanFilter`.
- `spans.py`: `ATTRIBUTE_EQ_FILTER_FIELDS` stays an explicit list of the eight. `SpanFilter` is the public contract and the catalog is storage plumbing, so adding a name should be a deliberate act rather than a set operation between the two. The contract tests below supply the drift protection.
- `spans.py`: the endpoint description now names only the fifteen served fields and says which operators they take.
- `openapi/`: regenerated. Pure deletion of the five properties plus the reworded description.
- SDK and CLI: regenerated with `make update-sdk`. `SpanFilterParam`, the spans resource, both checked-in CLIs, and `docs/cli/reference.mdx` no longer offer the five fields. `SpanFilterParam` now carries exactly the fifteen fields a live Intake serves.
- `plugins/nemo-insights`: the Analyst told its model that `dataset_name` and `evaluation_run_id` were valid span filters. The first returned 500 and the second is not a field at all. It also advertised a status value of `"ok"`, which does not exist: `SpanStatus` is `success`, `error`, `cancelled`, `unknown`, and a live Intake answers `status=ok` with `400 Filter field 'status' must be one of`. All three corrected.

## Tests

`services/intake/tests/test_spans_filter_contract.py` walks the chain that broke — filter JSON, then the API filter builder, then the SQL builder — for every field `SpanFilter` publishes. It holds both directions of the contract: no published field is unroutable, and no routed name has fallen out of the schema.

I confirmed the guard works by temporarily republishing `dataset_name`, which fails three tests with:

```
AssertionError: SpanFilter publishes ['dataset_name'], which _span_filter cannot route.
Either handle the field or remove it from the published schema.
```

## Test plan

- [x] `uv run --frozen pytest services/intake/tests --ignore=services/intake/tests/integration` — 277 passed
- [x] `uv run --frozen pytest services/intake/tests/integration/spans` — 108 passed
- [x] `uv run ruff check` — clean
- [x] `uv run --frozen ty check services/intake` — 7 diagnostics, none in changed files
- [x] `make update-sdk` — five fields gone, diff scoped to intake spans with no unrelated generator drift
- [x] Live probe of the status values: `success` 200, `error` 200, `ok` 400
- [x] Pre-existing `nemo-insights` test failures confirmed identical on `main`
- [ ] Re-run the 500 repro after deploy to confirm the five now return 400

## Related

Found while adding Intake trace query tools to Eval Author in #1212, which had inherited the same wrong filter vocabulary in its docstrings.

## Summary by CodeRabbit

- **Changes**
  - Updated span and span-group filtering to use supported fields such as project, evaluation, test case, agent, model, tool, and provider.
  - Removed dataset and prompt filtering options from the API, CLI commands, and documentation.
  - Clarified that non-time filters use exact matching, while `started_at` supports `gte` and `lte` comparisons.
  - Updated status filter examples and supported values.

- **Tests**
  - Added coverage ensuring published filters are supported, routable, and correctly translated into queries.


services/intake/src/nmp/intake/spans/api/spans.py
```diff
@@ -31,25 +31,19 @@
...
-        "prompt_version",
-    }
+# Which SpanFilter fields resolve to a span attribute, derived rather than listed.
+#
+# A hand-kept list drifts from the catalog, and the failure is severe: the repository
+# calls spec_for_field on every attribute filter, and an uncatalogued field raises
+# ValueError there, which surfaces as HTTP 500 rather than a rejected request. Deriving
+# the set means a SpanFilter field with no catalog entry falls through to the explicit
+# 400 below instead.
+ATTRIBUTE_EQ_FILTER_FIELDS = frozenset(SpanFilter.model_fields) & frozenset(
```
BrianNewsom: Could we keep this list explicit? I find the intersection a little hard to reason about: `SpanFilter` is the public API contract, while the catalog is storage plumbing. Listing the eight supported attribute filters here makes additions deliberate, and the new contract tests should give us the drift protection we need.

openapi/openapi.yaml
```diff
@@ -4608,8 +4608,9 @@ paths:
         schema:
           $ref: '#/components/schemas/SpanFilter'
         description: Filter spans by session_id, trace_id, parent_span_id, project,
-          evaluation context fields, source, kind, status, model, tool_name, provider,
-          agent_id, agent_name, prompt_name, prompt_version, and started_at.
+          evaluation_id, test_case_id, source, kind, status, model, tool_name, provider,
```
BrianNewsom, line 4611: We need to regenerate the checked-in Python SDK and CLI in this PR -  This changes `SpanFilter`, but both generated clients still advertise the five fields being removed, so the repo would temporarily ship two different contracts.

services/intake/tests/test_spans_filter_contract.py
```diff
@@ -0,0 +1,82 @@
...
+from nmp.intake.spans.api.spans import ATTRIBUTE_EQ_FILTER_FIELDS, _span_filter
+from nmp.intake.spans.api.spans_schemas import SpanFilter
+from nmp.intake.spans.span_attribute_catalog import spec_for_field
+
+# A value each field accepts, so a failure means the field and never the value.
+FILTER_VALUES: dict[str, Any] = {
+    "kind": "LLM",
+    "status": "success",
+    "started_at": {"$gte": "2026-01-01T00:00:00"},
+}
```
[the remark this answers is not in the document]
aleckhoury: The inconsistency is real, but it points the other way: the test is right and the docstrings were wrong. `SpanStatus` is `success`, `error`, `cancelled`, `unknown`, with no `ok` member, so `"success"` in the fixture is valid and `"ok"` would not be.

Confirmed against a live Intake:

```
status=success   HTTP 200  matched=3725
status=ok        HTTP 400  Filter field 'status' must be one of: success, error, cancelled, unknown
status=error     HTTP 200  matched=232
```

Good catch on the mismatch though, because the docs you cited were the broken side and an agent following them would have written a query that could not succeed. Both now read `("success"/"error"/"cancelled"/"unknown")`.

services/intake/src/nmp/intake/spans/api/spans.py
aleckhoury: Agreed, reverted to an explicit list of the eight. Your framing is the right one: `SpanFilter` is the public contract and the catalog is storage plumbing, and intersecting them buried that distinction in a set operation.

The only reason I derived it was to stop a name being routed to the catalog without an entry, which is what turned into the 500. The contract tests cover that directly now, so deriving was standing in for a test I had already written. I added the missing reverse direction too, `test_no_attribute_filter_is_routed_that_the_schema_does_not_publish`, so a name left behind after a schema change fails rather than sitting there as dead routing.

openapi/openapi.yaml
aleckhoury: Good catch, done. Ran `make update-sdk`, which dropped the five fields from `SpanFilterParam`, the spans resource, both checked-in CLIs, and `docs/cli/reference.mdx`. `SpanFilterParam` now carries exactly the fifteen fields a live Intake serves.

The diff stayed scoped to intake spans, so no unrelated generator drift came along for the ride.

BrianNewsom (approved):

PR #1248, opened 2026-08-11
fix(cli): make uv tool install the front door for nemo-platform

callingmedic911:
- `uv tool install "nemo-platform[all]"` is the documented install path now.
- fixed `nemo --version` to 0.0.0 for PyPi user.
- detect and show right command `uv tool ...` vs `pip ...` based on how user set up nemo-platform.
- update docs to promote `uv tool` as recommended path.
- fix `script/compile-wheel-constraints.sh`: a missing closing quote makes `bash -n` fail — so the constraints could never be regenerated.
- tightened  python version mismatch: nemo-platform-sdk declared `>=3.12`, the wrapper wheel it ships in supports 3.12-3.13.
___
**One of the problem I ran into:** `nvidia-nat-core` 1.8.0 caps `cryptography` at `<47` while `data-designer-engine` 0.8.0 requires `>=48.0.1`, so `nemo-platform[all]` built from current `main` does not resolve from PyPI. 

Considered path:
- Bump NAT to 1.9.0aX release. (Not stable?): Went with this option
- Downgrade data-designer. (Lose features?)
- Wait for NAT to be removed as dep. (Unsure about timeline)
___
<details>
<summary>Cursor's description</summary>

## Summary

`uv tool install "nemo-platform[all]"` is now the documented install and it resolves from PyPI. Getting there turned up two things the old wheel-test could not see: `nemo --version` reported `0.0.0` for every PyPI user, and the wheel's dependency closure was unresolvable outside this checkout, because CI's install step ran inside the repo and silently inherited `[tool.uv] override-dependencies`.

## Changes

- `_version.py` queries `nemo-platform` instead of `nemo-platform-sdk`. The SDK distribution only exists in the workspace (`nemo-platform-sdk` is 404 on PyPI) — PyPI users get the SDK bundled inside the wrapper wheel, so the lookup raised and fell back to `0.0.0`, which also reached telemetry `client_version` and the SDK client User-Agent. Both names report the same version in the workspace, so nothing changes for contributors. The edit is in the generator template, and `SDK_DISTRIBUTION_NAME` is untouched because it still names the distribution being built.
- New `local/install.py` detects a uv tool environment through the `uv-receipt.toml` uv writes at the environment root, so the missing-extra hint prints `uv tool install 'nemo-platform[all]'` there and `pip install 'nemo-platform[all]'` in a virtual environment. Used by `nemo services run`, the `nemo setup` preflight, and `local/services.py`. The Python 3.14 variant of that hint is gone: the wheel declares `requires-python <3.14`.
- **NAT moves to the 1.9 line** so the published wheel resolves. `nvidia-nat-core` 1.8.0 caps `cryptography` at `<47` while `data-designer-engine` 0.8.0 requires `>=48.0.1`; 1.9 is the first NAT line whose cap (`>=48,<49`) overlaps. That line ships only pre-releases today, and uv accepts a pre-release only when a requirement it reads names one, so all six NAT distributions in the closure are named rather than just the two the plugin imports. The lock, the wrapper extras, the license manifest, and `third_party/requirements-main.txt` follow.
- **`wheel-test` installs with `--no-config`.** The step runs inside the checkout, so uv was applying the repo's `override-dependencies` (notably `cryptography>=50` for CVE-2026-69247) and testing a dependency set no user can get. That is exactly why the venv leg passed while the new tool leg — which does not read project config — failed. `wheel-test` also gains the `uv tool install` matrix leg, and `script/test-nemo-cli.sh` now asserts `nemo --version` reports a real version instead of printing it unchecked.
- `script/compile-wheel-constraints.sh` has been syntactically invalid since it landed — a missing closing quote makes `bash -n` fail — so the constraints could never be regenerated. Fixed, moved to a 3.12 probe environment, and given `--no-config`. Regenerating picks up the NAT bump plus the drift that accumulated in the meantime, and drops the `litellm<1.92` cap: it and its py3.14 rationale predate the 3.12-3.13 range, and litellm is in neither wheel's closure, so the pin constrained a package the smoke test never installs.
- Install docs lead with `uv tool install "nemo-platform[all]"` and present the two paths as a tab set: the tool install is global with nothing to activate, the virtual environment is for importing the SDK from your own code and now uses `uv venv` + `uv pip install` rather than mixing tools. `SETUP.md` is labelled as the contributor path. The bundled SDK's `requires-python` and classifiers are tightened to the supported 3.12-3.13 range, and the wheel build action defaults to 3.12 rather than 3.11.

## Python version handling

`--python` stays out of the install commands. The prerequisites state 3.12-3.13, and the escape hatch lives in CLI troubleshooting, because uv enforces only the **lower** bound of `requires-python`: it planned to install `numpy==1.24.4`, which declares `<3.12`, into a 3.14 environment, and it preferred a managed 3.14 over the system 3.12 for a tool install of this wheel. NAT declares `<3.14` too, so 3.14 support is gated upstream regardless.

## Type of Change

## Quality Gates

## Verification

Targeted validation, re-run after merging `main`:

- End to end from `/tmp`, no checkout, no virtual environment, no repo config, against a wheel built from this branch: `uv tool install <wheel>` lands on 3.13 and `nemo --version` reports a real version; `nemo services run` prints the uv tool remediation, and the same check in a pip virtual environment prints the pip remediation; running the printed command verbatim resolves from PyPI (`cryptography==48.0.1`, `nvidia-nat-core==1.9.0a20260811`, `data-designer==0.8.0`), after which `script/test-nemo-cli.sh` boots all 17 services and reports `platform healthy; default+system workspaces present` (exit 0).
- Both CI legs reproduced locally with their exact arguments, the regenerated constraints file, and uv 0.9.14.
- `uv run pytest packages/nemo_platform_ext/tests` — 1762 passed, 4 skipped. `uv run pytest plugins/nemo-agents/tests` on NAT 1.9 — 990 passed, 1 failed: `test_is_port_free_returns_false_for_occupied_port`, which turns on kernel `SO_REUSEADDR` semantics on the agent VM and passes in CI.
- `uv run ruff check .`, `uv lock --check` (uv 0.9.14), `license find-missing`, `fern check`, and `validate-mdx` are clean. The license manifest and the requirements export were both cross-checked against the merged lock.
- CI on the latest head: every job passes except `Python integration tests`, which hit the pytest-xdist `[gw0] node down: Not properly terminated` crash and then the 1800s job timeout. The immediately preceding run passed that job on identical test code, so it needs a re-run.

## Notes for the reviewer

- **The NAT pins are dated pre-releases.** NAT publishes no stable release above 1.8.0, so `>=1.9.0a20260811,<1.10` is the floor; it accepts newer 1.9 alphas at install time and stable 1.9.0 when it ships. The comment in `plugins/nemo-agents/pyproject.toml` says to collapse back to `core` + `langchain` at that point, and NAT is expected to give way to Fabric before long.
- **A user install lands on `cryptography==48.0.1`, below the 50.0.0 that fixes CVE-2026-69247.** Publishing a `cryptography>=50` floor in wheel metadata is not an option — resolving the wheel with that floor added is unsatisfiable, since the closure pulls `>=48.0.1,<=49` through data-designer and `<49` through NAT. The repo's override keeps developers on 50; getting users there needs an upstream cap relaxation.
- **The alternative to the NAT bump was worse.** Relaxing our `data-designer==0.8.0` pin so users fall back to 0.7.0 also resolves, and a workspace-only `constraint-dependencies` floor would keep the lock on 0.8.0 — but 0.7.0 predates the filesystem-provider abstraction `data_designer_nemo` is built on: three symbols it imports at module top level arrived in 0.8.0, and `DirectorySeedReader` takes no arguments there, so `context.py` would need a second seed-reader path that development never exercises.

<div><a href="https://cursor.com/agents/bc-54fecadd-ee7d-4a49-b9f0-9df1e13550d4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-54fecadd-ee7d-4a49-b9f0-9df1e13550d4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Added isolated `uv tool install` options for the CLI and full platform services.
  * CLI errors now suggest the appropriate installation command for missing services.

* **Bug Fixes**
  * Improved troubleshooting for unsupported Python versions and unavailable dependency wheels.
  * CLI version checks now reject fallback or invalid values.

* **Documentation**
  * Updated setup and installation guidance with clearer environment-specific options.
  * Clarified supported Python versions as 3.12 and 3.13.
  * Added guidance for SDK installation in virtual environments.

</summary>


docs/get-started/setup.mdx
callingmedic911: Same as above:

I want to document both path. Tool is easy / quick  / global. Other one is manual / not global. Suggested is tool. Avoid over explaining, keep short and sweet.

docs/cli/index.mdx
```diff
@@ -28,25 +28,33 @@ If you previously installed the `nemo-microservices` package, uninstall it first
...
-Or with uv:
+</Tab>
+
+<Tab title="Virtual environment">
+
+Use when you also import the SDK from your own code. `nemo` works only while the environment is active.
 
 ```bash
-uv pip install "nemo-platform[all]"
+pip install "nemo-platform[all]"
```
mckornfield: should we just prefer the `uv` way, tool or not?

mckornfield (commented):
know we have the nat question still. I also think we should make `pip install` the last option, personally I'd prefer just `uv tool install` or `uv install` and then pip last (or not at all because it sucks)

.github/wheel-constraints/nemo-platform-services.txt
callingmedic911, line 1: The scrip that updated this file was broken, so there are a few version bumps that are unrelated after I fixed it

pyproject.toml
```diff
@@ -341,7 +341,7 @@ override-dependencies = [
   "xgrammar>=0.1.32",        # High – customizer + nmp-gpu-tasks
   "fastmcp>=3.2.0",          # GHSA-vv7q-7jx5-f767 (Critical) + GHSA-rww4-4w9c-7733 (High); overrides vendored sdk/python/nemo-platform <3 constraint
   "wandb>=0.28.1",           # wandb-core Go CVEs (nspect); unsloth + training images
-  "cryptography>=50.0.0,<51", # GHSA-g6cj-pr64-35w5; overrides nvidia-nat-core's <47 cap after changelog review
+  "cryptography>=50.0.0,<51", # GHSA-g6cj-pr64-35w5; overrides nvidia-nat-core's and data-designer-engine's <49 caps after changelog review
```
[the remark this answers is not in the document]
callingmedic911: I resolved the wheel with that floor added: unsatisfiable, because the closure pulls cryptography>=48.0.1,<=49 through data-designer (and <49 through NAT).

I know that we can bump it upstream but NAT is supposed to be removed soon in favour of Fabric, so I'm temporarily working around with what we have.

docs/cli/index.mdx
```diff
@@ -28,18 +28,20 @@ If you previously installed the `nemo-microservices` package, uninstall it first
...
 
-### Install in a Virtual Environment
+### Install as a Tool
 
-Install the PyPI `nemo-platform` wrapper package in a Python 3.12-3.13 virtual environment. If you are working from a source checkout, run `make bootstrap` from the repository root instead; the generated SDK package under `sdk/python/nemo-platform/pyproject.toml` does not define the wrapper extras.
+Install the PyPI `nemo-platform` wrapper package with [uv](https://docs.astral.sh/uv/). uv puts the `nemo` command on your PATH in its own isolated environment, so there is no virtual environment to create or activate. If you are working from a source checkout, run `make bootstrap` from the repository root instead; the generated SDK package under `sdk/python/nemo-platform/pyproject.toml` does not define the wrapper extras.
 
 ```bash
-pip install "nemo-platform[all]"
+uv tool install "nemo-platform[all]"
```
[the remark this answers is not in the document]
callingmedic911: Doesn't uv's installer puts ~/.local/bin on PATH and that's where tool executables land? If it's a unusual uv's case I don't want to document it.

.github/workflows/ci.yaml
```diff
@@ -1379,7 +1399,7 @@ jobs:
           # (see the nemo-platform CLI step above).
           cd "${RUNNER_TEMP}/wheelcheck"
           uv venv .venv --python "${PYTHON_VERSION}"
-          uv pip install --python .venv/bin/python \
+          uv pip install --no-config --python .venv/bin/python \
```
callingmedic911, line 1402: this step was never testing a user's install, it was testing ours.

It's working directory is the repo too, from the checkout. uv discovers [tool.uv] configuration from the nearest pyproject.toml. So uv pip install "<wheel>[services]" in CI was resolving against our development constraints (bunch of CVE deps override)

docs/cli/index.mdx
callingmedic911: I agree with you and went in this direction

.github/wheel-constraints/nemo-platform-services.txt
```diff
@@ -21,40 +22,46 @@ datasets==4.3.0
...
-nvidia-nat-config-optimizer==1.8.0
-nvidia-nat-core==1.8.0
-nvidia-nat-langchain==1.8.0
-openai==2.47.0
+nvidia-nat-atif==1.9.0a20260811
+nvidia-nat-config-optimizer==1.9.0a20260811
+nvidia-nat-core==1.9.0a20260811
+nvidia-nat-eval==1.9.0a20260811
+nvidia-nat-langchain==1.9.0a20260811
+nvidia-nat-opentelemetry==1.9.0a20260811
```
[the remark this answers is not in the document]
callingmedic911: fixed

mckornfield (approved):

crookedstorm (approved):
LGTM from devops

svvarom (approved):

PR #1139, opened 2026-08-06
fix(deployments): gate READY on workload reachability (AIRCORE-998)

maxdubrinsky:
## Summary

Deployments reported `READY` before the workload had bound its port. On the openshell backend a port was exposed (which reads as READY) as soon as the serve pid was alive; on docker a running container with no declared `readinessProbe` was treated as READY. A live pid or a running container has not necessarily called `bind()`, so a caller trusting READY hit repeated `502`s for the roughly 14s until `nat serve` finished starting. This gates READY on the port actually accepting a connection.

## Related Issue

AIRCORE-998.

## Changes

- OpenShell: probe the workload from inside the sandbox against `127.0.0.1:<port>` (gateway-independent) before exposing a port. Docker: TCP-connect the published port in the no-declared-probe branch of `check_readiness_probe`. Both honour a declared `readinessProbe` (httpGet/tcpSocket/exec) and otherwise default to a TCP connect; portless workloads stay ready.
- Not-yet-reachable reports `STARTING`, so the reconciler's existing `starting_timeout_seconds` is the progress deadline that eventually fails a never-binding workload. No new deadline, knob, or status; death detection unchanged. Exposing only once reachable keeps the openshell `endpoints -> READY` fast path sticky (no flapping).
- Hardening: HTTPS probes skip cert verification (k8s httpGet semantics); UDP-only and unresolvable declared ports fall back to the first TCP port or running-implies-ready; the exec-probe RPC and the docker socket connect are both bounded by the probe timeout so a hung probe cannot stall the serial reconcile loop.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
- [x] Documentation not applicable — justification: no user-visible API or CLI surface change. The `deploy-sandbox` SKILL text that documents the 502-retry workaround is tracked to follow via AIRCORE-981.

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [x] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
[...]

PR #1259, opened 2026-08-12
fix(nemo-deployments): gate OpenShell READY on a curl reachability probe

maxdubrinsky:
## Summary

The readiness gate added in #1139 never ran: it located an interpreter with `command -v python3 || command -v python`, but packaged agent images ship Python only in `/workspace/.venv`, which is not on the exec PATH, so the probe hit its `else exit 0` branch and the gate read that as reachable. READY was published about 14s before `nat serve` bound its port. This probes with `curl` instead, resolved at `/usr/bin/curl` (installed by the NAT and Fabric Dockerfiles, pinned by the sandbox policy) with a PATH fallback, and fails closed when curl is absent, so READY now means the port is reachable.

## Related Issue

AIRCORE-998 (NVBug 6567588). Reopened by QA after #1139 did not change the measured behavior.

## Changes

- Replace the Python readiness probe with a curl probe in `openshell/backend.py`; a missing curl exits nonzero (fails closed) instead of exiting 0.
- httpGet probe uses `curl --fail --insecure` (a 5xx reads as not-ready; a self-signed loopback cert is accepted).
- tcpSocket and default probe do a loopback connect classified by curl exit code (7 = refused = not-ready, any other exit = ready). `telnet://` was rejected because curl stays connected until `--max-time` on a live port.
- `deploy-sandbox` SKILL: drop the "retry a 502 for 30-60s" guidance now that READY implies reachability.
- Add a unit test for the fail-closed behavior and an integration test that asserts READY waits for the port to bind.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

- `uv run pre-commit run --files <4 changed files>`: passed (ruff, ruff format, ty, copyright headers, plugin-import check, merge-conflict check). Full `pre-commit run -a` not run locally (heavy container hooks such as helm-docs and OpenAPI generation are irrelevant to these files); left to CI.
- Unit: `pytest plugins/nemo-deployments/tests/unit/backends/openshell/test_backend.py` with the `openshell` extra: 71 passed, including a new fail-closed test.
- Integration: `pytest plugins/nemo-deployments/tests/integration/backends/openshell/test_openshell_backend.py` against a live gateway (v0.0.92, docker driver): 2 passed, including a new slow-bind test asserting READY waits for bind.
- Probe behavior confirmed directly via `openshell sandbox exec` (the backend's own exec path): unbound port -> curl exit 7 (not-ready); bound socket -> curl exit 0 (ready); with no Python on PATH the old probe exits 0 (READY, reproducing the bug) while the new probe exits 1 (STARTING, fixed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment readiness checks for HTTP and TCP services.
  * Deployments now remain in a starting state until the workload accepts connections.
  * Improved handling of unavailable readiness-check tools and HTTPS loopback checks.
  * Updated 502 troubleshooting guidance to recommend immediate serve-log inspection.

* **Tests**
  * Added coverage for delayed service startup, readiness transitions, and endpoint availability.


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -138,24 +138,9 @@ def _delivery_script(path: str, mode: int) -> str:
...
-# cert is not rejected on verification (matching Kubernetes httpGet HTTPS probe semantics);
-# the context is ignored for plain http, so one program covers both schemes.
-_HTTP_PROBE_PROGRAM = (
-    "import sys, ssl, urllib.request; "
-    "urllib.request.urlopen(sys.argv[1], timeout=float(sys.argv[2]), context=ssl._create_unverified_context())"
-)
-_TCP_PROBE_PROGRAM = (
-    "import sys, socket; socket.create_connection((sys.argv[1], int(sys.argv[2])), timeout=float(sys.argv[3])).close()"
-)
+_CURL_ABS = "/usr/bin/curl"
```
mckornfield, line 143: Ah I see we basically do a "which curl" locally, so 👍 

curious, why did the python bit not work? Path nonsense? I saw it elsewhere, think we should probably also make the path better, but this is fine

mckornfield (approved):

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
maxdubrinsky: Yeah basically. Claude wanted to do some insane python discovery that would involve us rooting around in any possible venv location for the right python instance and I was like "but curl exists"

PR #1267, opened 2026-08-12
feat(studio): run guardrail checks against an unsaved draft config

aray12:
## Summary

The Test and Validate tab could only exercise the *saved* guardrail config, so checking whether an edit worked meant saving it first — publishing a half-tested change to whatever already pointed at that config, and writing a config version the user did not want to keep. This adds a Draft/Saved run-target control: a Draft run sends the current form state inline to `/checks` instead of referencing the saved config by id. No backend change is needed — the guardrails service already accepts a whole `RailsConfig` in `guardrails.config`, and Studio simply never populated it.

Runs now also record what produced them, so a result keeps describing the config that ran even after that config changes.

https://github.com/user-attachments/assets/81699c72-baf2-4661-8760-49550e604b80

## Changes

- **Draft/Saved run target** (`GuardrailTestCasesEditor`) — a `SegmentedControl` beside "Run N Tests". Draft is selected when the form is dirty and disabled when it is pristine; picking Saved on a dirty form still exercises the saved config. The choice resets when navigating to a different config.
- **Inline draft dispatch** (`guardrailChecks.ts`) — `runGuardrailCheck`/`runGuardrailChecks` take an optional `draftConfig`. When present it becomes the effective config for both model resolution and the request, sent as `guardrails: { config: <RailsConfig> }`. Never alongside `config_ids`: the service's validator nulls those out for an object `config`, which would make the request's meaning non-obvious from the wire. A draft also takes precedence over a check's stored `guardrails`, so an explicitly chosen target is never silently discarded.
- **Honest run records** (`types.ts`) — `RunRecord` gains `is_draft` (in place of a `config_version`, which is meaningless for a draft) and `activated_guardrails`, a snapshot of the coverage the config declared at run time. Both optional, so records written before this change deserialize and render unchanged.
- **Result rendering** — Run History shows a "Draft" badge instead of `v{n}`, matching the wording of the run-target control; Rail Status prefers the run's own snapshot over deriving coverage from the currently-loaded config, falling back to today's behaviour when a record has no snapshot.
- **`useDraftRailsConfig`** — extracts the server-data-plus-form-values derivation the Configuration tab already performed inline, so both tabs share one definition of "the draft" rather than deriving it twice.
- **MSW `/checks` handler** — now honours an inline config so a draft-run test asserts the code's behaviour rather than a constant in the mock.
- `ActivatedGuardrail` moves from the side-panel module to `api/guardrail-checks/types.ts`, since it is now persisted data; re-exported from `railLabels.ts` for existing importers.

Note for reviewers: `guardrailChecks.ts` imports `getActivatedGuardrails` from the side-panel module, which is an api → components direction. It is deliberate — the alternative duplicates the detector/flow catalog and lets check-result labels drift from the config editor's. The clean fix is hoisting `detectors.ts`/`flowRegistry.ts` out of `GuardrailConfigTab/` into a neutral module, which is larger than this change warrants.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs describe the checks tab; the behaviour is discoverable from the control itself.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test` — 310 files, 2847 tests passed (full suite, pre-rebase).
- `pnpm --filter nemo-studio-ui test src/api/guardrail-checks src/routes/guardrails src/components/sidePanels/GuardrailCheckDetailSidePanel` — 13 files, 101 tests passed (re-run after rebasing onto current `main`).
- `pnpm --filter nemo-studio-ui test src/components/sidePanels/GuardrailCheckDetailSidePanel` — 3 files, 24 tests passed (re-run after the badge-label change in `f4947cf63`).
- `pnpm --filter nemo-studio-ui typecheck` — clean. `pnpm --filter @nemo/common typecheck` — clean.
- `pnpm lint:fix` then `pnpm lint` — clean, no reformatting required.
- `uv run pre-commit run -a` — ruff, ruff format, `ty`, config reference, copyright headers, UI lint-staged, and merge-conflict hooks all passed. **Two hooks did not run for environmental reasons, neither related to this change:** `Helm Docs Container` (no local Docker daemon; no Helm files are touched) and `Run uv lock with platform uv` (local uv is 0.9.28, the hook requires 0.9.14). `Check for uv.lock drift` passed and no `pyproject.toml` is modified, so the lockfile is unaffected. CI runs both.
- Pre-push hooks (copyright headers, UI typecheck, merge conflicts) passed.

Not run: manual end-to-end against a live platform, which needs a running instance. The new tests cover the wire format for both run targets, the recorded run fields, and the two rendering branches.

Existing coverage was deliberately left unmodified — the pre-existing `guardrailChecks.test.ts` cases still pass as written, which is the evidence that a Saved run sends a byte-identical request.

## New Features
- Run guardrail checks against saved configurations or unsaved drafts.
- Select Saved or Draft as the test target; unsaved edits default to Draft.
- Run history identifies draft runs and saved configuration versions.
- Guardrail status reflects the configuration snapshot used during each run.

## Bug Fixes
- Draft runs use the selected model and guardrails and reject drafts without a usable model.
- Preserved status display for older runs without configuration snapshots.

### DIFF
--- a/web/packages/studio/src/api/guardrail-checks/guardrailChecks.test.ts
+++ b/web/packages/studio/src/api/guardrail-checks/guardrailChecks.test.ts
@@ -133,6 +133,75 @@ describe('runGuardrailCheck', () => {
     );
     expect(recordedCheckRequests).toHaveLength(0);
   });
+
+  it('snapshots the saved config coverage onto the run', async () => {
+    const { run } = await runGuardrailCheck(WORKSPACE, snapshot('benign-greeting'));
+
+    // cfg-1 declares two input and two output flows, none of which the registry
+    // recognizes — so each falls back to its raw name.
+    expect(run.activated_guardrails?.map((g) => g.label)).toEqual([
+      'check pii',
+      'check toxicity',
+      'mask pii output',
+      'check output facts',
+    ]);
+    expect(run.is_draft).toBeUndefined();
+  });
+});
+
+describe('runGuardrailCheck against a draft', () => {
+  /** A draft that differs from cfg-1 in both its model and its rails. */
+  const DRAFT: RailsConfig = {
+    models: [{ type: 'main', engine: 'openai', model: 'gpt-4o-draft' }],
+    instructions: [{ type: 'general', content: 'Be extremely cautious.' }],
+    rails: { input: { flows: ['jailbreak detection'] } },
+  };
+
+  it('sends the draft inline instead of referencing the saved config by id', async () => {
+    await runGuardrailCheck(WORKSPACE, snapshot('benign-greeting'), DRAFT);
+
+    expect(recordedCheckRequests).toEqual([
+      {
+        model: 'gpt-4o-draft',
+        messages: [{ role: 'user', content: 'Hello there' }],
+        guardrails: { config: DRAFT },
+      },
+    ]);
+    // Never both: the service's validator would silently discard config_ids.
+    expect(recordedCheckRequests[0]?.guardrails).not.toHaveProperty('config_ids');
+  });
+
+  it('records the run as a draft with no config version', async () => {
+    const { run } = await runGuardrailCheck(WORKSPACE, snapshot('benign-greeting'), DRAFT);
+
+    expect(run.is_draft).toBe(true);
+    expect(run.config_version).toBeUndefined();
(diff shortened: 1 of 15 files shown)


nakolean (approved):

marcusds (approved):
I think a bit heavy handed on the comments, but otherwise approved.

PR #1272, opened 2026-08-13
feat(studio): connect Studio to Intake based eval schema, update modal

nv-odrulea:
## Summary

Studio now publishes agent-evaluation results to Intake and reads them back. Submitting a run creates an Intake Experiment, Evaluation, and config fileset, then sends spec.publication.intake so the run's telemetry lands in Intake.

The agent detail page's Evaluations tab reads those published evaluations via intake/evaluations endpoint with `filter[agent_name]`. Agent evaluations now shows a 3 tab view:
- Active Jobs (evaluation jobs)
- Completed Evaluations (intake evaluations)
- Experiments (tbd if this stays)

Screenshot 2026-08-13 at 2 18 17 PM

Screenshot 2026-08-13 at 2 18 28 PM

Screenshot 2026-08-13 at 2 18 35 PM

#### Intake Evaluation detail page

Screenshot 2026-08-13 at 2 19 18 PM

## Changes

- Submit — submitEvaluationJob.ts gains SubmitSelections.evaluationId and spreads publication.intake into both the agent-evaluate and dataset-driven bodies; job names now stem from the experiment.
- Modal — SubmitEvaluationModal.tsx moves off fileset-picking: "Use Example" creates fileset → Experiment (carrying metadata.eval_config_fileset) → Evaluation → job; "Choose Experiment" replaces "Choose Fileset" and validates that the selected Experiment owns a fileset with a root eval-config.json. Partial failures roll back what the submit created.
- Read — useAgentDetails.ts replaces the job fetch with useListEvaluations(filter[agent_name]); EvaluationsTab becomes two StudioDataView tables behind a segmented toggle (flat evaluations / grouped by experiment), dropping the redundant CTA and footer link.
- Tests/mocks — added an MSW handler for the Intake experiments list.

## Type of Change

## Quality Gates

- [x] Tests added or updated for changed behavior — builder tests cover publication present/absent and job-name precedence
- [ ] Documentation updated fo
- [x] Documentation not applicable — justification: no user-facing docs cover this flow. Caveat below.

## Verification

- [x] Pull request title follotional Commit format
- [x] Every commit includes an appropriate Signed-off-by: trailer
- [x] uv run pre-commit run -a passes, or any blocked checks are identified below
- [x] Targeted tests pass

## Targeted validation:

pnpm --filter nemo-studio-ui test          → 307 files, 2795 tests passed
pnpm --filter nemo-studio-ui t
pnpm lint:fix                              → clean
uv run pre-commit run -a                   → all pass except helm-docs

helm-docs fails with Please install helm-docs to run the pre-commit hook! — the binary is not installed locally. No Helm files are touched by this PR.

## New Features

- Submit evaluations using existing experiments or create new ones automatically.
- Reuse saved evaluation configurations from experiments.
- View active jobs, completed evaluations, or results grouped by experiment.
- Review scores, latency, cost, run counts, and timestamps.
- Navigate directly to evaluation and experiment details.
- Delete evaluations or experiments in bulk.
- Added an email-security triage sample agent with dataset- and task-driven evaluation examples.

## Bug Fixes

- Improved submission validation and failure cleanup.
- Preserved existing experiment resources when submissions fail.


web/packages/studio/src/components/evaluation/SubmitEvaluationModal.tsx
```diff
@@ -365,23 +444,58 @@ export const SubmitEvaluationModal: FC<SubmitEvaluationModalProps> = ({
...
+          : await submitAgentEvalJob(
+              workspace,
+              buildAgentEvalRequestBody(spec, selections) as AgentEvaluateJobRequest
+            );
+        if (!created?.name) throw new Error('Submission did not return a job name');
+        return { name: created.name, isDataset: isDatasetEvalSpec(spec) };
+      } catch (err) {
+        // Re-running an existing experiment seeds nothing, so there is nothing to unwind: its
+        // fileset predates this submit and its Evaluation is reused by the retry.
+        throw await discardSeeded(workspace, seeded, err);
```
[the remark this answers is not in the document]
nv-odrulea: Fixed — the leak is real, but correcting the stated consequence.

**The defect:** confirmed. In Choose-Experiment mode `seeded` stays `{}`, so `discardSeeded` deletes nothing and a failed job POST leaves an orphaned Evaluation (a `run_count: 0` row that still renders under that experiment in `ExperimentDataView`). In Use-Example mode this was already handled transitively, since deleting the Experiment cascades to its Evaluations.

**The correction:** "its name causes the retry to fail with a conflict" is not accurate. The name comes from `buildEvalJobName`, which appends `Math.random().toString(36).slice(2, 10)` (`submitEvaluationJob.ts:37`), so every retry generates a different name. There is no conflict — the retry succeeds and simply leaves a stray row behind. That makes this a resource leak rather than a blocked retry, so Minor rather than Major.

**Fix:** `seeded` now tracks `evaluationName`, and `discardSeeded` deletes the evaluation **only when the experiment is not itself being deleted** — otherwise the server-side cascade already removes it and the extra call would 404. So:

- Use Example: delete Experiment (cascades to the Evaluation) + fileset — unchanged
- Choose Experiment: delete only the Evaluation; the pre-existing Experiment and its fileset are untouched

Also corrected the `discardSeeded` docstring, which repeated the same incorrect name-conflict rationale.

No regression test here: modal tests are deferred to a follow-up PR by agreement with the author.

web/packages/studio/src/routes/agents/AgentDetailRoute/evaluations/formatRollups.ts
```diff
@@ -0,0 +1,38 @@
...
+  value: string;
+}
+
+/** Mean per evaluator. An evaluation names its own evaluators, so these vary row to row and
+ *  cannot each be a column. */
+export const evaluatorScores = (evaluation: AgentEvaluationRow): EvaluatorScore[] =>
+  Object.entries(evaluation.aggregate_scores ?? {}).map(([key, aggregate]) => ({
+    label: evaluatorLabel(key),
+    value: formatScore(aggregate?.mean),
+  }));
```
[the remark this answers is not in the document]
nv-odrulea: Fixed — valid catch.

`evaluatorLabel` drops the metric-type prefix whenever the score carries its own name, so two evaluators with the same score name collide: `llm-judge.accuracy` and `string-check.accuracy` both label as `accuracy`, producing duplicate React keys on the chips.

`EvaluatorScore` now carries the raw `aggregate_scores` key alongside the label, and `EvaluationsTable` uses `key={score.key}` for identity while still rendering `score.label`.

Not currently reachable with the shipped sample configs (their keys are `llm-judge.accuracy`, `llm-judge.attack_type_plausible`, `string-check.string-check` — all distinct after labelling), but it is latent and the fix is two lines.

web/packages/studio/src/routes/agents/AgentDetailRoute/useAgentDetails.ts
```diff
@@ -91,13 +97,28 @@ export const useAgentDetails = ({
...
   );
 
-  const agentEvals = useMemo(() => {
+  // The evaluation detail route is nested under an experiment, but an evaluation carries only
+  // ``experiment_ids``. One list call resolves them; the panel shows a handful of rows, so this
+  // is cheaper than a lookup per row.
+  const { data: experimentsResponse } = useListExperiments(
+    workspace,
+    { page_size: EXPERIMENT_PAGE_SIZE },
+    { query: { enabled: !!agentName && !!workspace } }
```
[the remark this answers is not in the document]
nv-odrulea: Fixed, though not by either suggested route — one of them isn't available.

**The defect is real, and the grouping half is the serious part.** Two unequal consequences:

- Flat table: already degraded gracefully — `onRowClick` checks `experimentName` and just doesn't navigate.
- Experiments view: `groupByExperiment` did `if (!evaluation.experimentName) continue`, which **silently dropped the evaluation from the rollup**, understating `evaluationCount` and `runCount` with no indication. That's the genuine data-integrity bug.

**"Fetch the referenced IDs directly" is not possible.** `getExperiment` takes a *name*, and `ExperimentFilter` exposes only `name`, `insight_id`, `is_deleted`, `metadata` — there is no id or id-list filter. We hold opaque ids (`experiment-group-…`), so there is no endpoint to resolve them one by one.

That left pagination, which is disproportionate here: the panel renders at most `RECENT_EVAL_LIMIT` rows, so paging an entire workspace's experiments to label a handful is a lot of requests for a summary view.

**What I did instead — remove the failure mode rather than mitigate it:**

- `groupByExperiment` now keys on `experiment_ids[0]`, which every evaluation carries, instead of the resolved name. No evaluation is ever dropped and the counts are always correct; an unresolved row loses only its label (falls back to the id) and its link.
- Added `sort: '-created_at'` to the experiments query, which was previously unsorted — so the 100-item window now matches the sibling evaluations query and misses are rare rather than arbitrary.
- Bulk delete rejects a selection containing unresolvable rows with a clear message, since deletion is by name.

web/packages/studio/public/sample-agents/email-security-triage/eval-config.dataset-driven.README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Five rows (three phishing, two benign), each with `subject`, `sender`, `body`, and `label`.
+`prompt_template` assembles them into the RFC-822 message the agent expects:
+
+```
+From: {{ item.sender }}
+Subject: {{ item.subject }}
+
+{{ item.body }}
+```
```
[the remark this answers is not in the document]
nv-odrulea: Fixed — tagged the block `jinja`, which is what it is (the `prompt_template` value, rendered per row).

Checked the sibling README while here; its verdict block was already tagged `yaml`, so this was the only MD040 hit in the two files added by this PR.

web/packages/studio/src/util/sampleAgents.ts
```diff
@@ -4,6 +4,26 @@
...
+
+const asObject = (value: unknown): Record<string, unknown> | null =>
+  value && typeof value === 'object' && !Array.isArray(value)
+    ? (value as Record<string, unknown>)
+    : null;
+
+/** The model a sample agent's config names, so the create modal can preselect it.
+ *
+ *  Mirrors the two shapes ``loadSampleAgentConfig`` writes back to:
+ *  - NAT (`nat-workflow-v1`): `llms.llm.model_name`
```
steramae-nvidia, line 23: Aren't we getting rid of NAT?

web/packages/studio/src/routes/agents/AgentDetailRoute/useAgentDetails.ts
```diff
@@ -58,18 +69,29 @@ export const useAgentDetails = ({
...
+    workspace,
+    {
+      filter: { agent_name: agentName ?? '' },
+      page_size: RECENT_EVAL_LIMIT,
+      sort: '-created_at',
+    },
+    { query: { enabled: !!agentName && !!workspace } }
+  );
+
+  const { data: agentJobsData } = useQuery({
```
steramae-nvidia, line 82: Can we make this a custom hook? Also fetchEvaluatorJobs seems pretty rough, is there any way we can prevent ourselves from exhaustively fetching evals?

web/packages/studio/src/routes/agents/AgentDetailRoute/evaluations/JobsTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+      onRowClick={(row) => navigate(destinationFor(row))}
+      attributes={{
+        DataViewRoot: { data: jobs },
+        DataViewTableContent: {
+          renderEmptyState: () => (
+            <TableEmptyState
+              icon={<ListChecks className="size-16" />}
+              header="No evaluation jobs yet"
+              emptyMessage="Runs appear here as soon as they are submitted, before any results are published."
+            />
```
steramae-nvidia, line 93: This will need to line up with Aaron's new changes to consolidating empty state

web/packages/studio/src/routes/agents/AgentDetailRoute/evaluations/ExperimentsTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+            onClick={() => setDeleteRows(selectedRows)}
+          >
+            <Trash /> Delete
+          </Button>
+        )}
+        attributes={{
+          DataViewRoot: { data: experiments },
+          DataViewTableContent: {
+            renderEmptyState: () => (
+              <TableEmptyState
```
steramae-nvidia, line 126: This will also need to be adjusted for the empty state consolidation

web/packages/studio/src/routes/agents/AgentDetailRoute/evaluations/EvaluationsTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+            onClick={() => setDeleteRows(selectedRows)}
+          >
+            <Trash /> Delete
+          </Button>
+        )}
+        attributes={{
+          DataViewRoot: { data: evaluations },
+          DataViewTableContent: {
+            renderEmptyState: () => (
+              <TableEmptyState
```
steramae-nvidia, line 138: Empty state consolidation

web/packages/studio/src/components/evaluation/experimentEvalConfig.ts
```diff
@@ -0,0 +1,67 @@
...
+    signal,
+  }: {
+    experimentId: string;
+    experimentName: string;
+    filesetName: string;
+    signal?: AbortSignal;
+  }
+): Promise<string> => {
+  const name = buildEvalJobName(experimentName);
+  await createEvaluation(
```
steramae-nvidia, line 61: why not  just use the mutate methods from the `useEvaluatorCreateEvaluateJob` hook? If you don't, then you'd lose out on the React Query hooks and cache

nv-odrulea: Different service endpoints, both SDK

`evaluatorCreateEvaluateJob` → `nemo-evaluator` → `POST /apis/evaluator/v2/workspaces/{workspace}/evaluate/jobs`


`createEvaluation` → `services/intake` → `POST /apis/intake/v2/workspaces/{workspace}/evaluations`

"Evaluation" is now overloaded. 

It can mean Intake's `Evaluation` entity, or the `EvaluateJob` that `nemo-evaluator` defines and runs on the shared jobs service.
 
Both are necessary for this integration. When a job finishes it publishes its results to Intake, and it needs an Intake Evaluation to publish into. That's the `publication.intake.evaluation_id` field on the job spec (it takes the evaluation's name, not its id). So the modal creates the "Intake Evaluation" first, then submits the job pointing at it.  Intake's experiments have their own evaluations which are totally independent of `nemo-evaluator`.

This is why I'm starting to use the names "intake evaluation" vs "evaluation jobs" to make the distinction.

web/packages/studio/src/routes/agents/AgentDetailRoute/useAgentDetails.ts
nv-odrulea: So, on jobs, `agent` is not a filterable field. `PlatformJobsListFilter` allows only `workspace`, `project`, `name`, `created_at`, `updated_at`, `status`, `source`. 

The agent lives at `spec.target.agent.name` (targetNameForEvalJob, utils.ts:42-51), and nested `spec.*` isn't filterable. So the filtering must happen client-side, which forces some amount of walking. This is a temporary hack. We can extract to a hook but that will likely be obsolete once we are able to filter by agent.

It's not totally unbounded today. `shouldStop` halts once 5 matching jobs are found (RECENT_EVAL_LIMIT = 5)

```
fetchEvaluatorJobs(workspace, signal, (all) =>
  all.filter((job) => targetNameForEvalJob(job) === agentName).length >= RECENT_EVAL_LIMIT)
```

Still, in the case where an agent has few or no eval jobs, this consumes all available jobs in workspace before it ever hits 5 matching records. I'll add a hard page cap of 5 as well so we don't go infinitely.

I expect this temporary logic will be replaced entirely once we are able to filter jobs by agent. For now, we just need to show running jobs under agent detail so we're tolerating this client side hackery.

web/packages/studio/src/util/sampleAgents.ts
nv-odrulea: We are. But it's not deprecated yet. (NAT is still default config type in `nemo-agents`)

Keeping this here temporarily helps me quickly spin up a NAT agent for troubleshooting as I'm trying to get the Fabric agent to parity. In this case, the `model` goes in a different place depending on whether you have a `nat-workflow-v1` vs `nemo-agents-spec-v1` config schema.  Just spares me having to vibecode this on the fly every time I want to experimentally run a NAT agent next to the Fabric agent to see what the delta is.

I'd prefer to remove all the NAT stuff in one standalone PR later when it's really deprecated. Until the Fabric agent is ready I think it's premature.

steramae-nvidia (approved):

web/packages/studio/src/routes/agents/AgentDetailRoute/useAgentDetails.ts
nv-odrulea: yes, adding stop predicate on max pages = 5, so we will not exhaust all possible pages in cases where there's an agent with zero matches. 

before:
"look through all pages until you find 5 matches"

after:
"look through up to 5 pages until you find up to 5 matches"

PR #1289, opened 2026-08-13
feat(customization): accum every training metric as time series

albcui:
## Summary

Training progress reporting was refactored to handle generic metrics from each training backend, and made sure that they are recorded as time series in the job's status details, under one rule, shared across all backends.

This is foundational work, which the new GRPO-specific metrics depend on.

## Changes

1\. Generalization of metrics

Previously, we accepted `loss, lr, grad_norm` as hardcoded metric names. Now, `report_train_step(loss=, lr=, grad_norm=)` becomes `report_train_step(metrics={...})` — backends hand over their framework's own dict and the phase supplies the prefix, so every metric is stored and reported as `<phase>_<name>`. `train_loss` and `val_loss` naturally come out of this rule as `phase="train"/"val"` and `name="loss"`, not special cases. Alongside it, we accumulate a time series for every reported metric under `{series: [{step, epoch, value}]}`.

Deleted the RL service's own 95-line copy of `TrainingProgressCallback` in favor of the shared one, and added a thin `services/rl/.../training/progress.py` that binds the RL service name — mirroring what unsloth and automodel already had.

> Note: Studio now reads `details.train_lr` / `details.train_grad_norm` instead of `details.lr` / `details.grad_norm`.

2\. Transport: what a report is allowed to say

The Jobs service merges `status_details` key-wise, but the merge is shallow -- a sent key replaces the stored value wholesale, so `metrics` needs to be either sent in full or omitted entirely. Hardened the checkpoint / epoch-end / training-start report methods to omit metrics rather than sending an empty copy.

3\. log_metrics throttling at NeMo RL logger level

NeMO RL library doesn't expose a config to control how often to call `log_metrics`. It currently calls `log_metrics` on every step, which is very noisy. Especially when considering how we need to send the entire accumulated metrics to the jobs service each time we need to make an update, which grows quadratically relative to how many times we call the service. This means we need to implement some kind of throttling mechanism to limit the number of reports to the jobs service. This is why we introduce `_MAX_REPORTS_PER_RUN=200`, which applies an upper limit to how many times we can call the jobs service. Which step ends up being reported depends on the max steps. This limit is applied independently to both the training and validation paths, so at most we make 400 requests to the jobs service.

3\. Backend-specific naming repairs

Two backends violated the new rule in opposite directions. Automodel's recipes pre-prefix some validation metrics, so names came back doubled (`val_val_acc1`) — strip_val_prefix takes it off wherever the recipe put one. And the RL validation branch required a chartable loss before reporting anything, which dropped every validation pass GRPO ran, since it scores on accuracy and avg_length and produces no loss at all.

4\. Docs and housekeeping

get-job-status.mdx and metrics.mdx document the naming rule and the series payload; the rest is headers, an unused reference, and CodeRabbit follow-ups.

## Type of Change

## Quality Gates

- [ ] Documentation updated for user-visible behavior — see note below

## Verification

- [x] Targeted tests pass

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Training progress supports additional numeric metrics, including learning rate and gradient norms.
  - Metric histories are preserved when training resumes and separated by training phase.
  - Validation reports can omit unavailable loss or checkpoint information.
  - Reinforcement-learning progress updates follow configured training schedules.

- **Bug Fixes**
  - Unsupported, infinite, or malformed metric values are excluded.
  - Final progress updates are reliably sent when training ends, including after interruptions.
  - Progress percentages are correctly bounded, and stored metric data is safely isolated.


docs/customizer/tutorials/metrics.mdx
```diff
@@ -27,10 +27,38 @@ The time to complete this tutorial is approximately 10 minutes.
 
 ## Available Metrics
 
-Each customization job tracks two key metrics:
+A customization job tracks every numeric metric its backend reports, not a fixed
```
gabwow: Is backend the right word here? DPO and GRPO are same backend, do they report the same metrics?

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
```diff
@@ -4,25 +4,191 @@
...
+    try:
+        return math.isfinite(float(value))
+    except (OverflowError, ValueError):
+        # An unbounded Python int overflows float(). The contract here is that
+        # any value a backend hands us gets classified without raising, so an
+        # absurd counter loses its series rather than the run losing reporting.
+        return False
+
+
+def _namespace(phase: str, metrics: Mapping[str, object]) -> dict[str, float | int]:
```
gabwow: Maybe "_reportable_metrics_for_phase" is better than introducing an overloaded concept of a namespace and also capturing the casting

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
```diff
@@ -206,14 +346,43 @@ def log_plot(self, figure: Any, step: int, name: str) -> None:
...
+        """Report whatever the throttles withheld, oldest step first.
+
+        Step order so a series' points are appended in the order they were
+        produced, rather than in whichever order the kinds happen to iterate.
+
+        Reachable from ``__del__``, so failures must not propagate; the reporter
+        already swallows and logs transport errors, and this guards the rest.
+        """
+        pending, self._pending = self._pending, {}
+        for kind, report in sorted(pending.items(), key=lambda item: item[1]["step"]):
```
gabwow: Are we worried about:
```
step 90: validation happens, throttled -> pending["validation"] = step 90
step 100: train happens, interval hits -> train step 100 is sent immediately
close(): flushes pending validation step 90
```

gabwow (approved):
General non-blocking question: do we want some sort of contract with studio now that we're returning general dictionaries?

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
```diff
@@ -9,23 +9,102 @@
...
+# the blob twice, so upload grows as the square of the report count -- see the
+# payload note in nmp.customization_common.training.callbacks. Capping the count is
+# what keeps that finite; the resolution argument is what makes 200 the right
+# number rather than merely a small one.
+#
+# Per path, so a run making full use of both is bounded by twice this. That is a
+# bound either way; what matters is that no path is unbounded, and applying one
+# budget across two cadences that fire independently would mean whichever ran
+# first starved the other.
+_MAX_REPORTS_PER_RUN = 200
```
albcui: The problem this solves is that the NeMo RL library currently doesn't expose any logging cadence controls, so it's logging every step. However, this could be very costly for us if we simply call the jobs service each step, because we now accumulate the metrics in a time series and send the whole payload each time. `_MAX_REPORTS_PER_RUN` is applying a limit to how many times we can communicate to the jobs service.

Relatedly, this issue is also in Automodel and Unsloth. Ideally, we consolidate the logging cadence control and logic in one place. I'm thinking of exposing a `ProgressReportingConfig.max_reports` field that applies to all training backends, and centralizing the interval resolution logic in the `customization_common` package.

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
```diff
@@ -33,73 +199,151 @@ class TrainingProgressCallback:
...
-        self._train_metrics.append({"step": step, "epoch": epoch, "value": loss})
+        """Record every metric as a point and report them as current values.
+
+        The one path both ``report_train_step`` and ``report_validation`` take.
+        ``phase`` namespaces the series (``train``/``val``); ``report_phase`` is
+        what the Jobs service records as the task's phase.
+        """
+        namespaced = _namespace(phase, metrics)
+        for name, value in namespaced.items():
+            self._series.setdefault(name, []).append({"step": step, "epoch": epoch, "value": value})
```
albcui: Metrics time series accumulation logic is consolidated and centralized in this one line.

docs/customizer/tutorials/metrics.mdx
albcui: The infrastructure for logging metrics is at the backend level. DPO and GRPO share the same logger, just different metrics. Previously, we were being opinionated about which metrics to report, now I'm just going to report all numeric metrics given to us. Maybe we should have an allowlist, but we can add that later. This means,  once this change lands, we should more or less get GRPO metrics for free.

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
albcui: Hmm is the worry that the flush would race against the train at step 100?

I think it should result in something like
```
"train_<name>": [100]
"val_<name>": [90]
```

However, I'm not a big fan of the throttling logic implemented at the `nemo_rl_logger.py` level. I think this logic should be centralized in the `customization_common`.

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
albcui: Good point, renamed to `_qualify_metric_names`.

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
gabwow: Yeah, the concern is pending is sorted against itself, but we could still have unordered timestamps post flush.

packages/nmp_customization_common/src/nmp/customization_common/training/reporting.py
```diff
@@ -0,0 +1,100 @@
...
+#: recorded; this only bounds how often the accumulator is sent.
+#:
+#: Ten seconds is chosen against what it costs and what it buys, both measured: a
+#: 594-step run over half an hour drops from 594 requests to 180 and from 36s to
+#: 11s blocked inside the training loop, while a progress bar that moves every
+#: ten seconds is as live as anyone reads one.
+DEFAULT_MIN_REPORT_INTERVAL_SECONDS = 10.0
+
+
+class ProgressReportingConfig(BaseModel):
```
albcui, line 42: Net new API object, which will be attached to each training backend job spec.

packages/nmp_customization_common/src/nmp/customization_common/training/reporting.py
```diff
@@ -0,0 +1,100 @@
...
+    # Every model this is embedded in forbids extras, inheriting it from
+    # NamespacedModel, and unsloth's schema module states the contract outright:
+    # "typos in the JSON shape become validation errors, not silently-ignored
+    # fields". This model is not a NamespacedModel -- deliberately, so it emits
+    # one shared OpenAPI component rather than three -- and so it did not inherit
+    # that, which made it the one place in the request body where
+    # `time_series_metric` was accepted and quietly did nothing.
+    model_config = ConfigDict(extra="forbid")
+
+    min_report_interval_seconds: float = Field(
```
albcui: Instead of throttling, and reducing the resolution of the log curves, I decided to keep the resolution (pay the full payload size throughout the life of a job), but rate limit the number of times we communicate to the Jobs service.

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
```diff
@@ -30,76 +273,298 @@ class TrainingProgressCallback:
...
+            reporter: Transport to the Jobs service.
+            time_series_metrics: Qualified metric names or glob patterns to
+                accumulate, or None for all of them.
+            min_report_interval_seconds: Least time between metric reports going
+                on the wire. 0 sends every one. Points are recorded regardless.
+            clock: Monotonic time source, injectable so the limiter can be tested
+                without sleeping.
+        """
         self._reporter = reporter
+        self._limiter = ReportRateLimiter(min_report_interval_seconds, clock)
```
albcui, line 296: New rate limiter decides when to fire a request to Jobs service

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
```diff
@@ -30,76 +273,298 @@ class TrainingProgressCallback:
...
+        #: Where training was when a metric was last observed, cleared once sent.
+        #: Doubles as the "a report is owed" flag ``close()`` reads.
+        self._current_progress: _Progress | None = None
+        #: Whether a point has been appended since the last send, which decides
+        #: whether the next report carries the series at all. Only the series are
+        #: gated: they are the expensive part, the scalars are a few numbers.
+        self._series_dirty = False
+        #: None becomes "*", which is what it already meant, so there is one code
+        #: path. ``[]`` stays empty -- a different and equally valid request.
+        self._time_series_metrics: tuple[str, ...] = _clean_patterns(time_series_metrics)
```
albcui, line 306: Not to be confused with the actual metrics stored. These are a list of metric names/patterns to signal to us which values need to be accumulated as time-series (as opposed to which values are scalars).

packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py
```diff
@@ -30,76 +273,298 @@ class TrainingProgressCallback:
...
+        #: bare number is a scalar and is replaced. Which one a name is gets
+        #: decided once, on first arrival, by :meth:`_records_history`, and is
+        #: never re-checked.
+        #:
+        #: Seeded from the server (see the module docstring). Everything seeded
+        #: arrives as a list and so stays a series, including a name the current
+        #: patterns would not select -- its history exists, and freezing it
+        #: half-written serves nobody. A seeded point recording no step is
+        #: dropped: nothing can place it on a curve.
+        self._metrics: dict[str, float | int | list[dict[str, float | int]]] = {
```
albcui, line 346: Holds the actual metrics we want to store in the job's status_details. Carries both time-series and scalar values  with the type as discriminator.

Things worth calling out on:
- previously, we took advantage of the Jobs service' shallow merge capability, where we report a single scalar metric key, and let the Jobs service update just that key's value. However, because we delay requests to the Jobs service now, we basically snapshot a copy of the status_details' `"metrics"` field, keep mutating it until the `update_task` request is allowed by the rate limiter, and submit the entire metrics payload altogether.
- This has some nice side effects. For example, if a previous request failed, we can self-heal by sending it again.
- We pay the cost of `isinstance()` for every metric key, but we gain simplicity by storing `self._metrics` more or less the same payload shape as what's stored in a job's `status_details.metrics`.

plugins/nemo-automodel/src/nemo_automodel_plugin/schema.py
```diff
@@ -92,6 +93,7 @@ class ScheduleSpec(AutomodelSchema):
     max_steps: int | None = Field(default=None, gt=0)
     val_check_interval: float | None = None
     seed: int | None = None
+    progress_reporting: ProgressReportingConfig = Field(default_factory=ProgressReportingConfig)
```
albcui, line 96: We should agree that `schedule` is right location to embed `ProgressReportingConfig`.

docs/customizer/tutorials/metrics.mdx
gabwow: @albcui I don't think this addresses my question. I think
"A customization job tracks every numeric metric its backend reports, not a fixed..." should be something like "A customization job tracks every numeric metric a given training  algorithm reports, not a fixed..."

packages/nmp_customization_common/src/nmp/customization_common/training/reporting.py
```diff
@@ -0,0 +1,100 @@
...
+    stays correct when a framework adds a metric.
+    """
+
+    # Every model this is embedded in forbids extras, inheriting it from
+    # NamespacedModel, and unsloth's schema module states the contract outright:
+    # "typos in the JSON shape become validation errors, not silently-ignored
+    # fields". This model is not a NamespacedModel -- deliberately, so it emits
+    # one shared OpenAPI component rather than three -- and so it did not inherit
+    # that, which made it the one place in the request body where
+    # `time_series_metric` was accepted and quietly did nothing.
```
gabwow: I think this comment can be more succinct and clearer.

packages/nmp_customization_common/src/nmp/customization_common/training/reporting.py
```diff
@@ -0,0 +1,100 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    min_report_interval_seconds: float = Field(
+        default=DEFAULT_MIN_REPORT_INTERVAL_SECONDS,
```
gabwow, line 54: question: why is the dimension seconds vs an aspect of training like epoch?

gabwow (changes requested):
a question and a comment

gabwow:
Assuming  the wording comments are addressed, feel free to resolve my comments and merge

gabwow (approved):

docs/customizer/tutorials/metrics.mdx
albcui: ah gotcha, sorry I misunderstood your question.

PR #1295, opened 2026-08-13
fix(evaluator): make Gym run failures legible and catch rollouts where no agent ran

SandyChapman:
## Summary

Running the [AALGO-485](https://linear.app/nvidia/issue/AALGO-485/validate-the-gym-runner-across-a-representative-sample-of-built-in) coverage sweep across five Gym environments found two ways a Gym run fails without saying anything useful, and one way it fails while reporting success. Before: a startup timeout said only `Gym servers not ready within 240.0s`; a collection failure named two logs, neither holding the cause; and an environment whose agent never started produced "completed" trials scoring 0.0. After: the timeout names the outstanding server and the knob to raise, the collection failure points at the log with the traceback, and a rollout where the model was never called is failed rather than scored.

Also makes the rollout coverage tests runnable without a model endpoint, which is what let the sweep establish any of this — that half of the suite had never executed.

## Related Issue

Tracked in Linear as AALGO-485. No GitHub issue.

Follows #1280 (AALGO-498, promptless Gym rows), now merged. Rebased onto `main`, so this is a single commit containing only this change. `gdpval` and `legal_agent_bench` collect rollouts end to end because of #1280 — without it they fail at task discovery.

## Changes

- **Startup timeouts name the server that did not come up.** New `_pending_servers` parses Gym's own repeated readiness line and reports the *last* one — earlier polls list servers that have since started, so reporting those sends the reader after healthy servers. The message also states how to extend the wait, because a cold cache is a legitimate cause rather than a symptom.
- **Collection failures point at `gym_env.log`.** `gym eval run` surfaces a server-side fault as a bare HTTP 500; the traceback lives in the startup log the message never named.
- **`_agent_never_ran`** — a rollout with no output *and* stated zero input-side token usage is marked `FAILED` with its reward withheld, and the run raises when that is the whole picture. Input-side only (`total_tokens` / `input_tokens` / `prompt_tokens`): a model call always sends a prompt, whereas zero *output* tokens merely describes an empty answer, which is a legitimate result.
- **`gym eval run` gets the `hydra.run.dir` redirect.** It builds its own argv and never received the one `_selection_args` carries, so every collection dropped `outputs/<date>/<time>/` into the caller's cwd — contrary to #1203's description.
- **`_StubPolicyServer`** — a stdlib OpenAI-compatible endpoint so the rollout tests no longer skip. It answers each prompt with *that row's* own ground truth, so a mis-attributed rollout scores 0 and fails the test; a canned reply would score identically to a correct run. `NEMO_GYM_POLICY_BASE_URL` remains an override for a real model.
- **Coverage cases record what each environment actually needs** — `legal_agent_bench`'s asset directories must be absolute and it needs ~8 minutes of startup; `gdpval` binds its resources-server under a different name.

### Design calls

- **Detection is conservative in one direction on purpose.** Both conditions must hold, and absence of a usage key is never read as zero. The failure mode to avoid is failing a healthy trial, not missing a broken one.
- **Only `FAILED` when it can be proven.** An empty answer with tokens spent stays `COMPLETED` — a model may legitimately answer with nothing and earn 0.0.
- **The raise requires the whole run to be empty**, and stands down when a failures-sidecar trial or an unattributed rollout exists: those carry their own diagnosis, and discarding them for a vaguer message would be a loss.
- **Per-case `startup_timeout_s` rather than a higher default.** A long global default turns a genuinely wedged server into a long wait for every environment.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the behaviour lives in this runtime's own docstrings, updated here at module, function and case-table level. No Fern page documents Gym run diagnostics.

## Verification

Targeted validation:

| Command | Result |
| -- | -- |
| `pytest packages/nemo_evaluator_sdk/tests -q` | **1581 passed, 10 skipped** |
| `pytest .../test_gym_environment_coverage.py -q -ra` (real Gym, live rollouts) | **9 passed, 1 skipped** |
| `uv run ruff check` | All checks passed |
| `uv run ruff format --check packages/nemo_evaluator_sdk/` | 233 files already formatted |
| `tools/lint/lint-python-types.sh` | exit 0 |
| `make vendor` | vendored mirror in sync, idempotent across runs |

**The `pre-commit run -a` box is left unchecked.** Four hooks fail for host-toolchain reasons; each is recorded below rather than dismissed.

### Coverage sweep

Run against `nemo-gym` 0.5.0 installed from PyPI on CPython 3.13.14, with its `bin` on PATH. Four of five environments collect real rollouts end to end; the fifth needs a GPU.

```
  mcqa[baseline]              validate ✓   rollout ✓  reward 1.0
  gpqa_diamond[tiny]          validate ✓   rollout ✓  reward 1.0
  gdpval[heavy-cpu-judge]     validate ✓   rollout ✓  (unblocked by #1280)
  legal_agent_bench[docker]   validate ✓   rollout ✓  (absolute asset paths + 1800s startup)
  wmt_translation[gpu]        validate ✓   rollout — skipped, no nvidia-smi (AALGO-494)
```

`mcqa` and `gpqa_diamond` assert a perfect score, which is the attribution check: each task is answered with its own ground truth, so any mis-pairing scores 0.

**Tests were checked by mutation, not coverage.** Disabling the detection fails 3 tests; widening the input-token key set fails 2; dropping the unattributed-rollout guard, the `hydra.run.dir` argument, the withheld reward, or the last-poll rule each fail 1.

### Reviewer note — please weigh in

`_agent_never_ran` went through five rounds of independent adversarial review, and rounds 1, 2, 4 and 5 each found another rollout shape where it misfired **in the destructive direction** — failing a healthy trial or aborting a valid run. Every one is fixed and regression-tested, and the last change moves from excluding shapes one at a time to an invariant (a call always consumes input tokens) that should retire the class. But the honest read is that inferring "the model was never called" from an unstandardised usage block is a heuristic, and it is wired to a destructive action.

Worth a second opinion on whether the runner should be able to **fail an entire run** on that inference, or whether it should stamp the trial and warn, leaving the decision to the caller. The detection itself earned its place — it is what caught `legal_agent_bench` silently scoring 0.0 — but the blast radius is a judgement call.

### Blocked checks (host toolchain, not this change)

| Hook | Failure | Why it is not this change |
| -- | -- | -- |
| `uv-lock` | requires uv 0.9.14, host has 0.9.30 | diff touches no `pyproject.toml`/`uv.lock`; `uv-lock-check`, which catches real drift, passes |
| `studio-lint-staged` | `mise ERROR No version is set for shim: pnpm` | diff touches no `web/` files |
| `Helm Docs` | helm-docs container unavailable locally | diff touches no Helm chart |
| `ty` (pre-commit) | 9 diagnostics, all `invalid-argument-type` on the pre-existing `_config` test helper | that rule is the first entry in CI's own ignore list (`tools/lint/lint-python-types.sh`, 204 repo-wide instances), and that script exits 0. The one diagnostic that *was* mine — a `log_message` override signature — is fixed rather than ignored |

### Follow-up found, not filed

`wmt_translation` hardcodes `/opt/Gym/.cache/comet-python` as its cache root, which exists only inside NVIDIA's containers; on a developer machine it fails with `PermissionError: /opt/Gym` long before any GPU requirement bites. Overridable via `WMT_TRANSLATION_COMET_PY_CACHE`, and it affects 2 of 106 environments (`wmt_translation`, `longmt_eval`). Worth an upstream Gym request.

## Summary by CodeRabbit

* **New Features**
  * Added a complete Gym-based agent evaluation workflow with task discovery, rollout execution, reward scoring, aggregate metrics, and result reporting.
  * Added support for promptless tasks, answer formatting, startup timeouts, and environment overrides.

* **Bug Fixes**
  * Improved diagnostics for startup, collection, logging, and partial failures.
  * Detects agents that never ran and prevents misleading rewards.

* **Testing**
  * Expanded coverage for Gym execution, failures, readiness, logging, and timeout handling.
  * Added a built-in policy stub for reliable testing without a model endpoint.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -848,7 +872,129 @@ async def _wait_for_servers(self, env_log: Path, env_proc: asyncio.subprocess.Pr
             if ready.search(text):
                 return
             await asyncio.sleep(2)
-        raise TimeoutError(f"Gym servers not ready within {cfg.startup_timeout_s}s; see {env_log}")
+        raise TimeoutError(self._startup_timeout_message(text, env_log))
+
+    def _startup_timeout_message(self, env_log_text: str, env_log: Path) -> str:
```
ngoncharenko: Considering the file is quite long, non-core logic utils can go to other files

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -848,7 +872,129 @@ async def _wait_for_servers(self, env_log: Path, env_proc: asyncio.subprocess.Pr
...
+        return False
+    usage = response.get("usage")
+    if not isinstance(usage, Mapping):
+        # No usage block to corroborate with; an empty output on its own is not enough to judge.
+        return False
+    # Zero usage has to be *stated*, not inferred from a key being absent. Two vocabularies are in
+    # play — the Responses API's `input_tokens`/`output_tokens` and Chat Completions'
+    # `prompt_tokens`/`completion_tokens` — and Gym's model servers speak both depending on the
+    # adapter. Reading only one set would make a real call reported in the other read as zero.
+    reported = [usage[key] for key in _TOKEN_USAGE_KEYS if isinstance(usage.get(key), (int, float))]
```
ngoncharenko: I wonder if we could import those keys' ids from the gym directly, so we point to the source of truth. Myb understanding, there's a reason why we don't do it, right?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
```diff
@@ -1321,6 +1489,40 @@ def _trials_from_rollouts(
...
+        # that did run.
+        #
+        # Measured against *every* trial, including those the failures sidecar contributed. A run
+        # that also recorded real failures is not the uniform "nothing ran" picture this raise is
+        # for: those trials carry their own diagnosis (a verifier timeout, say, from an attempt that
+        # did reach the model), and raising would discard it for a less specific message. Likewise
+        # `unattributed_successes` never became trials and so are invisible to `len(trials)`, but
+        # each is a result the environment produced and one may have run perfectly well — claiming
+        # nothing was measured while such a record exists would be false.
+        if empty_output_count == len(trials) and not unattributed_successes:
```
ngoncharenko: do we also need to consider `not unattributed_failures` case here?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Right now we don't have a dependency on the Gym Python package (we just shell out assuming the user installed it). This is because depending on Gym pulls in Ray which is specifically excluded due to CVEs. We'll either need a way to include Gym without Ray or to loosen our restriction on Ray.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: I broke it up into a package of its own. Thanks for the suggestion

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/gym_runtime.py
SandyChapman: Thanks. Fixed.

PR #1310, opened 2026-08-14
feat(evaluator): roll up trial errors into summary and trial

ngoncharenko:
## Summary

- `AgentEvalSummary` now reproduces Harbor's `exception_stats`. Trials carry a typed `TrialError` instead of a stringly-typed `metadata["exception_type"]`, and the summary rolls those up as `{error type: [trial_id, ...]}` — byte-identical to Harbor's shape (see examples below). 
- Before, the only rollup was inside `reward_payload_from_result`, which re-walked `result.trials`, keyed by **task**, and could not be derived from a persisted summary.

## Related Issue

AALGO-428 (P3.2). Unblocks AALGO-441 (P3.4).

## Examples

Summary output — Harbor's shape, no reconstruction:

```python
summary.error_trial_ids  # {"RuntimeError": ["alpha__b", "beta__a"], "TimeoutError": ["gamma__a"]}
summary.error_count      # 3
```

`_trial_error` is total by construction. `_trial_from_harbor_result` runs *outside* the only `try/except` in `build_trials_from_job_dir` (which guards `json.loads` alone), so a `ValidationError` here would abandon every remaining trial in the job dir:

```python
_trial_error("")                          # -> TrialError(type="UnknownException")   (not a raise)
_trial_error({"exception_type": 123})     # -> TrialError(type="UnknownException")
_trial_error({"exception_message": 42})   # -> message dropped, not coerced
# traceback truncated to _MAX_TRACEBACK_CHARS; unparseable occurred_at -> None
```

Two deliberate calls worth a look:

- **No status filter.** An errored Harbor trial is `PARTIAL`, not `FAILED`, so it is still scored. A trial that errored *and* scored 1.0 appears in the rollup **and** counts as a pass in `task_metric_values` — Harbor double-files it the same way.
- **`occurred_at` is not `format: date-time`.** RFC 3339 requires an offset; Harbor writes naive local time (`2026-08-13T17:22:32`) while stamping trial start in UTC. Claiming the format would make a JS client parse it into its own zone and silently shift the instant.

## Changes

- `TrialError` (`type`/`message`/`traceback`/`occurred_at`) + `AgentEvalTrial.error`; frozen, `extra="forbid"`
- `AgentEvalSummary.error_trial_ids` + `error_count`; `from_scores()` gains keyword-only `trials=`
- Harbor adapter emits the typed error and **stops writing** `metadata["exception_type"]`; `reward_payload_from_result` reads `trial.error.type` (keeps its task-keyed shape — AALGO-441 changes that)
- `_metric_row` exposes `trial.error`, so metrics have a typed path
- Read-side lift: a pre-`TrialError` bundle's `metadata["exception_type"]` still resolves to `.error`
- Regenerated `plugins/nemo-evaluator/openapi/openapi.yaml` (`AgentEvalTrial` is public API) and the vendored SDK mirror

### Where to focus review

| File | Why |
|---|---|
| `agent_eval/trials.py` | The new model + the legacy-bundle lift |
| `agent_eval/runtimes/harbor_runtime.py` | `_trial_error` must be **total** — see below |
| `agent_eval/results.py` | Rollup field + `_error_trial_ids` |

Everything else is tests, the vendored mirror (generated), and the regenerated spec.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: every new field carries a `Field(description=...)` that lands in the generated OpenAPI schema; no prose doc covers this surface yet.

## Verification

Targeted validation:

- `uv run --frozen pytest packages/nemo_evaluator_sdk/tests plugins/nemo-evaluator/tests plugins/nemo-optimization/tests -q -m "not integration"` — **2479 passed, 2 skipped, 42 deselected**
- `uv run ruff check packages plugins` — passed; `uv run ruff format --check packages plugins` — 1938 files already formatted
- `uv run --frozen ty check <changed files>` — **zero new diagnostics**; the 29 reported are pre-existing (10 `harbor_runtime.py`, 9 `test_persistence.py`, 9 `test_agent_evaluate.py`, 1 `test_evaluator.py`), each confirmed against a stashed baseline
- `make vendor` — all four mirrored files byte-in-sync with source
- `make refresh-openapi` — `TrialError` schema generated; `make update-web-sdk` produces no committable diff (`web/packages/sdk/generated/` is gitignored)
- `uv run pre-commit run -a` — **2 hooks blocked locally, both environment-only and outside this diff**: `Helm Docs` (helm-docs not installed) and `Run uv lock with platform uv` (local uv 0.9.30 vs pinned 0.9.14). This change touches no Helm files and no `pyproject.toml`/`uv.lock`. Ruff, ty, copyright-header, merge-conflict and plugin-import hooks all passed, and no hook modified the tree.

## Summary by CodeRabbit

* **New Features**
  * Added structured trial error details, including type, message, traceback, and timestamp.
  * Added error summaries grouped by error type, affected trial IDs, and total counts.
  * Preserved error information in evaluation results, metrics, and saved artifacts.
  * Added API support for reporting and validating trial errors.

* **Bug Fixes**
  * Improved Harbor error handling, timeout reporting, and malformed error payload support.
  * Errored trials remain scoreable while clearly surfacing failure details.

* **Documentation**
  * Updated Harbor examples and commands for the current execution workflow.


packages/nemo_evaluator_sdk/examples/harbor/hello_world_dataset/injected-runtime-error/environment/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.22
+
+# bash is required for Harbor's docker environment
+RUN apk add --no-cache bash
```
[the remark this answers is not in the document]
ngoncharenko: Skipping — pre-existing pattern, not introduced here.

This Dockerfile is **byte-identical** to `hello_world_dataset/hello-world/environment/Dockerfile`, which is already on `main`:

```
FROM alpine:3.22
RUN apk add --no-cache bash
WORKDIR /app
```

Trivy DS-0002 applies equally to both; only the copy shows up because only the copy is in this diff. Adding `USER` to one of two identical fixtures would break the invariant the fixture exists to preserve — its verifier comment says it is kept matching hello-world so Harbor's discovery scan treats them the same.

It is also a local, throwaway Harbor task container in an `examples/` dataset whose entire job is to sleep past a 1s agent timeout so `exception_info` gets populated.

Hardening example Harbor task images is a reasonable repo-wide change, but it should cover both fixtures in its own PR rather than diverge them here.

packages/nemo_evaluator_sdk/tests/agent_eval/fixtures/harbor_error_dataset/injected-runtime-error/instruction.md
```diff
@@ -0,0 +1,6 @@
+Intentional Harbor failure fixture for exception propagation testing.
+
+Attempt to create hello.txt with "Hello, world!" as the content.
+
+The reference solution script deliberately sleeps past the 1s agent timeout so
+Harbor records exception_info.
```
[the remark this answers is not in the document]
ngoncharenko: Skipping — MD041 is not enforced in this repo, and the sibling fixture has the same shape.

`.markdownlint.json` exists but is wired into nothing:

```
$ rg -n "markdownlint" .pre-commit-config.yaml tools/lint/ Makefile .github/workflows/
(no matches)
```

Its contents only disable two rules (`MD013`, `MD034`); no hook, lint script, Make target, or workflow ever invokes markdownlint.

The pre-existing `hello-world/instruction.md` on `main` also opens with body text rather than an H1, so adding one here would make the two task fixtures inconsistent for no enforced benefit.

packages/nemo_evaluator_sdk/tests/agent_eval/fixtures/harbor_error_dataset/injected-runtime-error/tests/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Unreachable when the environment Dockerfile fails to build; kept so the task
+# shape matches hello-world for Harbor discovery.
+mkdir -p /logs/verifier
+
+if [ -f /app/hello.txt ] && [ "$(tr -d '\n' < /app/hello.txt)" = "Hello, world!" ]; then
```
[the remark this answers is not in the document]
ngoncharenko: Skipping the comparison change — this verifier never decides the task's outcome.

`solution/solve.sh` sleeps 10s against `[agent] timeout_sec = 1.0`, so Harbor records `exception_info` and the trial is `PARTIAL` before the verifier's reward matters. That is the whole point of the fixture: it exercises exception propagation, not reward correctness.

The file is also copied verbatim from `hello-world/tests/test.sh` (same `tr -d '\n'`), and it is kept that way deliberately so the task shape matches for Harbor's discovery scan. Tightening the copy but not the original would diverge them.

You did catch a real problem one line up, though — the comment named the wrong failure mode ("Unreachable when the environment Dockerfile fails to build"). Fixed in 8c1e92f17; it now describes the agent timeout.

packages/nemo_evaluator_sdk/examples/harbor/hello_world_dataset/injected-runtime-error/environment/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.22
```
ngoncharenko: additional task that triggers exception, so we could have this along the happy path

packages/nemo_evaluator_sdk/examples/harbor/run_harbor_example.py
```diff
@@ -44,9 +49,25 @@
...
 HELLO_WORLD_DATASET_DIR = Path(__file__).resolve().parent / "hello_world_dataset"
+HELLO_WORLD_TASK_NAME = "harbor/hello-world"
+INJECTED_ERROR_TASK_NAME = "harbor/injected-runtime-error"
+
 
+def _task_names(*, inject_error_task: bool) -> list[str] | None:
+    """Default to the healthy task; include the permanent error fixture when requested."""
+    if inject_error_task:
+        return [HELLO_WORLD_TASK_NAME, INJECTED_ERROR_TASK_NAME]
+    return [HELLO_WORLD_TASK_NAME]
```
SandyChapman: I'm wondering if we should be editing the harbor example with this rather than having this just as an integration test fixture. I'm not sure it'd be useful for users trying to onboard their harbor evals to purposely inject errors.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -110,6 +112,11 @@
 _DIGEST_SKIP_DIRS = frozenset({".git", "__pycache__", ".venv", ".uv", ".mypy_cache", ".pytest_cache"})
 _DIGEST_CHUNK_BYTES = 1 << 20
 
+# Cap on the traceback carried into a trial. Bundles are portable and a traceback is diagnostic
+# text, not data anyone joins on, so it is bounded rather than faithful. A real Harbor traceback is
+# ~1.2 KB, so this keeps whole ones while refusing a pathological payload.
```
SandyChapman: Can you take a pass to remove or reduce the verbosity on comments? I'm not sure we need this comment for example.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
```diff
@@ -57,8 +66,79 @@ class AgentOutput(BaseModel):
     )
 
 
+# Type recorded when a producer reported a failure but named no usable type. The SDK's own fallback,
+# not Harbor's - Harbor's `ExceptionInfo.exception_type` is required and always populated, so this
+# fires only for hand-built or malformed payloads.
```
SandyChapman: This file is not harbor specific, so we shouldn't really be calling out Harbor in the code comments specifically as it's just one case of... well... not _many_, but at least a few.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
```diff
@@ -57,8 +66,79 @@ class AgentOutput(BaseModel):
...
+# ``AgentEvalTrial.metadata`` under this key. Read on load so bundles written before the typed field
+# still roll up; no longer written. Kept indefinitely -- bundles on disk are permanent.
+LEGACY_ERROR_TYPE_METADATA_KEY = "exception_type"
+
+
+class TrialError(BaseModel):
+    """What went wrong producing one trial, as the producer reported it.
+
+    Present means the *producer* reported a failure. It does **not** imply ``status is FAILED``: an
+    errored Harbor trial is deliberately :attr:`AgentEvalTrialStatus.PARTIAL` so it is still scored.
```
SandyChapman: `...an errored Harbor trial...` Same thing here.

ngoncharenko: done

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/trials.py
ngoncharenko: done

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
ngoncharenko: done

packages/nemo_evaluator_sdk/examples/harbor/run_harbor_example.py
ngoncharenko: Great point, done - moved to the integration test

packages/nemo_evaluator_sdk/tests/agent_eval/test_harbor_runtime.py
```diff
@@ -1433,5 +1601,91 @@ def test_absent_exception_info_leaves_the_trial_completed(tmp_path: Path) -> Non
...
+    [fresh] = build_trials_from_job_dir(
+        job_dir, [AgentEvalTask(id="t", intent="x", inputs={"instruction": "p"}, metrics=[HarborRewardMetric()])]
+    )
+    legacy = AgentEvalTrial.model_validate(
+        {"id": "t__aaa", "task_id": "t", "status": "partial", "metadata": {"exception_type": "RuntimeError"}}
+    )
+
+    assert fresh.error is not None and fresh.error.type == "RuntimeError"
+    assert "exception_type" not in fresh.metadata
+    assert legacy.error is None  # free-form metadata is never promoted to a typed error
```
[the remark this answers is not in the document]
ngoncharenko: This is intentional. error is the only carrier; metadata["exception_type"] is not promoted, no backward compatibilty is needed at this point since no real consumers of this HarborRunner at this point

SandyChapman (approved):
lgtm! thanks for the adjustments

PR #1316, opened 2026-08-14
fix(evaluator): detect structured output support from the endpoint

SandyChapman:
## Summary

Structured output mode was selected from `Model.format`, but support is a property of the endpoint, not of the label on the model. `integrate.api.nvidia.com` is served under the `nim` label yet rejects `nvext.guided_json` (400) and silently ignores root `guided_json`, while honouring OpenAI `response_format`. Preflight probed only the two `guided_json` placements for `nim`, resolved to `UNSUPPORTED`, and fell back to an unenforced "return JSON" prompt instruction. Judges kept scoring; nothing surfaced the downgrade.

Detection now probes every endpoint with the same ordered candidate list, `response_format` first. `Model.format` is deprecated, ignored, and defaults to `openai`.

## Changes

- `detect_structured_output_mode` probes all endpoints identically. Probe payloads are built through `InferenceStructuredOutput`, so a probe sends exactly what production sends.
- `Model.format` marked `deprecated`, default `nim` → `openai`; internal reads removed. Plugin `openapi.yaml` regenerated.
- New `session.py`: `begin_evaluation_session()`, a re-entrant run-scoped cache. Run scoping is what makes caching a *failed* probe safe — a transient failure dies with the run instead of persisting for the process.
- Five generation paths built a structured-output hook and issued requests without probing (target generation, agent evaluation, the multi-metric benchmark path, ProfBench, direct `compute_scores()`). All resolve first, and a hook used unresolved warns once so a future path is visible rather than silent.
- Probe budget raised from 128 tokens; a truncated probe now warns. See below — this was a live-only bug.
- The default `max_tokens=4096` cap was NIM-only; flipping the default format would have silently removed it. Now unconditional.
- Docs, skill references, the fabric notebook and the example-spec generator no longer instruct users to set `format`.

## Type of Change

## Quality Gates

## Verification

Targeted validation (rebased onto `main`, re-run after the rebase):

| Command | Result |
|---|---|
| `pytest packages/nemo_evaluator_sdk/tests` | 1674 passed, 10 skipped |
| `pytest plugins/nemo-evaluator/tests` (unit) | 837 passed |
| `pytest plugins/nemo-optimization/tests plugins/nemo-customizer/tests` | 105 passed |
| `uv run ruff check` / `ruff format --check` | clean |
| `tools/lint/lint-python-types.sh` | exit 0 |
| `make vendor` + `make refresh-openapi` | no drift against the rebased base |

New behavioural tests are mutation-verified: each was confirmed to fail with its corresponding fix removed.

`pre-commit run -a` is **not** fully green and the box is left unchecked. Three hooks fail for missing local toolchain, none governing a file in this diff: `Helm Docs` (binary absent; no `k8s/helm` changes), `uv-lock` (pins uv 0.9.14, local 0.9.30; no `pyproject.toml`/`uv.lock` changes, and the separate lock-drift check passes), `studio-lint-staged` (no pnpm shim; no `web/` changes). `tools/lint/lint-openapi.sh` needs `mapfile` (bash 4) on a bash 3.2 host; its check was reproduced portably — no drift across the committed specs.

### Live verification

Run against `integrate.api.nvidia.com`. Which encodings each model actually honours:

| Model | `response_format` | root `guided_json` | `nvext` |
|---|---|---|---|
| `meta/llama-3.3-70b-instruct` | yes | yes | ignored |
| `meta/llama-3.1-8b-instruct` | yes | yes | ignored |
| `nvidia/nemotron-3-nano-30b-a3b` | yes | ignored | ignored |
| `nvidia/llama-3.3-nemotron-super-49b-v1.5` | yes | ignored | ignored |

No model honoured `nvext.guided_json`, which is the encoding the old code chose for every `nim`-labelled model.

End-to-end judge with `format="nim"` explicitly set, confirming the label is ignored:

```
nvidia/nemotron-3-nano-30b-a3b  ->  openai_response_format   scores [4.0, 0.0]
meta/llama-3.1-8b-instruct      ->  openai_response_format   scores [nan, nan]
```

**Live testing found a bug that the unit tests could not.** The probe capped output at 128 tokens. Reasoning models spend that budget thinking and return truncated or empty content, which is indistinguishable from an endpoint ignoring the encoding — so every reasoning model resolved to `UNSUPPORTED` and silently lost enforcement:

```
nemotron-3-nano-30b-a3b       max_tokens=128   finish=length  reasoning=610ch  content='We need to output only a JSON…'
nemotron-3-nano-30b-a3b       max_tokens=2048  finish=stop    reasoning=2072ch content='{"__nmp_probe_score": 0}'
llama-3.3-nemotron-super-49b  max_tokens=128   finish=length  reasoning=592ch  content=''
llama-3.3-nemotron-super-49b  max_tokens=2048  finish=stop    reasoning=8281ch content='{"__nmp_probe_score": 0}'
```

A mocked probe always answers instantly, so no hermetic test could have caught this.

### Reviewer notes

1. **`RangeScore` produces NaN on `meta/llama-3.1-8b-instruct`.** Not this change: `RangeScore` derives an `integer` schema with `minimum`/`maximum`, and that model's grammar backend degenerates on it — emitting `{` followed by an unbounded run of tabs until it hits the token cap. The same schema typed as `number` returns `{"helpfulness": 4}`. Worth deciding whether `RangeScore` should derive `number`. Incidentally, the unconditional `max_tokens` cap is what bounds that runaway.
2. **`nvidia/nemotron-3-super-120b-a12b` is listed by `/v1/models` but 404s on chat completions.** Listed is not served — an argument for probing over any static capability table.
3. **The endpoint was unstable during testing** (404s, 500s, 504s, per-model timeouts). `meta/llama-3.3-70b-instruct` timed out at every budget including 128, so the raised probe budget is not the cause.
4. **`max_tokens`** — openai-format models that never set it now get a 4096 cap. Deliberate; may warrant a release note.
5. **Cache key is `(url, name)`**, so a session is assumed single-principal. No first-party path mixes credentials within a run, and keying on the API key would put a secret in a cache key. Recorded as an assumption in code.

### Deferred

- `new_hooks()` still returns a hook pre-set to a provisional encoding, with probing as a separate step each call site must remember — that shape is why five paths were missed. Making the hook unusable until probed is the structural fix; the warning is the interim mitigation.
- `nemo-optimization` still maps a user-facing `provider` onto `ModelFormat`, now a dead path.
- Probes run before `use_resilience_session()` opens, so they use the global scheduler and are absent from the session summary. Pre-existing; the judge's own preflight has always behaved this way.

## Summary by CodeRabbit

* **New Features**
  * Structured-output support is detected directly from model endpoints during preflight.
  * Endpoint capabilities are cached during evaluations to reduce repeated probing.
  * Evaluation workflows now run preflight checks before scoring.

* **Bug Fixes**
  * Improved structured-output compatibility, including fallback handling.
  * Applied default token limits consistently across model types.
  * Prevented duplicate concurrent capability probes.

* **Documentation**
  * Updated examples to remove explicit model format configuration.
  * Marked the model format setting as deprecated and ignored.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
@@ -166,17 +166,19 @@ class AgentEvaluator:
         runtime_config = resolved_config.model_copy(update={"run_id": run_id})
         started_at = datetime.now(UTC)
 
-        if (trials is None) == (target is None):
-            raise ValueError("provide exactly one of trials or target")
+        seam_error = "provide exactly one of trials or target"
+        if trials is not None and target is not None:
+            raise ValueError(seam_error)
 
         async with begin_evaluation_session():
-            # Branch on which seam was supplied so the type checker can narrow ``target`` to a
-            # concrete ``AgentEvalTarget`` without a cast.
+            # Branch on which seam was supplied so the type checker narrows each of ``trials`` and
+            # ``target`` to a concrete type without a cast. The final arm is the "neither" case.
             if trials is not None:
                 trial_list = list(trials)
-            else:
-                assert target is not None
+            elif target is not None:
                 trial_list = await self._generate_trials(tasks=task_list, target=target, config=runtime_config)
+            else:
+                raise ValueError(seam_error)
             scores = await self._score_trials(
                 tasks=task_list,
                 trials=trial_list,
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
@@ -46,10 +46,15 @@ from nemo_evaluator_sdk.agent_inference import (
     make_agent_inference_fn,
     new_agent_inference_client,
 )
-from nemo_evaluator_sdk.execution.metric_execution import generate_online_sample, run_sync
+from nemo_evaluator_sdk.execution.metric_execution import (
+    generate_online_sample,
+    resolve_target_structured_output_mode,
+    run_sync,
+)
+from nemo_evaluator_sdk.structured_output import structured_output_mode_session
 from nemo_evaluator_sdk.execution.samples import build_metric_input
 from nemo_evaluator_sdk.inference import InferenceFn
-from nemo_evaluator_sdk.metrics.protocol import Metric, validate_metric_result
+from nemo_evaluator_sdk.metrics.protocol import Metric, MetricWithPreflight, validate_metric_result
 from nemo_evaluator_sdk.metrics.utils import metric_type_name
 from nemo_evaluator_sdk.values import (
     Agent,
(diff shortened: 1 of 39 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
```diff
@@ -161,22 +166,23 @@ async def run(
...
-            run_id=run_id,
-        )
+
+        async with begin_evaluation_session():
+            # Branch on which seam was supplied so the type checker can narrow ``target`` to a
+            # concrete ``AgentEvalTarget`` without a cast.
+            if trials is not None:
+                trial_list = list(trials)
+            else:
+                assert target is not None
```
JashG: [nit] Do we need this assert?

skills/nemo-evaluator-plugin/assets/specs/llm_as_judge.json
```diff
@@ -36,7 +36,7 @@
             "name": "nvidia/nemotron-3-super-120b-a12b",
             "host_url": null,
             "api_key_secret": "NVIDIA_API_KEY",
-            "format": "nim"
+            "format": "openai"
```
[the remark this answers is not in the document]
SandyChapman: The generator no longer sets `format` — this is Pydantic serialising the field's default, which changed with the flip to `openai`. Removing it means excluding deprecated fields at serialisation, which affects every `Model` consumer, so I've left it as a follow-up rather than special-casing this one asset.

JashG (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/evaluator.py
SandyChapman: No, I updated to remove it.

PR #1321, opened 2026-08-14
fix: patch PyTorch header for ARM64 DeepEP

svvarom:
## Summary

Backport PyTorch's upstream `List_inl.h` type-spelling fix so DeepEP can compile on ARM64 with CUDA 13.3 while keeping the existing PyTorch 2.11, vLLM 0.25.1, DeepEP, and NeMo-RL pins unchanged.

## Root cause

PyTorch 2.11 spells `List<T>::operator[]`'s iterator offset type through a dependent `decltype(impl_->list)::difference_type`. CUDA 13.3's nvcc rejects that expression while DeepEP builds on ARM64.

PyTorch corrected the header upstream in [c2671b8](https://github.com/pytorch/pytorch/commit/c2671b853e1553b258a30c15d66f9647faae7aee), but the fix is not present in PyTorch 2.11.

## Changes

- add the exact upstream `List_inl.h` correction as a small cherry-pick
- apply it after the base `uv sync` installs PyTorch and before the vLLM extra builds DeepEP
- use `git apply --check` so missing headers, changed context, or an already-applied patch fail loudly
- assert the corrected expression exists after applying the patch
- document removal when the RL stack moves to PyTorch 2.13 or later

No CUDA or dependency pins and no lockfiles change.

## Validation

- `git diff --check`
- applied the patch successfully against the exact PyTorch v2.11.0 `List_inl.h`
- `make docker-print TARGET=nmp-rl-base-builder`
- `make docker-print TARGET=nmp-rl-training`

The cache-free ARM64 image build and RL smoke test are intentionally left to CI rather than completed locally. This targets the failure in [Build Automodel Docker images](https://github.com/NVIDIA-NeMo/Platform-Deploy/actions/runs/31823327198/job/94841620610).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RL container builds by applying a targeted PyTorch compatibility patch during dependency setup.
  * Fixed iterator offset type handling in PyTorch list access to improve build and runtime compatibility.

### DIFF
--- a/docker/rl/Dockerfile.nmp-rl-base
+++ b/docker/rl/Dockerfile.nmp-rl-base
@@ -194,12 +194,22 @@ uv sync ${UV_SYNC_MODE} --no-install-project
 # https://github.com/pytorch/pytorch/commit/c2671b853e1553b258a30c15d66f9647faae7aee.
 # Remove this when the RL dependency stack moves to PyTorch 2.13 or later.
 torch_include=/opt/nemo_rl_venv/lib/python3.13/site-packages/torch/include
-test -f "${torch_include}/ATen/core/List_inl.h"
+list_inl="${torch_include}/ATen/core/List_inl.h"
+test -f "${list_inl}"
+# UV_LINK_MODE=symlink above leaves every venv file a symlink into UV_CACHE_DIR, and `git apply`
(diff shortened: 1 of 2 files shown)


ironcommit (approved):

ironcommit:
merged already in #1332

PR #1353, opened 2026-08-17
feat(studio): toggle guardrail rails explicitly, starting with self check (ASTD-339)

aray12:
## Summary

The Configuration tab was read-only: it rendered a guardrail config across six panels but the only editable fields were two textareas. This makes rails editable by modelling each one explicitly — a rail is a named object you switch on and off, and one switch performs every edit the engine requires. Self check is implemented as the first rail; the structure is what the remaining rails plug into.

It also makes the tab's companion — Test and Validate — actually work. Running a guardrail's tests failed for any config without a `models[]` entry, so every config authored in Studio was born unable to run its own tests. The tab now owns a **Main Model** field, and new configs are seeded with a resolved workspace default.

https://github.com/user-attachments/assets/d43c650f-5b26-403f-9fc1-1d66e636f76d

## Changes

- Add `routes/guardrails/rails/`: a `RailDefinition` owns `isScopeEnabled` / `setEnabled` / `hasStoredSettings` / `clearSettings`, all pure over `RailsConfig`.
- Add `configOps.ts`, the immutable flow and prompt primitives every rail shares.
- Implement the self-check rail. One switch writes the `self check input` / `self check output` flows *and* the `self_check_input` / `self_check_output` prompts together, seeded from the platform's own defaults in `app/seeding.py`. The rail adds no `models[]` entry of its own and has no model picker in its settings panel, because self check runs on whichever model is already serving the request. (Distinct from the config-level Main Model added below: that names the model the *whole config's tests* run against, not a task LLM bound to one rail.)
- **The stage badges carry state.** `Input` / `Output` are solid green when that stage is running and solid gray when it is not, matching the on/off treatment in `DataPrivacyPanel`. Both stages stay listed either way, so a rail running on input only is distinguishable from one running on both without opening the settings panel — that difference is what the rail *is*. The visible text is the stage name in both states, so the state also goes in `aria-label` rather than living in colour alone.
- **Settings are not standardised.** `RailDefinition.renderSettings` is an optional render prop: a rail owns its trigger *and* whatever that trigger opens — a modal side panel, a popover, an inline expander, or nothing at all. `RailsList` renders the shared left column (switch, name, stage badges) and knows nothing about panels. Self check supplies `SelfCheckSettings`, which is modal because every Studio side panel carrying a footer is modal; a non-modal one sits in normal page stacking, where page-level fixed elements float over its actions.
- The panel's confirm button is **Apply**, not Save — it only updates the working copy; nothing reaches the server until the page's Save Guardrail.
- The form holds the guardrail document and nothing else. The instruction fields write through the same `config` controller the rails use, so there is one representation of `instructions` / `sample_conversation` rather than a flat copy the save path had to overwrite.
- Replace the read-only Overview / Pipeline / Detectors / Models & prompting / Behavior panels with a single read-only **Configuration JSON** view built on the shared `CodeEditor`, so the fields Studio cannot configure yet stay visible and provably preserved.
- Add Storybook stories, so the surface can be reviewed without a running platform.

Behaviour worth calling out: switching a rail **off removes its flows but keeps its prompts**, so a tuned prompt survives an off/on round trip. That leaves an "off but holding settings" state, which is when the row offers to discard them.

## Model selection for guardrail tests

Added in `319ba31f` and `9d890ac8`, after the review rounds below.

Clicking **Run N Tests** on `self-check` failed with `Unknown error`. Two independent defects, one masking the other:

- `resolveConfigModel` threw before any request was made — no `POST /checks` ever fired — because the config had no `models[]` entry. Studio offered no way to add one and `CreateGuardrailModal` created configs without one, so this was the default state of anything authored in Studio.
- `getErrorMessage` ended with `fallbackMessage ?? error.message`, so a plain `Error` always reported the call site's generic string and discarded its own. The precise message never reached the toast.

What changed:

- **Main Model field on the Configuration tab**, above General Instructions, using the shared `WorkspaceModelSelect`. It writes through the same `config` controller as every other field, so the document stays the form's only representation.
- **New configs are seeded** with a resolved default: prefer a named default, else the first model that has a `model_provider`, else nothing. Following `DataDesignerJobBuildRoute/models.ts`, a provider-less model is never seeded — auto-fill happens without the user asking, and seeding an undeployed model hands them a config whose only symptom is a failed test run. An empty required field beats a name that fails later. The duplicate path is untouched; it inherits the source's models.
- **`resolveConfigModel` no longer falls back to any model that declares a reference.** For the common shape — a config declaring only task LLMs — that fallback selected `system/nemoguard-8b-content-safety` as the *generation* model, producing a run that looks fine and means nothing. Only the `main` entry is eligible now.
- **Run is disabled with an explanation** when the targeted config has no main model, instead of failing once per check. It re-evaluates per run target, so a Draft that has one is runnable while the Saved config still isn't.
- `getErrorMessage` now prefers the error's own message. This is repo-wide in effect — all 20 call sites are toast or inline error displays where the real message is strictly better, and it is its own commit so it can be reverted alone.

**Where the model lives, and why the code says so at length.** The service treats a missing `main` entry as normal and always routes on `request.model` (`nemo_guardrails_plugin/rails.py:236` — *"model name: always `request_body['model']`"*), injecting a placeholder when a `main` entry omits a name (`llmrails_cache.py:185`). The name stored on the config is therefore how **Studio** remembers the user's choice, not how the service routes. `resolveConfigModel` carries a comment saying this explicitly, so nobody later deletes the field on the grounds that the service ignores it. This was a deliberate call — a run-time picker on the Test tab was the alternative.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs cover this tab; intent is captured in module-level comments and the new stories.

## Verification

Targeted validation, re-run at `9d890ac8`:

| Command | Result |
| --- | --- |
| `uv run pre-commit run -a` (via commit hooks) | Passed — copyright headers, UI lint-staged, DCO, merge-conflict checks |
| `pnpm --filter nemo-studio-ui typecheck` | Passed |
| `pnpm --filter @nemo/common typecheck` | Passed |
| `pnpm --filter nemo-studio-ui lint` | Passed (0 warnings) |
| `pnpm --filter nemo-studio-ui test` | Passed — 317 files, 2928 tests |
| `pnpm --filter @nemo/common test` | Passed — 120 files, 1496 tests |
| `pnpm --filter nemo-studio-ui build` | Passed at `6bf894a`; **not re-run** for the two model-selection commits |

New tests: `rails/configOps.test.ts` (flow/prompt primitives), `rails/selfCheck/selfCheck.test.ts` (the rail contract, including that a flow is never enabled without the prompt its action renders, and that `isScopeEnabled` answers per stage), `GuardrailConfigurationPanel.test.tsx` (switch, badge state, discard, panel apply/cancel, variable insertion, and the Main Model field's read/write), `GuardrailConfigTab/index.test.tsx` (tab shape; unsupported rails survive in the JSON), `GuardrailConfigTab/mainModel.test.ts` (the `main`-entry helpers, including that task LLMs survive a write and that clearing collapses the key), `routes/guardrails/defaultModel.test.ts` (default resolution, including that provider-less models are skipped), plus new cases in `CreateGuardrailModal/index.test.tsx` (seeding, the no-models path, and that a duplicate does not query the catalogue) and `GuardrailChecksTab/index.test.tsx` (the Run gate on both targets).

Six existing tests were updated rather than worked around:

- The Configuration tab's instruction field is now labelled **General Instructions** (title case, per the design and `web/.claude/skills/ux-guidelines`), so the detail-route and checks-tab tests that dirty the form through it were updated to match.
- `resolveConfigModel`'s `falls back to the first model that declares a reference` asserted the behaviour this PR removes; it is now a `throws` case under `only task LLMs`.
- Two assertions on the old `has no usable model to run checks against` string were updated to the new message.
- `getErrorMessage`'s `returns fallback message when provided` encoded the precedence being inverted; replaced by one case proving the message wins and one proving the fallback still applies to a message-less error.

**Manually verified** in a running Studio against a live backend: the rail switch writes flow + prompt together, the stage badges track which stages are actually running (including the input-only mixed state), the modal opens clear of page-level floating UI, Apply and Cancel behave, the variable chip inserts at the caret, the JSON panel tracks unsaved edits, and the draft actions stay `aria-hidden` on a pristine load.

For the model selection, end-to-end on a real model-less config: Run disabled carrying `title="Set a main model on the Configuration tab to run tests"` → picked a model → saved → `GET .../configs/self-check` returns `[{type: "main", engine: "nim", mode: "chat", model: "default/abacusai-dracarys-llama-3-1-70b-instruct"}]` → Run enabled → fired → persisted a RunRecord at `status: success`, `config_version: 3`. Dirtying the form then running against **Draft** persisted `is_draft: true` with `config_version: null`.

### Review follow-ups

CodeRabbit's two actionable findings are addressed in `4d183c5`:

- **Configuration JSON did not scroll.** Confirmed against a running Studio — with only a `max-height`, CodeMirror sized itself to the whole document (3278px for a real config) and spilled out of the 560px panel. Now a definite height, verified scrolling (1628px of content in a 394px viewport).
- **`hasStoredSettings` contract.** The implementation reports stored settings regardless of whether the rail is running, which contradicted its doc comment. Corrected the contract rather than the code: it stays a pure predicate, and `RailsList` owns the policy of offering discard only for a switched-off rail, since `clearSettings` would otherwise stop a live one.
- Nitpick applied: `scopes` and `variables` are `readonly`.

Stakeholder review is addressed in `6bf894a`:

- **The stage badges communicated nothing.** They rendered `rail.scopes` unconditionally, so they described capability while sitting exactly where state belongs — and a rail enabled for input only looked identical to one enabled for both. Badges now carry per-stage state, so the thing that defines what a rail does is legible from the row.
- To support that, `RailDefinition` trades rail-level `isEnabled` for per-scope `isScopeEnabled`, and `RailsList` derives the rail-level answer as `scopes.some(isScopeEnabled)`. Deriving it is the point: two independently declared predicates could disagree, and the switch and the badges would then contradict each other.

### Known limitations

- **A master off/on round trip re-enables every stage.** `setEnabled(data, true)` turns all stages on, so a user who had input only and toggles the rail off and back on gets both. Prompts survive; the flow selection does not. An absent flow leaves no trace in the config, so restoring the prior selection would mean inventing state to store it in — accepted as a tradeoff rather than carried. It was invisible before this change and is visible now that the badges track state.
- **Toggling a stage off leaves an empty `rails.<stage>: { flows: [] }` behind.** Deliberate — the stage object can carry `parallel` / `streaming` settings no rail owns, and dropping it would discard them — and it is inert to the engine, but it does add noise to the JSON. Pruning a stage that empties *and* holds nothing but `flows` is a candidate follow-up.
- **`scopes` now does double duty**: the ordered stage list *and* the iteration source for state lookup. Fine for self check; a rail whose stages are not uniform (content safety binds a task model per stage) may want richer entries than a bare `RailScope[]`.
- **Configs created before this change stay unrunnable until re-saved.** They have no `main` entry, so Run stays disabled with its explanation. Deliberately not auto-prefilled on load: a config that marks itself unsaved the moment you open it is worse than an empty required field. No migration is included.
- **The Main Model field sits on the Configuration tab but is scoped to testing.** Its help text says so, because at inference time IGW routes on the request's model rather than this field. If that reads as oddly hedged for a field at the top of the tab, moving it beside the Draft/Saved toggle on Test and Validate is a small change — the helpers and the resolver are both independent of where the field lives.
- **One test-coverage gap, stated rather than papered over.** The Run gate's "a Draft that gains a main model the Saved config lacks" direction has no unit test: it needs the real `WorkspaceModelSelect` driven against a live catalogue, which the route-level test does not stand up. The shipped test covers both static verdicts plus a dirty draft that keeps its model; the gaining direction was verified manually (above).
- **Seeding a brand-new config was verified by unit test only**, not by creating one against the live backend.

Unrelated observation while working here: studio's `vitest.setup.tsx` MSW `beforeAll` hook exceeds the default 10s `hookTimeout` locally for any test file with a heavy import graph — reproducible on untouched files, e.g. `components/ValueWithLabel/index.test.tsx`, and on anything importing `@studio/constants/constants`. Not touched in this PR; CI is unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added a streamlined guardrail configuration panel with enablement controls and scope-specific settings.
  * Added Self Checks for input and output protection, including editable prompts and template variables.
  * Added Apply, Cancel, and discard options for managing unsaved changes and stored settings.
  * Added an always-visible, read-only JSON view of the complete configuration.
  * Added automatic default model selection for newly created guardrails.
* **Improvements**
  * Preserves unsupported or unexposed configuration fields when editing and saving guardrails.
  * Guardrail test runs now require a configured main model and explain when runs are unavailable.
  * Error messages now prioritize useful details over fallback text.


web/packages/studio/src/routes/guardrails/rails/selfCheck/prompts.ts
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Default self-check prompts, copied from the platform's seeded `self-check` config
+ * (`services/guardrails/app/seeding.py`) so a guardrail built in Studio starts from the
+ * same policy as one built with the CLI.
```
walston, line 7: It's worth adding a note to the python that you're duplicating the flow in typescript in case someone updates it.

walston (approved):

PR #1354, opened 2026-08-17
fix(cli): don't print unqualified Setup complete when no default model was set

yamini:
## Summary

`nemo setup --auto` already detects when model discovery finds nothing and prints a warning (No default model set), but the function still fell through to an unconditional green "Setup complete!" banner if the platform health check passed. A user running setup against a working API key but the wrong inference endpoint (or any other path that leaves the model registry empty) saw a clean success message despite the platform being non-functional for inference. This gates the final banner on whether a default model was actually set.

## Related Issue

Linear: [NMP-14](https://linear.app/nvidia/issue/NMP-14/nemo-setup-auto-silently-continues-after-api-key-validation-failure)

## Changes

- `_run_auto_mode` now shows the existing green "Setup complete!" message only when a default model was set, otherwise it shows "Setup complete with warnings" and points back at the warnings already printed above (`sdk/python/nemo-platform/src/nemo_platform/cli/commands/setup.py`, `packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py`; these are two separately-maintained copies of the same function, both fixed identically)
- Added a test confirming the clean success message still shows when a default model is set
- Extended the existing "no default model" test to assert the warning banner shows and the plain success message does not

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: CLI output message change only, no documented behavior contract to update

## Verification

- [ ] `uv run pre-commit run -a` — ran pre-commit scoped to the 4 changed files instead of the full repo; ruff, ty, copyright header, and merge-conflict checks all passed
- [x] Targeted tests pass

Targeted validation:

- `uv run pytest sdk/python/nemo-platform/tests/vendored/nemo_platform_ext/cli/commands/test_setup.py -q` → 233 passed, 3 failed (pre-existing on `main`, unrelated `TestMaybeStartServices` tests, confirmed by running the same command against a clean checkout of `main`)
- `uv run pytest packages/nemo_platform_ext/tests/cli/commands/test_setup.py -q` → same result, same 3 pre-existing failures
- Manually exercised `_run_auto_mode` with no models discovered against the real (non-mocked) console output: confirmed it now prints "Setup complete with warnings" instead of the plain green success message
- I intentionally left `_validate_api_key`'s fail-open behavior on ambiguous status codes untouched; its docstring documents that as a deliberate design choice (avoid blocking setup when a provider is unreachable), and this PR only fixes the final banner not reflecting a degraded state the function already knew about

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto setup now verifies platform health before reporting completion.
  * Failed health checks stop setup and prevent a misleading success message.
  * Setup reports warnings when no default model is configured, and confirms success when one is available.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
@@ -2356,10 +2356,14 @@ def _run_auto_mode(
         headers=_platform_request_headers(cli_context),
     )
 
-    if _verify_platform_health(base_url):
+    if not _verify_platform_health(base_url):
+        raise typer.Exit(1)
+
+    if default_model:
         console.print(f"\n{CHECK} [green]Setup complete![/green]")
     else:
-        raise typer.Exit(1)
+        console.print(f"\n{WARN} [yellow]Setup complete with warnings.[/yellow]")
+        console.print("  No default model was set; review the warnings above before using the platform.")
 
 
 def _run_interactive_mode(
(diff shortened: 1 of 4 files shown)


anastasia-nesterenko (approved):
👏

PR #1364, opened 2026-08-18
feat(studio): add raw/md/json format toggle to span payloads

rrhyne:
## What

Span Input/Output payloads rendered one fixed way, which made JSON payloads hard to scan and markdown payloads render as escaped text.

This replaces `SpanPayloadBlock` with `SpanPayloadView`, which renders a payload in one of three formats — `raw` (verbatim), `md` (rendered markdown), or `json` (pretty-printed and syntax-highlighted) — plus a `SpanPayloadFormatToggle` that sits on the section trigger.

The two share state through `useSpanPayloadFormat`, called in `SpanMetadataAccordions` because the toggle renders in `slotEnd` while the payload renders in `slotContent`.

## Behavior

- Payloads open in `json` when they parse as JSON and `raw` otherwise, so the common case needs no click.
- The toggle hides itself when the span has no payload, and disables `json` (with an explanatory tooltip) for payloads that are not JSON.
- Selecting a format on a collapsed section also opens it.
- The choice is scoped to the payload it was made for, so selecting another span re-derives the default instead of inheriting the previous span's view.
- The trigger is a `<summary>`, so each button suppresses the row toggle.
- Payloads at or above 20,000 characters paint a spinner for one frame before mounting the renderer, and skip Shiki highlighting so the full text always appears.
- Kind-specific payloads (e.g. the retriever query) use `SpanPayloadView` without a toggle and take the same default.

## Note on the `key` workaround

`CodeSnippetCode` in `@nvidia/foundations-react-core` stores the Shiki-highlighted HTML in state and never clears it. When `language` goes `json` → `text`, `skipHighlight` becomes true and the highlight effect early-returns, so `codeToRender` still holds the previous highlighted markup and renders it via `dangerouslySetInnerHTML` instead of the new raw `value` — selecting `raw` from `json` left the pretty-printed JSON on screen.

Worked around by keying `CodeSnippet` on the language it is given, so it remounts with clean state on every format change. Keying on `language` rather than the format also covers the case where a large JSON payload falls back to `text` and would otherwise inherit a previous payload's markup. Worth fixing upstream — `CodeSnippetCode` should reset `highlightedCode` when `skipHighlight` flips true.

## Testing

- 13 unit tests in `SpanPayloadView.test.tsx`, including a regression test for the stale-markup bug above (verified it fails without the `key` and passes with it).
- Full `IntakeDetail` suite: 43 tests pass.
- `tsc --noEmit` and `eslint` clean.
- Verified manually in Studio against a live LLM span: `raw` renders the single-line payload, `json` pretty-prints it, `md` renders markdown, and all transitions between the three round-trip correctly.

## Summary by CodeRabbit

* **New Features**
  * Added Raw, Markdown, and JSON viewing options for span input, output, and retriever query payloads.
  * Added automatic format detection, JSON fallback handling, empty-state messaging, and deferred rendering for large payloads.
  * Format selections open the relevant section and persist for each payload.
  * Improved code highlighting when switching formats.
  * Preserved numeric formatting, including large integers, negative zero, and duplicate keys, in JSON views.

* **Documentation**
  * Documented payload format behavior, defaults, and rendering details.


web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,276 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// One module owns everything about viewing a span payload: the body, the format
+// toggle, and the format state they share. The toggle renders on an accordion
+// trigger while the body renders in that item's content slot, so the state has
+// to be lifted to their common parent via the hook — which means this file
+// exports a hook alongside its components and waives fast refresh's
+// component-only rule.
+/* eslint-disable react-refresh/only-export-components */
```
marcusds: Rather than this just move the other exports out to their own files: consts, utils, or types.ts.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,276 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// One module owns everything about viewing a span payload: the body, the format
```
marcusds: This file has a lot more comments than general that we normally have. We usually let the code try to speak for itself and only comment where needed.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,276 @@
...
+  payloadLabel: string;
+}
+
+/**
+ * Compact `raw | md | json` switch for a payload section header. Renders nothing
+ * when there is no payload, and disables JSON for payloads that are not JSON.
+ * The header may be a `<summary>` (the accordion row trigger), so each click
+ * stops propagation and the default toggle.
+ */
+export const SpanPayloadFormatToggle: FC<SpanPayloadFormatToggleProps> = ({
```
marcusds: I would split SpanPayloadFormatToggle out into its own file.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,276 @@
...
+ */
+const parseJsonPayload = (value: string | null | undefined): string | null => {
+  const trimmed = value?.trim();
+  // Cheap gate before JSON.parse: only objects and arrays are worth a JSON view,
+  // and it keeps every plain-text payload away from the parser entirely.
+  if (!trimmed || !(trimmed.startsWith('{') || trimmed.startsWith('['))) {
+    return null;
+  }
+  try {
+    return JSON.stringify(JSON.parse(trimmed), null, 2);
```
marcusds: This isn't lossless, is that OK?

marcusds (changes requested):

marcusds:
Can you also add a screenshot to the description?

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+// One module owns everything about viewing a span payload: the body, the format
+// toggle, and the format state they share. The toggle renders on an accordion
+// trigger while the body renders in that item's content slot, so the state has
+// to be lifted to their common parent via the hook — which means this file
+// exports a hook alongside its components and waives fast refresh's
+// component-only rule.
+/* eslint-disable react-refresh/only-export-components */
+
+import { MarkdownContent } from '@nemo/common/src/components/MarkdownContent';
```
marcusds: I'd consider moving the small section that uses MD into its own component with this import and than lazy loading it since this import alone adds 100KB

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Trimmed in 96776fb9. Comment lines went from 76/276 to 46/242, which lines up with the neighbors (`SpanFeedbackControls.tsx` is 14/122, `spanKeyValues.tsx` 21/441).

Kept the four bits the agent didn't think are recoverable from reading it:

- the `key={language}` remount on `CodeSnippet`, which is there because it keeps highlighted markup in state and never clears it
- why the render deferral is decided during render instead of in an effect
- why the disabled `json` button needs a focusable wrapper for its tooltip
- why format selection is keyed by payload text

Happy to cut further if any of those still read as noise.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Done in 96776fb9 — the `eslint-disable` is gone entirely. Split it three ways, following what's already in this directory:

- **`spanPayloadFormat.ts`** — types and pure helpers (`SpanPayloadFormat`, `SpanPayloadFormatState`, `parseJsonPayload`, `autoFormat`), same shape as `keyValueTypes.ts` / `keyValueFormatting.ts`
- **`useSpanPayloadFormat.ts`** — the hook, same as `useSpanAnnotationActions.ts`
- **`SpanPayloadView.tsx`** — components only, so `react-refresh/only-export-components` is satisfied on its own

I left `LARGE_PAYLOAD_RENDER_DEFER_CHAR_LIMIT` and `FORMAT_OPTIONS` in the view module: both are private with a single use site each, so they never tripped the rule and they read better next to the component using them. Say the word if you'd rather they move too.

Consumers updated in `SpanMetadataAccordions.tsx` and the test; `RetrieverSpanContent.tsx` only imports the component, so it was untouched. Pure code movement — lint, typecheck, and all 47 `IntakeDetail` tests pass.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Good catch — it wasn't, and worse than I'd assumed. Fixed in 111e5ffd.

`JSON.stringify(JSON.parse(text), null, 2)` routes every number through a float64, so:

| payload | rendered as | |
|---|---|---|
| `{"span_id":9007199254740993}` | `9007199254740992` | silently wrong |
| `{"trace_id":1234567890123456789}` | `1234567890123456800` | silently wrong |
| `{"score":1.0}` | `1` | cosmetic |
| `{"temp":1e2}` | `100` | cosmetic |
| `{"delta":-0}` | `0` | cosmetic |
| `{"a":1,"a":2}` | `{"a":2}` | key dropped |

The first two are the ones that bothered me: an int64 span or trace id inside a payload renders as a different id with nothing signalling it. And since `autoFormat` makes `json` the default for anything that parses, readers land on that view without choosing it — "raw is one click away" isn't much of a defence when raw isn't where you start.

Now parsed with a reviver that re-emits each number as the literal it was parsed from, using `JSON.rawJSON` and the reviver's `context.source` (TC39 JSON source text access). Everything above round-trips exactly except duplicate keys.

No feature detection needed: engines without the proposal pass no `context`, so the reviver falls through to the parsed value and the view degrades to precisely the old behaviour. Preserved path is Chrome/Edge 114, Firefox 135, Safari 18.4. TypeScript 6.0.3 declares neither API, so both are typed locally — worth knowing if this pattern spreads.

Duplicate keys still collapse to the last value; no reviver can undo that, since the object is built before any reviver runs. Noted in the doc comment, with `raw` as the exact view. If you'd rather it be lossless including duplicates, the alternative is to skip the round trip and re-indent the source text directly — happy to go there, it's just a fair bit more code for a case I've not seen in a real payload.

Added a test asserting the int64 and `1.0` cases.

rrhyne: Switched to re-indenting the source text with `jsonc-parser` instead — d787022 supersedes the reviver in 111e5ffd.

- Fully lossless now — duplicate keys included, which the reviver couldn't fix. `{"a":1,"a":2}` renders both entries.
- Browser-support question is gone. No Firefox 135 / Safari 18.4 floor, no feature degradation path to reason about.
- The TypeScript hack is gone. No local declarations for undeclared `JSON.rawJSON`, no cast on the reviver.

Implementation notes:

- `parseJsonPayload` now re-indents the source via jsonc-parser's `format()` and never converts values.
- We can't use jsonc-parser's `applyEdits` — it rebuilds the document per edit, so it's quadratic (6s on a 585KB payload). Replaced with an 11-line single-pass apply: 3.4ms, byte-identical output.
- `JSON.parse` stays purely as a validity gate, result discarded — `format()` happily formats trailing commas and comments, so something strict has to reject those.
- Added `isJsonPayload` so `useSpanPayloadFormat` doesn't format an entire payload just to answer a boolean for the toggle. That mattered more once formatting got its own cost.
- New dep: `jsonc-parser@3.3.1`, zero transitive deps, 6.5KB min+gzip. Worth flagging given your 100KB concern on `MarkdownContent` — it's small, but it's not nothing, and it's on the main chunk.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Done in 8783785 — toggle is in its own file now, with the tests left together since both describes share the same PayloadSection harness.

rrhyne:
> Can you also add a screenshot to the description?

Oh the irony.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Done!

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,124 @@
...
+// react-markdown and its remark/micromark chain are ~100KB, for a format most
+// readers never select. Load them when the markdown view is first rendered.
+const MarkdownContent = lazy(() =>
+  import('@nemo/common/src/components/MarkdownContent').then((module) => ({
+    default: module.MarkdownContent,
+  }))
+);
+
+/** Shown while a payload is too large to render immediately, or markdown is loading. */
+const PayloadPending: FC = () => (
```
marcusds: Similar as the other example, I like to keep one component per file.

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
```diff
@@ -0,0 +1,124 @@
...
+  parseJsonPayload,
+  type SpanPayloadFormat,
+} from '@studio/components/IntakeDetail/IntakeComponents/spanPayloadFormat';
+import { type FC, lazy, Suspense, useEffect, useMemo, useState } from 'react';
+
+const LARGE_PAYLOAD_RENDER_DEFER_CHAR_LIMIT = 20_000;
+
+// react-markdown and its remark/micromark chain are ~100KB, for a format most
+// readers never select. Load them when the markdown view is first rendered.
+const MarkdownContent = lazy(() =>
```
marcusds, line 16: Image

web/packages/studio/src/components/IntakeDetail/SpanMetadataAccordions.tsx
```diff
@@ -56,24 +63,34 @@ const UsageSection: FC<SpanSectionContext> = ({ usageEntries }) => (
   </Stack>
 );
 
-const InputSection: FC<SpanSectionContext> = ({ span }) => (
+const InputSection: FC<SpanSectionContext> = ({ span, inputFormat }) => (
```
marcusds, line 66: Same as above, I think having one component per file is much easier for readability and usability.

I like to put them in a ./components/ sub folder for these small components.

web/packages/studio/package.json
```diff
@@ -67,6 +67,7 @@
     "classnames": "catalog:",
     "handlebars": "catalog:",
     "hyparquet": "catalog:",
+    "jsonc-parser": "3.3.1",
```
marcusds: Adding new packages requires a special process, I'll reach out to you on Slack about this.

marcusds (approved):

web/packages/studio/src/components/IntakeDetail/IntakeComponents/SpanPayloadView.tsx
rrhyne: Resolved using a different method to be lossless other than duplicate keys.

PR #1372, opened 2026-08-18
fix(docker): remove vendored libsndfile codecs

ironcommit:
## Summary

The `automodel` and `rl` codec cleanup lists (`codec-file-removals.txt`) previously pinned exact `libsndfile_*.so` paths, including a content-addressed uv cache hash segment that changes across builds. `remove-listed-files.sh` only supported exact paths, so any drift in the cache hash or arch-specific filename silently broke cleanup. This change makes the removal script glob-aware and updates both file lists to use globs instead of brittle exact paths.

## Changes

- `docker/scripts/remove-listed-files.sh`: expand each listed path with `compgen -G` and remove every match instead of a single exact path; still fails if no absolute path is given.
- `docker/automodel/codec-file-removals.txt` and `docker/rl/codec-file-removals.txt`: replace the pinned `libsndfile_<arch>.so` / uv cache hash paths with glob patterns (`libsndfile_*.so`, `archive-v0/*/...`).
- Added `tests/unit/test_remove_listed_files.py` covering glob expansion, non-matching patterns, and rejection of relative paths.
- Added smoke tests (`test_soundfile_libsndfile_removed` in `tests/smoke_gpu/test_customizer_automodel.py`, `test_uv_cache_soundfile_libsndfile_removed` in `tests/smoke_gpu/test_rl_training.py`) that assert no `libsndfile_*.so` files remain in the built images.

## Type of Change

## Quality Gates

## Verification

Targeted validation:

- `uv run pytest tests/unit/test_remove_listed_files.py -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of bundled `libsndfile` libraries across Python environments, virtual environments, and archive directories.
  * Cleanup now handles all matching library versions and architectures instead of a single fixed file.

* **Tests**
  * Added checks to verify that unwanted `libsndfile` libraries are removed from supported installation and training environments.

### DIFF
--- a/docker/automodel/codec-file-removals.txt
+++ b/docker/automodel/codec-file-removals.txt
@@ -3,7 +3,7 @@
 #
 # Codec cleanup for release image scans. Remove this file and its
 # Dockerfile call when the upstream image contents no longer need this workaround.
-/opt/venv/lib/python3.12/site-packages/_soundfile_data/libsndfile_arm64.so
+/opt/venv/lib/python3.*/site-packages/_soundfile_data/libsndfile_*.so
 /usr/lib/aarch64-linux-gnu/libFLAC.so.12
 /usr/lib/aarch64-linux-gnu/libmp3lame.so.0
 /usr/lib/aarch64-linux-gnu/libogg.so.0
--- a/docker/rl/codec-file-removals.txt
+++ b/docker/rl/codec-file-removals.txt
@@ -3,4 +3,6 @@
 #
 # Codec cleanup for release image scans. Remove this file and its
 # Dockerfile call when the upstream image contents no longer need this workaround.
-/opt/uv_cache/archive-v0/TX3VFTh1tOX_1IKC/_soundfile_data/libsndfile_x86_64.so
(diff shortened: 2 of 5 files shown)


svvarom (approved):

tests/unit/test_remove_listed_files.py
```diff
@@ -0,0 +1,60 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import subprocess
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+SCRIPT = REPO_ROOT / "docker/scripts/remove-listed-files.sh"
```
mckornfield: wonder how useful this test is

mckornfield (approved):

tests/unit/test_remove_listed_files.py
ironcommit: removed

PR #1333, opened 2026-08-14
fix(agents): stamp ATIF trajectories with the deployment name

marcusds:
## Summary

The Fabric translator falls back to the agent config's own `name` when `telemetry.atif.agent_name` is unset:

```python
atif.setdefault("agent_name", config.name)   # fabric/translator.py:177
```

Every deployment built from a given config therefore emits the same `gen_ai.agent.name`, and no query keyed on the agent entity can match its own traces.

Measured on a live instance — agent entity `email-security-triage-6p05fw`, config name `email-security-triage`:

```sql
SELECT attributes_string['gen_ai.agent.name'] AS agent_name, count() AS spans, uniq(trace_id) AS traces
FROM intake.spans WHERE is_deleted = 0 GROUP BY agent_name

┌─agent_name────────────┬─spans─┬─traces─┐
│ email-phishing-agent  │  2514 │     23 │
│ email-security-triage │    36 │      2 │
└───────────────────────┴───────┴────────┘
```

Zero spans under the deployment name. Deployment now stamps it.

## Related Issue

Supports ASTD-424 (agent trace aggregation), which needs to select traces by agent entity. Follows #1322 (Studio-side workspace binding) and #1327 (deploy-time workspace binding).

## Changes

- New `bind_atif_agent_name(config, agent_name)` in `nemo_agents_plugin/utils.py`.
- `_NemoAgentsSpecConfigHandler.resolve_for_deployment` calls it. That method already received `agent_name` and explicitly discarded it (`del agent_name`); it now binds it. The translator's `setdefault` leaves an injected value alone.
- No-ops when `telemetry` or `telemetry.atif` is absent or not a mapping, so configs without ATIF are untouched.

### Behavioural notes

- Two deployments of the same config are now **distinguishable** in telemetry, where previously they collided. That is the point, but it is a visible change for anyone grouping by `gen_ai.agent.name`.
- Traces ingested before this change keep the old config name. Nothing migrates; expect a mixed window.
- `config.name` itself is untouched — only the ATIF telemetry block is stamped.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
[...]

PR #1374, opened 2026-08-18
feat(intake): agent identity and time-bucketed trace metrics [ASTD-424]

marcusds:
## Summary

ASTD-424 needs the agent entity page to render trace-derived KPIs and charts for **one specific agent**. That requires two things this PR delivers:

1. **Identity** — an agent entity's traces must be selectable. Agent id and name now flow from deploy → ATIF → `gen_ai.agent.id` → a real `trace_index` column.
2. **Aggregation** — a time-bucketed metrics endpoint over those traces.

Three gaps blocked it. `trace_index` carried no agent column, so agent-scoped selection meant scanning the `attributes_string` map. Nothing in the telemetry identified an *agent entity* — ATIF's `agent_name` comes from the config's own `name`, shared by every agent built from it. And there was no aggregation API at all; Studio's Monitor tab downloads NAT telemetry JSONL and reduces it in the browser.

Measured on a live instance before the fix — agent entity `email-security-triage-6p05fw`:

```sql
SELECT attributes_string['gen_ai.agent.name'] AS agent_name, count() AS spans
FROM intake.spans WHERE is_deleted = 0 GROUP BY agent_name

┌─agent_name────────────┬─spans─┐
│ email-phishing-agent  │  2514 │
│ email-security-triage │    36 │
└───────────────────────┴───────┘
```

Zero spans under the entity name, so no agent-keyed query could match.

> **Note for reviewers:** despite the `intake` scope, this also changes **`plugins/nemo-agents`** — every Fabric deployment now stamps `telemetry.atif.extra`. See `ef9b1e3e0a`.

## Related Issue

ASTD-424, under ASTD-391 (Agent Entity Page). Closes ASTD-433 (agent-name filter for traces and spans) — spans already supported it; traces did not, and now do. Follows #1322 and #1327, which fixed workspace binding on the same telemetry path.

### Identity

- **`3ab8ba075e`** — `trace_index` gains `agent_name` / `agent_id` / `agent_version` plus bloom-filter skip indexes, resolved from the attribute catalog exactly as `project` and `test_case_id` already are. Migration `ch_trace_index_0007_agent` reuses `_create_trace_index_schema`, which rebuilds and backfills from `spans`. The filter is plumbed through `TraceListFilter` → the trace_index `WHERE` builder → the public `TraceFilter`.
- **`ae8a0ecfc8`** — ATIF's agent block has no `id` field and forbids unknown keys, so `agent.extra` is the only channel a deploying platform has. Intake reads `nemo.agent.id` from there and maps it to `gen_ai.agent.id`.
- **`ef9b1e3e0a`** — `resolve_for_deployment` already received the entity name and explicitly discarded it (`del agent_name`). It now stamps `telemetry.atif.extra["nemo.agent.id"]`.

**`ORDER BY` is deliberately unchanged.** Leading with agent would demote `root_started_at` and regress workspace-wide time-range listing, the common case. Skip indexes plus monthly partitioning cover agent-scoped reads; a `PROJECTION` is the escalation if that proves insufficient.

### Metrics

```
GET /v2/workspaces/{workspace}/traces/metrics
    ?bucket=total|hour|day|week|month
    &timezone=America/Los_Angeles
    &filter[agent_name]=email-security-triage
    &filter[started_at][$gte]=2026-08-01T00:00:00Z
```

Returns run and failed-run counts; sum, mean, p90 and p99 rollups for each token category and cost; and mean, p50, p90, p95 and p99 latency per bucket. It reuses `TraceFilter`, so it composes with `evaluation_id` as well as agent scoping.

Composed from the same helpers as `_trace_hydration_sql`: `_metric_roots_sql` filters `trace_index` and carries bucket, status and latency; `_trace_aggregates_sql` — now parameterized by `extra_where_sql` / `extra_select_sql` — supplies per-trace token and cost sums to both callers. Reading through `current_spans_sql` is mandatory: `spans` is a `ReplacingMergeTree`, so summing raw rows would double-count re-ingested spans. The helper now reads with `FINAL`, applying ClickHouse's latest-write-wins behavior directly without materializing wide per-column `argMax` states. All current callers scope spans by `session_id` before applying broader trace predicates.

Tokens cannot be denormalized onto `trace_index` — child LLM spans arrive in different insert blocks than the root, so a `TO`-table materialized view cannot sum across them.

- **`3930a5d1a2`** adds `avg_latency_ms`. The ASTD-391 chart plots a latency average, and a mean is not derivable from percentiles: on real data the mean run is 69410 ms against a 25447 ms median, a 2.7× gap from right skew.
- **`9a4f34bfcd`** exposes `models` / `providers` on the `Trace` response. The repository has always computed them; the API schema dropped them at the boundary.

## Mapping to the ASTD-391 prototype

Pulled the Figma Make source to check coverage rather than eyeballing the screenshot. The prototype's timeframe selector is `'day' | 'week' | 'month' | 'range'`, which maps 1:1 onto `bucket=day|week|month` and `bucket=total` + a `started_at` range.

| Prototype element | Status |
| --- | --- |
| Chart: cost / tokens / latency per bucket | ✅ |
| Avg token count · per run | ✅ `total_tokens / trace_count` |
| Avg cost | ✅ `cost_usd / trace_count` |
| Avg latency · ms/tok | ⚠️ ratio-of-sums only — see below |
| Details → Agent ID | ✅ |
| Details → Models | ✅ |
| "Average" tile | ❌ no unit or qualifier — undefined in the design itself |
| Open Insights, Experiments, Benchmarks, Evaluations tab | ➖ other services |

Two product questions remain, neither blocking this PR but both blocking Studio wiring:

- **`ms/tok` semantics.** Ratio-of-sums is derivable (`avg_latency_ms × trace_count / total_tokens` ≈ 1.52 ms/token on live data); mean-of-ratios is not. These diverge sharply on skewed data.
- **What the "Average" tile means.**

## Design decisions worth reviewing

**Identity is not stamped onto `gen_ai.agent.name`.** That was the first approach (#1333, closed). `gen_ai.agent.name` doubles as the span `name` field (`atif_mapping.py:297,359`), so overwriting it renames what users see in every trace list, and flows into `trace_index.root_name`, the traces API, and Experimentalist's display fallback. Too much blast radius for an identity problem. `gen_ai.agent.id` feeds nothing display-facing.

No spec change was needed in `nemo-relay` or NAT — `extra` is the sanctioned passthrough, verified working against the real exporter.

**The stamp carries the entity *name*, not the entity id.** Both are valid identities; name was chosen deliberately. Note the prototype's Details card displays an entity id (`agent-6pyp1nWQaerfWw1m2`), so whoever wires Studio should filter by `agent_name`.

**Bucket types are cast.** `toStartOfWeek` / `toStartOfMonth` return `Date` while `toStartOfDay` / `toStartOfHour` return `DateTime`. Pydantic accepts both but yields *naive* datetimes for the former and timezone-aware for the latter, so chart clients would see inconsistent offsets.

**No pagination, deliberately.** A chart wants its whole series. The response is bounded by retention — 90 days × hourly = 2160 buckets max, roughly 750 KB worst case — so there is a hard ceiling rather than an open-ended result. `page`/`page_size`/`sort` were being silently accepted and ignored by the shared list validator; they now return 400.

## Two naming traps

- **`agent_id` means two different things in this repo.** Container metadata's `agent_id` (`container/metadata.py:203`) is a truncated SHA-256 of config + pyproject + build env — a content hash that changes on every rebuild. `nemo.agent.id` here is the agent *entity*, stable across rebuilds.
- **Two `error_count`s.** The trace rollup's counts failed *spans* within a trace; the metrics one counts failed *runs* by root status. Both correct for their purpose; commented at the definition rather than renamed, since the span-level one is existing API surface.

## Compatibility

Additive. Audited rather than assumed:

- `nemo.agent.id` as an ATIF extras key is **new** — the `com.nemo.agent.id` hits in the tree are a Docker label, different namespace
- **Nothing read `agent.extra`** before this change
- **Nothing consumes `gen_ai.agent.id`** — two producers (Analyst, a seed script), zero readers
- `gen_ai.agent.name` is untouched, so span names, trace lists and existing consumers are unaffected
- New response fields, filter fields and the endpoint are all additive

Two caveats that are *not* "nothing changes":

- **The migration drops and rebuilds `trace_index`** rather than `ALTER`ing. Not zero-downtime; reads during the rebuild see an incomplete table. This is the established pattern — migrations `0004`, `0005` and `0006` all re-run the same function — and `spans` is the durable source of truth, but on a large deployment that backfill `INSERT` is not instant.
- **A non-dict `telemetry.atif.extra` is replaced.** Such a value would already fail `RelayAtifConfig.extra: JsonObject | None`, so it was broken regardless, but the replacement is silent. A user-supplied dict is preserved and merged.

Traces ingested before this change have no `agent_id`; nothing migrates, so agent-keyed queries see only new data.

Fabric only. NAT-format agents export to the Files service and never reach Intake — out of scope by decision.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the new endpoint is self-describing via OpenAPI; the rest is internal schema and deploy-time config resolution.

| Area | Covers |
| --- | --- |
| `test_spans_clickhouse_migrations.py` | new columns, indexes, MV select expressions; catalog bag-key assertions, since the MV bakes keys in at creation and a rename needs a new migration |
| `test_traces_api.py` | agent filter mapping, public schema descriptions, `models`/`providers` round-trip and omission in summary mode |
| `test_traces_clickhouse_repository.py` | filter compiles to a `trace_index` column with no `candidate_spans` fallback; all bucket expressions; `total` collapses without leaking its sentinel; rollup reads deduplicated spans through `FINAL`; join keys on full trace identity |
| `test_trace_metrics_api.py` | timezone validation, response mapping, `bucket_start` omitted for `total`, pagination params rejected |
| `test_atif_v17.py` | id reaches `gen_ai.agent.id` on every span while `gen_ai.agent.name` keeps the config name; absent/empty/non-string extras yield no id |
| `test_utils.py`, `test_agent_config_formats.py` | stamping, existing extras preserved, unusable extras replaced, absent telemetry tolerated; through the real resolver, stored config not mutated |

## Verification

| Command | Result |
| --- | --- |
| `pytest services/intake/tests` (unit) | 335 passed |
| `pytest plugins/nemo-agents/tests/unit` | 1001 passed, 3 failed (pre-existing) |
| `ruff check` / `ruff format --check` on both trees | clean |
| `uv run --frozen pytest services/intake/tests/test_traces_clickhouse_repository.py services/intake/tests/test_evaluation_session_clickhouse_repository.py services/intake/tests/test_traces_api.py services/intake/tests/test_trace_metrics_api.py services/intake/tests/test_clickhouse_executor.py -q` | 60 passed |
| `uv run --frozen pytest services/intake/tests/integration/spans/test_otlp_ingest_simple.py::test_otlp_reingest_same_batch_deduplicates_before_and_after_merge -q` | 1 passed against isolated ClickHouse |
| `uv run --frozen ty check services/intake/src/nmp/intake/repository/clickhouse/trace.py` | clean |
| `tsc --noEmit` (studio) | clean |
| `script/copyright_fixer.py` over every changed file | Processed 44 files, updated 0 |

### Verified against live ClickHouse 26.3, not just mocks

The unit tests use a fake client, so the SQL never executes there. Run against a real instance with real data:

- Migration applied: **25 `trace_index` rows before, 25 after**, agent columns backfilled from `spans`
- All five bucket granularities execute; hourly splits the day's 23 traces into 12 and 11
- The rollup reproduces the known per-agent figures exactly: **23 traces / 2514 spans / 658 tool calls / 968087 input tokens** for `email-phishing-agent`
- The `206f77e045` refactor returns **identical** figures, confirming it is behaviour-preserving
- Every endpoint variant exercised over HTTP, including a 400 for an unknown timezone

### Rollup performance validation

Measured against ClickHouse 26.3 with 3.8 million spans. Every `FINAL` result matched the corresponding `argMax` result exactly.

| Caller | `argMax` | `FINAL` |
| --- | ---: | ---: |
| Trace metrics | 8.91 s / 3.53 GB | 0.33 s / 150 MB |
| Trace hydration, 10 traces | 25 ms / 13 MB | 21 ms / 22.5 MB |
| Evaluation metric sort, 10,001 sessions | 653 ms / 387 MB | 177 ms / 122 MB |
| Evaluation hydration, 100 sessions | 181 ms / 22.3 MB | 149 ms / 25.7 MB |

This is what caught the `Date` vs `DateTime` bucket inconsistency; the mocked tests could not have.

The full identity chain was also exercised in-process against the **real** components — `bind_atif_agent_id` → `nemo_relay.AtifExporter` → `trajectory_to_spans`:

```
span: name=email-security-triage   gen_ai.agent.id=email-security-triage-6p05fw
```

### Local pre-commit limitations

`uv run pre-commit run -a` completed all code-related hooks successfully: Ruff, Ruff format, ty, generated config reference, uv lock checks, copyright headers, forbidden imports, merge-conflict detection and Flox-lock checks. Host-tooling hooks were blocked because this machine does not have `helm-docs` or `yq`, and the isolated worktree does not have Studio's `lint-staged` dependency. This change does not touch Helm, toolchain versions or Studio.

### Pre-existing failures, not introduced here

- `test_port_allocation.py` (2) — `PermissionError` on `socket.bind` under the local sandbox.
- `test_cli_list_output.py::TestDeploymentsListOutput::test_deployments_list_defaults_to_table` — fails identically with these changes stashed.

## Not verified

- **No end-to-end run against a deployed agent** — redeploying the sample agent and confirming new spans carry `gen_ai.agent.id`. The chain is proven in-process with the real relay exporter and real intake mapping; only the deployed-container leg is unexercised.
- **Studio is not wired to this endpoint.** The Monitor tab still parses Files-service JSONL; consuming `/traces/metrics` is separate work.

## Summary by CodeRabbit

* **New Features**
  * Added bucketed, timezone-aware trace metrics at `/traces/metrics`.
  * Metrics now include run counts and token, cost, and latency rollups with percentile statistics.
  * Added agent-name filtering and model/provider details for trace responses.
  * Added CLI support for retrieving trace metrics, including a default seven-day window.
* **Updates**
  * Trace metrics now use session-aware aggregation and enforce query limits.
  * Agent ID filtering and response fields were removed.
* **Documentation**
  * Updated API, CLI, and permission documentation for the new metrics endpoint and fields.


services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -200,6 +206,53 @@ async def get_trace(self, *, workspace: str, trace_id: str, mode: TraceMode) ->
...
+            filters=filters,
+            bucket=bucket,
+        )
+        # Reuses the same per-trace span rollup as trace hydration, rescoped from a
+        # page of trace refs to whatever the filter selected.
+        rollups_sql, rollup_parameters = _trace_aggregates_sql(
+            spans_table,
+            extra_where_sql="span_versions.trace_id IN (SELECT trace_id FROM roots)",
+            extra_select_sql="countIf(trace_spans.kind = 'TOOL') AS tool_call_count",
+        )
```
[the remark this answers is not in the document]
marcusds: Claude says: Already addressed in fa48f1d0ad, which predates this review — the finding was raised independently and fixed then. Current code carries both predicates:

```python
extra_where_sql=(
    # trace_id alone drives the bloom filter; the tuple keeps a trace_id
    # shared across source formats from pulling in unrelated spans.
    "span_versions.trace_id IN (SELECT trace_id FROM roots)\n"
    "                AND (span_versions.source_format, span_versions.trace_id) "
    "IN (SELECT source_format, trace_id FROM roots)"
),
```

Declining the proposed diff specifically, because it drops the scalar `trace_id IN`. That predicate is what the `idx_trace_id` bloom-filter skip index can use for granule pruning; a tuple `IN` is unlikely to be usable for index analysis, so removing it trades a real scan reduction for nothing.

It also matches this file's existing pattern — `_PAGE_TRACE_REFS_WHERE_SQL`, feeding the same `_trace_aggregates_sql`, keeps both for the same reason:

```python
_PAGE_TRACE_REFS_WHERE_SQL = (
    "span_versions.trace_id IN %(page_trace_ids)s\n"
    "                AND (span_versions.source_format, span_versions.trace_id) "
    "IN %(page_trace_keys)s"
)
```

Verified against live ClickHouse after the change: identical figures (23 traces, 2514 spans, 658 tool calls, 968087 input tokens).

services/intake/src/nmp/intake/spans/api/trace_metrics.py
```diff
@@ -0,0 +1,90 @@
...
+from nmp.intake.spans.api.traces import _trace_filter
+from nmp.intake.spans.api.traces_schemas import TraceFilter
+
+router = APIRouter(dependencies=[Depends(require_workspace_access)])
+API_TAG = "Traces"
+_ALLOWED_QUERY_PARAMS = frozenset({"bucket", "timezone", "filter"})
+
+
+@router.get(
+    "/v2/workspaces/{workspace}/trace-metrics",
```
BrianNewsom: Since this operation aggregates the traces resource, can we expose it as `/v2/workspaces/{workspace}/traces/metrics` rather than `/trace-metrics`? That keeps the API hierarchy consistent with the existing `/traces` endpoints and the generated `traces.get_metrics` SDK organization.

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
```diff
@@ -316,6 +320,29 @@ def rebind_intake_ingest_workspace(config: dict[str, Any], workspace: str) -> di
...
+    if not isinstance(telemetry, dict):
+        return config
+    atif = telemetry.get("atif")
+    if not isinstance(atif, dict):
+        return config
+    extra = atif.get("extra")
+    if not isinstance(extra, dict):
+        extra = {}
+        atif["extra"] = extra
+    extra[NEMO_AGENT_ID_EXTRA_KEY] = agent_name
```
BrianNewsom: Isn’t the goal here to aggregate all runs of the same agent, whether they originate from customer code or a Platform deployment? ATIF already provides `agent.name`, which seems like the natural grouping and filtering key. Can you clarify why that isn’t sufficient and what distinct use case requires a separate `agent_id`?

services/intake/src/nmp/intake/repository/clickhouse/trace.py
```diff
@@ -200,6 +206,59 @@ async def get_trace(self, *, workspace: str, trace_id: str, mode: TraceMode) ->
...
+            bucket=bucket,
+        )
+        # Reuses the same per-trace span rollup as trace hydration, rescoped from a
+        # page of trace refs to whatever the filter selected.
+        rollups_sql, rollup_parameters = _trace_aggregates_sql(
+            spans_table,
+            extra_where_sql=(
+                # trace_id alone drives the bloom filter; the tuple keeps a trace_id
+                # shared across source formats from pulling in unrelated spans.
+                "span_versions.trace_id IN (SELECT trace_id FROM roots)\n"
```
BrianNewsom: I ran some load testing locally against a workspace with 3.8M spans. The current broad `argMax` rollup took 47s and 3.5GiB, while reading `spans FINAL` took 1.44s and 168MiB.

Trace-ID-only `FINAL` was slower for narrow hydration because ClickHouse expanded three bloom-selected granules to 498. However, carrying `session_id` from the selected roots and filtering by `workspace + session_id + (source_format, trace_id)` allowed the primary key to engage: a 10-trace hydration query took 24ms versus 26ms for the existing `argMax` path.

Can we include `session_id` in the selected root identity and use it in the span lookup?

services/intake/src/nmp/intake/spans/api/trace_metrics.py
```diff
@@ -0,0 +1,90 @@
...
+    ),
+    timezone: str = Query(
+        default="UTC",
+        description="IANA timezone the buckets are aligned to, e.g. America/Los_Angeles.",
+    ),
+    parsed: ParsedFilter = Depends(make_filter_dep(TraceFilter)),
+) -> TraceMetrics:
+    _validate_query_params(request)
+    _validate_timezone(timezone)
+    filters = _trace_filter(workspace, parsed)
```
BrianNewsom: Can we default an unspecified time range to the last seven days and add appropriate ClickHouse execution, memory, and rows-read limits? Bounding the number of returned buckets does not bound the amount of data scanned.

services/intake/src/nmp/intake/spans/api/trace_metrics_schemas.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TraceMetricBucketParam(StrEnum):
+    TOTAL = "total"
+    HOUR = "hour"
+    DAY = "day"
+    WEEK = "week"
+    MONTH = "month"
+
+
+class TraceMetricPointResponse(BaseModel):
```
BrianNewsom, line 46: Could we use a reusable metric-rollup shape rather than continuing to add flat fields? For example, a complete daily response could look like:

```json
{
  "bucket": "day",
  "data": [
    {
      "bucket_start": "2026-08-18T00:00:00Z",
      "run_count": 143,
      "failed_run_count": 7,
      "tool_call_count": 658,
      "span_count": 2514,
      "input_tokens": {
        "sum": 968087,
        "mean": 6769.84,
        "p90": 12004,
        "p99": 18442
      },
      "output_tokens": {
        "sum": 42112,
        "mean": 294.49,
        "p90": 681,
        "p99": 1203
      },
      "cached_tokens": {
        "sum": 300000,
        "mean": 2097.90,
        "p90": 8000,
        "p99": 16000
      },
      "total_tokens": {
        "sum": 1010199,
        "mean": 7064.33,
        "p90": 12680,
        "p99": 19310
      },
      "cost_usd": {
        "sum": 12.42,
        "mean": 0.0869,
        "p90": 0.19,
        "p99": 0.42
      },
      "latency_ms": {
        "mean": 69410,
        "p50": 25447,
        "p90": 121000,
        "p95": 180300,
        "p99": 310000
      }
    }
  ]
}
```

`bucket=total` can keep the same envelope and return one data point without `bucket_start`. I’d omit a latency count and use `run_count`/`failed_run_count` rather than the more ambiguous `trace_count`/`error_count`.

I also tested the additional p90/p99 calculations locally, including 2,160 bucket groups. They did not materially increase query memory or execution time. Using combined aggregates such as `quantiles(0.9, 0.99)(input_tokens)` avoids maintaining separate quantile states.

services/intake/src/nmp/intake/spans/api/trace_metrics_schemas.py
```diff
@@ -0,0 +1,52 @@
...
+    bucket_start: datetime | None = Field(
+        default=None,
+        description="Start of the bucket in the requested timezone. Omitted when bucket=total.",
+    )
+    trace_count: int = Field(ge=0, description="Traces (agent runs) started in this bucket.")
+    error_count: int = Field(ge=0, description="Traces whose root span ended in error.")
```
BrianNewsom: Are `error_count`, `tool_call_count`, and `span_count` required by the current product/design? The prototype appears focused on run count, tokens, cost, and latency. If these additional counts do not have an immediate consumer, I’d prefer keeping the initial API minimal and adding them later as additive fields rather than anticipating future requirements in the first public contract. `run_count` still seems necessary as the denominator for the per-run metrics.

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
marcusds: In my head it made sense to do it by deployment, but I can't think of why now. We can remove this and if needed in the future add it back

marcusds: Oh, it was I was worried about scenarios agent.name could be missing or be non-uninque.

How would we handle these scenarios if name isn't set?

But that being said if non-platform run, how would we manage no agent id? Fallback to agent name and still have this issue?

services/intake/src/nmp/intake/spans/api/trace_metrics.py
marcusds: Done

services/intake/src/nmp/intake/repository/clickhouse/trace.py
marcusds: Done

services/intake/src/nmp/intake/spans/api/trace_metrics.py
marcusds: Default for time range is set but I am not sure what is realistic values.

Claude picked these when asked:
```
METRIC_QUERY_SETTINGS = {
    "max_execution_time": 30,          # seconds
    "max_memory_usage": 4 * 1024**3,   # 4 GiB
    "max_rows_to_read": 200_000_000,
}
```

But that is the upper limit of what you encountered in the comment above, so I don't think this is right.

services/intake/src/nmp/intake/spans/api/trace_metrics_schemas.py
marcusds: Done. Previous UX mocks had tool_call, but I think it is fair to still remove that unless needed in the future.

services/intake/src/nmp/intake/spans/api/trace_metrics_schemas.py
marcusds: Done

services/intake/src/nmp/intake/spans/api/trace_metrics.py
marcusds: On second thought, I think these should be set global in env so users can configure. If this is running on a huge beefy server with TBs of ram, who are we to say they can't use a hundred gigs on a query.

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
BrianNewsom: > Oh, it was I was worried about scenarios agent.name could be missing or be non-unique.

Yeah certainly can happen.  non-unique is sort of an interesting question. This name describes the identity of the agent and is used to "group" like executions. So non-uniqueness is kind of the point?

> But that being said if non-platform run, how would we manage no agent id? Fallback to agent name and still have this issue?

This is actually part of why I think we should use name vs. ID. Name is "more" standard (it's in openinference otel and atif) whereas agent.id is /only/ present in genai semantic convention OTEL, which in my experience is least common.

So instead of falling back, I was thinking we use name here.

agent.id is a valid key and ultimately we should support it, but I think name is probably sufficient for the scope of this change in my view.

services/intake/src/nmp/intake/spans/api/trace_metrics.py
BrianNewsom: Yeah I think you're right. Let's pull out the Clickhouse limits in this review and we can deal with that separately. Sorry for the misdirect here!

services/intake/src/nmp/intake/spans/api/trace_metrics_schemas.py
BrianNewsom: Awesome, thank you

plugins/nemo-agents/src/nemo_agents_plugin/utils.py
marcusds: Makes sense. Changes is already made.

services/intake/src/nmp/intake/spans/api/trace_metrics.py
marcusds: Done

BrianNewsom (approved):

PR #1383, opened 2026-08-18
fix: address review feedback for submit-only migration

mckornfield:


docs/data-designer/cli.mdx
```diff
@@ -30,41 +31,17 @@ def load_config_builder() -> dd.DataDesignerConfigBuilder:
     return config_builder
 ```
 
-The same configuration source can usually be used with `run` or `submit`. Resource choices determine whether it is compatible with NeMo Services execution; see [Execution Modes](/documentation/design-synthetic-data/execution-modes).
-
-## Run Versus Submit
+The same configuration source is used with `submit`. Resource choices determine whether it is compatible with NeMo Services execution; see [Execution Modes](/documentation/design-synthetic-data/execution-modes).
 
-`run` executes the Data Designer workload locally, in the CLI process. This can be fully local, but it is not an offline-only mode. A local run can still use the Files API, Secrets API, and Inference Gateway API from a running NeMo Services cluster when the configuration references the corresponding resources.
+## Submit Versus Platform Execution
```
mikeknep, line 36: These don't make sense semantically

docs/data-designer/cli.mdx
```diff
@@ -7,7 +7,7 @@ description: ""
 ---
 <a id="data-designer-cli"></a>
 
-The NeMo Data Designer plugin adds the `nemo data-designer` command group. Use it to execute Data Designer workloads locally in the CLI process or submit them to NeMo Services.
+The NeMo Data Designer plugin adds the `nemo data-designer` command group. Use it to submit preview and create workloads to NeMo Services.
```
mikeknep, line 10: Probably makes more sense something like...
> Use it to generate high-quality synthetic data on NeMo Platform.

(Side note, I see "NeMo Services" already exists... is that a legit proper noun?)

docs/data-designer/cli.mdx
```diff
@@ -30,41 +31,17 @@ def load_config_builder() -> dd.DataDesignerConfigBuilder:
...
 
-`run` executes the Data Designer workload locally, in the CLI process. This can be fully local, but it is not an offline-only mode. A local run can still use the Files API, Secrets API, and Inference Gateway API from a running NeMo Services cluster when the configuration references the corresponding resources.
+## Submit Versus Platform Execution
 
 `submit` sends the workload to NeMo Services. The Data Designer API and Jobs API coordinate execution, job lifecycle, logs, and artifact persistence. The NeMo Services deployment may itself be local or remote.
 
 | Command | Workload execution | NeMo Services required? |
 |---------|--------------------|-------------------------|
-| `preview submit` | Local CLI process | Optional |
-| `create submit` | Local CLI process | Optional |
```
mikeknep, line 44: lol at this table as it exists on the target branch, just totally borked. But also this entire table is pointless now, since the platform services (again that "NeMo Services" noun!) are always required.

docs/data-designer/cli.mdx
mckornfield: lol no clue on NeMo Services, but it shows up a bunch

PR #1365, opened 2026-08-18
feat(agents): Initial AgentInvocationJob skeleton

mikeknep:
## Summary
Provides the initial `NemoJob` plumbing for the new **Agent Invocation job**. This first PR is deliberately lightweight; we resolve the agent working directory, snapshot it as a named job result, and exit.

## Changes
- Adds initial `AgentInvocationJob(NemoJob)` implementation
- Wires up API factory routes
- Unit tests + one integration test

While here: fixes a bug in the copyright fixer that was mistakenly adding copyright headers to openapi.yaml files. The fixer script was not recognizing the `.copyrightignore` file.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
- [x] Documentation not applicable — justification: The job does not do anything useful yet; we'll document it once it actually "does the thing"

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [x] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

## Summary by CodeRabbit

* **New Features**
  * Added agent invocation jobs that resolve configured agents and execute them on CPU.
  * Added support for staging base files and mounted artifacts in agent working directories.
  * Added validation for file references and mount paths, including overlap and path-safety checks.
  * Added task, service routes, and API operations for managing invocation jobs, results, logs, and status.

* **Bug Fixes**
  * Improved copyright-ignore handling for explicitly targeted files.

* **Tests**
  * Added comprehensive unit and integration coverage for invocation, workspaces, validation, execution, and failure scenarios.

PR #1388, opened 2026-08-18
feat(agents): Call Fabric from the job

mikeknep:
## Summary
Building on https://github.com/NVIDIA-NeMo/nemo-platform/pull/1365, this PR adds the meat of the `AgentInvocationJob`: calling Fabric and persisting the output results.

## Changes
- Updates the job implementation to call Fabric and persist results after Fabric finishes
- Updates some common `nemo_agents_plugin.fabric.invocation` helpers
- Adds e2e tests

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification:

## Verification

### DIFF
--- a/e2e/test_nemo_agents_invoke_job.py
+++ b/e2e/test_nemo_agents_invoke_job.py
@@ -1,0 +1,369 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""E2E tests for Fabric-backed agent invocation jobs."""
+
+from __future__ import annotations
+
+import io
+import json
+import tarfile
+from typing import Any
+
+import pytest
+from nemo_agents_plugin.entities import NEMO_AGENTS_SPEC_CONFIG_FORMAT
+from nemo_platform import NeMoPlatform
+from nmp.testing import MockProviderResponse, add_mock_provider
+from nmp.testing.e2e import wait_for_platform_job
+
+from e2e.agents_deploy_helpers import (
+    TEST_AGENT_RESPONSE,
+    _mock_backed_fabric_agent_config,
+    delete_agent_if_exists,
+    unique_name,
+)
+
+pytestmark = [
+    pytest.mark.subprocess_only,
+    pytest.mark.e2e_config(
+        "e2e/configs/local-subprocess.yaml",
+        harness={"backend": "subprocess"},
+    ),
+    pytest.mark.timeout(600),
+]
+
+
+def _agents_url(sdk: NeMoPlatform, workspace: str, path: str) -> str:
+    return f"{str(sdk.base_url).rstrip('/')}/apis/agents/v2/workspaces/{workspace}/{path.lstrip('/')}"
+
+
+def _job_diagnostic_message(sdk: NeMoPlatform, job: Any, workspace: str, prefix: str) -> str:
+    parts = [prefix]
+    if job.status_details:
+        parts.append(f"Status details: {job.status_details}")
+    if job.error_details:
+        parts.append(f"Error details: {job.error_details}")
+    try:
+        logs = sdk.jobs.get_logs(workspace=workspace, name=job.name)
+        if logs.data:
+            parts.append(f"Job logs ({len(logs.data)} entries):")
+            for entry in logs.data:
+                parts.append(f"  - {entry.message}")
+    except Exception as error:
+        parts.append(f"Could not fetch job logs: {error}")
+    return "\n".join(parts)
+
+
+def _list_invoke_job_results(sdk: NeMoPlatform, workspace: str, job_name: str) -> dict[str, Any]:
+    response = sdk._client.get(_agents_url(sdk, workspace, f"jobs/invoke/{job_name}/results"))
+    assert response.status_code == 200, f"Failed to list invoke job results for {job_name}: {response.text}"
+    return response.json()
+
+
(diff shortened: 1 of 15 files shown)


e2e/test_nemo_agents_invoke_job.py
```diff
@@ -0,0 +1,369 @@
...
+
+import pytest
+from nemo_agents_plugin.entities import NEMO_AGENTS_SPEC_CONFIG_FORMAT
+from nemo_platform import NeMoPlatform
+from nmp.testing import MockProviderResponse, add_mock_provider
+from nmp.testing.e2e import wait_for_platform_job
+
+from e2e.agents_deploy_helpers import (
+    TEST_AGENT_RESPONSE,
+    _mock_backed_fabric_agent_config,
```
mikeknep: Ought to make this public in the helper module, and/or refactor a bit. I'll clean this up

e2e/test_nemo_agents_invoke_job.py
```diff
@@ -0,0 +1,369 @@
...
+    delete_agent_if_exists,
+    unique_name,
+)
+
+pytestmark = [
+    pytest.mark.subprocess_only,
```
mikeknep: Not sure if this needs to be subprocess-only; ideally not, of course. I'll explore if it works on the other harnesses (or if not, why not)

plugins/nemo-agents/src/nemo_agents_plugin/jobs/invoke.py
```diff
@@ -90,6 +114,15 @@ async def to_spec(  # type: ignore[override]
         except NemoEntityNotFoundError as exc:
             raise ValueError(f"Agent '{request.agent}' not found.") from exc
 
+        _validate_agent_config_format(agent.config_format)
+        resolved_agent_config = resolve_agent_config_for_deployment(
```
mikeknep, line 130: We may want to change this helper function name, as it's no longer just for deployments

benmccown: I can refactor/rename when I hit the `AgentInvocationJob` work related to env/envspec if you'd like.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/invoke.py
```diff
@@ -141,23 +174,115 @@ async def compile(  # type: ignore[override]
...
-            materialize_agent_workdir(step_config.workdir, sdk.files, local_workdir)
-            result_ref = ctx.results.save(INPUT_WORKDIR_RESULT_NAME, local_workdir)
+            materialize_agent_workdir(step_config.workdir, sdk.files, fabric_dirs.workspace)
+
+        input_workdir_ref = ctx.results.save(INPUT_WORKDIR_RESULT_NAME, fabric_dirs.workspace)
+
+        try:
+            result = asyncio.run(
+                invoke_agent_config_request_once(
+                    AgentConfigInvocationRequest(
```
benmccown, line 204: Claude caught a possible issue. Not sure if this is actually valid?

---

`AgentInvocationJob.run()` (jobs/invoke.py) constructs `AgentConfigInvocationRequest` without setting `timeout_seconds`, so it defaults to `None`. That flows into `run_fabric_agent_once`'s `asyncio.wait_for(..., timeout=request.timeout_seconds)` (fabric/runtime.py), where `None` means unbounded wait — no fallback/default is applied anywhere in that path.

Nothing above this layer enforces a per-invocation timeout either:
- `NemoJob.run()` is invoked directly by the scheduler with no `asyncio.wait_for`/deadline wrapping.
- `PlatformJobStep.lifecycle.staleness_timeout_seconds` defaults to `0` (disabled), and `AgentInvocationJob.compile()` doesn't set it.

The only backstop is the execution profile's `ttl_seconds_active` (default 24h, often 1h in local configs), which maps to k8s `activeDeadlineSeconds` (or the docker/subprocess/volcano equivalent). That's an infra-wide ceiling unrelated to this specific invocation, not a per-job timeout — a stuck Fabric call can occupy a job slot for up to that long before anything intervenes.

Suggest threading a `timeout_seconds` value through from job config (e.g. `AgentInvocationJobConfig`) into `AgentConfigInvocationRequest` rather than relying on the infra-level deadline.

benmccown (approved):
One comment, otherwise LGTM!

plugins/nemo-agents/src/nemo_agents_plugin/jobs/invoke.py
mikeknep: Seems legit to me, thanks. Addressed in 9b04b2ebf. I set the default to 1 hour... 🤷 we can modify any time

e2e/test_nemo_agents_invoke_job.py
mikeknep: Decided to just make the one function public for now, punt on wider e2e helper refactoring. Addressed in e81d20385

e2e/test_nemo_agents_invoke_job.py
mikeknep: Pretty sure these constraints were unnecessary. Passes locally without the marks; CI will be the true arbiter. Dropped in f06af47d2

plugins/nemo-agents/src/nemo_agents_plugin/jobs/invoke.py
mikeknep: Sure, let's punt it to then, since there may be some larger changes involved. Thanks!

PR #1156, opened 2026-08-07
feat(customizer): add grpo support with gym environments

soluwalana:
## Summary

Adds NeMo Platform GRPO training support with NeMo Gym environments (Mode B / OpenSandbox). Operators convert Prime Intellect hub envs offline via `pi-to-gym-conversion` into `adapter-wheels-v1` FileSets + Gym JSONL, upload them, and submit GRPO jobs; the compiler downloads the environment FileSet, emits sandboxed Gym host config with master-time egress allowlists, and runs the real GRPO driver instead of the Penguin stub.

## Related Issue

- Linear: [AALGO-418](https://linear.app/nvidia/issue/AALGO-418) (NemoPlatform GRPO backend)
- Epic: [AALGO-340](https://linear.app/nvidia/issue/AALGO-340)
- Out of scope for this PR: Studio UI, AALGO-439 skills/docs as a deliverable (skill hyperparam docs were lightly updated for accuracy)

## Changes

- **OpenSandbox install/verify** — helm values, install script, and verification helpers under `packages/nmp_customization_common/values/opensandbox/`
- **Files / metadata** — `FilesetPurpose.ENVIRONMENT` + `EnvironmentMetadataContent` for environment packages
- **CLI convert** — `pi-to-gym-conversion` (`nmp-rl`): hub → `adapter-wheels-v1` + Gym JSONL (requires `pip` + `verifiers`; fails closed if dataset generation yields 0 rows)
- **Schemas** — `GRPOTraining` + optional `environment` on RL job specs; environment manifests in `nmp.rl.schemas.environment` / job types in `nmp.rl.schemas.job`
- **Plugin** — transform accepts environment refs; `minimal_grpo.json` fixture; contract tests
- **Compiler / config** — third FileSet download, sandbox paths, `sandboxed_gym_default` / `sandbox_cluster_capable` fail-closed, egress env placeholders
- **Runtime** — `sandbox_config` helpers, GRPO config compilation, replaced Penguin `grpo_driver` with Gym GRPO + optional Mode A bootstrap / Mode B master egress fill
- **Docs / skills** — RL plugin README, `services/rl` README, customizer hyperparams references

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [x] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
[...]

PR #1400, opened 2026-08-19
feat(evaluator): run Gym evaluations inside a sandboxed Gym host

SandyChapman:
## Summary

Lets an Evaluator Gym evaluation run its environment **inside a sandbox** instead of as a
subprocess tree in the job container. Today `GymAgentTaskRunner` needs the `gym` CLI on PATH and a
Gym checkout as its working directory, and whatever an environment's `resources_server` code does,
it does with the job's credentials. That is fine for a vetted environment on trusted hardware and
unacceptable once the environment arrives as a user-supplied FileSet.

Whether a run is sandboxed is a **deployment** decision, not a job field: the same submitted
`GymRunnerTarget` runs colocated on a dev box and sandboxed on a shared cluster. Customizer settled
the same question the same way for GRPO (`NMP_RL_SANDBOXED_GYM_DEFAULT`), so one submit contract
covers both.

This also carries Sam Oluwalana's `sandboxed_gym` package (extracted from #1154) into the
workspace, with his commits and authorship intact.

## Related Issue

Supersedes the integration half of #1154. Adjacent to #1156 (GRPO) and #1315 (Gym e2e).

## ⚠️ Two host-provider paths — one is untested

Host provider paths

**`docker` is validated end to end**: real episode broker on a real socket, real container, real
NeMo-Gym 0.5.0, real `mcqa` grading, one COMPLETED trial attributed by the `_ng_task_index` we
stamped. It exists so the path can be executed and debugged without a cluster.

**`opensandbox` has never been executed.** Not once. The live test
(`test_a_real_opensandbox_host_serves_rollouts`) is skipped unless `RUN_SANDBOXED_GYM_LIVE` and
cluster credentials are set, and no OpenSandbox API call has ever been made from this code. PVC
mounts, enforced egress, the readiness probe against a real host and image pull are all unexercised.
`OpenSandboxDriver` type-checks against the real SDK types, which is a compile-time check, not a
runtime one.

**The Docker path proves the code, not the isolation.** A container is not kata-qemu, and that
provider *records* the egress allowlist rather than applying it — deliberately, and readable back
off the provider, so the difference stays visible rather than being asserted away. Anyone reading
"validated" should read it as "the wiring works", not "the boundary holds".

Running the real one needs a Kubernetes cluster: `kubectl`/`helm` install, and for the kata profile
a `RuntimeClass/kata-qemu` plus kata-labelled H100 nodes. The `crun` profile is the cheaper first
target — it exercises PVC mounts, the readiness probe and the driver's API calls without special
hardware.

## Architecture

Architecture

The trusted side holds the OpenSandbox credential and the episode broker; the sandbox gets a
per-run token and nothing else. A local PoC found that putting stock OpenSandbox credentials on the
untrusted side lets `provider_options` escalate to socket or host mounts, which is the hole the
broker closes (sandboxed-GRPO RFC §6.7).

The **episode tier is dashed because nothing has provisioned one**. It is only used by environments
that ask for nested sandboxes; an ordinary evaluation never creates one.

## Changes

**Package integration.** `sandboxed_gym` becomes a workspace member. That required removing its
`nemo-gym` dependency first: Gym floors at CPython 3.13.14 and pulls `mlflow-skinny>=3.15.1`, which
`services/unsloth` contradicts — and uv does not fail on that, it resolves backwards to nemo-gym
0.2.1, a release predating the sandbox work. The imports were also from `nemo_gym.sandbox.broker`,
which exists only on an unmerged fork. The 290-line wire contract and 191 lines of provider types
are vendored instead; `BROKER_PROTOCOL_VERSION` detects drift.

**NeMo-Gym removed from both provisioning tiers.** `OpenSandboxDriver` drives the `opensandbox` SDK
directly, replacing a 1541-line provider of which seven methods were used. Gym now appears in
exactly one module — `runtime/gym_host_runtime.py`, which runs *inside* the Gym image.

**Directory transfer added to the broker contract** (protocol `1` → `2`). Nothing below the broker
has a directory primitive, and `download_dir` cannot even enumerate without an `exec`, so it is
synthesized at the backend seam — one trusted implementation, paths quoted once, and a backend with
a native transport can override it.

**Service-side execution.** `EvaluatorConfig` gains the sandbox settings; the gates raise rather than
fall back, so a cluster that cannot sandbox refuses the run instead of quietly running user
environment code beside the job's credentials.

**Credentials move to secret references.** `GymRunnerTarget.env_secrets` is `{ENV_NAME: SecretRef}`,
resolved service-side through the path metric and endpoint secrets already take. A sandboxed run
whose `env_vars` looks credential-shaped is refused, naming the keys.

**Selection maps onto Gym's own mechanism.** `--config`/`--model-type`/`--resources-server` are sugar
over `config_paths`; Gym merges what it is handed and *then* loads those paths itself, so the
selection travels as relative paths that Gym resolves inside the sandbox. We never need Gym's YAML
schema.

## Findings that only surfaced by running it

- `RolloutCollectionHelper.run_examples` reads `row["agent_ref"]["name"]` with **no fallback**, while
  `gym eval run` resolves the agent from config during preprocessing — so a dataset that works under
  the CLI arrives unroutable at a sandboxed host. `agent_ref_name` now stamps rows that lack one.
  The name Gym wants is the instance an environment's config defines (`mcqa_simple_agent`), not the
  component (`simple_agent`); routing to the component times out.
- Gym spawns every server through `uv`; without it on PATH all four die with "Process ... finished
  unexpectedly" and nothing further.
- The serve config granted **no egress at all**, so a real host would have come up healthy with a
  route to the broker and nothing else, then failed at the first rollout. Every test passed through
  it, because a stubbed host needs no egress.
- The environment and workspace mounts **resolved to the same directory** — one claim, both sub-paths
  empty — so a run could modify the environment it was evaluating.
- `wait_ready` polled a dead container until the ready timeout, turning a two-second crash into a
  fifteen-minute wait with the traceback unread in `docker logs`.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

| Command | Result |
| --- | --- |
| `pytest plugins/nemo-evaluator/tests` | 901 passed, 26 skipped |
| `pytest packages/sandboxed_gym/tests` | 177 passed, 2 skipped (was 0 executed — the package was not installed) |
| `pytest packages/nemo_evaluator_sdk/tests/agent_eval` | 724 passed, 10 skipped |
| `tools/lint/lint-all.sh` | 16/17 pass; `lint-openapi` needs bash 4+ (`mapfile`) and cannot run on macOS — its assertion verified by hand |
| `ty check` | All checks passed |
| `uv lock --check` | in sync; no existing package moves |
| Live Docker run | broker on a real socket → container → NeMo-Gym 0.5.0 → `mcqa` → 1 COMPLETED trial |

`uv run pre-commit run -a` was not run to completion locally: the UI hook needs a `pnpm` shim this
machine lacks. The individual gates it wraps are listed above and CI is green.

## Open questions

1. **`BrokerSandboxProvider` has no consumer.** It implements the AALGO-321 seam against the broker,
   built when we expected a harness to run *inside* an episode sandbox. The Gym path went the other
   way. Adopt it for the Fabric harness path, or drop it before merge?
2. **`docker/gym-host/Dockerfile` duplicates `nmp-gym-tasks`** (#1315, landed while this was in
   flight). Mine runs the host runtime rather than the CLI, but it could build *from* that image
   instead of installing NeMo-Gym again.
3. **@soluwalana** — the vendoring and the NeMo-Gym removal reshape your package; and
   `values/opensandbox` now exists on both this branch and #1156, already diverged. It probably
   belongs on #1156.

## Summary by CodeRabbit

* **New Features**
  * Added sandboxed Gym evaluation with configurable Docker and OpenSandbox hosting.
  * Added secure episode brokering for execution, file transfers, lifecycle management, and egress controls.
  * Added Gym agent routing and secret-backed environment variables.
  * Added CLI serving modes, session descriptors, health checks, and rollout endpoints.
  * Added offline model stub and deployment configurations for supported sandbox runtimes.
  * Added an option to preserve an image’s configured entrypoint during agent deployment.
* **Documentation**
  * Added package documentation, setup guidance, serving examples, and verification scripts.
* **Tests**
  * Added broad unit, integration, security, and end-to-end coverage.


packages/sandboxed_gym/src/sandboxed_gym/broker.py
```diff
@@ -0,0 +1,160 @@
...
+        self._server = None
+        self._loop: asyncio.AbstractEventLoop | None = None
+        self._thread: threading.Thread | None = None
+        self._socket: socket.socket | None = None
+
+    def _reserve_socket(self) -> tuple[socket.socket, int]:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if self._config.port is not None:
+            sock.bind(("", self._config.port))
```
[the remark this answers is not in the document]
SandyChapman: Intentional, and required for the feature to work. The broker exists to be reached from *inside* the episode sandbox, across a network boundary — `_resolve_advertise` publishes a cluster Service DNS name or the node IP, and the uvicorn server two lines down binds `0.0.0.0` for the same reason. Loopback would make the broker unreachable by the only caller it has.

The exposure is guarded rather than left open: `start()` mints a fresh `secrets.token_urlsafe(32)` per broker, and every one of the six routes in `build_broker_app` carries `dependencies=[auth]`, compared with `hmac.compare_digest` and fronted by a request-body size limit. Policy (`approved_images`, TTL and concurrency caps) is trusted-side config the job cannot influence.

Proposing this be dismissed as "used in tests / won't fix".

packages/sandboxed_gym/src/sandboxed_gym/netutil.py
```diff
@@ -0,0 +1,41 @@
...
+    sock: socket.socket,
+    port_range_low: int,
+    port_range_high: int,
+    max_retries: int = 50,
+) -> int:
+    """Bind ``sock`` to a random free port in ``[low, high)``."""
+    for _ in range(max_retries):
+        port = random.randint(port_range_low, port_range_high - 1)
+        try:
+            sock.bind(("", port))
```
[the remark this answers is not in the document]
SandyChapman: Same finding as broker.py:54 — `bind_socket_in_range` is the port-search helper the broker calls at broker.py:57, so it binds on the broker's behalf and inherits both its reachability requirement and its token guard. Rationale and dismissal proposal are in the broker.py thread.

packages/sandboxed_gym/src/sandboxed_gym/runtime/gym_host_runtime.py
```diff
@@ -0,0 +1,284 @@
...
+        raise RuntimeError(f"{GYM_GLOBAL_CONFIG_ENV_KEY} must be a JSON object")
+    return parsed
+
+
+def _free_port_in_range(low: int, high: int) -> int:
+    for port in range(low, high + 1):
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            try:
+                sock.bind(("", port))
```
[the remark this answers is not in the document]
SandyChapman: Different context from the other two, and a weaker case than they are.

`_free_port_in_range` is a probe: it binds inside a `with` block purely to find a free port, then closes it. The durable listener is Gym's own head server, which `_allocate_head_server_port` configures with `host: "0.0.0.0"` so the sandboxed job can reach it — that surface belongs to Gym rather than to this package, and unlike the broker it carries no token guard; it is reachable only from the job's own sandbox network.

Proposing dismissal on the same grounds, but flagging the distinction so it is not waved through by association with the broker threads.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -344,6 +346,16 @@ def _resolve_target(
             )
             return fabric_runtime, None, None
         if isinstance(target, GymRunnerTarget):
+            # Sandboxing is a deployment decision, not a job field: the same target runs colocated
+            # on a trusted box and sandboxed on a shared cluster. The gates raise rather than fall
+            # back, so a cluster that cannot sandbox refuses the run instead of quietly running user
+            # environment code beside this job's credentials.
+            if should_sandbox(EvaluatorConfig(), target):
```
JashG: This runs in the Gym tasks container, so by calling `EvaluatorConfig()` I don't think platform-defined env vars would propagate to the container (i.e. `EvaluatorConfig` will fallback to its default values in the pydantic model).

Could compilation resolve and validate the trusted sandbox settings in the evaluator service, serialize them into an internal step configuration, and have `_resolve_target()` consume it?

packages/sandboxed_gym/src/sandboxed_gym/orchestrator.py
```diff
@@ -0,0 +1,280 @@
...
+        max_request_bytes=sandbox.max_request_bytes,
+        max_response_bytes=sandbox.max_response_bytes,
+        ttl_s=sandbox.ttl_s,
+        ready_timeout_s=sandbox.ready_timeout_s,
+        resources=sandbox.resources,
+        runtime_http_port=sandbox.runtime_http_port,
+        allow_internet=sandbox.allow_internet,
+        public_dns_allow=sandbox.network_policy.public_dns_allow,
+        resolver_addresses=sandbox.network_policy.resolver_addresses,
+        entrypoint=(tuple(sandbox.entrypoint) if sandbox.entrypoint else tuple(default_gym_host_entrypoint())),
```
JashG: If I'm understanding this right, `packaged_gym_host_script()` finds the installed script in the outer Evaluator job container and returns its absolute path - I think under `/app/.venv/lib/python3.13/site-packages`. That path is then placed in `GymHostSpec.entrypoint` and passed unchanged to OpenSandbox, so the Gym container receives a path that lives in the outer container.

Maybe we set `entrypoint` to `None` if `sandbox.entrypoint` is not defined, so we fallback to the image's default startup command?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def serve_config(target: GymRunnerTarget, config: EvaluatorConfig, *, job_id: str) -> dict[str, Any]:
+    """Assemble the ``SandboxedGymServeConfig`` payload for one evaluation.
+
+    Job-derived and deployment-derived settings meet here and nowhere else: the target supplies the
+    environment selection, the plugin config supplies the cluster facts, and the broker token and
+    rollout URL are minted by the session itself.
+    """
+    return {
```
JashG, line 252: This payload doesn't forward `target.env_vars` or the resolved `target.env_secrets `to the Gym host container, so sandboxed evals might lose required credentials. I think we'd want to pass those into `boostrap_env`?

I think we need add an explicit Gym host environment field in this payload, populated with those
requested variables, that gets merged into `GymHostSpec.bootstrap_env` in `build_gym_host_spec()`.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
```diff
@@ -0,0 +1,266 @@
...
+        "job_id": job_id,
+        "host_provider": config.sandbox_host_provider,
+        "sandbox": {
+            "image": config.sandbox_runtime_image,
+            # One claim, two sub-paths. The environment mount is read-only and the workspace is not,
+            # so they must not resolve to the same directory.
+            "environment_pvc_claim": config.sandbox_job_storage_pvc_claim,
+            "environment_sub_path": config.sandbox_environment_sub_path,
+            "workspace_pvc_claim": config.sandbox_job_storage_pvc_claim,
+            "workspace_sub_path": config.sandbox_workspace_sub_path,
```
JashG: It looks like every run uses the same deployment-wide `workspace_sub_path`. i.e. writable state across concurrent eval runs would write to the same folder. Should these be job-scoped paths, such as `workspace/<job_id>/...`?

SandyChapman: yes, I think you're right. I'll address this

packages/sandboxed_gym/src/sandboxed_gym/orchestrator.py
SandyChapman: will take a look at this. good catch.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: thanks, will take a look at this.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
SandyChapman: Credential-shaped env_vars are refused on purpose — the sandbox runs untrusted environment code. But you're right that the rest get dropped silently, which is the wrong failure mode. I'll fix it.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Fixed. The compiler now resolves a `SandboxPlan` service-side and passes it in the step environment; `_resolve_target` reads that instead of constructing `EvaluatorConfig`. The capability gates run at submit time now rather than mid-run.

packages/sandboxed_gym/src/sandboxed_gym/orchestrator.py
SandyChapman: Fixed. `entrypoint` is `None` when unset, so the runtime image starts itself from its own `CMD`. The image-relative fallback could never fire anyway, since the packaged file always exists where the orchestrator runs.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
SandyChapman: Fixed. Both now travel in `SandboxedGymServeConfig.host_env` and merge into `bootstrap_env`, and an unresolved `env_secrets` entry fails by name. Credential-shaped `env_vars` stay refused, since the sandbox runs the environment's own code.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
SandyChapman: Fixed, it's `{sub_path}/{job_id}` now. The environment mount stays deployment-wide since it's read-only.

JashG (approved):

PR #1403, opened 2026-08-19
feat(agents): adding deployment-scoped session entity

mmogallapalli:
## Summary

Introduces persisted, deployment-scoped `AgentSession` entities and workspace-level APIs for multi-turn conversations with deployed Fabric-backed agents. Invocations can now supply a persisted session ID so the gateway can validate its deployment binding and the serving shim can lazily create, reuse, and clean up the corresponding Fabric runtime.

## Related Issue

Closes [AIRCORE-1035](https://linear.app/nvidia/issue/AIRCORE-1035/create-deployment-scoped-agentsession-entity-and-api).

## Changes

- Add the persisted `AgentSession` entity with a required deployment ID, `active` and `closed` lifecycle states, and optional activity and expiration timestamps.
- Register sessions with the entity framework and add the corresponding request, response, OpenAPI, authorization, and SDK schemas.
- Add workspace-level APIs to create, list, get, close, and delete sessions, including deployment existence and workspace validation.
- Add optimistic version checking when deleting sessions to avoid deleting a concurrently replaced entity.
- Update the deployed-agent gateway to accept an optional Platform session ID and validate that the session belongs to the deployment being invoked.
- When invoking by agent name with a session ID, route through the gateway to the exact deployment bound to that session; preserve existing deployment selection when no session ID is supplied.
- Use the Platform session ID as the Fabric serving shim's logical runtime key, lazily creating a runtime on first use and reusing it for later turns.
- Add best-effort Fabric runtime cleanup when a persisted session is closed or deleted.
- Add entity, API, authorization, gateway, SDK, and Fabric session-management tests.
- Keep the initial model deployment-scoped without user ownership or job-session support.
## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

## Summary by CodeRabbit

- **New Features**
  - Added agent session management to create, list, retrieve, close, and delete sessions.
  - Sessions track deployment association, lifecycle status, and activity timestamps.
  - Added filtering, pagination, authorization, and concurrency safeguards.
  - SDK invocations can include a session ID for persistent routing.

- **Bug Fixes**
  - Improved validation and error handling for unavailable, mismatched, or closed sessions.
  - Added reliable cleanup of associated runtimes when sessions close or are deleted.
  - Improved deployment routing to consistently honor active session associations.


plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
```diff
@@ -97,10 +106,31 @@ async def open_session(self) -> FabricRuntimeSession:
             raise
 
     async def resolve_session(self, session_id: str | None) -> FabricRuntimeSession:
-        """Open a new session or resolve an existing session by its opaque ID."""
+        """Resolve a session, lazily starting a runtime under a supplied Platform ID."""
         if session_id is None:
             return await self.open_session()
-        return await self._session_registry.get(session_id)
+        if session_id in self._closed_session_ids:
+            raise FabricSessionNotFoundError(f"Fabric session '{session_id}' was not found.")
```
mikeknep, line 113: Is there a reason we're storing **closed** session IDs and not **open** ones? I would think that the open sessions are:
- more knowable (we only know sessions that were closed within the lifetime of this deployment, not sessions "orphaned" by a restart)
- smaller (# of closed sessions >= # of open sessions)

mikeknep: I guess storing closed sessions means you don't need to pop open session IDs out of a set when they close... 🤔

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -29,6 +29,11 @@
 
     # Invocation (routes through the agents gateway)
     result = nemo.agents.invoke(agent="calculator", input="What is 2+2?")
+    result = nemo.agents.invoke(
```
mikeknep, line 32: Oof, not a problem for this PR specifically, but are we overloading this `invoke` verb/command? 😬

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
```diff
@@ -0,0 +1,341 @@
...
+    permissions=[SessionPerms.CLOSE],
+)
+async def close_session(
+    workspace: str,
+    name: str,
+    entity_client: NemoEntitiesClient = Depends(get_entity_client),
+) -> AgentSession:
+    """Close a session. Closing an already-closed session is idempotent."""
+    try:
+        session = await entity_client.get(AgentSession, name=name, workspace=workspace)
```
mikeknep, line 169: Do we need this initial `get` call? It introduces a race condition, e.g. two concurrent requests to close the same session. Seems like we could probably just make the update call regardless?

Oh... unless we need the full session object to make that call. Ugh. I guess our update calls are `PUT` style, not `PATCH` style? Hmm, leaving this comment here, but this might be unavoidable.

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
```diff
@@ -0,0 +1,341 @@
...
+        )
+    except NemoEntityNotFoundError as exc:
+        raise HTTPException(
+            status_code=404,
+            detail=f"Session '{name}' not found in workspace '{workspace}'.",
+        ) from exc
+    except NemoEntityConflictError as exc:
+        raise HTTPException(
+            status_code=409,
+            detail=f"Session '{name}' is being modified concurrently.",
```
mikeknep, line 257: Hmm. I'm not sure how this would happen in practice 🤔

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/gateway.py
```diff
@@ -324,7 +436,13 @@ async def _proxy(
         target_url = f"{target_url}?{request.url.query}"
 
     # Build forwarded headers — strip hop-by-hop and platform-internal headers
-    headers = {k: v for k, v in request.headers.items() if k.lower() not in _HOP_BY_HOP}
+    headers = {
+        k: v
+        for k, v in request.headers.items()
+        if k.lower() not in _HOP_BY_HOP and k.lower() != SESSION_ID_HEADER.lower()
```
mikeknep: I was curious what the `_HOP_BY_HOP` headers are, and it looks like that constant already has not just HBH but also some other platform internal headers. Maybe we should just add `SESSION_ID_HEADER` to that existing constant? Also maybe just rename the constant? Though I do see it used in one other place, in `_stream_with_headers`... not sure if that needs different logic than this `_proxy` method?

mmogallapalli: Good call, I just broke up the constants to separate both hop_by_hop and platform internals and then created two explicit constants for request and response headers we want to strip. I didn't want to add `SESSION_ID_HEADER` directly since we want to keep it on the response side.

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
mmogallapalli: This was added to handle concurrent delete and close requests. In this case we need to read the session to obtain its `db_version`. If another request closes/updates the sessions before the delete executes then we will hit a 409. Two concurrent deletes will hit a 404 as expected.

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
mmogallapalli: Good catch, yeah I think the initial get is unavoidable for the reason you said
I added a re-fetch block if a `NemoEntityConflictError` is hit such that we can return that "earlier" close success, but default to the existing 409 behavior if another operation is the reason for the error.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mmogallapalli: We store open sessions in `SessionRegistry._sessions`. This set is track the closed ("tombstoned") sessions for the current deployment process, such that we don't accidentally recreate a runtime for a late or retry request after a session has been closed.

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/gateway.py
mikeknep: I like the separate, explicit REQ and RESP header constants 👍

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
mikeknep: Ah I see, I didn't notice the `expected_db_version` here. 👍

plugins/nemo-agents/src/nemo_agents_plugin/api/v2/sessions.py
mikeknep: Looks like update will also automatically send the `expected_db_version`. Neat. And the re-fetch addition makes sense too. Neat as well. Thanks!

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
mikeknep: As discussed on Slack: yes we are, and we need to fix it. Not a blocker here

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mikeknep: Gotcha, makes sense!

mikeknep (approved):

PR #1409, opened 2026-08-19
fix(agents): stage agent spec fileset in subprocess deployments

marcusds:
## Summary

Subprocess deployments never read the `{agent}-spec` fileset, so an agent's skills, MCP servers, and prompts were missing in the default deploy mode. Before this change `_create_fabric_deployment` created an empty base directory and wrote only `agent.yaml` into it, while container deployments already staged the fileset; after it, a subprocess deployment receives the same artifacts a container deployment does.

## Related Issue

Tracked in Linear as ASTD-449. There is no corresponding GitHub issue.

## Changes

- Thread the agent name through `create_deployment` into `_create_fabric_deployment` — it was previously discarded by `del agent, ...` — and download the `{agent}-spec` fileset into the deployment base directory using the same platform SDK files client the container path uses.
- Stage before writing the resolved config, so the injected gateway URL still wins over any `agent.yaml` carried in the fileset, matching the substitution the container path performs.
- Keep an unavailable fileset non-fatal, matching the container fallback, so existing config-only agents continue to deploy. A config referencing `skills.paths` that were not staged still fails the deployment.
- Add `stage_fabric_spec_dir` alongside the existing `stage_fabric_spec_config_files`, and generalize the skill-path check into `validate_referenced_skill_paths`, which now takes paths relative to the agent base directory so container `config_files` and an on-disk base directory share one definition.
- Do not apply the 900 KB staged-byte cap on this path: it bounds ConfigMap and env delivery for container deployments, neither of which is involved in a subprocess deployment.
- Clear previously staged files before downloading. The base directory is deterministic and reused: when the controller cannot find a running process it returns the deployment to `pending` (`controller.py:335`) and the next reconcile re-creates it over the same directory, so a file dropped from the fileset would otherwise survive locally, and a stale skill tree could satisfy validation for a deployment that staged nothing. Directories named by `environment.workspace` and `environment.artifacts` are preserved so a restart does not discard a prior run's telemetry.
- Match `skills.paths` entries on path components rather than string prefixes. `PurePosixPath(".")` stringifies to `"."` and no staged path starts with `"./"`, so the shared validator rejected `skills.paths: ["."]` even though packaging resolves it to the agent root and accepts it when `SKILL.md` is present (`container/fabric_validator.py:90`). Component matching also stops `skills/review` from matching a sibling named `skills/review-notes`.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing interface change; `nemo agents deploy --mode subprocess` behaves as its existing documentation already describes.

## Verification

Targeted validation:

- `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_fabric_artifact_staging.py` — 20 passed. Reverting only the source file fails 4 of the 6 added cases, including the one asserting that a stale skill tree no longer satisfies validation when the fileset is unavailable.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_runner_in_memory.py` — 24 passed.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit` — 1004 passed, 3 failed. All three failures reproduce on an unmodified tree: two `test_port_allocation` cases cannot `bind()` under the local sandbox, and `test_cli_list_output.py::TestDeploymentsListOutput::test_deployments_list_defaults_to_table` asserts `"12345" not in output` against a fixture timestamp containing `T20:01:00.123456`.
- `uv run ruff check plugins/nemo-agents/` — passed; `uv run ruff format --check` on the changed files — passed.
- `uv run --frozen ty check plugins/nemo-agents/src/nemo_agents_plugin/runner/` — passed.
- `uv run pre-commit run -a` — every hook passed except two blocked by the local environment and unrelated to these files: `Helm Docs` (helm-docs not installed locally) and `Run uv lock with platform uv` (requires uv 0.9.14, local is 0.9.30). `Check for uv.lock drift` passed and no hook modified the tree. The checklist item above is left unchecked because the full sweep did not run clean locally; CI covers both hooks.
- End to end against a local platform: created an agent from `plugins/nemo-agents/examples/nemo-agent-config/calculator-agent/agent-with-mcp.yaml`, deployed it with `--mode subprocess`, and inspected the deployment base directory. With this change the directory contains `mcps/calculator.py` plus the agent's other spec files; with the change stashed, the same deployment produced `agent.yaml` alone. The deployed `agent.yaml` carried the injected gateway `base_url`, confirming the resolved config overwrote the fileset copy.

Pre-existing and out of scope: invoking that MCP example returns a 503 from `nemo_fabric_adapters/common/lifecycle.py` during adapter start, before any staged file is read. It reproduces identically with this change stashed, and a plain non-MCP calculator agent invokes successfully on the same machine, so the deepagents harness itself is working.

## Summary by CodeRabbit

* **New Features**
  * Fabric-based agent deployments now stage required specifications and related artifacts locally before launch.
  * Deployments support MCP artifacts, agent configuration files, optional filesets, and both subprocess and container execution.
  * Referenced skills are validated against staged files, including root-level skill paths.

* **Bug Fixes**
  * Staging removes stale files while preserving runtime directories and cleans up incomplete deployments after failures.
  * Improved path and symlink validation prevents access outside the agent’s permitted directory.
  * Internal specification files are excluded from the staged artifact set.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/runner/fabric_artifact_staging.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/runner/fabric_artifact_staging.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+import posixpath
 import shutil
 import tempfile
 from collections.abc import Collection
@@ -98,25 +99,35 @@ def _runtime_dir_names(agent_config: dict[str, Any]) -> set[str]:
         value = environment.get(key)
         if not isinstance(value, str) or not value:
             continue
-        rel = PurePosixPath(value)
+        rel = PurePosixPath(posixpath.normpath(value))
         if rel.is_absolute():
             continue
-        parts = [part for part in rel.parts if part != ".."]
-        if parts and parts[0] != ".":
-            names.add(parts[0])
+        # normpath resolves "foo/../workspace" to "workspace"; a leading ".." escapes base_dir.
+        if rel.parts and rel.parts[0] != "..":
+            names.add(rel.parts[0])
     return names
 
 
 def _clear_staged_tree(base_dir: Path, preserved: set[str]) -> None:
+    """Remove previously staged content, leaving *preserved* runtime directories.
+
+    Deletion failures are fatal: a surviving skill tree would otherwise satisfy
+    validation for a deployment that staged nothing.
+    """
     if not base_dir.is_dir():
         return
     for child in base_dir.iterdir():
         if child.name in preserved:
             continue
-        if child.is_dir() and not child.is_symlink():
-            shutil.rmtree(child, ignore_errors=True)
-        else:
-            child.unlink(missing_ok=True)
+        try:
+            if child.is_dir() and not child.is_symlink():
+                shutil.rmtree(child)
+            else:
+                child.unlink(missing_ok=True)
+        except OSError as exc:
+            raise FabricArtifactStagingError(
+                f"Failed to remove stale agent spec artifact {child.name!r} from {base_dir}: {exc}"
+            ) from exc
 
 
 def _staged_relative_paths(base_dir: Path) -> set[PurePosixPath]:
@@ -152,7 +163,10 @@ async def stage_fabric_spec_config_files(
                 agent_yaml_path=agent_yaml_path,
                 rewritten_agent_yaml=config_yaml,
             )
-            _validate_referenced_skill_paths(rewritten_agent_config, config_files, agent_yaml_path)
+            # Validate against what the fileset delivered, not config_files, which
+            # carries the generated agent.yaml whether or not the fileset supplied one.
+            delivered = {path for path in _staged_relative_paths(tmp_path) if path.name != AGENT_SPEC_FILENAME}
+            validate_referenced_skill_paths(rewritten_agent_config, delivered)
             _validate_staged_size(config_files, fileset_name)
             return config_files
     except (FileNotFoundError, PlatformNotFoundError, PluginClientNotFoundError) as exc:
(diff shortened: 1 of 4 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/runner/fabric_artifact_staging.py
```diff
@@ -38,6 +40,89 @@ async def download(
...
+    _clear_staged_tree(base_dir, _runtime_dir_names(agent_config))
+
+    if not agent_name or sdk is None:
+        validate_referenced_skill_paths(agent_config, _staged_relative_paths(base_dir))
+        return
+
+    fileset_name = agent_spec_fileset_name(agent_name)
+    try:
+        await sdk.download(local_path=str(base_dir), fileset=fileset_name, workspace=workspace)
+    except (FileNotFoundError, PlatformNotFoundError, PluginClientNotFoundError) as exc:
```
mckornfield: any reason we shouldn't just catch any Exception?

plugins/nemo-agents/src/nemo_agents_plugin/runner/fabric_artifact_staging.py
```diff
@@ -38,6 +42,136 @@ async def download(
...
+    names: set[str] = set()
+    for key in ("workspace", "artifacts"):
+        value = environment.get(key)
+        if not isinstance(value, str) or not value:
+            continue
+        rel = PurePosixPath(posixpath.normpath(value))
+        if rel.is_absolute():
+            continue
+        # normpath resolves "foo/../workspace" to "workspace"; a leading ".." escapes base_dir.
+        if rel.parts and rel.parts[0] != "..":
```
mckornfield, line 113: `..` always worries me, sec wise lol

mckornfield (approved):

plugins/nemo-agents/src/nemo_agents_plugin/runner/fabric_artifact_staging.py
marcusds: I don't think we want to catch things like auth errors etc here.

PR #1411, opened 2026-08-19
feat(eval-author)!: replace the CLI with skills for Harbor eval discovery

aleckhoury:
## Summary

Harbor tasks live in the customer's repository, so a tool that proposes changes to an eval suite has to write to that repository, and customers would not grant that however it was sandboxed. This replaces the `nemo agents eval-author` CLI with skills the customer's own coding agent reads and follows, so the work happens under their agent's existing permissions and nothing gets installed. Before, establishing whether a repository's Harbor evaluations run required `nemo agents eval-author discover`, which needs the platform installed and a workspace resolved; now a copyable skill directory does it against a local checkout.

**This is a breaking change.** The `nemo agents eval-author` command group is gone, along with its four placeholder verbs. `nemo agents` now lists only `analyst` and `experimentalist`. The `nemo-eval-author-plugin` distribution is also gone: the directory ships skills and a test, so it no longer builds a package at all.

### Added

Two skills under `plugins/nemo-eval-author/skills/`, in a core-plus-sub-flow shape:

- **`eval-author`** (core): owns the standard that governs every sub-flow — a provider's own validators judge each recorded fact, rather than the agent inferring it from file layout. Also owns the shared vocabulary (check, required, advisory, rung, proven, provider), the boundaries, and the routing table. Granted `[Read, Grep, Glob]` only, because it routes and explains rather than executing.
- **`eval-author-discover`** (sub-flow): the discovery steps, deferring the standard to the core rather than restating it. Runs three phases in one invocation: probe for Harbor, inventory the repository with the standard library, then run Harbor's eight-rung validation ladder in-process when Harbor is importable. Ends by saving a report to `.eval-author/discovery.md`, so findings outlive the run.

Bundled scripts (`eval-author-discover/scripts/`):

- `discover.py` — entry point. Owns phase order, report assembly, and the exit code; nothing provider-specific. Prints JSON and writes no files.
- `_checks.py` — the check-result contract. Plain `dataclass` and plain strings, so it carries no dependency and its JSON needs no conversion step.
- `providers/harbor/_probe.py` — Harbor capability detection via `find_spec`, so a missing Harbor costs no import.
- `providers/harbor/_inventory.py` — finds job configs, datasets, and task directories. Standard library, plus PyYAML when present and a regex fallback when not.
- `providers/harbor/_ladder.py` — runs Harbor's validators. Imported only after the probe reports Harbor available.

### Removed

The CLI is removed in full, not deprecated, because the skill supersedes it and a command group whose only working verb is replaced has nothing left to offer:

- `cli.py` and the `nemo.cli.agents` entry point, including the `audit`, `propose`, `run`, and `doctor` verbs, which were placeholders that exited nonzero.
- The `discovery/` package (`run.py`, `scan.py`, `validate.py`, `report.py`) that implemented `discover`.
- Their tests: `tests/discover/` (4 files), `test_cli.py`, `harbor_fixtures.py`, and `conftest.py`, which existed to configure `litellm` for the agent tests that moved to Experimentalist in #1413.
- `.env.example`, which documented model variables for that same agent.

### The directory stops being a Python package

The `src/<module>/skills/` layout exists so a plugin's skills can be imported and shipped through the `nemo.skills` entry point, which Eval Author deliberately does not use. Once the agent code moved to Experimentalist, that tree held one `py.typed` marker for a package with no modules, so the skills moved to the plugin root and the directory now declares `[tool.uv] package = false` and builds nothing.

**Dependencies go from seven to zero.** Being a non-package also removes it from the `experimentalist` dependency group, from `[tool.uv.sources]`, and from two pieces of config that were already dead: a `ty` source path for a directory that no longer exists, and an `empty-body` override scoped to nooa agent classes that left with the agent. `uv.lock` records it as `source = { virtual = ... }`.

The contract test held the last coupling: it imported `nemo_insights_plugin` for a drift guard comparing the bundled check contract against the platform's. That guard existed so a skill report would read like one from the Eval Author CLI, which this PR deletes, so it went too. The five tests that make Harbor judge a fixture suite now skip when Harbor is absent instead of failing, which retires the `tests/discovery_exclusions.py` entry entirely: the suite runs on `pytest` and `pyyaml` alone, matching the boundary the bundled scripts already hold.

### Design decisions worth a reviewer's attention

- **The report is saved by the agent, not by a script.** The CLI uploaded its `discovery.md` to a fileset, which these skills cannot do and should not: they talk to no platform service. Rather than move that plumbing into the script, the skill states where the report goes (`.eval-author/discovery.md`, leading with the JSON as front matter) and lets the agent write it. Where a file belongs in someone's repository is a judgement, so it stays in the open where the user can see it. The report is left visible and uncommitted — worth committing so a teammate skips the discovery pass, but that is the user's call, and their `.gitignore` is not ours to edit.
- **The sub-flow trades a blanket no-writes grant for `Write` without `Edit`.** It creates its own report and never rewrites a file that predates it, which is the permission customers actually declined. The core stays read-only, since it only routes. A test enforces both halves.
- **Harbor is a required capability, not a declared dependency.** It is the only import beyond the standard library, it is reached only after the probe, and a repository holding Harbor evaluations has Harbor by construction. Without it the script still emits an inventory, so nothing crashes; it just proves nothing.
- **Validation is delegated, never reimplemented.** `harbor job start --print-config` exits 0 on a config naming a nonexistent dataset path and a nonexistent agent, so only 2 of the 8 rungs are reachable from the CLI. The ladder therefore runs in-process, where the same config produces `resolution` and `agent` failures.
- **Without Harbor, nothing is proven.** The report still comes back so an agent can orient in an unfamiliar repository, but every finding carries `"proven": false` and the exit code is 1.
- **Provider code sits under `scripts/providers/harbor/`, not `scripts/harbor/`.** A directory named `harbor` on `sys.path` is importable as a namespace package, which makes `find_spec("harbor")` succeed on a machine with no Harbor and the probe claim an install that is not there. A test guards this.
- **One line in the vendored wrapper is edited by hand.** `make vendor` left the `eval-author` entry point in `packages/nemo_platform/pyproject.toml` pointing at the deleted `cli` module, and dropped that table's "Generated" marker comment. `nemo agents` skips entry points that fail to import, so this would have shipped as invisible dead metadata rather than a visible break. Removing the line lets vendor reclaim the table, restore the marker, and stop rewriting it; I confirmed vendor does not re-add it.

Tests: `plugins/nemo-eval-author/tests/test_skill_contract.py`, 23 cases covering frontmatter completeness against `docs/contributing/skills-spec.mdx`, the core/sub-flow routing contract, the dependency boundary (an AST walk that fails on any import outside the standard library, a sibling, or Harbor's own dependencies), the lazy-ladder-import guarantee, the provider name-collision guard, and discovery behavior on valid, broken, and Harbor-free repositories.

## Type of Change

- [ ] Documentation only

## Quality Gates

Documentation touched, because removing a command group makes four passages wrong: `docs/agents/insight-driven-optimization.mdx` described the `nemo agents eval-author` namespace, listed it in a setup verification snippet, named it in the command reference intro, and gave it its own reference section. The plugin README is rewritten around the skills, the `plugins/README.md` uninstall table no longer lists a package that cannot be installed, and the Experimentalist plugin's README, `AGENTS.md`, and `eval_author/README.md` no longer point at a CLI that exists.

## Verification

Targeted validation:

| Command | Result |
|---|---|
| `uv run --frozen pytest plugins/nemo-eval-author/tests -q` | 23 passed |
| `uv run --isolated --no-project --with pytest --with pyyaml pytest plugins/nemo-eval-author/tests -q` | 18 passed, 5 skipped |
| `uv run --frozen pytest plugins/nemo-experimentalist/tests -q` | 995 passed, 44 skipped |
| `uv run ruff check plugins/nemo-eval-author tests/discovery_exclusions.py` | All checks passed |
| `uv run ruff format --check plugins/nemo-eval-author tests/discovery_exclusions.py` | 7 files already formatted |
| `uv run --frozen ty check` | Passes; no diagnostic names a file this PR touches |
| `uv run pre-commit run -a` | Every hook passes except `Helm Docs`, see below |
| `uv sync --frozen --all-packages` | Succeeds, and uninstalls the stale `nemo-eval-author-plugin` editable |
| `make vendor` | No changes, confirming the plugin is out of the bundling path |
| DCO audit over `origin/main..HEAD` | 7 of 7 commits carry a matching `Signed-off-by:` |

The isolated run is the load-bearing one: it installs nothing but `pytest` and `pyyaml`, so it proves the tests hold the same boundary as the skills they guard. The five skips are the Harbor-judging cases.

`Helm Docs` fails, and it is not from this PR. The hook regenerates `k8s/helm/README.md` into a state that differs from what `main` has committed, then `Fix copyright headers` re-adds a header the regeneration dropped, leaving a duplicate. `git diff origin/main HEAD -- k8s/` is empty for this branch, so no commit here touches that tree.

Verification specific to removing the command group, since the failure modes here are quiet rather than loud:

- `uv run --frozen nemo agents --help` lists only `analyst` and `experimentalist`. The eval-author group is gone from the CLI surface.
- Root test discovery still collects the plugin's tests. Previously `tests/discovery_exclusions.py` gated them on `find_spec("nemo_eval_author_plugin")`, which cannot work now that nothing is importable; the skip markers make the gate unnecessary, so the tests are collected in every environment and degrade on their own.
- `make vendor` is idempotent: a second run leaves the tree clean, which is the condition `lint-sdk-vendored` checks.

Behavioral verification of the skill itself, run against a scratch repository holding one valid Harbor task and one job config:

- With Harbor importable: exit 0, `proven: true`, `runnable: true`, all eight rungs pass, and the report returns `cd <repo> && harbor job start -c harbor-job.yaml`.
- With Harbor unavailable (`python -S`, which drops site-packages): exit 1, `proven: false`, `harbor_importable: false`, and every finding other than the `harbor` check itself marked unproven.
- The two guard tests were confirmed non-vacuous by making them fail on purpose: creating a `scripts/harbor/` directory fails both the static collision test and the Harbor-free behavior test with the real crash it predicts (`ModuleNotFoundError: No module named 'harbor.agents'`), and removing the sub-flow's reference to the core fails the deferral test.

Base drift: `main` moved 18 commits ahead and the branch conflicted on `uv.lock` alone, with both `pyproject.toml` files auto-merging. Resolved by merging `origin/main` with `--signoff` and regenerating the lock from the merged inputs via `script/uv-lock.sh` rather than hand-editing it. `mergeable` is now `MERGEABLE`.

Known limitation, not addressed here: `docs/contributing/skills-spec.mdx` asks for a `tests.json` with four-mode routing tests per skill, and neither skill ships one. It matters more than usual because splitting one skill into two creates the routing ambiguity those tests exist to catch. No CI workflow currently references `skill-test.py` or `skill-cli-lint.py`, and 15 of the 19 existing plugin skills also lack `tests.json`, so this is consistent with the current state of the tree rather than a regression. Happy to add it here if a reviewer would rather it land together.


docs/agents/insight-driven-optimization.mdx
```diff
@@ -285,10 +285,9 @@ evaluation suite for validating candidates that address the Insight. The
 Experimenter invokes the Eval Author workflow in Insight mode and reads the
 `eval_author` section of the experiment configuration.
 
-The plugin also exposes the canonical `nemo agents eval-author` command
-namespace with `discover`, `audit`, `propose`, `run`, and `doctor` verbs. These
-standalone verbs are currently scaffolding and exit with a nonzero status until
-their implementations are available.
+The Eval Author has no command namespace. Work on the evaluation suites in your
+own repository runs through the `eval-author` skills, which your agent reads and
+follows. See the [Eval Author README](https://github.com/NVIDIA-NeMo/nemo-platform/tree/main/plugins/nemo-eval-author).
```
aleckhoury: Lets just remove this block for now

packages/nemo_platform/pyproject.toml
```diff
@@ -728,11 +716,10 @@ nemo-anonymizer-plugin = { source = "../../plugins/nemo-anonymizer/src/nemo_anon
...
 nemo-deployments-plugin = { source = "../../plugins/nemo-deployments/src/nemo_deployments_plugin", module = "nemo_deployments_plugin", inherit = { "entry-points" = ["nemo.*"] } }
-# Eval Author, Experimentalist, and Insights ship together: Experimentalist imports
-# EvalAuthor and EvalAuthorConfig at module scope, Eval Author still borrows
-# Experimentalist helpers, and both read the Insights profile contract. Bundling all
-# three into one distribution collapses that dependency cycle into intra-wheel extras.
-nemo-eval-author-plugin = { source = "../../plugins/nemo-eval-author/src/nemo_eval_author_plugin", module = "nemo_eval_author_plugin", inherit = { "entry-points" = ["nemo.*"] } }
+# Eval Author is deliberately absent. It ships skills and no code, declares no entry
+# points, and has nothing a distribution can expose, so bundling it into
+# `nemo-platform[all]` would ship files no mechanism can find. Add it back when the
+# skills register under `nemo.skills`.
```
aleckhoury: Just remove the whole block. No explanation needed

plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/eval_author/README.md
```diff
@@ -17,8 +17,9 @@ This package lives inside the Experimentalist plugin and uses its evaluator, sta
 and trace helpers directly. Experimentalist insight mode imports and runs Eval Author
 before optimization begins.
 
-The `nemo agents eval-author` command group lives in the separate
-[Eval Author plugin](../../../../nemo-eval-author/README.md).
+The customer-facing path is a skill rather than an agent, and the
+[Eval Author package](../../../../nemo-eval-author/README.md) holds those skills. It
+has no CLI and shares no code with this agent.
```
aleckhoury: Just remove this whole block

plugins/nemo-experimentalist/AGENTS.md
```diff
@@ -29,8 +29,9 @@ Experimentalist imports nothing from `nemo-eval-author-plugin`, so the package c
   ten borrows to zero, so it went away with them.
 - `tests/test_contract_dependency.py` asserts that this plugin never declares
   `nemo-eval-author-plugin` as a dependency, which is what keeps the cycle broken.
-- `plugins/nemo-eval-author/` keeps the `nemo agents eval-author` command group and its
-  `discovery/` package, whose one remaining borrow is `make_client`.
+- `plugins/nemo-eval-author/` ships the customer-facing skills and nothing else. Its CLI
+  and `discovery/` package went away with the pivot to skills, and with them the last
+  borrow from this plugin, so neither side imports the other now.
```
aleckhoury: Just remove this whole block

plugins/nemo-experimentalist/AGENTS.md
```diff
@@ -41,10 +42,10 @@ registered under the `nemo.cli.agents` entry-point group, which the `nemo-agents
...
 `nemo experimentalist` alias.
 
-Analyst and Eval Author follow the same rule: `nemo agents analyst run` (was
-`nemo insights analyze`) and `nemo agents eval-author <verb>`. Prefer
-`ctx.command_path` over a hardcoded path when a message quotes the command back
-to the user.
+The Analyst follows the same rule: `nemo agents analyst run` (was `nemo insights
+analyze`). Prefer `ctx.command_path` over a hardcoded path when a message quotes the
+command back to the user. Eval Author had a command group under this rule and no
+longer does; it ships skills instead.
```
aleckhoury: Just remove the Eval Author reference here entirely

plugins/nemo-experimentalist/README.md
```diff
@@ -78,9 +78,9 @@ single leader, so complementary strengths stay alive across rounds.
   analyze traces or host an Insight API.
 - [Eval Author](src/nemo_experimentalist_plugin/eval_author/README.md) builds the
   Insight-specific evaluation suite, and Insight mode invokes it automatically. It
-  ships inside this plugin; the separate
-  [Eval Author plugin](../nemo-eval-author/README.md) owns the
-  `nemo agents eval-author` commands.
+  ships inside this plugin. The separate
+  [Eval Author package](../nemo-eval-author/README.md) is the customer-facing path,
+  and it ships skills rather than an agent or a CLI.
```
aleckhoury: Just remove the second sentence here starting with "The separate...or a CLI"

BrianNewsom (approved):

PR #1446, opened 2026-08-21
chore(deps): add ipywidgets to the dev group

SandyChapman:
## Summary

Adds `ipywidgets` to the `dev` dependency group. Notebook examples that drive rich-based progress displays — Harbor's trial runner reaches one through the evaluator SDK — currently emit a `UserWarning` and fall back to static reprints. Dev-only, so nothing changes for consumers of the published packages.

## Changes

- `pyproject.toml`: add `ipywidgets>=8.1.9` to the `dev` group, alongside `papermill` and `ipykernel`, which are already there for the same reason
- `uv.lock`: regenerated with uv 0.9.14 (adds `ipywidgets`, `jupyterlab-widgets`, `widgetsnbextension`)

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Tests not applicable — justification: dependency-group addition with no code change; correctness is enforced by the lockfile drift check rather than by a test
- [x] Documentation not applicable — justification: dev-only dependency, no user-visible behavior

## Verification

Targeted validation:

- `uv lock` (uv 0.9.14) — resolved 572 packages; added `ipywidgets v8.1.9`, `jupyterlab-widgets v3.0.17`, `widgetsnbextension v4.0.16`
- `uv lock --check` (uv 0.9.14) — passed, no drift
- `uv run pre-commit run -a` — all hooks passed **except two that are blocked in my local environment, not by this change**:
  - **Helm Docs** — fails with `Please install helm-docs to run the pre-commit hook!`; the tool is not installed locally. This diff touches no `k8s/` or `helm/` files, so the hook has nothing of mine to check.
  - **Run uv lock with platform uv** — fails because the `uv` on PATH inside the hook environment is 0.9.30 while the repo requires 0.9.14. I generated and verified the lockfile with 0.9.14 explicitly (from the Flox environment). Notably **`Check for uv.lock drift` passed**, which is the check that would catch an actually-wrong lockfile.

  Leaving the box unchecked rather than claiming a pass — CI has both tools and will exercise them properly.

## Summary by CodeRabbit

* **Chores**
  * Added the `ipywidgets` package to the development environment.


mckornfield (approved):
🤔 wonder if we will introduce the unfriendly CVE bot, even though it is only for dev things, stupid scanners

PR #1451, opened 2026-08-21
fix(evaluator): publish the real ATIF trajectory to Intake

SandyChapman:
## Summary

Intake received **one synthetic step per trial**, so a trace showed no input and no output in Studio no matter which agent ran. Harbor's ATIF-capable agents already write an ATIF-v1.7 trajectory to `agent/trajectory.json`, and Intake ingests that same schema — so the trajectory now flows end to end. A codex run lands as ~10 spans carrying the task prompt, the agent's reasoning, its tool calls and their results, and the final answer.

> **Stacked on [#1450](https://github.com/NVIDIA-NeMo/nemo-platform/pull/1450)** (`fix(evaluator): let a producer declare its trace format`) — based on that branch, not `main`. Intake can only read the trajectory as ATIF once it's labelled correctly. Review #1450 first.

## Changes

- `intake/mapping.py`: `atif_steps_from_trial` reads the trial's trace evidence into ATIF steps
- `intake/publish.py`: sends the recorded steps, with a fallback (below)
- `agent_eval/runtimes/harbor_runtime.py`: `_final_agent_message` supplies the trial's `output_text`
- Vendored SDK mirror regenerated
- Tests for step mapping, the fallback, and final-answer extraction

### The fallback, and why it isn't an enumeration

The SDK's ATIF read model is deliberately **more permissive** than Intake's ingest schema — `step_id` and `tool_call_id` are optional there and required here — so a trajectory that parses locally can still be refused with a 422.

Rather than enumerating those divergences (and drifting out of sync with them as either side evolves), publish catches `UnprocessableEntityError`, retries without the recorded steps, and keeps the trial's scores. Losing a run's measurements over a detail of its trace is the wrong trade. The retry is logged at warning level so it isn't silent.

### Idempotency

`step_id` is normalised to the one-based sequence ingest validates, and a step with no timestamp takes the run's start. This matters because Intake's spans table is a `ReplacingMergeTree` keyed on `(workspace, session_id, start_time, id)` — unstable timestamps would make re-publishing duplicate rather than replace.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing API or CLI surface change; the user-visible effect is richer trace content in Studio for runs that already publish

## Verification

Targeted validation:

- `pytest plugins/nemo-evaluator/tests/intake packages/nemo_evaluator_sdk/tests/agent_eval/test_harbor_runtime.py` — **173 passed**
- `ruff check` / `ruff format --check` — pass
- **Mutation-checked** — each behaviour was broken on purpose and the matching test confirmed red:
  - send no recorded steps → `test_a_rejected_trajectory_falls_back_instead_of_failing_the_trial` fails
  - remove the `UnprocessableEntityError` fallback → same test fails
  - drop `_final_agent_message` → `test_atif_trajectory_supplies_the_final_answer` and `test_final_answer_ignores_trailing_non_agent_steps` fail
- **End to end against a live Harbor run**: a trial published 72 ATIF steps (versus the single synthetic step before), and the trace renders with real content in Studio.
- `uv run pre-commit run -a` — all hooks passed **except two blocked locally, not by this change**: **Helm Docs** (not installed; no `k8s/`/`helm/` files here) and **Run uv lock with platform uv** (PATH `uv` 0.9.30 vs required 0.9.14; no dependency change here, and `Check for uv.lock drift` passed).

## Known limitations (deliberately not addressed here)

Two Intake-side schema gaps surfaced while building this. Both are filed rather than worked around, because the producer side has the information and ATIF has nowhere to put it:

- **ASE-880** — ATIF `step_id` is validated at ingest then discarded, so Studio orders spans by `start_time`. Two steps sharing a timestamp render out of order. A publish-side timestamp nudge was implemented and then **reverted**: it writes a time the producer never recorded, into a store whose purpose is being the record of what happened, and only fixes traces published through this one path.
- **ASE-881** — ATIF has no field for terminal status or error, so a timed-out run is indistinguishable from a genuine reward-0.0 failure.

## Independent review

An independent model (Codex) was run over this branch. Two findings, both reproduced.

**Fixed — intermediate reasoning reported as the trial's answer.** `_final_agent_message` scanned backwards for the last *non-empty* agent message, so a trajectory ending on an empty agent step returned the one before it (`_final_agent_message -> 'working'` where the last agent message was `''`). Fixed in the follow-up commit; only the last agent step can be the answer, and an empty one means no answer. Mutation-checked — reverting the fix turns the new test red.

**Not fixed here — a mode-switching re-publish can leave two trace roots.** Verified as real, but it is an Intake-side design question rather than something this PR should decide.

If a trial is published once *with* recorded steps and later *without* them (or vice versa), the two publishes disagree about the root span's `start_time`:

- the root span ID is stable, derived from workspace + session only (`atif_mapping.py:217-224`)
- but the root's `start_time` comes from the trajectory's own steps (`_trajectory_started_at`, `atif_mapping.py:287-301`)
- and the spans table is `ORDER BY (workspace, session_id, start_time, id)` (`clickhouse_migrations.py:142`)

Same ID plus a different `start_time` is a different sort key, so the `ReplacingMergeTree` keeps **both** rows rather than replacing. Two roots for one session.

This requires a mode switch between publishes — the trace file disappearing, or a 422 that stops happening — so it is off the normal path. Flagged rather than fixed because the general fix is a decision about whether `start_time` belongs in that sort key, which is the same question ASE-880 raises about `step_id` being validated at ingest and then discarded. Better decided together than papered over here. Stable rich-to-rich and degraded-to-degraded re-publishes replace correctly.

**Cleared by that review:** step-ID renumbering across absent/duplicate/non-sequential source IDs; `None` timestamps (removed by `exclude_none=True` before `setdefault`); empty step arrays (the SDK parser requires at least one step); tool calls without `tool_call_id` (rejected by Intake, triggering the intended fallback); and that the 422 retry cannot double-publish through the current endpoint, since validation and mapping both run before the first persistence call (`spans/ingest/atif.py:114-138`).

## Summary by CodeRabbit

- **New Features**
  - Harbor evaluations now report the final agent message from available trajectories.
  - ATIF trial data can include complete recorded trajectories rather than only a final output.
- **Bug Fixes**
  - Trajectory processing now ignores trailing non-agent steps and preserves intentionally empty final messages.
  - Unsupported, malformed, or unavailable trajectory data is handled safely.
  - Publishing retries with a simplified trajectory when a complete trajectory is rejected, improving reliability.

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
@@ -29,13 +29,16 @@ Design constraints (see AALGO-289):
 
 from __future__ import annotations
 
+import logging
 import math
+from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Literal
+from typing import Literal, cast
 
 from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEvalTaskScore
 from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial
+from nemo_evaluator_sdk.values.evidence import EVIDENCE_FORMAT_ATIF, EVIDENCE_TRACE
 from nemo_platform.types.intake.evaluation_context_param import EvaluationContextParam
 from nemo_platform.types.intake.evaluator_result_create_params import EvaluatorResultCreateParams
 from nemo_platform.types.intake.evaluator_result_data_type import EvaluatorResultDataType
@@ -43,6 +46,9 @@ from nemo_platform.types.intake.ingest.atif_agent_param import AtifAgentParam
 from nemo_platform.types.intake.ingest.atif_create_params import AtifCreateParams
 from nemo_platform.types.intake.ingest.atif_final_metrics_param import AtifFinalMetricsParam
 from nemo_platform.types.intake.ingest.atif_step_agent_param import AtifStepAgentParam
+from nemo_platform.types.intake.ingest.atif_step_param import AtifStepParam
+
+logger = logging.getLogger(__name__)
 
 # --- Shared conventions -----------------------------------------------------
 
@@ -52,12 +58,6 @@ ATIF_SCHEMA_VERSION: Literal["ATIF-v1.7"] = "ATIF-v1.7"
 #: Default ``agent.version`` when the run target carries none. Neither Model nor
 #: Agent has a version field today, and ATIF requires one (design doc §3.9 #6).
 DEFAULT_AGENT_VERSION = "unknown"
-
-# Evidence-descriptor keys. ATIF is carried as a ``format`` on ``kind="trace"``,
-# *not* as a distinct ``kind``. These are string literals until D1 (AALGO-281)
-# promotes them to shared descriptor-key constants on the SDK evidence types.
-EVIDENCE_KIND_TRACE = "trace"
-TRACE_FORMAT_ATIF = "atif"
 
 
 def session_id_for(run_id: str, trial_id: str) -> str:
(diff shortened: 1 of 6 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
```diff
@@ -131,7 +167,7 @@ def trial_to_atif_ingest(
         "schema_version": ATIF_SCHEMA_VERSION,
         "session_id": session_id_for(run_id, trial.id),
         "agent": agent,
-        "steps": [step],
+        "steps": list(steps) if steps else [step],
```
SandyChapman, line 170: Prior to this change, we only had a single step in the trace for input/output.

JashG (approved):

PR #1464, opened 2026-08-21
feat(agents): package an agent container image from the platform [ASTD-459]

marcusds:
## Summary

Docker and k8s deployments need a container image, and only `nemo agents package` produced one. That command builds from a directory on the submitter's machine, so an agent created through Studio — whose source of truth is the `{agent}-spec` fileset — had no way to fill the image field. This adds `agents.package-agent`, a platform job serving `POST /apis/agents/v2/workspaces/{workspace}/jobs/package`, which stages that fileset into a temporary build context and runs the same Fabric build the CLI runs, returning the image tag to hand to `agents deploy --image`.

The build runs as a **host subprocess** rather than in a container, matching the four agents jobs that already exist: the Fabric Dockerfile uses BuildKit cache mounts, so it needs a real Docker CLI. Where no subprocess execution profile is registered — notably `runtime = kubernetes` — the submission is rejected at POST time instead of failing opaquely at schedule time.

## Related Issue

ASTD-459

## Changes

**`refactor(agents): decouple container packaging from typer`**

- Add `container/errors.py`: `AgentPackagingError` plus `ContainerToolingUnavailableError`, `AgentConfigValidationError`, `ManagedFileConflictError`, `ImageBuildError`, `ImagePublishError`.
- Replace `typer.Exit` / `typer.echo` in `container/builder.py` and `container/publisher.py` with those exceptions and an `on_progress` callback defaulting to the module logger. The CLI passes `typer.echo` and maps the exceptions back to exit code 1, so its observable behaviour is unchanged.
- Widen two `**kwargs` annotations in `test_container.py` that `ty` rejects when the file is checked on its own.

**`feat(agents): package an agent container image from the platform`**

- `jobs/package_agent.py` — `PackageAgentJob`. `to_spec` resolves the `Agent` entity at submit time so a missing or non-Fabric agent fails the POST; `compile` verifies the subprocess execution profile and emits a host-subprocess step; `run` stages the spec fileset, writes `agent.yaml` from the stored config, and builds.
- `tasks/package/__main__.py` — task entrypoint, passing both the sync and async task SDKs.
- Register the collection in `service.py` and under `nemo.jobs`; regenerate `plugins/nemo-agents/openapi/openapi.yaml` (submit, list, get, status, logs, results, download, cancel, delete).
- Add `[container]` to the workspace's `nemo-agents-plugin` dependency so `python-on-whales` is present in the runtime that executes the job.
- README section for packaging a platform-resident agent, with a limitations table.

### Scope

Fabric (`nemo-agents-spec-v1`) agents only — NAT workflows package from a source checkout and stay on the CLI, rejected with a clear message. The image is left in the host daemon; pushing to a registry from the platform is deliberately deferred, so k8s deployments still publish via `nemo agents package --publish` locally.

### Notes for reviewers

- **Job name is `package-agent`, REST path is `/jobs/package`.** Naming the job `package` makes the generated sub-group shadow the existing `nemo agents package` command on the same Typer app, taking the local packaging flags with it. `job_collection_path` pins the API path so the REST surface is unaffected.
- **`base_image_url` / `base_image_tag` / `python_version` / `uv_version` are validated with a strict grammar.** They are interpolated into the rendered Dockerfile unescaped (`ARG BASE_IMAGE_URL={{ base_image_url }}`, `autoescape=False`), and unlike the CLI the submitter here is a remote API caller; a value containing a newline appends build instructions the host daemon then executes. The patterns are enforced by pydantic at the request boundary and appear as `pattern` on the generated OpenAPI schema.
- **`compile` failures surface as 422, not 400.** The job framework maps `PlatformJobCompilationError` to 422 and `PlatformJobDependencyUnavailableError` to 503; there is no path to 400 from a compiler. This follows the same mechanism `nemo-evaluator` uses for its Harbor subprocess-profile check.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

| Command | Result |
|---|---|
| `uv run pytest plugins/nemo-agents/tests/unit` | 1181 passed, 2 failed — both `test_port_allocation.py`, which `bind()`s a socket and is blocked by the local sandbox. Verified passing (13/13) with the sandbox disabled; unrelated to this change. |
| `uv run pytest .../test_package_agent_job.py` | 28 passed (new) |
| `uv run pytest .../test_container.py` | 137 passed |
| `uv run pytest .../test_fabric_package_validation.py` | 26 passed |
| `uv run ruff check plugins/nemo-agents/` | All checks passed |
| `uv run ruff format --check plugins/nemo-agents/` | 202 files already formatted |
| `uv run --frozen ty check` (repo-wide, as CI runs it) | 684 diagnostics, all pre-existing; none in the files this PR adds or edits |
| `uv run --frozen script/generate-openapi-spec.sh` | Completed; `plugins/nemo-agents/openapi/openapi.yaml` regenerated and committed |
| `uvx uv@0.9.14 lock --check` | `Resolved 571 packages` — lock is current under the pinned uv |
| `uv sync --frozen --all-packages` | `+ python-on-whales==0.81.0`; `import python_on_whales` now succeeds in the workspace venv |
| `uv run nemo agents package --help` | Local packaging flags intact (regression check for the naming issue above) |
| `uv run nemo agents package-agent --help` | Job group present alongside it |

Blocked pre-commit hooks, both environmental and neither related to this change:

- **`uv-lock`** — requires exactly uv 0.9.14 on `PATH`; this machine has 0.9.30. Skipped for the commit and push. Its substantive requirement is satisfied: `uv.lock` was regenerated with `uvx uv@0.9.14 lock`, `uvx uv@0.9.14 lock --check` passes, and the `uv-lock-check` drift hook passes.
- **`helm-docs`** — the `helm-docs` binary is not installed locally. No Helm files are touched by this PR.

Not yet exercised: an end-to-end submit against a running platform with a Docker daemon. The build path itself is the code `nemo agents package` already uses in production; what is new and untested end-to-end is the fileset-staging step feeding it.

## Summary by CodeRabbit

* **New Features**
  * Added package-agent APIs for creating, monitoring, managing, and retrieving packaging jobs and results.
  * Added workspace-scoped container image naming and publishing through job results.
  * Added task and service support for package-agent jobs.
  * Added validation for image names, configuration, Dockerfiles, and managed files.
  * Added progress reporting and clearer packaging error messages.

* **Documentation**
  * Updated packaging and deployment guidance, including dependencies and `.dockerignore` requirements.


plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+_HOST_BUILD_REQUIREMENT = (
+    "Agent packaging builds an image with the host Docker CLI, so it requires a "
+    "registered subprocess execution profile. Build locally instead with "
+    "`nemo agents package --agent <agent.yaml>` and pass the resulting tag to "
+    "`nemo agents deploy --image`."
+)
+
+
+class PackageAgentInput(BaseModel):
+    """Submitter-facing shape: an agent name plus the packaging knobs."""
```
mckornfield: `submitter-facing shape` lol

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+            ],
+        )
+
+    @staticmethod
+    async def _require_subprocess_profile(profile: str, async_sdk: object) -> None:
+        """Reject the submission unless *profile* resolves to a host subprocess backend."""
+        if async_sdk is None:
+            raise PlatformJobDependencyUnavailableError(
+                f"Unable to resolve execution profile '{profile}': no platform client was injected. "
+                "Retry the submission."
```
mckornfield: this always says retry, but are these really retryable? for this and the below command

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+
+class PackageAgentJob(NemoJob):
+    """Build a container image for a platform-resident agent."""
+
+    # NOT "package": the generated job sub-group mounts onto the same Typer app
+    # that already owns `nemo agents package`, and the later registration wins —
+    # naming this "package" makes the local packaging flags unreachable.
+    # The REST collection path is pinned below so the API surface stays /jobs/package.
+    name: ClassVar[str] = "package-agent"
+    job_collection_path: ClassVar[str | None] = "/jobs/package"
```
mckornfield, line 148: interesting, is this a placeholder or a real path?

mckornfield (approved):

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def run(
+        self,
+        config: dict,
+        *,
+        ctx: JobContext | None = None,
+        async_sdk: AsyncNeMoPlatform | None = None,
+    ) -> dict:
+        """Stage the agent's spec fileset into a temp build context and build the image."""
+        del ctx
```
mmogallapalli: `run()` discards `ctx` and returns the image tag as a regular dictionary, but `run_task()` only logs that value and converts it into an exit code. It does not publish the value through the Platform job results API.

As a result, callers polling `/results` will not be able to retrieve the generated image tag described in the README. Could we write the result payload to `ctx.storage.ephemeral`, save it through `ctx.results.save(...)`, and return the resulting reference?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+        from nemo_agents_plugin.container.builder import build_fabric_agent_image
+
+        cfg = PackageAgentSpec.model_validate(config)
+        with tempfile.TemporaryDirectory(prefix="nemo-agent-package-") as tmp:
+            build_dir = Path(tmp)
+            asyncio.run(self._stage(cfg, build_dir, async_sdk))
```
mmogallapalli, line 300: the recursively uploaded fileset can contain a custom `.dockerignore`, and the builder intentionally preserves user-owned `.dockerignore` files. Validation checks the staged files on disk before Docker applies those exclusions.

For example, a `.dockerignore` containing `agent.yaml` or a referenced skill directory could pass validation, produce a successful build, and then fail when the image starts. For this server-managed path, should we ignore the incoming `.dockerignore` and always generate the managed one, or validate that it cannot exclude required artifacts?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+)
+
+
+class PackageAgentInput(BaseModel):
+    """Submitter-facing shape: an agent name plus the packaging knobs."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    agent: str = Field(description="Name of the Agent entity to package.")
+    tag: str | None = Field(
```
mmogallapalli, line 90: `tag` comes from a remote workspace-scoped request and is forwarded directly to the host Docker daemon. Docker tags are daemon-global, so a caller could submit an existing tag, such as a Platform service image or an image owned by another workspace, and repoint it to the newly built image.

Can the server-side path enforce a Platform-owned tag namespace, derive the tag itself, or reject collisions instead of allowing arbitrary daemon-wide tags?

pyproject.toml
```diff
@@ -192,7 +192,7 @@ enabled-plugins = [
   "nemo-auditor-plugin",
   "nemo-safe-synthesizer-plugin",
   "nemo-switchyard",
-  "nemo-agents-plugin",
+  "nemo-agents-plugin[container]",
```
mmogallapalli: adding `nemo-agents-plugin[container]` to the root `enabled-plugins` group installs `python-on-whales` for repository service environments, but it does not propagate into the published `nemo-platform[nemo-agents-plugin]` extra.

The generated extra in `packages/nemo_platform/pyproject.toml` still does not include `python-on-whales`, so a PyPI-installed Platform can expose this job but fail at runtime with `ContainerToolingUnavailableError`.

Can we wire this dependency into the published package contract as well and add a metadata test confirming the released extra includes the Docker client dependency?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Added a namespace, now using  nemo-agents/{workspace}/{tag}

pyproject.toml
marcusds: Done

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Server-managed path the incoming .dockerignore is discarded and the managed one wins. The CLI still honours one you wrote yourself.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Good catch, fixed.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Fixed, but only on this one

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: I blame Claude.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -0,0 +1,312 @@
...
+    )
+    uv_version: str | None = Field(
+        default=None,
+        pattern=VERSION_PATTERN,
+        description="uv version baked into the image (e.g. '0.9.14').",
+    )
+    allow_root: bool = Field(default=False, description="Run the agent as root instead of the 'agent' user.")
+    sandbox_runtime: str | None = Field(
+        default=None, description="Render an image compatible with a sandbox runtime (e.g. 'openshell')."
+    )
```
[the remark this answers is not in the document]
marcusds: Declining — the premise doesn't hold. sandbox_runtime never reaches the template as text.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Real, `/apis/agents/v2/workspaces/{workspace}/jobs/package`

mmogallapalli (approved):

PR #1469, opened 2026-08-21
feat(agents): publish a packaged agent image to a registry

marcusds:
> **Stacked on #1464.** Base is `astd-459-package-an-agent-container-image-from-the-platform/mschwab`, so the diff here is the single publish commit. Review and merge #1464 first; GitHub will retarget this to `main` automatically.

## TL;DR

`agents.package` built an image and left it in the platform host's local daemon. A Kubernetes cluster cannot pull from there, so `--mode k8s` still required someone to run `nemo agents package --publish` on their laptop.

Setting `registry` on the job now tags and pushes after the build, and the result carries the remote reference. Authentication stays on the host — no credentials cross the API.

## Summary

This closes the last gap between "the platform can build an agent image" and "a cluster can run it". `POST /jobs/package` accepts `registry` and, optionally, `push_tag`; when `registry` is set the job pushes the image it just built and reports the remote tag on `published`. With `registry` omitted the behaviour is exactly as before.

## Related Issue

None. Follow-on to ASTD-459 (#1464), which deliberately scoped publishing out.

## Changes

- `jobs/package_agent.py`: add `registry` and `push_tag` to `PackageAgentInput`; `run()` calls the existing `container.publisher.docker_push` after a successful build and returns `published`.
- Validate both against the Docker reference grammar, reusing `IMAGE_REPOSITORY_PATTERN` and adding `IMAGE_REFERENCE_PATTERN` for the `name:tag` form.
- Reject `push_tag` without `registry` at the API boundary via a model validator, rather than silently ignoring it.
- Regenerate `plugins/nemo-agents/openapi/openapi.yaml`.
- README: document the flag, the `published` field, and the shared-identity caveat below.

## Authentication, and its one real limitation

The platform host must already be authenticated to the target registry — an operator runs `docker login` there once. This is the same assumption `container/publisher.py` has always made, and the same one the CLI makes today.

The tradeoff is deliberate and worth a reviewer's attention: **every workspace pushing from a given host shares that host's registry identity.** That is fine for a single-tenant or dev deployment and wrong for real multi-tenancy. Per-workspace credentials would mean threading a Secret through `compile()` as an `EnvironmentVariableFromSecret` and calling `docker.login` in the task — the mechanism `nemo-evaluator` already uses for API keys — but it puts credential handling into this path and deserves its own change. The README states the limitation rather than leaving it implicit.

Keeping credentials out of the request body also means there is no new secret-bearing field on a job spec that gets persisted and returned by `GET /jobs/package/{name}`.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

| Command | Result |
|---|---|
| `uv run pytest plugins/nemo-agents/tests/unit` | 1189 passed, 2 failed — both `test_port_allocation.py`, which `bind()`s a socket and is blocked by the local sandbox. Verified passing (13/13) with the sandbox disabled; unrelated to this change. |
| `uv run pytest .../test_package_agent_job.py` | 36 passed (8 new: push skipped without a registry, tag forwarded, explicit `push_tag` honoured, push failure propagates, plus input validation) |
| `uv run ruff check plugins/nemo-agents/` | All checks passed |
| `uv run ruff format --check plugins/nemo-agents/` | 202 files already formatted |
| `uv run --frozen ty check` on the changed files | All checks passed |
| `uv run --frozen script/generate-openapi-spec.sh` | Completed; `registry` and `push_tag` appear on `PackageAgentInput` with their `pattern` constraints |

All pre-commit hooks pass on the commit itself, including `ty`. Two are blocked in this environment and were skipped for `git push` only, neither related to this change:

- **`uv-lock`** — requires exactly uv 0.9.14 on `PATH`; this machine has 0.9.30. No dependency files are touched here and the `uv-lock-check` drift hook passes.
- **`helm-docs`** — the binary is not installed locally. No Helm files are touched.

**Not exercised end-to-end.** The unit tests mock `docker_push`, so the real tag-and-push round trip against a live registry has not been run. Doing so needs a build that completes, which on a source checkout is blocked by the contract-version pin that #1466 diagnoses — so this is best verified once one of: a released `nemo-platform` is installed, or a hand-built image is tagged against a local `registry:2`.

## Summary by CodeRabbit

* **New Features**
  * Package-agent jobs can publish built images to a container registry.
  * Added registry and custom remote image tag options, with automatic tag generation when omitted.
  * Job results now report the published image tag.

* **Bug Fixes**
  * Added validation for registry references and image tags.
  * Custom tags must include a registry and remain within the caller’s workspace namespace.

* **Documentation**
  * Clarified authentication requirements and execution hosts for local and submitted jobs.
  * Confirmed registry credentials aren’t accepted through the API.


plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -129,6 +131,29 @@ class PackageAgentInput(BaseModel):
...
+    registry: str | None = Field(
+        default=None,
+        pattern=IMAGE_REPOSITORY_PATTERN,
+        description=(
+            "Push the built image to this registry (e.g. 'nvcr.io/my-org'). The host running "
+            "this command must already be authenticated to it; credentials are never accepted "
+            "over this API."
+        ),
+    )
+    push_tag: str | None = Field(
```
mmogallapalli, line 144: `push_tag` accepts any image reference and is passed verbatim to `docker_push`, which operates using the host’s shared Docker credentials. A caller in one workspace can therefore overwrite any repository or tag authorized for that host, including another workspace’s images.

Please enforce an operator-approved registry and a workspace-scoped destination such as `<registry>/nemo-agents/<workspace>/...`, or require per-workspace credentials and authorization before allowing arbitrary destinations.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -321,7 +347,16 @@ def run(
                 on_progress=logger.info,
             )
 
-        payload = {"image": image, "agent": cfg.agent}
+        published = ""
+        if cfg.registry:
```
mmogallapalli, line 380: The build returns a mutable daemon tag, after which `docker_push` performs separate `docker.tag` and `docker.push` operations. Because subprocess jobs can run concurrently, another job can rebind either the local tag or shared destination tag between those operations, causing this job to publish the wrong image while reporting success.

Please use an immutable image ID or digest and serialize publishing per destination—or otherwise make the operation race-safe—and add a regression test covering this interleaving.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: fixed in 8e051e7, push_tag now has to stay nested under nemo-agents/{workspace}/

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Fixed in f70e4e0e0c by pushing the resolved image ID instead of the mutable tag name; didn't add locking for concurrent pushes to the same destination tag since that's inherent to any registry, not specific to this job.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
```diff
@@ -137,6 +166,23 @@ class PackageAgentSpec(PackageAgentInput):
...
+        if self.push_tag and self.workspace:
+            scoped_pattern = rf"(^|/){re.escape(TAG_NAMESPACE)}/{re.escape(self.workspace)}/[^/]+$"
+            if not re.search(scoped_pattern, self.push_tag):
+                raise ValueError(
+                    f"'push_tag' must be nested under '{TAG_NAMESPACE}/{self.workspace}/' (e.g. "
+                    f"'<registry>/{TAG_NAMESPACE}/{self.workspace}/<name>'). Docker tags are "
+                    "daemon-global while the auth boundary here is the workspace; an unscoped "
+                    "push_tag would let this workspace overwrite another workspace's image on "
+                    "the shared host."
+                )
```
mmogallapalli: My agent suggested tightening this check a little more. Right now, `push_tag` has to use the correct workspace namespace, but it can still point to a different registry than the one supplied in `registry`.

```suggestion
        if self.push_tag and self.registry and self.workspace:
            expected_prefix = (
                f"{self.registry.rstrip('/')}/"
                f"{TAG_NAMESPACE}/{self.workspace}/"
            )
            remote_name = self.push_tag.removeprefix(expected_prefix)
            if (
                not self.push_tag.startswith(expected_prefix)
                or not remote_name
                or "/" in remote_name
            ):
                raise ValueError(
                    f"'push_tag' must be nested under '{expected_prefix}' (e.g. "
                    f"'{expected_prefix}<name>'). Docker tags are daemon-global while the "
                    "auth boundary here is the workspace; an unscoped push_tag would let this "
                    "workspace overwrite another workspace's image on the shared host."
                )
```

mmogallapalli (approved):
gave one suggestion for tightening the `push_tag` check, otherwise LGTM

plugins/nemo-agents/src/nemo_agents_plugin/jobs/package_agent.py
marcusds: Fixed in 7895d73 — push_tag now has to start with <registry>/nemo-agents/{workspace}/, not just contain the workspace segment.

PR #1486, opened 2026-08-24
feat(agents): add environment support to ExecuteAgentJob

benmccown:
## Summary

Gives `ExecuteAgentJob` the same `AgentEnvironment` support that PR #1379 added to `AgentDeployment`. An execute job can now reference an `AgentEnvironment` (a `"workspace/name"` ref or inline); at create time its `EnvironmentSpec` is merged into the resolved agent config and its `ComputeSpec` and secret-env references are snapshotted onto the job step. Before, an execute job ran only the bare agent config with no environment dependencies, resources, or secret-backed env vars; after, the run gets the merged config, executor resources from compute, and secret-backed step env vars.

## Changes

- `ExecuteAgentJobConfig`: new `environment: str | AgentEnvironmentInline | None` input field.
- `ExecuteAgentStepConfig`: new snapshot fields `environment` (raw request value), `compute: ComputeSpecInline | None`, and `secrets: dict[str, str]`, mirroring `AgentDeployment`.
- `to_spec`: resolves and merges the referenced environment (reusing the shared `environment_resolution` helpers), snapshots compute/secrets/raw environment, and validates the **merged** config so an `EnvironmentSpec` selecting a non-local provider is rejected. `EnvironmentResolutionError` is surfaced as a `ValueError` on the create path (matching the existing "Agent not found" pattern).
- `compile`: injects each secret ref as a secret-backed step env var via `EnvironmentVariable(from_secret=...)`, guarded against a reserved set (the `NEMO_JOB_*`/`NMP_TASK_CONFIG` job-substrate vars plus the deployment-container reserved names), and maps agents `ComputeResources` onto the executor `ResourcesSpec` (`cpu`/`memory` pass through; `nvidia.com/gpu` -> `num_gpus`; any other resource key is rejected).
- `run`: unchanged — Fabric reads the merged config and inherits the substrate-populated process env (MCP secrets resolved via env-var-name indirection).
- Added unit tests in `test_execute_job.py` (environment ref/inline snapshot, no-environment empty snapshot, missing-ref -> ValueError, non-local-provider rejection, MCP secret indirection, compile secret env + compute resources, no-compute omits executor resources, unsupported resource key raises, reserved-name collision raises). Also refreshed a pre-existing stale snapshot assertion in the create-route test (the Fabric `environment` block gained `env`/`connection`/`metadata` defaults in #1379; that assertion failed on `main` independently of this change).
- Regenerated the nemo-agents plugin OpenAPI spec (`make refresh-openapi`); the diff is scoped to the three new fields (`AgentEnvironmentInline`/`ComputeSpecInline` schemas already existed). The nemo-agents plugin is not part of Stainless SDK generation, so no SDK regen.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal job wiring; no user-facing docs surface changed (environment support is not exposed in the hand-written CLI, matching the deployment path).

## Verification

Targeted validation:

- `uv run ruff check plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py plugins/nemo-agents/tests/unit/test_execute_job.py` — passed.
- `uv run ruff format --check` (same files) — passed (already formatted).
- `uv run --frozen ty check` on changed source + test files — passed.
- `bash tools/lint/lint-python-types.sh` (full-repo `ty check`, mirrors CI's `lint-python-types`) — passed. (Initial CI run flagged `step["executor"]["resources"]` in a compile test: the step `executor` is a union of executor TypedDicts and `SubprocessExecutionProviderParam` has no `resources` key. Fixed by casting the executor to `dict[str, Any]` before subscripting, matching the existing `step["config"]` pattern.)
- `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_execute_job.py --import-mode=importlib -p no:cacheprovider -q` — 50 passed.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit --import-mode=importlib -p no:cacheprovider -q` — 1335 passed (no regressions).
- `make refresh-openapi` — completed; diff scoped to `plugins/nemo-agents/openapi/openapi.yaml`.
- `uv run pre-commit run -a` — relevant hooks passed (ruff, ruff format, ty, config-reference, copyright headers, plugins-not-import-nmp-common, merge-conflict check, uv.lock drift check). Three hooks failed only for local toolchain/environment reasons unrelated to this change and not touching changed files: `helm-docs` (binary not installed; no Helm changes), `uv-lock` (local uv 0.9.21 vs required 0.9.14; no dependency changes and the `uv.lock` drift check passed), and `studio-lint-staged` (local Node engine mismatch / corepack pnpm download; no `web/` changes). CI will re-run these on pinned toolchains.

## Summary by CodeRabbit

* **New Features**
  * Added optional stored or inline environment configuration for agent execution jobs.
  * Environment settings can include compute resources and secret-backed variables.
  * Configured resources and secrets are applied automatically during execution.

* **Bug Fixes**
  * Added validation for unsupported resources and reserved environment names.
  * Invalid reserved-secret configurations now return clear compilation errors or HTTP 422 responses.
  * Environment settings are consistently resolved and preserved for each execution step.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
@@ -66,6 +66,7 @@ from nemo_platform_plugin.jobs.constants import (
     PERSISTENT_JOB_STORAGE_PATH_ENVVAR,
     TASK_CONFIG_ENVVAR,
 )
+from nemo_platform_plugin.jobs.exceptions import PlatformJobCompilationError
 from nemo_platform_plugin.jobs.image import get_qualified_image
 from nemo_platform_plugin.refs import ENTITY_REF_PATTERN, parse_entity_ref
 from pydantic import BaseModel, Field
@@ -284,10 +285,26 @@ class ExecuteAgentJob(NemoJob):
                 command=["nemo_agents_plugin.tasks.execute"],
             ),
         )
-        # Snapshotted compute -> executor resources. Injected only when the
-        # environment supplied a compute spec, so the default CPU sizing is left
-        # to the jobs backend otherwise.
-        resources = _compute_to_resources(step_config.compute)
+        # Snapshotted compute -> executor resources, and secret-env refs ->
+        # secret-backed step env vars. Both validate the snapshot and raise
+        # ``ValueError`` on a bad shape (an unsupported resource key, or a secret
+        # env name colliding with a reserved job env var). Surface those as a
+        # ``PlatformJobCompilationError`` so the jobs create route maps them to a
+        # descriptive 422 rather than an opaque 500 (the route's compile wrapper
+        # only translates ``PlatformJobCompilationError``, not bare ``ValueError``).
+        try:
+            # Snapshotted compute -> executor resources. Injected only when the
+            # environment supplied a compute spec, so the default CPU sizing is
+            # left to the jobs backend otherwise.
+            resources = _compute_to_resources(step_config.compute)
+            # Snapshotted secret-env refs -> secret-backed step env vars. The jobs
+            # substrate materializes each value into the process env under
+            # ENV_NAME; Fabric and its MCP servers read it by name (env-var-name
+            # indirection - see environment_resolution).
+            environment = _secret_environment(step_config.secrets)
+        except ValueError as exc:
+            raise PlatformJobCompilationError(str(exc)) from exc
+
         if resources is not None:
             executor["resources"] = resources
 
@@ -297,11 +314,7 @@ class ExecuteAgentJob(NemoJob):
                     name="execute-agent",
                     executor=executor,
                     config=step_config.model_dump(mode="json"),
-                    # Snapshotted secret-env refs -> secret-backed step env vars.
-                    # The jobs substrate materializes each value into the process
-                    # env under ENV_NAME; Fabric and its MCP servers read it by
-                    # name (env-var-name indirection - see environment_resolution).
-                    environment=_secret_environment(step_config.secrets),
+                    environment=environment,
                 )
             ],
         )
--- a/plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
@@ -15,7 +15,17 @@ from typing import Any, ClassVar, cast
 from nemo_agents_plugin.agent_config import AgentConfig
 from nemo_agents_plugin.agent_config_formats import resolve_agent_config_for_deployment
 from nemo_agents_plugin.config import AgentsConfig
-from nemo_agents_plugin.entities import NEMO_AGENTS_SPEC_CONFIG_FORMAT, Agent
+from nemo_agents_plugin.entities import (
+    NEMO_AGENTS_SPEC_CONFIG_FORMAT,
+    Agent,
+    AgentEnvironmentInline,
+    ComputeSpecInline,
+)
+from nemo_agents_plugin.environment_resolution import (
+    EnvironmentResolutionError,
+    merge_environment_spec_into_agent_config,
+    resolve_environment,
+)
 from nemo_agents_plugin.fabric.invocation import (
     AgentConfigInvocationRequest,
     FabricDirectories,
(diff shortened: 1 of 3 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -85,6 +164,14 @@ class ExecuteAgentStepConfig(BaseModel):
     request: ExecuteAgentJobConfig
     agent: ResolvedAgentConfig
     workdir: AgentWorkdir | None = None
+    # Immutable snapshot of the resolved AgentEnvironment, mirroring
+    # AgentDeployment. ``agent.config`` already holds the merged config;
+    # ``compute`` and ``secrets`` are snapshotted for the executor. ``environment``
+    # retains the raw request value for provenance. Once created, the job is not
+    # kept in sync with the underlying environment entities.
+    environment: str | AgentEnvironmentInline | None = None
```
mikeknep: Should we drop `str` from this union type and only store a snapshot of the materialized environment at create-time, since the agent environment entity is mutable and might be changed sometime after the job?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -279,6 +414,95 @@ def _has_workdir_inputs(workdir: AgentWorkdir) -> bool:
...
+    if reserved:
+        raise ValueError(
+            f"Secret env var(s) {', '.join(reserved)} collide with reserved job env var name(s). "
+            f"Reserved names: {', '.join(sorted(_RESERVED_ENV_VAR_NAMES))}."
+        )
+
+    return [
+        EnvironmentVariable(name=env_name, from_secret=EnvironmentVariableFromSecret(name=ref))
+        for env_name, ref in secrets.items()
+    ]
```
mikeknep, line 438: I don't see any validation anywhere that the `ref`s here do actually exist as secret entities and that the user is authenticated to use them. Should that be happening in `nemo_agents_plugin.environment_resolution._collect_secrets`?

benmccown: That does happen, just happens within the Jobs service (using the user's identity) and not at the nemo-agents API layer. We could potentially lift it up into this layer if you think that's sensible (I'm on the fence, I could go either way), though I'd probably hit that in a quick followup PR scoped to that change, if you want me to execute on it. LMK what you think!

mikeknep: Ahh right, OK makes sense to happen at the Jobs layer, that sounds fine to me. I guess my only follow-up question (and not a blocker for this PR) is: does Deployments have the same feature?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
benmccown: That's actually what we do. On the `ExecuteAgentStepConfig` that we store, we persist the fully qualified environment spec, which is ultimately persisted on the Job entity. Nothing ever re-fetches the `AgentEnvironment`. The `compile` and `run` steps only read the baked-in config/compute/secrets. The str we're persisting is really just a dead pointer/reference for posterity.

mikeknep: > On the `ExecuteAgentStepConfig` that we store, we persist the fully qualified environment spec

This specifically is the `merged.config` we set as `ExecuteAgentStepConfig.agent.config` in `to_spec` below, right?

> The str we're persisting is really just a dead pointer/reference for posterity.

The `ExecuteAgentStepConfig` already has the full `request: ExecuteAgentJobConfig`. So do we need this separate `environment` attribute at all?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
benmccown: > It's similar but not identical:
>
> - **Jobs** validate secret refs **synchronously at create time** under the **caller's identity** (`validate_job_secrets` in the jobs dispatcher): a missing ref or one the user can't access fails the create request with a clear not-found / no-access error.
> - **Deployments** resolve secret refs **later, at reconcile/deploy time** in the nemo-deployments plugin (`resolve_secret_ref` → `access_secret`); a failure surfaces as a **deployment status failure** rather than a rejected create call. And that resolution runs under the **deployment controller's identity**, not the requesting user's — so it enforces existence but doesn't do the same per-user authZ gate at request time.
>
> So the strict "does *this user* have access, checked up front" behavior is actually stronger on the Jobs side. 

Probs worth creating an issue and separate PR to address on the Deployment side, if that sounds good? I'll take care of creating it.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
benmccown: agent injected reply:

> Yes to both.
>
> **On the merged config** — correct, that's `merged.config`, set as `ExecuteAgentStepConfig.agent.config` in `to_spec`. Worth being precise about the layering, since "environment spec" can be ambiguous: we don't persist the `EnvironmentSpec` as a standalone object. `merge_environment_spec_into_agent_config` folds it into the platform-owned `AgentConfig` (env block, plaintext `env`, model-provider override), with secret refs lifted out into the separate `secrets` field (never embedded as plaintext). That merged `AgentConfig` is what's stored and what `compile`/`run` read. It's a genuine translation, not a passthrough — `AgentConfig` is our contract, and it's only later compiled into the external `nemo-fabric` `FabricConfig` by `translate_agent_config` (which also owns harness→adapter-id mapping, IGW credential binding, runtime env, telemetry). So the EnvironmentSpec→AgentConfig merge deliberately happens once, at create time, which is what gives us the immutable snapshot + early error surfacing.
>
> **On the separate `environment` attribute** — you're right, it was redundant. `to_spec` set it to exactly `request.environment`, and we already persist the whole `request: ExecuteAgentJobConfig`, so it carried nothing new. I've dropped it and now read provenance off `request.environment`. (`compute`/`secrets` stay — those are resolved snapshots that don't exist on the raw request.) This is a small divergence from `AgentDeployment`, which keeps a top-level `environment` because the deployment entity doesn't embed the raw request the way the job step config does. Done in 15f436f25.

mikeknep (approved):

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
benmccown: agent injected reply:

> Filed as [AIRCORE-1067](https://linear.app/nvidia/issue/AIRCORE-1067/deployments-validate-secret-refs-at-create-time-under-the-requesting) to bring the deployment path to parity (create-time, user-scoped secret-ref validation). Not blocking this PR. Resolving this thread.

PR #1504, opened 2026-08-24
chore(customizer): add missing grpo input configs + monitoring metrics

anubhutivyas:
## Summary
Adds the GRPO hyperparameters that had no API to set them, and the metric series a GRPO run needs to be monitored on reward rather than loss. Also fixes `normalize_rewards`, which was accepted by the API and then silently ignored — the advantage estimator reads a config block the compiler never wrote, so advantages were always normalized regardless of what the job asked for.

## Related Issue
https://linear.app/nvidia/issue/ASTD-388/add-grpo-to-customizer-ui

## Changes
- **Fix: `normalize_rewards` had no effect.** NeMo-RL's advantage estimator reads `grpo.adv_estimator.normalize_rewards`; the compiler wrote only the top-level `grpo.normalize_rewards`, which has no reader. The compiler now writes the `adv_estimator` block as well as the top-level pair, so the config matches a resolved recipe either way. Code in RL - https://github.com/soluwalana/RL/blob/nmp/customizer/nemo_rl/algorithms/advantage_estimator.py#L44

- **New GRPO hyperparameters**, threaded through the public schema, the job compiler, the internal step config, and the NeMo-RL config:

| Field | Compiles into |
| --- | --- |
| `ratio_clip_c` (dual clip) | `loss_fn` |
| `use_on_policy_kl_approximation` | `loss_fn` |
| `use_importance_sampling_correction` | `loss_fn` |
| `advantage_clip_low` / `advantage_clip_high` | `grpo` |
| `use_leave_one_out_baseline` | `grpo.adv_estimator` |
| `top_k` | `policy.generation` |

- **New metric series for GRPO.** A GRPO run recorded curves for reward, entropy, two KL diagnostics, length and truncation. It now also records `policy_kl_error`, `js_divergence_error` and `sampling_importance_ratio`, `total_reward/stddev` (the spread around the reward mean, which nothing else reported), and `baseline_reward/pct_{0,1,mixed}` (the share of prompt groups whose rollouts disagreed — GRPO scores each rollout against others in its group, so when `pct_mixed` reaches zero the run has stopped learning, and that shows before the reward curve flattens). `"train_reward"` became `"*_reward"` so one entry covers both phases.

- **`val_reward`.** GRPO reports its mean validation reward under the name `accuracy`, which is right for the math environment it was named after and misleading for anything else. The adapter now also reports the value as `reward`, so it arrives as `val_reward`. Copying rather than renaming keeps `val_accuracy` working for jobs that already have curves stored under it, and leaves `checkpointing.metric_name="val:accuracy"` reading the name NeMo-RL owns.

- **Run facts.** `report_training_start` takes an optional `run_facts` dict, sent once and never restated. GRPO reports `training_type` and `rollouts_per_step`; DPO reports `training_type` so the field is not half-populated.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Added GRPO controls for top-k sampling, reward normalization, leave-one-out baselines, advantage clipping, dual clipping, KL approximation, and importance-sampling correction.
  - Added configurable sandbox rollout chunk sizes, with a default of 8.
  - Training status now includes algorithm and rollout metadata.
  - Expanded training metrics with reward, policy-drift, baseline, timing, and diagnostic information.

- **Validation**
  - Added configuration checks to prevent invalid advantage-clipping ranges.


services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
```diff
@@ -336,11 +388,31 @@ def log_metrics(
...
+
+        A ``reward`` the algorithm reported itself is left alone. Nothing reports one
+        today, and overwriting a real one would be worse than not copying. A
+        non-numeric value is skipped too, since it could not be charted anyway.
+        """
+        source = self._validation_reward_metric
+        if source is None or "reward" in metrics:
+            return metrics
+        value = metrics.get(source)
+        return {**metrics, "reward": value} if is_chartable(value) else metrics
```
albcui: Is this really necessary? Why must we rename `"accuracy"` -> `"reward"`? What's wrong with just logging it as `"val_accuracy"` as-is?

anubhutivyas: Because it's named as `Validation reward` in this ticket - https://linear.app/nvidia/issue/ASTD-388/add-grpo-to-customizer-ui#ac6-job-monitoring-shows-0574e9bf. So, instead of Studio team doing additional work, would be straightforward to just show all metrics as is. Might be confusing to see different names when users see in UI vs when they run `get-status`

albcui: Hmm I think there are a couple of better ways to solve this:
1. Change this in RL library
2. A more generic mechanism to map library metric names to what we want the user to see

anubhutivyas: Had a sync with UI team, they are using "accuracy", so removed the additional "reward" metrics for validation.

albcui (approved):

services/rl/src/nmp/rl/tasks/training/backends/nemo_rl/nemo_rl_logger.py
```diff
@@ -228,6 +274,7 @@ def __init__(
         self._max_steps = max_steps
         self._num_epochs = num_epochs
         self._steps_per_epoch = steps_per_epoch
+        self._run_facts = dict(run_facts or {})
```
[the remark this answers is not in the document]
anubhutivyas: updated the order in `packages/nmp_customization_common/src/nmp/customization_common/training/callbacks.py)`, shouldn't be an issue

PR #1507, opened 2026-08-24
feat: add cli config context-delete and TLS support

ironcommit:
## Summary

Adds `up` and `test` as first-class run modes to the authentik harness (`run.sh`), so developers can stand up a durable Compose or Kubernetes auth-idp environment without immediately running tests. Also extends K8s startup timeouts, eliminates gateway port collisions between concurrent test runs, and records lifecycle state so reusable instances and NeMo contexts can be cleaned up consistently.

### New `up` / `test` run modes
- `run.sh up compose` starts the reusable Compose stack in detached mode, waits for the gateway health endpoint, registers a NeMo context, and exits without running pytest.
- `run.sh up k8s` creates or reuses a kind/k3d cluster, installs the Helm chart, starts a background `kubectl port-forward`, registers a NeMo context, and exits without running pytest. Uses the stable default port `18082`.
- `run.sh test compose` / `run.sh test k8s` run the contract tests without adding reusable contexts to the user's NeMo config.

### Lifecycle cleanup and context management
- Added `.generated/instances` lifecycle state files for reusable Compose and Kubernetes instances.
- `down compose` and `down k8s` clean only the selected backend; bare `down` still cleans both.
- `down --key <key>` targets the derived Compose/Kubernetes instance names without tearing down the default Compose stack first.
- `clean` removes all recorded lifecycle instances.
- Added `nemo config delete-context --prune-orphans` and uses it for Authentik teardown.

### K8s port-collision avoidance
- `test k8s` now picks a free ephemeral port by default (via `choose_free_tcp_port`) so it can run while an `up k8s` session holds the stable port. `NMP_AUTHENTIK_K8S_GATEWAY_PORT` still overrides both modes.
- K8s port-forward PID reuse and teardown now validate that the recorded PID is still the expected `kubectl port-forward svc/nemo-platform-envoy` process before reusing or killing it.
- Port-forward reuse also checks the selected local gateway port and restarts stale port-forwards when the port changes.

### TLS certificate authority support
- Authentik `up` contexts now save the gateway CA bundle in NeMo config using the Kubernetes-style `certificate_authority` field.
- Saved context CAs are applied through config resolution, SDK bootstrap, OAuth discovery, CLI auth discovery/login flows, OAuth provider cache keys, no-auth clients, token refresh, and workload token exchange.
- Device flow and OAuth discovery requests now build HTTPX TLS kwargs like the SDK migration branch: no `verify` argument by default, and `verify=<CA path>` only when a context or env CA bundle is configured.
- Direct token refresh and workload token-exchange requests now pass `verify` only with a concrete CA bundle path; otherwise they omit `verify` and rely on httpx's default certificate validation.
- CLI help and docs now show absolute CA paths.

### Extended K8s startup timeouts
- Helm wait timeout: `10m` -> `20m` (configurable via `NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT`).
- Helm upgrade command timeout now derives from `NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT` plus a 300-second margin; the default remains `1500s`.
- Pytest per-test timeout for Kubernetes contract tests: `900s` -> `2400s`, auto-applied via `conftest.py`.
- Startup probe `failureThreshold` raised to `80` for both API and core controller pods.

### Improved port-forward diagnostics
- Port-forward stdout/stderr is now captured to a log file under `NMP_AUTHENTIK_K8S_LOG_DIR`.
- Early-exit error messages include the log tail for faster debugging.
- `down` now stops any running K8s port-forward process before deleting the cluster.
- K8s CA bundle extraction now fails early if the Envoy TLS secret is missing the `ca.crt` entry.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [ ] Existing tests cover changed behavior - justification:
- [ ] Tests not applicable - justification:
- [ ] Documentation not applicable - justification:

## Verification

Targeted validation:

- `flox activate -d tools/python -- uv run --frozen pytest packages/nmp_common/tests/sdk_factory/test_sdk.py::test_get_platform_sdk_uses_workload_identity_when_token_file_configured packages/nmp_common/tests/sdk_factory/test_sdk.py::test_get_task_sdk_uses_workload_identity_when_token_file_configured packages/nemo_platform_ext/tests/cli/commands/test_auth_password_grant.py::test_login_password_grant_with_flags packages/nemo_platform_ext/tests/cli/commands/test_auth_password_grant.py::test_login_password_grant_with_env -q` - 4 passed
- `flox activate -d tools/python -- uv run --frozen pytest packages/nmp_common/tests/sdk_factory/test_sdk.py packages/nemo_platform_ext/tests/cli/commands/test_auth_password_grant.py packages/nemo_platform_ext/tests/cli/commands/test_auth.py packages/nemo_platform_ext/tests/client/test_client.py packages/nemo_platform_ext/tests/auth/test_utils.py -q` - 182 passed
- `flox activate -d tools/python -- uv run --frozen pytest tests/auth_idp/static/test_authentik_kubernetes_demo.py packages/nemo_platform_ext/tests/auth/test_utils.py packages/nemo_platform_ext/tests/auth/test_device_flow.py packages/nemo_platform_ext/tests/auth/test_token_provider.py packages/nemo_platform_ext/tests/auth/test_workload_exchange.py packages/nemo_platform_ext/tests/client/test_client.py packages/nemo_platform_ext/tests/config/test_config.py packages/nemo_platform_ext/tests/cli/commands/test_auth.py -q` - 339 passed
- `flox activate -d tools/python -- uv run --frozen pytest tests/auth_idp/static/test_authentik_kubernetes_demo.py tests/auth_idp/k8s/test_authentik_kubernetes_live.py -q` - 65 passed
- `uv run --frozen pytest tests/auth_idp/static/test_authentik_kubernetes_demo.py packages/nemo_platform_ext/tests/cli/commands/test_config.py packages/nemo_platform_ext/tests/config/test_config.py packages/nemo_platform_ext/tests/client/test_client.py packages/nemo_platform_ext/tests/auth/test_token_provider.py packages/nemo_platform_ext/tests/auth/test_workload_exchange.py -q` - 279 passed
- `uv run --frozen pytest packages/nemo_platform_ext/tests/auth/test_token_provider.py packages/nemo_platform_ext/tests/auth/test_workload_exchange.py -q` - 51 passed
- `bash -n contrib/auth/authentik/run.sh`
- `contrib/auth/authentik/run.sh down --key dev --dry-run`
- `contrib/auth/authentik/run.sh up k8s --dry-run --skip-image-load`
- `flox activate -d tools/python -- uv run ruff format ...`
- `flox activate -d tools/python -- uv run ruff check ...`
- `uv run --frozen python -m py_compile sdk/python/nemo-platform/src/nemo_platform/_client.py`
- `flox activate -d tools/python -- uv run --frozen pytest packages/nemo_platform_ext/tests/client/test_tls.py packages/nemo_platform_ext/tests/auth/test_device_flow.py packages/nemo_platform_ext/tests/auth/test_utils.py packages/nemo_platform_ext/tests/auth/test_token_provider.py packages/nemo_platform_ext/tests/auth/test_workload_exchange.py packages/nemo_platform_ext/tests/client/test_client.py packages/nemo_platform_ext/tests/cli/commands/test_setup.py -q` - 399 passed
- `flox activate -d tools/python -- uv run --frozen pytest packages/nemo_platform_ext/tests/client/test_tls.py packages/nemo_platform_ext/tests/auth/test_device_flow.py packages/nemo_platform_ext/tests/auth/test_utils.py packages/nemo_platform_ext/tests/auth/test_token_provider.py packages/nemo_platform_ext/tests/auth/test_workload_exchange.py packages/nemo_platform_ext/tests/client/test_client.py -q` - 163 passed
- `flox activate -d tools/python -- uv run ruff check packages/nemo_platform_ext/src/nemo_platform_ext/client/tls.py packages/nemo_platform_ext/src/nemo_platform_ext/auth/device_flow.py packages/nemo_platform_ext/src/nemo_platform_ext/auth/helpers.py packages/nemo_platform_ext/tests/client/test_tls.py packages/nemo_platform_ext/tests/auth/test_device_flow.py` - passed
- `flox activate -d tools/python -- uv run ruff format --check packages/nemo_platform_ext/src/nemo_platform_ext/client/tls.py packages/nemo_platform_ext/src/nemo_platform_ext/auth/device_flow.py packages/nemo_platform_ext/src/nemo_platform_ext/auth/helpers.py packages/nemo_platform_ext/tests/client/test_tls.py packages/nemo_platform_ext/tests/auth/test_device_flow.py` - passed
- `flox activate -d tools/python -- uv run pre-commit run -a` - passed

Note: local shell `uv` is 0.9.18, while the repository hook requires 0.9.14. The full pre-commit suite passed under the pinned Flox Python toolchain.

## Summary by CodeRabbit

* **New Features**
  * Added separate `up` and `test` workflows for Compose and Kubernetes environments, with reusable instances, configurable gateway ports, automatic free-port selection, and improved cleanup.
  * Added local TLS certificate authority configuration for secure deployments and authentication.
  * Added a command to delete saved contexts, with optional cleanup of unused records.
  * Added clearer diagnostics, including port-forward logs and recent failure output.

* **Bug Fixes**
  * Improved startup reliability with longer readiness and Helm operation timeouts.
  * Added startup probes for core authentication services.

* **Documentation**
  * Updated setup, TLS, context management, command, and port guidance.

### DIFF
--- a/contrib/auth/authentik/run.sh
+++ b/contrib/auth/authentik/run.sh
@@ -9,8 +9,10 @@ AUTHENTIK_ROOT="${SCRIPT_DIR}"
 REPO_ROOT="$(cd -- "${SCRIPT_DIR}/../../.." && pwd)"
 
 ACTION=""
+TARGET=""
 COMPOSE_DIR="${AUTHENTIK_ROOT}/compose"
 DRY_RUN="false"
+INSTANCE_KEY=""
 IMAGE_SELECTED="false"
 IMAGE_REGISTRY="${IMAGE_REGISTRY:-my-registry}"
 BAKE_TAG="${BAKE_TAG:-local}"
@@ -20,15 +22,17 @@ TEST_PLATFORM=""
 TEST_PLATFORM_SET="false"
 TEST_DOCKER_TARGET="nmp-api-docker"
 COMPOSE_DIR_SET="false"
-REUSE_COMPOSE_PROJECT_NAME="${NMP_AUTHENTIK_COMPOSE_PROJECT_NAME:-authentik-e2e-reuse}"
-REUSE_COMPOSE_GATEWAY_PORT="${NMP_AUTHENTIK_COMPOSE_GATEWAY_PORT:-18083}"
-REUSE_COMPOSE_GATEWAY_TLS_VOLUME="${NMP_AUTHENTIK_COMPOSE_GATEWAY_TLS_VOLUME:-authentik-e2e-${REUSE_COMPOSE_GATEWAY_PORT}-gateway-tls}"
-REUSE_COMPOSE_WORKLOAD_NETWORK_NAME="${NMP_AUTHENTIK_COMPOSE_WORKLOAD_NETWORK_NAME:-authentik-e2e-${REUSE_COMPOSE_GATEWAY_PORT}-workload}"
-REUSE_K8S_CLUSTER_NAME="${NMP_AUTHENTIK_K8S_REUSE_CLUSTER_NAME:-nmp-authentik-reuse}"
-K8S_GATEWAY_PORT="${NMP_AUTHENTIK_K8S_GATEWAY_PORT:-18082}"
+REUSE_COMPOSE_PROJECT_NAME="${NMP_AUTHENTIK_COMPOSE_PROJECT_NAME:-}"
+REUSE_COMPOSE_GATEWAY_PORT="${NMP_AUTHENTIK_COMPOSE_GATEWAY_PORT:-}"
+REUSE_COMPOSE_GATEWAY_TLS_VOLUME="${NMP_AUTHENTIK_COMPOSE_GATEWAY_TLS_VOLUME:-}"
+REUSE_COMPOSE_WORKLOAD_NETWORK_NAME="${NMP_AUTHENTIK_COMPOSE_WORKLOAD_NETWORK_NAME:-}"
+REUSE_K8S_CLUSTER_NAME="${NMP_AUTHENTIK_K8S_REUSE_CLUSTER_NAME:-}"
+DEFAULT_K8S_GATEWAY_PORT="18082"
+K8S_GATEWAY_PORT="${NMP_AUTHENTIK_K8S_GATEWAY_PORT:-}"
 K8S_JUNIT_XML="${NMP_AUTHENTIK_K8S_JUNIT_XML:-report-auth-idp-kubernetes.xml}"
 HELM_NAMESPACE="${HELM_NAMESPACE:-${NMP_AUTHENTIK_K8S_NAMESPACE:-nemo-authentik}}"
 HELM_RELEASE="${HELM_RELEASE:-${NMP_AUTHENTIK_K8S_HELM_RELEASE:-authentik-demo}}"
+HELM_WAIT_TIMEOUT="${HELM_WAIT_TIMEOUT:-${NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT:-20m}}"
 K8S_CLUSTER_NAME="${NMP_AUTHENTIK_K8S_CLUSTER_NAME:-}"
 K8S_RUNTIME="${NMP_AUTHENTIK_K8S_RUNTIME:-kind}"
 K8S_RUNTIME_SET="false"
@@ -36,8 +40,11 @@ K8S_KEEP_CLUSTER="${NMP_AUTHENTIK_K8S_KEEP_CLUSTER:-0}"
 K8S_REUSE_CLUSTER="${NMP_AUTHENTIK_K8S_REUSE_CLUSTER:-0}"
 K8S_SKIP_IMAGE_LOAD="${NMP_AUTHENTIK_K8S_SKIP_IMAGE_LOAD:-0}"
 K8S_SKIP_IMAGE_LOAD_SET="false"
+K8S_EXPORT_KUBECONFIG="${NMP_AUTHENTIK_K8S_EXPORT_KUBECONFIG:-0}"
+K8S_EXPORT_KUBECONFIG_SET="false"
 K8S_NGC_EXISTING_SECRET="${NMP_AUTHENTIK_K8S_NGC_EXISTING_SECRET:-}"
 K8S_IMAGE_PULL_SECRET="${NMP_AUTHENTIK_K8S_IMAGE_PULL_SECRET:-}"
+AUTHENTIK_WORKSPACE="${NMP_AUTHENTIK_WORKSPACE:-authentik-demo}"
 
 diagnostics_dir() {
     local mode="$1"
@@ -89,26 +96,75 @@ write_diagnostics_metadata() {
     } >"${output}/run-metadata.txt"
 }
 
+choose_free_tcp_port() {
+    python3 -c 'import socket
+with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+    sock.bind(("127.0.0.1", 0))
+    print(sock.getsockname()[1])'
+}
+
+validate_k8s_gateway_port() {
+    if [[ ! "${K8S_GATEWAY_PORT}" =~ ^[0-9]+$ ]]; then
+        die "NMP_AUTHENTIK_K8S_GATEWAY_PORT must be an integer TCP port"
+    fi
+    if ((K8S_GATEWAY_PORT < 1 || K8S_GATEWAY_PORT > 65535)); then
+        die "NMP_AUTHENTIK_K8S_GATEWAY_PORT must be between 1 and 65535"
+    fi
+}
+
+configure_k8s_gateway_port() {
+    if [[ "${TARGET}" != "k8s" ]]; then
+        return
+    fi
+    if [[ "${ACTION}" != "up" && "${ACTION}" != "test" ]]; then
+        return
+    fi
+    if [[ -z "${K8S_GATEWAY_PORT}" ]]; then
+        case "${ACTION}" in
+            up)
+                if [[ -n "${INSTANCE_KEY}" ]]; then
+                    K8S_GATEWAY_PORT="$(choose_free_tcp_port)"
+                else
+                    K8S_GATEWAY_PORT="${DEFAULT_K8S_GATEWAY_PORT}"
+                fi
+                ;;
+            test)
+                K8S_GATEWAY_PORT="$(choose_free_tcp_port)"
+                ;;
+        esac
+    fi
+    validate_k8s_gateway_port
+}
+
 usage() {
     cat <<'EOF'
 Usage:
-  contrib/auth/authentik/run.sh run-local [options]
-  contrib/auth/authentik/run.sh down [options]
+  contrib/auth/authentik/run.sh up compose [options]
+  contrib/auth/authentik/run.sh up k8s [options]
+  contrib/auth/authentik/run.sh test compose [options]
+  contrib/auth/authentik/run.sh test k8s [options]
+  contrib/auth/authentik/run.sh down [compose|k8s|all] [options]
+  contrib/auth/authentik/run.sh clean [compose|k8s|all] [options]
   contrib/auth/authentik/run.sh prepare-local [options]
-  contrib/auth/authentik/run.sh compose [options]
   contrib/auth/authentik/run.sh render-blueprint [options]
-  contrib/auth/authentik/run.sh k8s [options]
 
 Runs the local Authentik reference example or its auth-idp test suite.
 
 Actions:
-  run-local              Start NeMo, Authentik, and the gateway in the foreground.
-  down                   Remove the example Compose stack, reused test Compose stack,
-                         and reused Kubernetes test cluster.
+  up compose             Start the reusable Compose auth-idp stack and skip tests.
+  up k8s                 Create or reuse the Kubernetes auth-idp cluster, install
+                         the Helm chart, start a local gateway port-forward, and
+                         skip tests.
+  test compose           Build the local test image if needed, then run Compose
+                         auth-idp tests.
+  test k8s               Run the Helm-based kind/k3d Kubernetes E2E test.
+  down compose           Remove managed Authentik Compose instances and contexts.
+  down k8s               Remove managed Authentik Kubernetes instances and contexts.
+  down all               Remove managed Compose and Kubernetes instances.
+                         Bare "down" defaults to "down all".
+  clean                  Remove stale recorded instances using down semantics.
   prepare-local          Create generated local inputs without starting Compose.
-  compose                Build the local test image if needed, then run Compose auth-idp tests.
   render-blueprint       Copy the checked-in Authentik blueprint into generated inputs.
-  k8s                    Run the Helm-based kind/k3d Kubernetes E2E test.
 
 Image options:
   --image IMAGE          Use an existing nmp-api image.
@@ -120,10 +176,12 @@ Test options:
                          on gateway port 18083.
                          For k8s, use cluster nmp-authentik-reuse with
                          the selected runtime, creating it if needed and
-                         keeping it after the run.
-                         The k8s runner uses gateway port 18082 by default
-                         to avoid the tutorial's 18081 port. Override with
-                         NMP_AUTHENTIK_K8S_GATEWAY_PORT.
+                         keeping it after the run. The up actions use these
+                         reusable resources by default.
+                         The k8s up action uses gateway port 18082 by default
+                         to avoid the tutorial's 18081 port. The k8s test
+                         action chooses a free local port by default. Override
+                         either with NMP_AUTHENTIK_K8S_GATEWAY_PORT.
   --platform PLATFORM    Platform for the default local test image build.
                          Applies to compose and k8s.
                          Default: current machine architecture.
@@ -132,24 +190,35 @@ Test options:
   --skip-image-load      Do not load the nmp-api image into the reused cluster.
                          For a fresh cluster, use only with an explicit pullable
                          --image.
+  --export-kubeconfig    Also merge and switch the Kubernetes context into the
+                         default kubeconfig. By default, k8s up writes an
+                         isolated kubeconfig under .generated only.
+
+Instance options:
+  --key KEY              Manage a named up/down instance. The default instance
+                         uses contexts authentik-compose and authentik-k8s.
+                         A key derives names such as authentik-compose-KEY,
+                         authentik-k8s-KEY, and matching local resources.
 
 Other options:
-  --compose-dir DIR      Compose directory for run-local/down. Default: contrib/auth/authentik/compose.
+  --compose-dir DIR      Compose directory for down/compose.
+                         Default: contrib/auth/authentik/compose.
   --dry-run              Print commands without running them.
   -h, --help             Show this help.
 
 Examples:
-  contrib/auth/authentik/run.sh run-local
-  contrib/auth/authentik/run.sh run-local --image my-registry/nmp-api:local
+  contrib/auth/authentik/run.sh up compose
+  contrib/auth/authentik/run.sh up compose --key dev
+  contrib/auth/authentik/run.sh up k8s
+  contrib/auth/authentik/run.sh up k8s --key dev
+  contrib/auth/authentik/run.sh test compose
+  contrib/auth/authentik/run.sh test k8s
+  contrib/auth/authentik/run.sh down compose
+  contrib/auth/authentik/run.sh down compose --key dev
+  contrib/auth/authentik/run.sh down k8s
+  contrib/auth/authentik/run.sh clean
   contrib/auth/authentik/run.sh prepare-local
-  contrib/auth/authentik/run.sh compose
-  contrib/auth/authentik/run.sh compose --reuse
-  contrib/auth/authentik/run.sh compose --image my-registry/nmp-api:local
   contrib/auth/authentik/run.sh render-blueprint
-  contrib/auth/authentik/run.sh k8s
-  contrib/auth/authentik/run.sh k8s --runtime k3d
-  contrib/auth/authentik/run.sh k8s --reuse
-  contrib/auth/authentik/run.sh k8s --reuse --skip-image-load
   contrib/auth/authentik/run.sh down
 EOF
 }
@@ -159,6 +228,11 @@ die() {
     echo >&2
     usage >&2
     exit 2
+}
+
+fail() {
+    echo "error: $*" >&2
+    exit 1
 }
 
 image_ref() {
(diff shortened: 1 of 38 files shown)


contrib/auth/authentik/run.sh
```diff
@@ -438,33 +668,611 @@ run_in_repo() {
...
+        kind)
+            run_command kind export kubeconfig --name "${cluster_name}"
+            ;;
+        k3d)
+            run_command k3d kubeconfig merge "${cluster_name}" \
+                --kubeconfig-merge-default \
+                --kubeconfig-switch-context
+            ;;
+    esac
+    run_command kubectl config use-context "${context}"
```
tylersbray, line 949: > **agent** (`pr-review`): **P2** — `up k8s` mutates the developer's default kubectl context

`k8s_update_default_kubeconfig` (L925-939) runs `kind export kubeconfig` / `k3d kubeconfig merge --kubeconfig-merge-default --kubeconfig-switch-context` and then `kubectl config use-context`, and `k8s_up` calls it unconditionally (L1243).

Every other step in the k8s path already works against the isolated kubeconfig at `.generated/k8s/<cluster>/kubeconfig.yaml` — `k8s_helm_install`, `k8s_load_image`, `k8s_write_ca_bundle`, `k8s_wait_for_authentik`, and `k8s_start_port_forward` all pass `--kubeconfig`/`--kube-context` explicitly. So the harness does not need the default kubeconfig at all, but `run.sh up k8s` silently rewrites `~/.kube/config` and repoints the developer's current context at the Authentik cluster — including on reuse runs where no cluster was created. A developer who runs `up k8s` in one terminal and then `kubectl get pods` against dev-blue in another gets the wrong cluster with no warning.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 3 of 3 — ops / harness lifecycle. Remote-diff review at `dc85521`. Not reported by CodeRabbit or CodeQL.

<details>
<summary>Prompt for AI agents</summary>

```text
In contrib/auth/authentik/run.sh, stop mutating the user's default kubeconfig by
default. Make the k8s_update_default_kubeconfig call in k8s_up (around L1243)
opt-in:

1. Add a flag and env toggle, e.g. EXPORT_KUBECONFIG="${NMP_AUTHENTIK_K8S_EXPORT_KUBECONFIG:-0}"
   near the other K8S_* defaults, plus an --export-kubeconfig argument in the
   option parser and an entry in usage() under "Other options".
2. In k8s_up, guard the call:
       if [[ "${EXPORT_KUBECONFIG}" == "1" ]]; then
           k8s_update_default_kubeconfig "${cluster_name}" "${context}"
       fi
3. Leave k8s_update_default_kubeconfig itself unchanged.
4. In the success output of k8s_up, keep printing the "Kubeconfig: ${kubeconfig}"
   line and add a hint showing how to use it, e.g.
       echo "Use it with: KUBECONFIG=${kubeconfig} kubectl -n ${HELM_NAMESPACE} get pods"
5. Update the dry-run assertions in
   tests/auth_idp/static/test_authentik_kubernetes_demo.py that currently expect
   "kubectl config use-context kind-nmp-authentik-reuse",
   "kubectl config use-context kind-nmp-authentik-dev", and the k3d
   "kubeconfig merge ... --kubeconfig-switch-context" line: assert those strings
   are absent by default, and add a case that runs "up k8s --dry-run
   --export-kubeconfig" and asserts they are present.

Verify with: bash -n contrib/auth/authentik/run.sh and
contrib/auth/authentik/run.sh up k8s --dry-run --skip-image-load
```

</details>

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/config.py
```diff
@@ -143,6 +150,7 @@ def set_config(
       nemo config set --base-url https://api.example.com
       nemo config set --context staging --base-url https://nmp.staging.example.com
       nemo config set --context production --base-url https://nmp.example.com --activate
+      nemo config set --context local-tls --base-url https://localhost:8443 --certificate-authority ./ca.crt
```
tylersbray: > **agent** (`pr-review`): **P2** — relative CA path survives here, and saved paths are never normalized

Two related problems on this example line.

1. **Missed occurrence.** CodeRabbit's earlier comment on this PR asked for a stable absolute path in the `--certificate-authority` examples and listed three sites: `cli/commands/config_help.py:15-16`, `docs/cli/configuration.mdx:43`, and `docs/cli/connect-to-deployments.mdx:92-97`. Those three were fixed to `/path/to/ca.crt`, but this fourth occurrence — the `set_config` docstring example rendered by `nemo config set --help` — still shows `./ca.crt`.

2. **Root cause is unfixed.** The example is only a symptom: nothing normalizes the value on write. `set_config` puts the raw string into `params["certificate_authority"]` (L212-213) and `ConfigFile.ensure_context` assigns it verbatim to `cluster.certificate_authority` (`config/models.py:269-270`). `httpx` resolves `verify=<path>` against the *current process* cwd, and `run.sh` passes the same value to `curl --cacert`. So a context saved as `./ca.crt` from one directory silently fails TLS verification from any other directory, with an error that points at the certificate rather than at the path. Absolutizing on write fixes it for every entry point at once and makes the docs example a non-issue.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 2 of 3 — security / TLS propagation. Remote-diff review at `dc85521`. Item 1 is a partially-applied CodeRabbit finding; item 2 is new.

<details>
<summary>Prompt for AI agents</summary>

```text
Fix the certificate-authority path handling in two places.

1. packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/config.py,
   in the set_config docstring examples (around L153): change
     --certificate-authority ./ca.crt
   to
     --certificate-authority /path/to/ca.crt
   so it matches the wording already used in cli/commands/config_help.py and
   docs/cli/configuration.mdx.

2. Normalize the value when it is persisted so relative paths cannot be saved.
   In config.py set_config, where params["certificate_authority"] is populated
   (around L212-213), expand and absolutize a non-empty value before storing:
     if certificate_authority is not None:
         params["certificate_authority"] = (
             str(Path(certificate_authority).expanduser().resolve())
             if certificate_authority
             else None
         )
   Preserve the existing behavior where an explicit empty string clears the
   saved CA, and keep "certificate_authority is not None" as the has_options /
   params trigger so clearing still works.

3. Add tests in packages/nemo_platform_ext/tests/cli/commands/test_config.py
   next to test_set_certificate_authority_updates_selected_context_cluster:
   - running "config set --context X --certificate-authority ./ca.crt" from a
     tmp cwd persists an absolute path
   - a "~/..." value is expanded
   - an empty value still clears cluster.certificate_authority

Do not change the resolution-time behavior in config/config.py resolve(); only
normalize on write.
```

</details>

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -320,11 +320,11 @@ def _check_platform_reachable(base_url: str, timeout: float = 5.0) -> bool:
     Local ``nemo services run`` publishes ``/status``. Hosted deployments may
     only expose ``/cluster-info`` on ingress, so try both.
     """
-    verify = client_verify_from_env()
+    tls_config = httpx_tls_config_from_env()
```
tylersbray: > **agent** (`pr-review`): **P2** — `nemo setup` HTTPS probes still ignore the saved context CA

This PR threads `cluster.certificate_authority` through config resolution, SDK bootstrap, OAuth discovery, device flow, token refresh, and workload exchange. `nemo setup` was converted to the new `httpx_tls_config_from_env()` helper but is still called with no argument, so it only ever honors the `NMP_CLIENT_SSL_CERT_FILE` env var and never the CA saved in the active context:

- `_check_platform_reachable` (this line) and `_check_controller_health` (L485) — `httpx_tls_config_from_env()` with no CA
- `_hosted_platform_without_status` (L466-469) — inherits whatever the caller resolved
- `_agent_exists` (L1564), `_agents_api_ready` (L1577), and the demo-agent create/deploy calls (L1628, L1639, L1663) — bare `httpx.get`/`httpx.post` with no TLS kwargs at all

The user-visible path this breaks is the one this PR adds: `run.sh up k8s` registers context `authentik-k8s` with a saved `certificate_authority`, and after `nemo config use-context authentik-k8s` every `nemo auth`/SDK call works, but `nemo setup` reports the platform unreachable (and demo-agent deployment fails) unless the user also exports `NMP_CLIENT_SSL_CERT_FILE`. The same gap exists in `cli/commands/quickstart/cli.py` (L94, L319) and `cli/commands/services/cli.py` (L103); those are pre-existing, but they become reachable now that a private CA is a supported configuration rather than an env-var-only escape hatch.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 2 of 3 — security / TLS propagation. Remote-diff review at `dc85521`. Distinct from CodeRabbit's (already-fixed) `discover_nmp_config` verify-parameter comment on `client/factory.py`.

<details>
<summary>Prompt for AI agents</summary>

```text
In packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py,
propagate the active context's certificate_authority into the HTTPS probes,
mirroring what cli/commands/auth.py already does with
context.cluster.certificate_authority.

1. Add a keyword-only certificate_authority parameter (default None) to
   _check_platform_reachable (L317) and _check_controller_health (L475), and
   pass it into httpx_tls_config_from_env(certificate_authority) at L323 and
   L485. Forward it to _hosted_platform_without_status via the existing
   tls_config argument (no signature change needed there).
2. Add the same parameter to _check_platform_reachable_with_retries (L335) and
   forward it at L347.
3. Update the call sites (L368, L991, L1111, L1169, L2209) to resolve the CA
   once from cli_context.get_sdk_context().cluster.certificate_authority where a
   CLIContext is in scope, and pass it through. Where no CLIContext is
   available, keep the current None default rather than loading config again.
4. Give the agent helpers the same treatment: _agent_exists (L1561),
   _agents_api_ready (L1574), _deploy_demo_agent / _deploy_demo_agent_impl and
   the httpx calls at L1628, L1639, L1663 should accept a certificate_authority
   argument and pass **httpx_tls_config_from_env(certificate_authority) to each
   httpx.get/httpx.post, matching the pattern in auth/helpers.py
   discover_nmp_config.
5. Keep env precedence intact: httpx_tls_config_from_env already lets
   NMP_CLIENT_SSL_CERT_FILE win over the context value, so do not reorder that.

Add coverage in packages/nemo_platform_ext/tests/cli/commands/test_setup.py that
a context with a saved certificate_authority and no NMP_CLIENT_SSL_CERT_FILE set
(monkeypatch.delenv) results in httpx being called with verify=<context CA> for
the reachability and controller-health probes.

Leave quickstart/cli.py and services/cli.py out of scope for this PR unless you
also add tests for them; if you skip them, note it in the PR description.
```

</details>

contrib/auth/authentik/run.sh
```diff
@@ -438,33 +668,611 @@ run_in_repo() {
...
+    local args
+
+    [[ "${pid}" =~ ^[0-9]+$ ]] || return 1
+    kill -0 "${pid}" 2>/dev/null || return 1
+    args="$(ps -p "${pid}" -o args= 2>/dev/null || true)"
+    [[ "${args}" == *"kubectl"* ]] || return 1
+    [[ "${args}" == *"port-forward"* ]] || return 1
+    [[ "${args}" == *"svc/nemo-platform-envoy"* ]] || return 1
+    if [[ -n "${expected_port}" ]]; then
+        [[ "${args}" == *"${expected_port}:8080"* ]] || return 1
```
tylersbray: > **agent** (`pr-review`): **P3** — port-forward reuse check matches the port as a substring

Follow-up on CodeRabbit's earlier request to validate the recorded port before reusing a port-forward: the check landed, but it is a substring match against the full argv, so it accepts ports that merely end the recorded one. `expected_port=82` matches the argv fragment `18082:8080`, and `expected_port=8082` matches `18082:8080` as well. This is the guard that decides whether to reuse or restart the forward, so a false positive means `wait_for_https_ready` polls a port nothing is listening on and the run fails with a readiness timeout instead of restarting the forward.

Low likelihood in practice (`choose_free_tcp_port` returns high ephemeral ports and the collision needs a suffix relationship), but the fix is one line and it is the safety rail for the whole reuse path.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 1 of 3 — correctness. Remote-diff review at `dc85521`. Follow-up on the implementation of CodeRabbit's port-reuse comment, not a duplicate of it.

<details>
<summary>Prompt for AI agents</summary>

```text
In contrib/auth/authentik/run.sh, k8s_port_forward_pid_is_running (around
L1131-1145), make the expected-port check match a whole argv token instead of a
substring. The port-forward mapping is always passed as its own argument, so pad
the haystack and the needle with spaces:

-        [[ "${args}" == *"${expected_port}:8080"* ]] || return 1
+        [[ " ${args} " == *" ${expected_port}:8080 "* ]] || return 1

Keep the surrounding kubectl / port-forward / svc/nemo-platform-envoy checks and
the PID-format and kill -0 guards exactly as they are.

Then update the string-matching assertion in
tests/auth_idp/static/test_authentik_kubernetes_demo.py in
test_authentik_kubernetes_port_forward_pid_reuse_checks_process_command, which
currently asserts the literal:
    '[[ "${args}" == *"${expected_port}:8080"* ]]'
so it matches the new padded form.

Verify with: bash -n contrib/auth/authentik/run.sh
```

</details>

tests/auth_idp/runtime_kubernetes.py
```diff
@@ -55,8 +77,9 @@
 CLUSTER_DELETE_TIMEOUT_SECONDS = 180
 ROLLOUT_STATUS_TIMEOUT = "240s"
 ROLLOUT_COMMAND_TIMEOUT_SECONDS = 300
-HELM_WAIT_TIMEOUT = "10m"
-HELM_UPGRADE_COMMAND_TIMEOUT_SECONDS = 900
+HELM_WAIT_TIMEOUT = os.environ.get("NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT", "20m")
+HELM_UPGRADE_COMMAND_GRACE_SECONDS = 300
+HELM_UPGRADE_COMMAND_TIMEOUT_SECONDS = _duration_seconds(HELM_WAIT_TIMEOUT) + HELM_UPGRADE_COMMAND_GRACE_SECONDS
```
tylersbray: > **agent** (`pr-review`): **P3** — a bad `NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT` fails at import time

Follow-up on CodeRabbit's request to derive this from `HELM_WAIT_TIMEOUT`: the derivation landed, but `_duration_seconds` now runs at module import, and it raises `ValueError(f"invalid duration: {value}")` for anything it cannot parse. `HELM_WAIT_TIMEOUT` comes straight from the environment (L80), so `NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT=20min` or a stray quote turns into a collection error in `tests/auth_idp/runtime_kubernetes.py` — every test in the auth-idp suite errors out, and the traceback points at a module-level constant rather than at the environment variable the operator set. `run.sh` forwards this variable into pytest (L1505), so the harness is a live source of the bad value.

Also note `_duration_seconds` rejects `0` (`total == 0` raises) and accepts helm-invalid orderings like `30s5m`. Neither matters much, but the error path is what turns a typo into a confusing failure.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 3 of 3 — ops / harness lifecycle. Remote-diff review at `dc85521`. Follow-up on the implementation of CodeRabbit's timeout-derivation comment.

<details>
<summary>Prompt for AI agents</summary>

```text
In tests/auth_idp/runtime_kubernetes.py, make an invalid
NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT fail with a clear, actionable message
instead of breaking module import.

Around L79-82, wrap the derivation so the offending environment variable is
named in the error:

    HELM_WAIT_TIMEOUT_DEFAULT = "20m"
    HELM_WAIT_TIMEOUT = os.environ.get(
        "NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT", HELM_WAIT_TIMEOUT_DEFAULT
    )
    HELM_UPGRADE_COMMAND_GRACE_SECONDS = 300
    try:
        _helm_wait_seconds = _duration_seconds(HELM_WAIT_TIMEOUT)
    except ValueError as exc:
        raise RuntimeError(
            "NMP_AUTHENTIK_K8S_HELM_WAIT_TIMEOUT must be a helm duration such as "
            f"'20m', '1h30m', or a plain number of seconds; got {HELM_WAIT_TIMEOUT!r}"
        ) from exc
    HELM_UPGRADE_COMMAND_TIMEOUT_SECONDS = (
        _helm_wait_seconds + HELM_UPGRADE_COMMAND_GRACE_SECONDS
    )

Add unit coverage for _duration_seconds in the existing static test module
(tests/auth_idp/static/test_authentik_kubernetes_demo.py already imports this
module) covering: "20m" -> 1200, "1h30m" -> 5400, "900" -> 900, and that an
invalid value such as "20min" raises ValueError from _duration_seconds.

Do not change the existing default of 20m or the 300-second grace margin.
```

</details>

contrib/auth/authentik/run.sh
```diff
@@ -438,33 +668,611 @@ run_in_repo() {
...
+        printf "+ write %q\n" "${ca_bundle}"
+        return
+    fi
+
+    encoded="$(
+        kubectl --kubeconfig "${kubeconfig}" --context "${context}" \
+            -n "${HELM_NAMESPACE}" get secret nemo-platform-envoy-tls -o "jsonpath={.data.ca\\.crt}"
+    )"
+    if [[ -z "${encoded}" ]]; then
+        die "secret nemo-platform-envoy-tls in ${HELM_NAMESPACE} has no ca.crt entry"
```
tylersbray: > **agent** (`pr-review`): **P3** — runtime failures print the entire `--help` text

`die` (L221-226) prints the error, then the full `usage()` output, then exits 2. That is right for argument errors, but this PR routes genuine runtime failures through it too, so a missing Kubernetes secret key buries a one-line diagnostic under ~80 lines of usage text — exactly when the operator wants the error to be the last thing on screen.

Other runtime `die` sites added or reachable in this PR: the base64 decode failure just below (L1105), `k8s_wait_for_port_forward_ready` (L1172) — which prints a 40-line log tail and *then* the usage block — and `wait_for_https_ready`'s missing-`curl` check (L727).

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 3 of 3 — ops / harness lifecycle. Remote-diff review at `dc85521`. Not reported by CodeRabbit or CodeQL.

<details>
<summary>Prompt for AI agents</summary>

```text
In contrib/auth/authentik/run.sh, separate usage errors from runtime failures.

1. Add a fail() helper next to die() (around L221-226) that reports an error and
   exits without printing usage:

    fail() {
        echo "error: $*" >&2
        exit 1
    }

   Leave die() unchanged for argument/usage errors.

2. Switch the runtime failure sites from die to fail:
   - k8s_write_ca_bundle: the missing ca.crt entry (L1097) and the base64 decode
     failure (L1105)
   - k8s_wait_for_port_forward_ready: the readiness timeout (L1172), so the
     port-forward log tail stays the last thing printed
   - wait_for_https_ready: the "curl is required" check (L727)
   - ensure_gateway_tls_certificate: the "openssl is required" check (L527)

   Keep die() for everything in the argument parser and for the
   validate_* functions (validate_k8s_runtime, validate_instance_key,
   validate_k8s_gateway_port, validate_test_lifecycle), and for the "down target
   must be compose, k8s, or all" case.

3. If any dry-run test asserts on exit code 2 for these runtime paths, update it
   to 1. Check tests/auth_idp/static/test_authentik_kubernetes_demo.py for
   returncode assertions before changing anything.

Verify with: bash -n contrib/auth/authentik/run.sh
```

</details>

sdk/python/nemo-platform/src/nemo_platform/_client.py
```diff
@@ -226,7 +226,7 @@ def __init__(
         """
         env_base_url = os.environ.get("NEMO_PLATFORM_BASE_URL")
         bootstrap_base_url = base_url if base_url is not None else env_base_url
-
+        client_verify = client_verify_from_env()
```
tylersbray, line 229: > **agent** (`pr-review`): **P3** — confirm this Stainless-generated file's edit survives regeneration

This hunk (and the matching async one at L628) hand-edits a Stainless-generated file so that `client_verify` is seeded from the environment before bootstrap and then replaced by `client_init_kwargs.client_verify` (L256, L655) — the mirror of the change in `client/enhanced.py`.

I verified the change itself is sound at this HEAD: `_client.py` bootstraps through `nemo_platform_ext.client.factory.build_client_init_kwargs`, and this PR adds `client_verify` to that `ClientInitConfig`, so there is no attribute mismatch.

The concern is durability. Per `AGENTS.md`, `make update-sdk` regenerates this file via `./sdk/stainless.sh sync`, and the only post-generation step in the Makefile is `post-generation update-license-headers`. `sdk/**` is also excluded from CodeRabbit review by path filter, so nothing here would catch a silent revert. If this customization is not represented in the Stainless config or custom-code source, the next SDK regeneration drops it and the saved-CA path quietly stops working for direct SDK users — with no test failure, since the ext-side tests exercise `enhanced.py`.

**Reviewer:** `cursor-agent` (Cursor Auto-routed model), triple-review pass 1 of 3 — correctness. Remote-diff review at `dc85521`. No bot flagged this because `sdk/**` is excluded from CodeRabbit by path filter.

<details>
<summary>Prompt for AI agents</summary>

```text
This is a verification task, not a code change. Confirm that the client_verify
bootstrap edit in sdk/python/nemo-platform/src/nemo_platform/_client.py (sync
path around L229/L256, async path around L628/L655) survives SDK regeneration.

1. Read AGENTS.md "SDK Generation" and inspect sdk/stainless.sh sync plus any
   Stainless config in the repo (look for .stainless*, sdk/*.yml, or a
   custom-code / patches directory) to determine where hand-written additions to
   generated SDK files are declared.
2. If the customization is declared there, confirm the declared source matches
   what this PR added, and note in the PR description that it is covered.
3. If it is not declared, choose one:
   a. Add it to the Stainless custom-code source so regeneration reproduces it, or
   b. Move the logic out of the generated file: have _client.py delegate the
      verify decision to a hand-maintained module (the same way it already
      imports build_client_init_kwargs from nemo_platform_ext.client.factory and
      client_verify_from_env from nemo_platform_plugin.client.tls), so the
      generated file needs no edit.
4. Either way, add a regression test that constructs a client from a config with
   a saved cluster certificate_authority and NMP_CLIENT_SSL_CERT_FILE unset, and
   asserts DefaultHttpxClient was called with verify=<context CA>, so a future
   regeneration that drops this breaks a test instead of failing silently.

Do not run make update-sdk as part of this task.
```

</details>

contrib/auth/authentik/run.sh
```diff
@@ -438,33 +668,611 @@ run_in_repo() {
...
+        "${AUTHENTIK_ROOT}/helm"
+        --namespace
+        "${HELM_NAMESPACE}"
+        --create-namespace
+        --wait
+        --wait-for-jobs
+        --timeout
+        "${HELM_WAIT_TIMEOUT}"
+        --set
+        "nemo-platform.api.image.repository=${registry}/nmp-api"
```
mckornfield, line 1047: would this be cleaner as just a tmpdir values file?

contrib/auth/authentik/run.sh
mckornfield, line 1: overall comment is I saw some python snippets in here, and 1000 lines of bash I know would be even more in Python, but wonder if a couple of composable python scripts into a CLI would be an improvement 🤷 or 1000 lines of bash, I know it's a lot of binary calls too

packages/nemo_platform_ext/src/nemo_platform_ext/client/enhanced.py
```diff
@@ -420,7 +421,6 @@ async def main() -> None:
         if base_url is None:
             raise RuntimeError("NeMoPlatform client initialization failed: base_url is required")
 
-        client_verify = client_verify_from_env()
```
mckornfield, line 429: was this just moved up?

packages/nemo_platform_ext/src/nemo_platform_ext/client/tls.py
```diff
@@ -6,11 +6,51 @@
...
+
+
+def client_certificate_authority_from_env(
+    certificate_authority: str | None = None,
+    env: Mapping[str, str] | None = None,
+    *,
+    cert_file_envvars: Sequence[str] = (NMP_CLIENT_SSL_CERT_FILE_ENVVAR,),
+) -> str | None:
+    """Return the configured CA bundle path, if one is configured."""
+    for envvar in cert_file_envvars:
```
mckornfield, line 28: so it looks but the default is just one env var?

sdk/python/nemo-platform/src/nemo_platform/_client.py
mckornfield: should we not ignore this one? this one doesn't appear to be "generated"

mckornfield (approved):

contrib/auth/authentik/run.sh
ironcommit: yeah, it's mostly binary calls, and python is terrible with that.  I'd love to rip the script out, it's just so darn helpful.

packages/nemo_platform_ext/src/nemo_platform_ext/client/enhanced.py
ironcommit: yup

packages/nemo_platform_ext/src/nemo_platform_ext/client/tls.py
ironcommit: by default, yes.

svvarom (approved):
workflow changes LGTM

PR #1520, opened 2026-08-25
feat(anonymizer): make run remote jobs

ironcommit:
## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

* **New Features**
  * Added unified `nemo anonymizer preview` and `run` commands.
  * Added streaming previews through the CLI, Python SDK, and HTTP API.
  * Added local CSV/Parquet uploads through filesets for remote execution.
  * Added request inspection, dry runs, job monitoring, and artifact downloads.
  * Added example preview and run specifications.

* **Changes**
  * Full anonymization runs now execute through NeMo Platform Jobs.
  * Preview and run requests require model configurations for platform execution.

* **Documentation**
  * Updated tutorials, SDK references, workflows, and input/artifact guidance.

### DIFF
--- a/docs/anonymizer/index.mdx
+++ b/docs/anonymizer/index.mdx
@@ -11,7 +11,7 @@ The Anonymizer service detects personally identifiable information (PII) in text
 
 ## Overview
 
-The service wraps the open-source [NVIDIA NeMo Anonymizer library](https://github.com/NVIDIA-NeMo/Anonymizer) and exposes it through the NeMo Platform's Python SDK and CLI. The library still owns PII detection, replacement, rewrite, and config validation. The platform adds inference routing through the Inference Gateway, fileset-backed inputs, plugin-service execution for streaming preview, and a Jobs-worker path for full anonymization runs.
+The service wraps the open-source [NVIDIA NeMo Anonymizer library](https://github.com/NVIDIA-NeMo/Anonymizer) and exposes preview through the CLI, NeMo Platform Python SDK, and HTTP API, with CLI and SDK support for full anonymization jobs. The library still owns PII detection, replacement, rewrite, and config validation. The platform adds inference routing through the Inference Gateway, fileset-backed inputs, CLI upload/download conveniences, plugin-service execution for streaming preview, and a Jobs-worker path for full anonymization runs.
 
 ## How It Works: Library + Platform
 
@@ -62,14 +62,23 @@ preview_result.trace_dataset     # detection trace
 preview_result.display_record(0) # render a record with entity highlights
 ```
 
-For a full anonymization run, execute the job locally or submit it to the Jobs worker:
+The same preview request can be saved as YAML and streamed from the CLI:
 
 ```bash
-nemo anonymizer run run --spec-file /path/to/run-spec.yaml      # in-process
-nemo anonymizer run submit --spec-file /path/to/run-spec.yaml   # NeMo Services job
+nemo anonymizer preview --spec-file plugins/nemo-anonymizer/examples/redact.yaml --num-records 2 --fileset anonymizer-inputs
+nemo anonymizer preview --spec-file plugins/nemo-anonymizer/examples/redact.yaml --num-records 2 --fileset anonymizer-inputs --output-file preview.ndjson --quiet
 ```
 
-The SDK equivalent of `run submit` is `sdk.anonymizer.run(request)`, which returns an `AnonymizerJobResource` you can poll with `wait_until_done()` and pull artifacts from with `download_artifacts()`.
+For a full anonymization run, submit the job to the Jobs worker:
+
+```bash
+nemo anonymizer run --spec-file plugins/nemo-anonymizer/examples/redact.yaml --fileset anonymizer-inputs          # NeMo Platform job
+nemo anonymizer run --spec-file plugins/nemo-anonymizer/examples/redact.yaml --fileset anonymizer-inputs --watch  # submit and stream logs/status
+nemo anonymizer run --spec-file plugins/nemo-anonymizer/examples/redact.yaml --fileset anonymizer-inputs --watch --output-dir ./anonymizer-artifacts  # download artifacts after success
+nemo anonymizer run --spec-file plugins/nemo-anonymizer/examples/redact.yaml --fileset anonymizer-inputs --dry-run  # print schema/request without submitting
+```
+
+The SDK equivalent of `nemo anonymizer run` is `sdk.anonymizer.run(request)`, which returns an `AnonymizerJobResource` you can poll with `wait_until_done()` and pull artifacts from with `download_artifacts()`.
 
 **The platform handles:** Inference routing through the Inference Gateway, fileset-backed inputs, and authentication.
 
@@ -80,9 +89,9 @@ When using Anonymizer as a NeMo Platform service:
 | Feature           | Standalone Library                                  | NeMo Platform Service                                                                                       |
 |-------------------|-----------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
 | **Inference**     | Direct calls to NVIDIA Build defaults               | Routes through the Inference Gateway via `model_configs`                                                        |
-| **Execution**     | Local Python process                                | Streaming preview runs in the plugin service; full runs execute either in the local CLI (`run run`) or on the Jobs worker (`run submit`) |
-| **Input sources** | Local file, `http(s)` URL                           | Local file (`run run` only), `http(s)` URL, or NeMo Platform Fileset                                        |
-| **Artifacts**     | Local filesystem                                    | Local artifact directory (`persistent/results/artifacts`) for `run run`; NeMo Platform job artifact storage for `run submit` |
+| **Execution**     | Local Python process                                | Streaming preview runs in the plugin service; full runs execute on the Jobs worker |
+| **Input sources** | Local file, `http(s)` URL                           | CLI can upload local CSV/Parquet files to `--fileset`; SDK/API use `http(s)` URLs or NeMo Platform filesets |
+| **Artifacts**     | Local filesystem                                    | NeMo Platform job artifact storage for full `run`; CLI can download run artifacts with `--output-dir` |
 | **Authentication**| Direct API keys                                     | NeMo Platform Secrets service                                                                               |
 
 ## Replacement Strategies
@@ -103,11 +112,11 @@ See the [library documentation](https://github.com/NVIDIA-NeMo/Anonymizer/tree/m
 
 This package is a thin wrapper around the [NVIDIA NeMo Anonymizer library](https://github.com/NVIDIA-NeMo/Anonymizer). It does **not** re-document detection, replacement, or rewrite semantics. It adds:
 
-- A `nemo anonymizer` CLI with `validate`, `preview`, and `run` command groups.
+- A `nemo anonymizer` CLI with `preview`, `validate`, and `run` commands.
 - An `sdk.anonymizer` SDK accessor (`AnonymizerResource`, `AsyncAnonymizerResource`).
-- A streaming `anonymizer.preview` function that emits `preview_dataset`, `trace_dataset`, and `failed_records` frames from the plugin service.
-- An `anonymizer.run` job that writes `dataset.parquet`, `trace.parquet`, `metadata.json`, and optional `failed_records.json`. The job can execute in the local CLI process (`nemo anonymizer run run`) or on the NeMo Platform Jobs worker (`nemo anonymizer run submit` / `sdk.anonymizer.run`).
-- Fileset input handling (`fileset://<workspace>/<fileset>#<path>`).
+- A streaming preview API that emits `preview_dataset`, `trace_dataset`, and `failed_records` frames from the plugin service.
+- An `anonymizer.run` job that writes `dataset.parquet`, `trace.parquet`, `metadata.json`, and optional `failed_records.json`. The job executes on the NeMo Platform Jobs worker through `nemo anonymizer run` or `sdk.anonymizer.run`.
+- Fileset input handling (`fileset://<workspace>/<fileset>#<path>`) and CLI upload of local CSV/Parquet inputs when `--fileset` is supplied.
 - Inference Gateway routing for model providers referenced from `model_configs`.
 
 ## Next Steps
@@ -116,7 +125,7 @@ This package is a thin wrapper around the [NVIDIA NeMo Anonymizer library](https
 
 <Card title="Tutorials" href="/documentation/anonymize-data/tutorials">
 
-Walk through preview (`anonymizer.preview`) and job execution (`anonymizer.run`) end to end.
+Walk through preview (`nemo anonymizer preview` / `anonymizer.preview`) and job execution (`anonymizer.run`) end to end.
 
 </Card>
 <Card title="SDK Resources" href="/documentation/anonymize-data/sdk-resources">
--- a/docs/anonymizer/tutorials/index.mdx
+++ b/docs/anonymizer/tutorials/index.mdx
@@ -79,8 +79,8 @@ When using Anonymizer as a NeMo Platform service:
 | Feature        | Difference                                              | Details                                                                                |
 |----------------|---------------------------------------------------------|----------------------------------------------------------------------------------------|
 | **Inference**  | Routes through the Inference Gateway                    | Configure providers once and reference them by name from `model_configs`.              |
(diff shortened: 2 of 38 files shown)


docs/anonymizer/tutorials/index.mdx
```diff
@@ -79,8 +79,8 @@ When using Anonymizer as a NeMo Platform service:
 | Feature        | Difference                                              | Details                                                                                |
 |----------------|---------------------------------------------------------|----------------------------------------------------------------------------------------|
 | **Inference**  | Routes through the Inference Gateway                    | Configure providers once and reference them by name from `model_configs`.              |
-| **Input data** | Filesets and HTTP(S) URLs (local paths only in local CLI execution) | Use `sdk.files.filesets.create` / `sdk.files.upload`, then reference with `#<path>`. |
-| **Artifacts**  | Local or platform-managed                               | `run run` writes to `persistent/results/artifacts` locally; `run submit` stores artifacts in NeMo Platform job storage. |
+| **Input data** | Filesets and HTTP(S) URLs for full runs; local paths for local preview | Use `sdk.files.filesets.create` / `sdk.files.upload`, then reference with `#<path>`. |
```
mikeknep: We should drop the run/submit verbs from the `preview` NemoFunction as well, and only support previewing via the API

packages/nemo_platform_plugin/src/nemo_platform_plugin/verbs.py
```diff
@@ -0,0 +1,368 @@
...
+            service=service,
+            verb=verb,
+            lifecycle=cast(JobVerbLifecycle, lifecycle),
+            job_cls=job_cls,
+            input_schema=job_cls.input_spec_schema or job_cls.spec_schema,
+        )
+    return registrations
+
+
+def install_remote_job_verb(
```
mikeknep: Does this compose nicely with the existing CLI command signature override mechanism? Data Designer currently overrides the default in order to support the upstream library's friendlier argument for configs (loading from Python files). We need to keep that, but it'd probably be nice to pick up `--watch` "for free" as well.

docs/anonymizer/tutorials/run.mdx
```diff
@@ -90,42 +91,53 @@ spec_path.write_text(yaml.safe_dump(request.model_dump(mode="json", exclude_none
...
 
-Choose one execution path. Option A runs in the local CLI process. Option B submits the same request to the NeMo Platform Jobs worker.
-
-### Option A: Run Locally
+Submit the spec to the NeMo Platform Jobs worker:
 
 ```bash
-nemo anonymizer run run --spec-file /tmp/anonymizer-run.yaml
+nemo anonymizer run \
+  --spec-file /tmp/anonymizer-run.yaml \
```
mckornfield: any reason to use `/tmp` over just a relative directory?

docs/anonymizer/index.mdx
```diff
@@ -80,9 +83,9 @@ When using Anonymizer as a NeMo Platform service:
...
 | **Inference**     | Direct calls to NVIDIA Build defaults               | Routes through the Inference Gateway via `model_configs`                                                        |
-| **Execution**     | Local Python process                                | Streaming preview runs in the plugin service; full runs execute either in the local CLI (`run run`) or on the Jobs worker (`run submit`) |
-| **Input sources** | Local file, `http(s)` URL                           | Local file (`run run` only), `http(s)` URL, or NeMo Platform Fileset                                        |
-| **Artifacts**     | Local filesystem                                    | Local artifact directory (`persistent/results/artifacts`) for `run run`; NeMo Platform job artifact storage for `run submit` |
+| **Execution**     | Local Python process                                | Streaming preview runs in the plugin service; full runs execute on the Jobs worker |
+| **Input sources** | Local file, `http(s)` URL                           | Local file for `preview run`; `http(s)` URL or NeMo Platform Fileset for full `run` |
```
mckornfield: think I'm with mike's comment above, `preview run` looks kinda strange to me, why wouldn't it just be `preview`?

packages/nemo_platform_plugin/src/nemo_platform_plugin/verbs.py
```diff
@@ -0,0 +1,368 @@
...
+            "workspace",
+            str,
+            typer.Option("default", "--workspace", help="Workspace scope.", rich_help_panel="Submission"),
+        ),
+        kw(
+            "print_request",
+            bool,
+            typer.Option(
+                False,
+                "--print-request",
```
mckornfield: idk why this seems weird to me as an option. seems like something that would go under `verbose`?

plans/2026-08-24-anonymizer-remote-run-cli.md
mckornfield, line 1: do we want to commit this?

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/skills/anonymizer/workflows/interactive.md
```diff
@@ -10,14 +10,14 @@ Source of truth for this workflow: `docs/anonymizer/tutorials/index.mdx`, `docs/
...
    - If the output is a path, use `<path> anonymizer` as the command prefix for all `nemo anonymizer …` invocations in this workflow.
    - If the output is `CLI_NOT_FOUND`, STOP and follow the Troubleshooting section in SKILL.md. Do not continue.
-2. **Confirm the plugin service is mounted (only if the user wants `preview submit` or `run submit`).** Run `curl -s http://localhost:8080/openapi.json | jq -r '.paths | keys[]' | grep '^/apis/anonymizer/'`. If nothing prints, the plugin service isn't loaded — `nemo setup` does not auto-mount it. Tell the user to run `nemo services run` (no `--services` flag) and rerun the check. Local previews (`preview run`) and local runs (`run run`) do **not** need the plugin service mounted.
+2. **Confirm the plugin service is mounted (only if the user wants `preview submit` or a full `run`).** Run `curl -s http://localhost:8080/openapi.json | jq -r '.paths | keys[]' | grep '^/apis/anonymizer/'`. If nothing prints, the plugin service isn't loaded — `nemo setup` does not auto-mount it. Tell the user to run `nemo services run` (no `--services` flag) and rerun the check. Local previews (`preview run`) do **not** need the plugin service mounted.
```
mckornfield: have these docs gotten any "human" touch?

docs/anonymizer/index.mdx
mikeknep: To be clear, this is not just a matter of CLI aesthetics. We should deliberately **not** support things like local files as input sources, which was only possible in the service-less `run` mode. Doing so has been a headache for both DD and Anon: for example not just dealing with local files as seed data, but also trying to reconcile a model provider reference as **either** "something known to IGW" or "a model config defined in the local standalone library's YAML config files". It was complicated to implement, and confusing to explain.

If we don't drop the legacy verbs / local `run` mode for `preview`, we have to keep all that complexity, but only for preview and not for run, which would be weird in a different way because that would mean users could make "valid, preview-able" configs that then fail when submitted for full-scale execution on jobs.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/skills/anonymizer/SKILL.md
```diff
@@ -36,21 +36,20 @@ Read **only** the workflow file that matches the selected mode, then follow it:
...
+- SDK/API preview and run require an `http(s)` URL or a fileset reference (`<workspace>/<fileset>#<path>` or `fileset://...`) in `data.source`.
 - If a spec file matching the user's description already exists in the working directory, ask whether to edit it or create a new one.
 
 # Usage Tips and Common Pitfalls
 
 - **Replacement strategies need a discriminated payload.** Hand-written YAML specs must include `kind: redact` (or `annotate` / `hash` / `substitute`) inside the `replace` block.
 - **Substitute and rewrite need LLM-backed model aliases.** For plugin-service / Jobs execution they must be backed by providers declared in `model_configs`. For library-level details, refer to the [Anonymizer library docs](https://github.com/NVIDIA-NeMo/Anonymizer/tree/main/docs) or library skills.
-- **Spec files are YAML, not JSON.** `nemo anonymizer preview run --spec-file <path>` and `nemo anonymizer run run --spec-file <path>` both load YAML.
+- **CLI spec files are YAML, not JSON.** `nemo anonymizer preview --spec-file <path>` and `nemo anonymizer run --spec-file <path>` load YAML. Direct HTTP preview still uses JSON.
```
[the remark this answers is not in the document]
ironcommit: not applicable

plans/2026-08-24-anonymizer-remote-run-cli.md
ironcommit: removed

packages/nemo_platform_plugin/src/nemo_platform_plugin/verbs.py
ironcommit: It was actually my attempt to replace the explain verb.  "explain" would print out the input schema, so it can be programmatically built and validated.  I don't think we really need this right now, but I wanted to preserve the functionality.  --explain also doesn't make much sense.

docs/anonymizer/tutorials/run.mdx
ironcommit: replaced.

packages/nemo_platform_plugin/src/nemo_platform_plugin/verbs.py
ironcommit: That's the idea, but I'm still working out the details.

docs/anonymizer/tutorials/index.mdx
ironcommit: I thought about this, but I have an idea of how I want to do this.  So I'm going to give it a try.

docs/anonymizer/index.mdx
ironcommit: preview run has been replaced properly now.    And the local provider stuff has been removed.

plugins/nemo-anonymizer/examples/redact-preview.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+config:
+  replace:
+    kind: redact
+    format_template: "[REDACTED_{label}]"
+data:
+  source: "plugins/nemo-anonymizer/examples/anonymizer-input.csv"
+  text_column: biography
+  id_column: id
+num_records: 2
```
mikeknep: This is the only difference between the `-preview.yaml` and `-run.yaml` files, and I don't think is particularly necessary since the data source is only two lines anyways. Maybe drop this line and consolidate into a single `examples/redact.yaml` file?

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/jobs/run.py
```diff
@@ -54,10 +56,10 @@ async def to_spec(
         workspace: str,
         entity_client: object,
         async_sdk: AsyncNeMoPlatform,
-        is_local: bool,
+        **_: object,
```
mikeknep: TIL, never seen this before. Is this added because 1) the platform-wide plumbing will continue to supply `is_local` to every `to_spec` call, but 2) this implementation no longer needs that argument? Neat

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/app/context.py
mikeknep, line 1: This same pattern exists on the DD side and I punted it to a follow-up PR, but nice to see here for Anonymizer 👍 I will make these and more DD changes after this lands

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/sdk/resources.py
```diff
@@ -179,19 +179,21 @@ def preview(
...
+        returns an ``AnonymizerPreviewResult`` once the stream finishes.
         """
         with _PreviewFrameCollector() as collector:
-            for frame in self._preview(request=request, workspace=workspace):
+            for frame in self.preview_stream(request, workspace=workspace):
                 collector.accept(frame)
             return _build_preview_result(collector)
 
-    def _preview(
+    def preview_stream(
```
mikeknep, line 190: My first pass through the PR, the diff in this module seemed like trivial refactoring noise for no particular purpose—"maybe for unit testing purposes" I thought. I see now that this method is being used in the CLI. I'll open a separate comment thread for thoughts on the CLI changes, but from the perspective of this SDK resource: I don't particularly like making this method public because there's no reason for end users of the SDK to use this method—they should always just use `preview`, but having two public methods introduces some confusion ("which should I use when/why?").

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/cli.py
```diff
@@ -92,3 +142,626 @@ def validate_command(
     anonymizer = _make_local_anonymizer(model_configs=str(model_configs) if model_configs else None)
     anonymizer.validate_config(anonymizer_config)
     typer.echo("Config is valid.")
+
+
+def _install_anonymizer_run_verb(
```
mikeknep, line 222: Something seems wrong with the `job_verb` concept if it requires plugin authors write and maintain a nearly 200-line method with this much domain-agnostic / generic-platform-boilerplate stuff

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/cli.py
```diff
@@ -92,3 +142,626 @@ def validate_command(
...
+
+
+@plugin_verb(
+    service="anonymizer",
+    verb="preview",
+    kind="service",
+    lifecycle="stable",
+    input_schema=PreviewRequest,
+)
+def preview_command(
```
mikeknep: Similar reaction here: this seems like way too much "stuff" for the plugin author to own. I'm also still wondering why this PR is moving `preview` away from the `NemoFunction` abstraction and making it a fully custom command instead... I can't tell if the "local file with `--fileset`" feature is what's driving that or something else; if it's just that I definitely don't think the local file convenience is worth the tradeoff.

I'm probably somewhat biased because I worked with Razvan on some of the existing stuff (e.g. `NemoFunction`, CLI signature and renderer overrides) because they were originally built primarily for Data Designer, whereas this is "new and different". So I'm trying to not fall victim to NIH syndrome. But that said, I do feel like when I look at the Data Designer plugin code today, the vast majority of the various customizations are relatively high-level and are in service of DD-specific aesthetics/ergonomics; there's no messing around with things like resolving base urls and paths and auth headers, or validating the logical permutations of generic arguments like `watch | dry_run | output_dir`.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/jobs/run.py
ironcommit: yeah, is_local was removed.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/sdk/resources.py
ironcommit: I guess I'm a sucker for streamable apis.  One is streaming, the other is sync.  I can make that more clear in the doc strings.

mikeknep: I don't think end users would ever care to iterate through an `Iterator[PreviewFrame]` themselves, at least in the Anonymizer (and Data Designer) preview endpoints. The preview frames are low-level implementation details of how the API sends preview result data over HTTP. The `_PreviewFrameCollector` handles each of the different frame types the only real way that is valuable: log the log frames, convert the data-related frames to pandas DataFrames, accept-but-ignore the heartbeat and done frames, raise on error frames. `preview` ultimately returns the friendly, interactive `AnonymizerPreviewResult`.

plugins/nemo-anonymizer/src/nemo_anonymizer_plugin/cli.py
ironcommit: What I'm trying to do here should be easy, but it's not.  It's a lot for the plugin owner, I agree.  I'm in favor of keeping it, just as a potential use case for improving our shared code.  I can take a look at what we might want to do here, but I think having plugin developers just write their own typer apps is really the simplest possible thing.  Trying to work with shared cli generators is never going to work for everyone.

ironcommit:
closed in favor of https://github.com/NVIDIA-NeMo/nemo-platform/pull/1637

PR #1531, opened 2026-08-25
fix(studio): unify entity icons behind a single canonical registry [ASTD-447]

aahunt-nv:
Resolves: [ASTD-447](https://linear.app/nvidia/issue/ASTD-447)

## Summary

Implements **Surface 3** from the audit — nav icon vs. empty-state icon divergence. (The two Agent sub-pages in Surface 1 are being removed separately.)

The audit found 6 match / 9 mismatch / 4 no-nav-icon across 19 entities. The cause is structural: there were two independent icon sources with nothing linking them.

- `web/packages/studio/src/routes/groups/*Routes.tsx` — each route group imported its own `lucide-react` glyph for `slotIcon`.
- `web/packages/common/src/components/EntityEmptyState/registry.ts` — each descriptor carried its own `icon` field.

Hand-fixing the 9 would drift again on the next entity added, so this makes the mismatch unrepresentable instead:

1. New `web/packages/common/src/constants/entityIcons.ts` holds `ENTITY_ICONS` — the one canonical entity → glyph map.
2. `icon` is **removed** from `EmptyStateDescriptor`. `EntityEmptyState` resolves `ENTITY_ICONS[entity]`. A descriptor can no longer specify a conflicting glyph.
3. All 14 route groups and `WorkspaceSideNav` render from the same map. No `slotIcon` in Studio imports its glyph locally anymore.

`ENTITY_ICONS` is a superset of the empty-state registry, since an entity can have an icon before it has empty-state copy. `EmptyStateEntityKey` narrows `EntityKey`, and `EntityEmptyState` accepts only the narrower type — so "every empty state resolves an icon" stays a compile-time guarantee.

## Icon decisions

Where an entity had a nav glyph, that glyph wins — it's the persistent, most-frequently-seen representation.

| Entity | Icon | Change |
| --- | --- | --- |
| `agents` | `Bot` | was `HatGlasses` |
| `agentMonitorRuns` | `Logs` | monitor runs are agent telemetry |
| `baseModels` | `Boxes` | was `FolderOpen` fallthrough |
| `customModels` | `Metronome` | was `BrainCircuit` |
| `virtualModels` | `Waypoints` | was `Radar`, which frees `Radar` for providers |
| `datasets` | `Database` | new entry — see below |
| `filesets` | `FileStack` | was `Database` |
| `anonymizerJobs` | `UserPen` | was `VenetianMask` |
| `dataDesignerJobs` | `Form` | was `Lightbulb`, which collided with Insights |
| `safeSynthesizerJobs` | `DatabaseCheck` | was `ShieldCheck` |
| `guardrails`, `guardrailChecks` | `ShieldCheck` | tests inherit the config glyph |
| `evaluationResults`, `agentEvaluations`, `evaluationSessions` | `ChartBar` | was `FlaskConical` |
| `evalComparison` | `GitCompare` | was `ChartNetwork` |
| `telemetryTraces`, `insightTraces` | `ListTree` | was `Anchor` |
| `telemetrySpans` | `Logs` | was `Anchor` |

Unchanged: `deployments` (`Rocket`), `inferenceProviders` (`Radar`), `experiments`/`insightExperiments` (`FlaskConical`), `optimizerInsights` (`Lightbulb`), `jobs` (`ListChecks`), `filesetFiles` (`FolderOpen`), `secrets` (`LockKeyhole`), `members` (`UsersRound`).

Four calls worth surfacing:

- **Guardrails and its tests share `ShieldCheck`.** A config and its test cases are one thing to the user, so the tests inherit the config glyph rather than claiming a second shield. This retires `ShieldKeyhole` and leaves the shield family meaning guardrails only — Safe Synthesizer moves to `DatabaseCheck`, its nav glyph.
- **`jobs` keeps `ListChecks`.** It is a persistent sidebar item; `guardrailChecks` renders in exactly one empty state nested two levels into a config, so the nav item has the stronger claim.
- **`agentMonitorRuns` shares `Logs` with `telemetrySpans`.** The Monitor list is populated from the `nemo-agent-telemetry` fileset, so it is the same underlying data — `Bot` is reserved for the agent itself.
- **`datasets` gets a canonical entry** even though it is a nav grouping today, because it is becoming an entity in its own right. It would otherwise be the one nav glyph left outside the map.

## Testing

- New `entityIcons.test.ts`: every empty-state entity resolves an icon, aliases point at their parent, and **no glyph is reused across unrelated entities** — the regression guard for the `ShieldCheck`/`Radar` collisions. Intentional aliases are declared in one table with a reason each, so adding a collision means editing that table on purpose.
  Verified non-vacuous: pointing `jobs` at `ShieldCheck` fails with `expected [['guardrails','jobs']] to deeply equal []`.
- `pnpm --filter @nemo/common test` — 1513 passed.
- `pnpm --filter nemo-studio-ui test` — 2960 passed. `AssistantChatRoute/api.test.ts` fails with `localStorage.clear is not a function`; verified the identical 18 failures on `main`, so pre-existing and unrelated.
- `typecheck` clean on both packages (catches any missed `descriptor.icon`), `lint` clean (catches orphaned `lucide-react` imports).

## Icon stroke weight

Goal: nav icons render at a true **1px** stroke, and scaling an icon up scales its stroke with it.

`.lucide` in `web/packages/studio/src/index.css` already sizes every icon to 16px. This sets `stroke-width: 1.5` and **removes** the adjacent `vector-effect: non-scaling-stroke` rule.

The `1.5` looks wrong for a 1px target, so the mechanic is worth stating: `stroke-width` is authored in lucide's 24-unit viewBox, not in pixels, so the rendered width is `stroke-width * (rendered size / 24)`. At the 16px default that makes `1.5` render as exactly 1px, and the weight now tracks the icon size:

| Icon size | Rendered stroke |
| --- | --- |
| 16px (nav default) | **1.00px** |
| 20px | 1.25px |
| 24px | 1.50px |
| 32px | 2.00px |
| 48px | 3.00px |

Dropping `non-scaling-stroke` is what buys that second property. It pins the stroke to the *unscaled* coordinate space, i.e. one absolute width at every size — measured at **0.125px whether the icon was 16px or 48px**. That is both far too faint to read and explicitly not what we want, since a 48px icon should not wear a 16px hairline.

An earlier commit on this branch authored `stroke-width: 1` while that rule was still in place, which is why it did not produce a 1px stroke; the follow-up commit fixes it. This supersedes the override that [#1369](https://github.com/NVIDIA-NeMo/nemo-platform/pull/1369) removed a week ago, so the flip-flop in `git log` is intentional rather than an accidental revert.

Measured, not assumed: my first pass at this reasoned about `non-scaling-stroke` backwards. The numbers above come from the running dev server, rasterizing a probe element that inherits the real `.lucide` cascade and summing per-pixel alpha coverage across the stroke at each size. Confirmed in the same session that the shipped nav icons compute to `stroke-width: 1.5px` with `vector-effect: none`.

## Summary by CodeRabbit

* **UI Improvements**
  * Standardized icons across sidebar navigation and empty states for a more consistent visual experience.
  * Updated icons for agents, models, datasets, deployments, evaluations, experiments, jobs, guardrails, and other workspace areas.
  * Refined icon stroke rendering to maintain consistent appearance at different sizes.

* **Quality**
  * Added coverage to verify icon assignments, entity coverage, aliases, and intentional icon reuse.


steramae-nvidia (approved):

web/packages/common/src/constants/entityIcons.ts
```diff
@@ -0,0 +1,96 @@
...
+  deployments: Rocket,
+  virtualModels: Waypoints,
+  inferenceProviders: Radar,
+
+  // Data
+  datasets: Database,
+  filesets: FileStack,
+  filesetFiles: FolderOpen,
+  anonymizerJobs: UserPen,
+  dataDesignerJobs: Form,
```
steramae-nvidia, line 67: nit: Can we change this to be a [palette](https://lucide.dev/icons/palette)? [Data Designer](https://github.com/NVIDIA-NeMo/DataDesigner) uses this icon on their repo

aahunt-nv: We can follow up on this. We need to align a few icons with other open-sourced tools.

PR #1532, opened 2026-08-25
feat(agents): add environment/spec/compute-spec CRUD to CLI + SDK, and --environment to deploy

benmccown:
## Summary

Surfaces the AgentEnvironment / AgentEnvironmentSpec / AgentComputeSpec entities through the nemo-agents plugin's hand-written CLI and SDK, so the whole request/fulfill environment flow is usable without raw HTTP. Before: environment specs/environments could only be created via the REST API, and `nemo agents deploy` had no way to attach an environment. After: full CRUD command groups exist and `deploy` accepts `--environment`.

## Changes

- `nemo agents deploy --environment <ws/name>` — deploy an agent under an AgentEnvironment (merged into the agent config; compute/secret refs snapshotted at create time).
- New CLI subgroups (plural, matching the existing `deployments` group), each with `create` / `list` / `get` / `delete`:
  - `nemo agents environment-specs`
  - `nemo agents environments`
  - `nemo agents compute-specs`
  - `create` accepts `--spec-file` (JSON/YAML) or inline `--spec` (JSON string); `environments create` also takes `--environment-spec` / `--compute-spec` ref flags for the common case.
- Plugin SDK (`nemo_agents_plugin.sdk`): new sync subresources `nemo.agents.environment_specs`, `nemo.agents.environments`, `nemo.agents.compute_specs` (create/list/get/delete), plus an `environment=` parameter on `nemo.agents.deployments.create(...)`.
- Unit tests for the new CLI commands and SDK resources.

Scope notes: agents remain out of the Stainless-generated SDK by design — this extends the plugin's own hand-written CLI/SDK only. An async SDK variant is out of scope (the plugin registers a sync resource today). No API routes/schemas changed, so OpenAPI and the generated CLI reference are unaffected (verified: `generate-cli-commands` and the reference-docs generator produce no diff).

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs enumerate the nemo-agents plugin CLI (its commands are hand-written, not in the generated CLI reference); command help strings are inline.

## Verification

Targeted validation:

- `uv run --frozen pytest plugins/nemo-agents/tests/unit/ -q` → 1356 passed (includes 12 new CLI/SDK tests).
- `uv run ruff check` + `uv run ruff format --check` on the four changed files → clean.
- `uv run --frozen ty check` on the changed src files → All checks passed.
- `uv run pre-commit run -a` → all relevant hooks passed (ruff, ruff format, ty, config-reference doc, copyright headers, plugin-import guard, merge-conflict check). Two hooks failed for environment-only reasons unrelated to this change and modified no files: `Helm Docs` (helm-docs binary not installed locally; no Helm files changed) and `Run uv lock with platform uv` (local uv 0.9.21 vs pinned 0.9.14; no pyproject.toml changed and `uv.lock` drift check passed).
- Manual smoke test against a local platform: `environment-specs create --spec-file`, `environments create --environment-spec` (ref) and `--spec` (inline compute), and `deploy --environment` all created the entities and produced the correct create-time merge/snapshot (secrets lifted out of config, spec-wins env, compute box applied).

## Summary by CodeRabbit

* **New Features**
  * Added CLI and SDK support for creating, listing, retrieving, and deleting environment specs, environments, and compute specs.
  * Added inline JSON or JSON/YAML file input with validation for specification creation.
  * Deployments can reference saved environments or inline environment definitions.
  * Added workspace-aware resource management and specification references.

* **Changes**
  * Removed the `ethos migrate` command.
  * Updated agent creation and fileset handling with spec-package guidance.

* **Tests**
  * Added coverage for resource workflows, deployment configuration, validation, and request handling.


plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -304,4 +345,137 @@ def delete(self, name: str, workspace: str | None = None) -> None:
...
+    def __init__(self, parent: AgentsResource) -> None:
+        self._parent = parent
+
+    def create(self, *, name: str, workspace: str | None = None, **spec: Any) -> dict[str, Any]:
+        """Create a compute spec.
+
+        Args:
+            name: Unique compute-spec name within the workspace.
+            workspace: Target workspace.
+            **spec: ComputeSpecInline fields (``resources``, ``description``).
```
mikeknep, line 460: Wondering if instead of `**spec` kwargs, we should actually accept a `spec: ComputeSpecInline`. (And apply the same pattern for the other resources.) One of the advantages of being open source is we can share types more easily—the "backend" types aren't as hidden from the client. WDYT?

(I'm open to the argument that the kwargs approach fits better with existing resources/methods in the agents SDK.)

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -304,4 +345,137 @@ def delete(self, name: str, workspace: str | None = None) -> None:
...
+
+        Returns:
+            The created AgentComputeSpec as a dict.
+        """
+        payload: dict[str, Any] = {"name": name, **spec}
+        return self._parent._post(f"/v2/workspaces/{self._parent._workspace(workspace)}/compute-specs", payload)
```
mikeknep, line 466: nit: the `self._parent._workspace` reach is a little concerning, but I get it; probably fine, and I don't have an alternative to suggest

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -1215,6 +1250,315 @@ def deployments_wait(
...
+        no_truncate: Optional[bool] = typer.Option(
+            None,
+            "--no-truncate",
+            help="Don't truncate long values.",
+            rich_help_panel="Output Options",
+        ),
+    ) -> None:
+        """List environments."""
+        base_url = _resolve_base_url(base_url)
+        resp = _api_request("GET", base_url, f"/apis/agents/v2/workspaces/{workspace}/environments")
```
mikeknep, line 1460: Not for this PR (this diff fits the existing pattern), but: it feels like we should be reusing the SDK somehow here, e.g. so that we don't have to maintain paths like this in two places 🤔

plugins/nemo-agents/tests/unit/test_cli.py
```diff
@@ -772,3 +772,180 @@ def handler(req: httpx.Request) -> httpx.Response:
     assert "Request: GET http://test/apis/agents/v2/workspaces/default/agents" in result.stderr
     assert "Target: agents API route /apis/agents/v2/workspaces/default/agents" in result.stderr
     assert "nemo config view" in result.stderr
+
+
+# ---------------------------------------------------------------------------
+# environment / environment-spec / compute-spec commands (AIRCORE-1073)
```
mikeknep: nit: drop the ticket ref (also in the SDK test file)

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -960,6 +993,8 @@ def deploy(
             payload["name"] = name
         if image:
             payload["image"] = image
+        if environment:
+            payload["environment"] = environment
```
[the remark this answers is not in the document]
benmccown: Good catch — fixed. `deploy` now rejects a blank/whitespace `--environment` (exits 2 with an error) before building the payload, and only forwards the value when it is non-empty. Added a CLI test for the empty case.

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -141,6 +153,27 @@ def deployments(self) -> "_DeploymentResource":
...
+        if self._environment_specs is None:
+            self._environment_specs = _EnvironmentSpecResource(self)
+        return self._environment_specs
+
+    @property
+    def compute_specs(self) -> "_ComputeSpecResource":
+        """Sub-resource for AgentComputeSpec CRUD (``nemo.agents.compute_specs``)."""
+        if self._compute_specs is None:
+            self._compute_specs = _ComputeSpecResource(self)
+        return self._compute_specs
```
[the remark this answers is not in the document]
benmccown: Done — switched to direct (unquoted) return annotations on all four subresource properties (including the pre-existing `deployments`, for consistency), since the module imports `from __future__ import annotations`.

plugins/nemo-agents/tests/unit/test_cli.py
benmccown: Dropped the ticket refs from both the CLI and SDK test files.

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
benmccown: Good point. Kept `**spec` for this PR since it matches the existing agents SDK convention (`agents.create` / `deployments.create` are dict/kwargs-first), but I agree sharing the `*Inline` types is worth doing — and if we adopt typed specs it should be consistent across the whole agents SDK, not just these three. Captured as a Linear follow-up to decide and apply.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
benmccown: Agreed — maintaining these paths in both the CLI and the SDK is not ideal. Folded into the same Linear follow-up (have the CLI call the plugin SDK so the routes/payloads are defined once).

mikeknep (approved):

PR #1545, opened 2026-08-26
feat(guardrails): filter virtual models by guardrail config, refuse deleting one in use (ASTD-496)

aray12:
## Summary

Studio can author guardrail configs, but nothing could answer "which virtual models apply this one?", and deleting a config never checked whether anything was using it. Attach-time validation already rejects a dangling config reference, so the reference was only validated on one side. This adds the reverse lookup as a filter and makes the delete path enforce it.

Before: `DELETE /configs/{name}` always deleted, silently leaving any VirtualModel that applied the config pointing at a config IGW can no longer resolve. After: that delete returns 409 naming the VirtualModels, and the VirtualModel list can be filtered by guardrail config.

## Changes

- `filter[guardrail_config]=<workspace>/<name>` on `GET .../virtual-models`, matching VirtualModels whose request, response, or post-response middleware applies that stored config.
- `DELETE .../configs/{name}` returns 409 when a VirtualModel still applies the config, listing the referencing virtual models.
- New system-managed `guardrail_config_ids` field on the VirtualModel entity, derived from all three middleware pipelines by a `model_validator` so it re-derives on construction and on every entity-store read and cannot drift from the pipelines. `model_copy(update=...)` is the one path that skips validation, so the PATCH handler refreshes explicitly.
- `guardrail_config_membership_filter` lives next to the entity in `nemo_platform_plugin` so both IGW and the Guardrails service query the shared schema the same way.
- Regenerated OpenAPI specs.

### On not adding a migration

Pre-existing VirtualModels have no `guardrail_config_ids` key. Rather than backfill, the membership filter ORs an exact `$contains` against the denormalized list with a fallback that scans the raw pipelines, gated on the list being absent so it applies only to rows written before the field existed. Every row carrying the list is matched exactly.

This follows the one precedent in the repo: Intake spans both of its evaluation-membership representations the same way (`_group_membership_filter`, "The OR keeps un-migrated rows queryable, no migration") rather than rewriting rows. The alternative — a data migration in the entities service — would mean teaching a deliberately schema-agnostic store about IGW's `data` blob, and there is no precedent for it: both existing Alembic revisions are pure DDL on real columns.

The fallback arms are a superset, since `$contains` matches an array of objects as raw serialized text. That imprecision is confined to the list filter on un-migrated rows: the delete guard narrows with the filter and then re-checks each candidate against its parsed middleware calls, comparing the `config_type` discriminator as well as the reference. So a 409 is never raised on a collision, and a same-named config of another type never blocks a delete. The scan is cross-workspace, since config references are fully qualified.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: the new filter param and entity field are described in the regenerated OpenAPI spec, which is the reference for both; no narrative doc covers VirtualModel filters.

12 tests added: filter across all three pipeline phases, PATCH re-derivation, detach clearing the list, and for the delete guard — output rails, cross-workspace references, and the same-name-different-`config_type` case that must *not* block a delete. One existing test (`test_delete_config_not_found_during_delete`) was updated because delete now makes one additional entity-store call; its assertions are unchanged.

## Verification

Targeted validation, re-run after rebasing onto current `main`:

| Command | Result |
| --- | --- |
| `pytest services/guardrails/tests services/core/inference-gateway/tests/unit packages/nemo_platform_plugin/tests` | 2150 passed, 8 skipped |
| `pytest plugins/nemo-guardrails/tests/integration/` | 57 passed |
| `uv run ruff check` (changed packages) | passed |
| `uv run ruff format --check` (changed packages) | 316 files already formatted |
| `uv run --frozen ty check` (changed files) | 5 diagnostics, all pre-existing — see below |
| `python -m script.generate_openapi_spec` | spec regenerated, validated, no dangling `$ref`s |

### Integration-test fallout, fixed in this PR

The guard initially broke 15 tests across six files in `plugins/nemo-guardrails/tests/integration/`. They delete their config in `finally`, while `IGWLoopbackHarness` deletes the VirtualModels it created in *its* cleanup — which runs afterwards. So a route still applied the config at the moment it was deleted, and the service correctly returned 409.

Fixed by `detach_guardrail_config`, called from each teardown before the delete. It strips middleware calls referencing the config from the harness's VirtualModels, found via a new `IGWLoopbackHarness.virtual_models` accessor. It detaches rather than deletes so the harness's bookkeeping stays intact — it still deletes the routes itself, and logs no spurious "failed to delete" warning.

Two caching tests needed a different fix and are worth a look:

- `test_entity_config_delete_blocks_next_request` and `test_entity_config_recreate_reflected_in_next_request` delete a *referenced* config on purpose, to prove IGW fails closed and then recovers when the config returns. That is exactly the state this PR now prevents through the API.
- The state remains reachable in production — an entity removed out-of-band, or a reference predating this guard — so rather than drop the coverage, they now build it the way it actually arises: deleting the entity straight from the entity store via `sdk.entities.delete_entity_by_name`.
- If you would rather the API offered a `force`/cascade escape hatch instead, say so and I will swap these over; I avoided expanding the API surface without a stated need.

All 57 integration tests in that suite now pass locally.

### `lint-python-sdk` also fails

Expected, and the same root cause as the SDK note below: the OpenAPI spec in this PR is current but the Stainless-generated SDK has not been regenerated, so the sync check fails.

- **`uv run pre-commit run -a` was not run repo-wide.** The per-commit hooks ran and passed (DCO, merge conflicts); the full-repo sweep was not executed. Pushes used `--no-verify`: the local pre-push hook fails on `helm-docs` (not installed) and `uv-lock` (uv 0.9.28 vs the pinned 0.9.14), and this branch touches no chart, `uv.lock`, or `pyproject.toml` for either to check. CI runs them properly.
- **Python SDK not regenerated.** This touches `services/*/src/*/api/`, so per `AGENTS.md` it warrants `make update-sdk`; that needs `STAINLESS_API_KEY`, which was unavailable. The OpenAPI specs in this PR *are* current — regenerated again after the rebase onto current `main` and confirmed byte-identical to what the PR already contains. It is the Stainless step that is outstanding.

Pre-existing failures, re-verified on the current base after the rebase (same counts as on a clean tree):

- `packages/nemo_platform_plugin/tests/test_client_auth.py` — 7 failures, local base-URL config leakage.
- `services/core/entities/tests/integration/test_workspaces_crud_with_auth.py` — 18 failures, needs a live IdP.
- `services/core/inference-gateway/tests/integration/test_inference.py` — errors at setup, needs a live service.
- `ty` diagnostics: `RailsRegistryDep = None` / `ConfigRegistryDep = None` defaults and `create_config`'s `data` type, all outside the changed lines.

## Summary by CodeRabbit

- **New Features**
  - VirtualModel listings can now be filtered by guardrail configuration, including through the CLI.
  - VirtualModel responses now show associated guardrail configuration IDs.
  - Guardrail configuration associations are tracked across workspaces and middleware stages.

- **Bug Fixes**
  - Deleting a guardrail configuration still applied to VirtualModels now returns `409 Conflict` with referencing model details.
  - Associations remain accurate when VirtualModels are created or updated.

### DIFF
--- a/docs/cli/reference.mdx
+++ b/docs/cli/reference.mdx
@@ -3804,8 +3804,9 @@ JSON-only fields:
   created_at: \{gte: str, lte: str}
   updated_at: \{gte: str, lte: str}
 
-Filter virtual models by workspace, project, name, default_model_entity, created_at, and updated_at.
+Filter virtual models by workspace, project, name, default_model_entity, guardrail_config, created_at, and updated_at.
 * `--filter.default-model-entity`
+* `--filter.guardrail-config`
 * `--filter.name`
 * `--filter.project`
 * `--filter.workspace`
--- a/openapi/ga/individual/platform.openapi.yaml
+++ b/openapi/ga/individual/platform.openapi.yaml
@@ -2482,6 +2482,9 @@ paths:
                 $ref: '#/components/schemas/DeleteResponse'
         '404':
           description: Config does not exist.
+        '409':
+          description: Config is applied by one or more VirtualModels and cannot be
+            deleted.
         '422':
           description: Unable to delete config due to validation error while processing
             request.
@@ -3466,7 +3469,7 @@ paths:
         schema:
           $ref: '#/components/schemas/VirtualModelFilter'
         description: Filter virtual models by workspace, project, name, default_model_entity,
-          created_at, and updated_at.
+          guardrail_config, created_at, and updated_at.
       responses:
         '200':
           description: Paginated list of virtual models
@@ -20059,6 +20062,15 @@ components:
         override_proxy:
           title: Override Proxy
           type: string
+        guardrail_config_ids:
+          items:
+            type: string
+          type: array
+          title: Guardrail Config Ids
+          description: System-managed. Guardrail configs applied by this VirtualModel's
+            middleware, as "workspace/name" references. Derived from the middleware
+            pipelines on every write and ignored if supplied on a create or update
+            body. Filter on it with filter[guardrail_config].
         id:
           type: string
           title: Id
@@ -20174,6 +20186,12 @@ components:
           allOf:
           - $ref: '#/components/schemas/DatetimeFilter'
           description: Filter by update date.
+        guardrail_config:
+          description: Filter by a guardrail config reference in "workspace/name"
+            form. Matches VirtualModels whose request, response, or post-response
+            middleware applies that stored guardrail config.
+          title: Guardrail Config
+          type: string
       title: VirtualModelFilter
       type: object
     VirtualModelInferenceConfig:
--- a/openapi/ga/openapi.yaml
+++ b/openapi/ga/openapi.yaml
@@ -2482,6 +2482,9 @@ paths:
                 $ref: '#/components/schemas/DeleteResponse'
         '404':
           description: Config does not exist.
+        '409':
+          description: Config is applied by one or more VirtualModels and cannot be
+            deleted.
         '422':
           description: Unable to delete config due to validation error while processing
             request.
@@ -3466,7 +3469,7 @@ paths:
         schema:
           $ref: '#/components/schemas/VirtualModelFilter'
         description: Filter virtual models by workspace, project, name, default_model_entity,
-          created_at, and updated_at.
+          guardrail_config, created_at, and updated_at.
       responses:
         '200':
           description: Paginated list of virtual models
@@ -20059,6 +20062,15 @@ components:
         override_proxy:
           title: Override Proxy
           type: string
+        guardrail_config_ids:
+          items:
+            type: string
+          type: array
+          title: Guardrail Config Ids
+          description: System-managed. Guardrail configs applied by this VirtualModel's
+            middleware, as "workspace/name" references. Derived from the middleware
+            pipelines on every write and ignored if supplied on a create or update
+            body. Filter on it with filter[guardrail_config].
         id:
           type: string
           title: Id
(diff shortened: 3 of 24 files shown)


aray12:

- #1545 ◀
    - #1546
        - #1552

aray12:
Pushed a fix for the integration-test failures (`e4797bea2`).

**What broke:** 15 tests across six files in `plugins/nemo-guardrails/tests/integration/`. They delete their guardrail config in `finally`, but `IGWLoopbackHarness` deletes the VirtualModels it created in *its* cleanup, which runs afterwards — so a route still applied the config at the moment it was deleted, and the new guard correctly returned 409. The guard was right; the teardown ordering had been relying on deletion being unguarded.

**Fix:** a shared `detach_guardrail_config` helper, called from each teardown before the delete. It strips middleware calls referencing the config from the harness's VirtualModels, found via a new `IGWLoopbackHarness.virtual_models` accessor. Detaching rather than deleting keeps the harness's bookkeeping intact, so it still deletes the routes itself and logs no spurious "failed to delete" warning.

**The bit worth reviewer attention** — two tests in `test_middleware_config_caching.py` deliberately delete a *referenced* config to prove IGW fails closed and then recovers when the config comes back. That is precisely the state this PR now prevents through the API, so detaching first would have gutted what they test.

The state is still reachable in production (an entity removed out-of-band, or a reference predating this guard), and IGW's fail-closed behaviour still matters, so rather than delete the coverage I changed how they construct it: `sdk.entities.delete_entity_by_name` removes the config straight from the entity store, bypassing the service-level guard.

If you'd rather the delete endpoint grew a `force`/cascade option and these tests used that, I'm happy to swap — I avoided widening the API surface without a stated need, since the guard was the explicit ask.

All 57 tests in that suite now pass locally; 2150 unit tests still green.

`lint-python-sdk` remains red and is not something I can fix here: the OpenAPI spec in this PR is current, but `make update-sdk` needs `STAINLESS_API_KEY`.

JashG:
This change LGTM

JashG (approved):

PR #1549, opened 2026-08-26
feat(studio): Add intake traces for nemo assistant

dmariali:
## Summary by CodeRabbit

- **New Features**
  - Default assistant models are now resolved from the active platform configuration.
  - API parameters can be provided as JSON objects or JSON-formatted strings.
  - Improved skill discovery through the shared skills directory.
  - Added platform workspace and base URL configuration for MCP integrations.
  - Enabled ATIF trajectory telemetry export with configured ingestion settings.

- **Bug Fixes**
  - Improved MCP environment and platform skill-path compatibility.
  - Ensured packaged workspace modules and skills load correctly at startup.
  - Added clearer validation for unsupported agent configuration formats.


agents/nemo-studio-assistant-spec/AGENT-SPEC.md
```diff
@@ -43,7 +43,7 @@ The mission is grounded in the current deployment proof of concept and its imple
 ## Model
 
 - Mode: cloud
-- Family: NVIDIA Nemotron 3 Super 120B A12B
+- Family: Meta Muse Glimmer 30B
```
steramae-nvidia: Why switch off of a nemotron model?

agents/nemo-studio-assistant/src/nemo_studio_assistant/register.py
```diff
@@ -59,7 +59,7 @@ def _get_client(workspace: str) -> NeMoPlatform:
     if not request_workspace:
         raise ValueError("workspace is required")
     if request_workspace not in _clients:
-        base_url = os.environ.get("NEMO_BASE_URL") or os.environ.get("NMP_BASE_URL")
+        base_url = os.environ.get("NMP_BASE_URL") or os.environ.get("NEMO_BASE_URL")
```
steramae-nvidia, line 62: Shouldn't we prioritize NEMO vs NMP?

agents/nemo-studio-assistant/agent.yaml
```diff
@@ -62,26 +62,21 @@ harnesses:
 models:
   default:
     provider: nvidia
-    model: nvidia-nemotron-3-5-lightning-30b-a3b
```
steramae-nvidia, line 65: It feels bad to have to remember to update these in two places, what do we need to do in order to only have a single place for this?

steramae-nvidia (approved):

agents/nemo-studio-assistant/agent.yaml
dmariali: Initially I was trying to just pull whatever the user's preset default model was, I should switch it back to that so we don't need to manually set this. Just inference is such a pain in the butt sometimes! will update

PR #1399, opened 2026-08-19
feat(studio): Experiments Panel for Agent Details

steramae-nvidia:
Screenshot 2026-08-26 at 2 06 53 PM

Signed-off-by: Sean Teramae <steramae@nvidia.com>

## Type of Change

- [ ] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [ ] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
- [ ] Documentation not applicable — justification:

## Verification

- [ ] Pull request title follows the repository's Conventional Commit format
- [ ] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [ ] Targeted tests pass, or tests are marked not applicable above
- [ ] No secrets, API keys, or credentials are included

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Added a Recent Experiments panel to agent details with evaluation scores, trends, descriptions, counts, and seven-day comparisons.
  - Added loading and empty states, with options to run evaluations or open experiment details.
  - Agent details now show more complete experiment information.

- **Style**
  - Improved metric panel alignment and label readability.

- **Documentation**
  - Updated the experiments documentation link.

- **Tests**
  - Added coverage for experiment trends, evaluation states, navigation, and empty/loading views.

PR #1371, opened 2026-08-18
feat(experiments): Add new flags to Experiment entity

shanaiabuggy:
## Summary

Adds presentation and comparison settings to Experiments. Experiments can be marked as favorites, configured to show results over time, and assigned one baseline Evaluation by name.

The baseline relationship is stored on the Experiment rather than as a boolean on the Evaluation because an Evaluation can belong to multiple Experiments and may be the baseline for only some of them.

## Changes

- Add `is_favorite` and `evaluate_over_time` to the Experiment entity and create, update, response, and filter APIs.
- Add `baseline_evaluation_name` to the Experiment entity and update, response, and filter APIs.
- Require a baseline Evaluation to already exist and be a live member of the Experiment.
- Do not accept `baseline_evaluation_name` during Experiment creation because the Evaluation must be created first.
- Prevent removing or deleting an Evaluation while it is selected as a live Experiment's baseline.
- Allow callers to clear a baseline by setting `baseline_evaluation_name` to `null`.
- Preserve new flag values when older clients omit them from full Experiment updates.
- Treat missing flag values on legacy Experiment rows as `false`, including while filtering.
- Regenerate OpenAPI, Python SDK, CLI, and associated reference artifacts.

## Type of Change

- [x] Code change with documentation updates

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Documentation updated for user-visible behavior

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

Targeted validation:

- `UV_CACHE_DIR=/tmp/nemo-platform-uv-cache uv run --frozen pytest services/intake/tests/integration/test_experiments_crud.py -q -k 'experiment_flags or new_flags or baseline'` — 7 passed
- `make lint-fix` — passed
- Stainless schema mapping report — no warnings
- Stainless SDK freshness check — passed

## Summary by CodeRabbit

- **New Features**
  - Experiments now support favorites, evaluation-over-time settings, and selectable baseline evaluations.
  - Added filtering by favorite status, evaluation-over-time visibility, and baseline evaluation.
  - CLI commands support creating, updating, and listing experiments with these options.
- **Bug Fixes**
[...]

PR #1553, opened 2026-08-26
feat(studio): surface experiment presentation flags in Studio

rrhyne:
image

Screenshot 2026-08-26 at 6 26 29 PM

## Summary

Surfaces the Experiment `is_favorite` and `show_evaluations_over_time` flags (added to the API in #1371) in Studio. Both can now be set when creating or editing an experiment, and both change how an experiment renders in the list: favorites pin to the top with a green tag, and experiments flagged to graph over time show their evaluation trend inline on the card.

## Changes

**Create and edit modals**

- Add an `ExperimentFlagSwitch` component holding the toggle markup and the label/hint copy for both flags, so the create and edit modals render them identically and the copy has a single source.
- Wire both flags into `ExperimentCreateModal` via react-hook-form and into `ExperimentEditModal` via its existing local state.
- Remove the Create experiment / NeMo Assistant / CLI command tabs from the create modal; the latter two held static placeholder snippets.

**Experiments list**

- Pin favorites above the paginated list, each marked with a green `Favorite` tag. The list endpoint cannot sort by `is_favorite`, so favorites are fetched as their own filtered query and the main list is narrowed to non-favorites to avoid rendering twice. Both queries share the active search and column filters.
- Render the evaluation trend inline on the right of a card flagged with `show_evaluations_over_time`, replacing the evaluation count. The left of the card is unchanged.
- Add a `chartOnly` flag to `MetricTrendPanel` that drops the surrounding panel and header, for embedding where the container already shows the title and description. The existing panel path is unchanged.

## Notes for reviewers

- `is_favorite: false` matches rows stored as false *and* legacy rows with no stored value, so narrowing the main list does not hide experiments created before the flag existed.
- The pinned favorites query is capped. If a workspace exceeds the cap, the main list stops excluding favorites so the ones past it stay reachable rather than disappearing.
- The evaluator pills are buttons inside an interactive card, so the trend region stops click and keydown propagation; without it, selecting a series would navigate to the experiment.
- Trend values render through `formatEvaluatorScore` as bare floats rather than percentages, matching the rest of the app — evaluator scores carry no scale metadata, so a percentage would misreport an integer score. The Figma mock shows a percentage; this deliberately differs.
- #1399 touches the same `MetricTrendPanel` block this PR hoists into a variable, so expect a small mechanical conflict whichever lands second.

## Type of Change

## Quality Gates

- [x] Existing tests cover changed behavior — justification: the existing `MetricTrendPanel` and `ExperimentEditModal` suites cover the components this PR modifies and both still pass. No suite existed for `ExperimentRoute` or `ExperimentCard`.
- [x] Documentation not applicable — justification: no user-facing docs describe the experiments list or create modal.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui typecheck` — passed
- `npx eslint` over `ExperimentCreateModal`, `ExperimentEditModal`, `ExperimentFlagSwitch`, `ExperimentRoute`, `MetricTrendPanel` — passed
- `npx vitest run src/components/charts/MetricTrendPanel` — 6 passed
- `npx vitest run src/components/ExperimentEditModal` — 1 passed
- `uv run pre-commit run -a` — not run
- Not yet exercised in a browser; the modals and list page have not had a visual pass.

## New Features
- Added Favorite and “evaluations over time” settings when creating or editing experiments.
- Added evaluation trend charts with metric selection and seven-day change indicators.
- Added a pinned section for favorite experiments above the main experiment list.
- Simplified experiment creation by displaying the form directly.

## Bug Fixes
- Improved empty-state messaging for empty experiment lists.
- Improved trend loading and display for incomplete or unavailable evaluation data.
- Experiment names now provide direct navigation to experiment details.


web/packages/studio/src/components/charts/MetricTrendPanel/index.tsx
```diff
@@ -50,6 +50,12 @@ interface Props {
   formatDelta?: (delta: number) => string;
   chartHeight?: number;
   isPending?: boolean;
+  /**
+   * Render only the value, pills, and chart, dropping the surrounding panel and its header.
+   * Use when the embedding container already shows the title and description — `title` is
+   * still required, as it names the pill group for screen readers.
+   */
+  chartOnly?: boolean;
```
steramae-nvidia: Perhaps we should instead pull the chart out into its own component and wrap it in this panel?

steramae-nvidia (approved):

web/packages/studio/src/components/charts/MetricTrendPanel/index.tsx
rrhyne: Should be resolved now!

rrhyne:
@coderabbitai Both fixed in a78b2df9a.

- **Nested interactive controls:** the card is no longer a `role="button"`. The experiment name is now the link (`Anchor asChild` + router `Link`), so the evaluator buttons are no longer nested inside a button; card-wide click stays as a mouse-only convenience.
- **Stuck skeleton:** react-query reports a disabled query as `isPending` forever, so a card with no experiment id read as "loading". `showTrend` now requires an id and the pending state is gated on the query actually running.

Also resolved the #1399 conflict, keeping the `MetricTrend` extraction and carrying its styling changes across.

PR #1582, opened 2026-08-27
feat(studio): suggest starter tests on the empty Test and Validate tab (ASTD-498)

aray12:
## Summary
Add two one-click "Add <Label> Test" suggestions (PII, Jailbreak) below "Add Another Test" when a guardrail config has no test cases yet, seeded with hardcoded content matched to `SELF_CHECK_INPUT_PROMPT`'s policy bullets. Each fires the same create-check mutation "Add Another Test" already uses.

image

## Changes
- New `suggestedTestCases.ts` with the two suggestion definitions (label, icon, content).
- `GuardrailTestCasesEditor.tsx`: render the suggestion cards, gated on the tab having zero checks; each card has an inline "Add <Label> Test" button.
- Test coverage for showing/hiding the suggestions and creating a check from one.


aray12:

- #1582 ◀

PR #1590, opened 2026-08-27
fix(igw): rewrite served_model_name in upstream error response bodies

benmccown:
## Summary

IGW's response model-rewrite (`_rewrite_model_field`) only ran on the happy path. When an upstream returned 4xx/5xx, `fetch_proxy_response` raised `HTTPException` with the raw upstream error body and the rewrite never ran, so the upstream `served_model_name` leaked in error cases (rate-limit messages, content-policy blocks, "model not found", etc.). This change scrubs the served name out of error bodies too, so callers see the model entity reference in errors exactly as they do on success.

## Changes

- Add `_rewrite_model_field_in_error_body(error_body, served_model_name, restored_model_id)`: parses JSON-object bodies through the existing `_rewrite_model_field` (plus a substring pass so a served name embedded in a free-form error *message* string is caught, not just a structured `model` field), and falls back to a plain substring replace for plain-text / non-object / unparseable bodies. Empty or non-matching bodies are returned unchanged.
- Thread optional `served_model_name` / `restored_model_id` into `fetch_proxy_response`; its `response.status >= 400` branch runs the rewrite before raising. Because that status check precedes the SSE branch, a non-2xx returned mid-stream is scrubbed the same way. The 401/403/404→502 wrap path is covered too.
- Wire the two names in at the `virtual_model_proxy` call site (reusing the same entity ref the happy path restores).
- Tests: helper unit tests (structured JSON, message-string JSON, plain text, empty/no-match, JSON-scalar fallback) and `fetch_proxy_response` error cases (OpenAI 400 model-not-found, Anthropic 400, 429 rate-limit, plain-text 5xx, 401→502 wrap, no-op without names), plus an end-to-end `virtual_model_proxy` error-path test.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: internal error-handling fix; no user-facing docs describe the leaked field.

## Verification

Targeted validation:

- `uv run ruff check <proxy.py> <test_proxy.py>` → All checks passed
- `uv run ruff format --check ...` → 2 files already formatted
- `uv run --frozen ty check tests/unit/test_proxy.py` → only pre-existing SDK-typing diagnostics (the `created_at`/`updated_at` `datetime` vs ISO-string convention every existing `virtual_model_proxy` test in this file already shares); no new real diagnostics.
- `uv run --frozen pytest services/core/inference-gateway/tests/unit` → 478 passed, 5 skipped.
- Note: the ticket named `make test-unit-core-inference-gateway`, which does not exist as a Make target; ran the service's unit tests directly instead.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved proxy error responses by replacing eligible internal model identifiers with the served model name.
  - Updated rewriting across structured JSON, plain-text messages, scalar responses, and array-based error bodies.
  - Applied consistent handling to common authorization, rate-limit, client, server, and provider-specific errors.
  - Streaming requests now receive clearer model references in applicable error responses.
  - Added safeguards to avoid rewriting invalid or missing model identifiers, improving error-response reliability.

### DIFF
--- a/services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
+++ b/services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
@@ -4,6 +4,7 @@
 import asyncio
 import json
 import logging
+import re
 import uuid
 from collections.abc import AsyncIterable
 from dataclasses import dataclass
@@ -27,6 +28,7 @@ from nemo_platform_plugin.inference_middleware import (
     InferenceMiddlewareError,
     InferenceResponse,
 )
+from nemo_platform_plugin.refs import ENTITY_REF_PATTERN
 from nemo_platform_plugin.secrets.client import AsyncSecretsClient
 from nmp.common.entities.utils import parse_model_entity_ref
 from nmp.core.inference_gateway.api.backend_format import resolve_backend_format
@@ -57,6 +59,13 @@ ResponseResult = Union[dict[str, Any], AsyncIterator[dict[str, Any]]]
 logger = logging.getLogger(__name__)
 
 DEFAULT_CHUNK_SIZE = 4096
+
+# A model name safe to blind-substring-replace inside an error body: the entity-ref
+# shape (``[\w.-]`` segments with an optional ``/``), which both the entity store
+# (ENTITY_REF_PATTERN) and HuggingFace repo ids conform to. Its character set has no
+# JSON-structural characters, so replacing it can't corrupt a JSON body. See
+# _rewrite_model_field_in_error_body.
+_SAFE_MODEL_NAME = re.compile(ENTITY_REF_PATTERN)
 
 # OpenAPI extra configuration for proxy endpoints that accept arbitrary JSON bodies.
 # This is used to generate proper requestBody schema in OpenAPI spec for POST/PUT/PATCH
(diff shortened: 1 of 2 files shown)


mmogallapalli (approved):
LGTM

services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
```diff
@@ -644,6 +663,58 @@ def _rewrite_model_field(payload: Any, served_model_name: str, restored_model_id
...
+    except (json.JSONDecodeError, ValueError):
+        return error_body.replace(served_model_name, restored_model_id)
+
+    if isinstance(parsed, dict):
+        _rewrite_model_field(parsed, served_model_name, restored_model_id)
+        reserialized = json.dumps(parsed)
+        # Also scrub the served name out of any free-form message strings that
+        # _rewrite_model_field (which only touches structured `model` fields)
+        # would miss.
+        return reserialized.replace(served_model_name, restored_model_id)
```
mmogallapalli: Looks like the rewrite as constructed could inadvertently append the fixed model name again. For example:

```python
served_model_name = "model-1"
restored_model_id = "default/model-1"
```

Given this error body:

```json
{"model": "model-1", "error": {"message": "model-1 is unavailable"}}
```

`_rewrite_model_field` first changes the structured field to:

```json
{"model": "default/model-1", "error": {"message": "model-1 is unavailable"}}
```

The subsequent global `replace()` then processes the already-rewritten field, producing:

```json
{"model": "default/default/model-1", "error": {"message": "default/model-1 is unavailable"}}
```

_________________________________________________________________________________________

**[P1 — blocking]** The second replacement can turn `default/model-1` into `default/default/model-1`, while an empty `served_model_name` causes `str.replace("", ...)` to corrupt the entire body.

I suggest traversing the original parsed value once:

```python
def _replace_model_names(value: Any, served_name: str, restored_id: str) -> Any:
    if isinstance(value, str):
        return value.replace(served_name, restored_id)
    if isinstance(value, dict):
        return {
            key: _replace_model_names(item, served_name, restored_id)
            for key, item in value.items()
        }
    if isinstance(value, list):
        return [
            _replace_model_names(item, served_name, restored_id)
            for item in value
        ]
    return value


def _rewrite_model_field_in_error_body(
    error_body: str,
    served_model_name: str,
    restored_model_id: str,
) -> str:
    if not error_body or not served_model_name:
        return error_body

    try:
        parsed = json.loads(error_body)
    except (json.JSONDecodeError, ValueError):
        return error_body.replace(served_model_name, restored_model_id)

    rewritten = _replace_model_names(
        parsed,
        served_model_name,
        restored_model_id,
    )
    return json.dumps(rewritten)

mmogallapalli (changes requested):

services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py
benmccown: agent injected reply:

> Good catch — you're right on both counts. Fixed in `eaa00586e`.
>
> The root problem was the two-pass approach (structured rewrite, then a blanket `str.replace` over the re-serialized JSON): the second pass re-hit the field the first pass had already rewritten, so `model-1` → `default/model-1` → `default/default/model-1` whenever the restored id contains the served name (the common case). The empty-`served_model_name` corruption you noted was the same "blind replace is fragile" family.
>
> Rather than the recursive per-value walk, I went with the simplest thing that is actually correct for our data: a **single** substring pass over the raw body. It only ever touches the original string, so it can't double-apply, and it's format-agnostic (JSON, plain text, or unparseable all handled the same). To keep a blind replace from corrupting a JSON body, both names are first validated against the codebase's own `ENTITY_REF_PATTERN` (the entity-ref / HuggingFace model-id shape, whose character set has no JSON-structural chars) — if either isn't a plain model id (or is empty), the scrub is skipped and a warning logged rather than risk mangling the body. This is best-effort UX scrubbing, not a guarantee.
>
> Added regression tests: the double-application case (served name in both a structured field and a message string, asserting single application + still-valid JSON), an empty-served-name no-op, and a JSON-unsafe-char skip-and-passthrough. Thanks for the careful read.

mmogallapalli (approved):

PR #1592, opened 2026-08-27
refactor(agents): typed SDK spec params and CLI reuse of the SDK

benmccown:
## Summary

The new agents SDK resources (`environment_specs`, `environments`, `compute_specs`) took loose `**spec` kwargs, and the hand-written CLI duplicated request-building. This PR makes the SDK create methods accept the shared backend `*Inline` models for type safety and discoverability, and has the CLI reuse the SDK so request paths/payloads live in one place. Both follow-ups were deferred from a prior review.

## Changes

- **Typed SDK create params.** `environment_specs.create`, `environments.create`, and `compute_specs.create` now accept the shared `EnvironmentSpecInline` / `AgentEnvironmentInline` / `ComputeSpecInline` models via a `spec=` argument (plain dicts also accepted). Typed models are serialized with `exclude_unset=True` so only caller-set fields are sent — matching the old `**spec` semantics. The loose `**spec` kwargs are kept for back-compat and override the typed model when both are supplied.
- **SDK auth-header threading.** An optional `default_headers` attribute on the platform object is now attached to every SDK request, so an authenticated caller (the CLI) can reuse the SDK against a secured platform.
- **CLI reuses the SDK.** The `environment-specs` / `environments` / `compute-specs` create commands build their requests through the plugin SDK instead of hand-rolling request paths and JSON bodies. HTTP errors are translated to the same rich CLI messages as before.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: internal SDK/CLI plumbing; docstrings/usage examples updated in-code, no user-facing docs affected.

## Verification

Targeted validation:

- `uv run --frozen ty check <sdk.py> <cli.py> <test_sdk.py>` → All checks passed!
- `uv run ruff check` / `uv run ruff format --check` on changed files → clean
- `uv run --frozen pytest plugins/nemo-agents/tests/unit` → 1362 passed (includes 7 new SDK/CLI tests for typed params, back-compat kwargs, header threading, and CLI→SDK delegation)

## Summary by CodeRabbit

* **New Features**
  * Added SDK-backed creation of environment specifications, environments, and compute specifications through the CLI.
  * Added support for typed or dictionary-based resource specifications.
  * Added consistent authentication header handling across SDK requests.

* **Bug Fixes**
  * CLI SDK errors now use consistent formatted messages and exit codes.
  * Resource specifications preserve explicitly provided fields and honor input precedence rules.
  * Improved handling of structured specification data and detailed HTTP errors during resource creation.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
@@ -447,7 +447,7 @@ class _EnvironmentSpecResource:
         Returns:
             The created AgentEnvironmentSpec as a dict.
         """
-        payload: dict[str, Any] = {"name": name, **_spec_to_dict(spec), **spec_kwargs}
+        payload: dict[str, Any] = {**_spec_to_dict(spec), **spec_kwargs, "name": name}
         return self._parent._post(f"/v2/workspaces/{self._parent._workspace(workspace)}/environment-specs", payload)
 
     def list(self, workspace: str | None = None) -> List[dict[str, Any]]:
@@ -561,7 +561,7 @@ class _ComputeSpecResource:
         Returns:
             The created AgentComputeSpec as a dict.
         """
-        payload: dict[str, Any] = {"name": name, **_spec_to_dict(spec), **spec_kwargs}
+        payload: dict[str, Any] = {**_spec_to_dict(spec), **spec_kwargs, "name": name}
         return self._parent._post(f"/v2/workspaces/{self._parent._workspace(workspace)}/compute-specs", payload)
 
     def list(self, workspace: str | None = None) -> List[dict[str, Any]]:
--- a/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
@@ -29,11 +29,16 @@ Usage (once the SDK hub is wired up)::
 
     # Environments / specs (the request/fulfill split)
     spec = nemo.agents.environment_specs.create(
-        name="ben", env={"LOG_LEVEL": "debug"},
-        secrets={"GITHUB_PERSONAL_ACCESS_TOKEN": "default/ben-pat"},
+        name="ben",
+        spec=EnvironmentSpecInline(
+            env={"LOG_LEVEL": "debug"},
+            secrets={"GITHUB_PERSONAL_ACCESS_TOKEN": "default/ben-pat"},
+        ),
     )
     env = nemo.agents.environments.create(name="repo-research-ben", environment_spec="default/ben")
-    cs = nemo.agents.compute_specs.create(name="big", resources={"limits": {"cpu": "2"}})
+    cs = nemo.agents.compute_specs.create(
+        name="big", spec=ComputeSpecInline(resources=ComputeResources(limits={"cpu": "2"})),
+    )
     dep = nemo.agents.deployments.create(agent="calculator", environment="default/repo-research-ben")
 
     # Invocation (routes through the agents gateway)
@@ -60,8 +65,14 @@ import re
 from typing import Any, List, Mapping
 
 import httpx
+from nemo_agents_plugin.entities import (
+    AgentEnvironmentInline,
+    ComputeSpecInline,
+    EnvironmentSpecInline,
+)
 from nemo_agents_plugin.session_protocol import SESSION_ID_HEADER
 from nemo_platform_plugin.sdk import NemoPluginSDKResources
+from pydantic import BaseModel
 
 _DEFAULT_WORKSPACE = "default"
 _DEFAULT_TIMEOUT = 30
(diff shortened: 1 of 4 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -386,21 +422,32 @@ class _EnvironmentSpecResource:
...
+                ``secrets``, ``mcp``, ``provider``, ``model_provider_override``,
+                ``workspace_path``, ``artifacts_path``, ``connection``,
+                ``metadata``, ``settings``, ...). Merged over ``spec`` when both
+                are given. Prefer the typed ``spec=`` argument.
 
         Returns:
             The created AgentEnvironmentSpec as a dict.
         """
-        payload: dict[str, Any] = {"name": name, **spec}
+        payload: dict[str, Any] = {"name": name, **_spec_to_dict(spec), **spec_kwargs}
```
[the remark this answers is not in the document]
benmccown: agent injected reply:

> Keep the explicit `name` authoritative.

Good catch — fixed in 940fe45c4. Moved `"name": name` to the end of the payload in both `environment_specs.create` and `compute_specs.create` (environments was already name-last), so a `name` key smuggled in via a dict `spec`/kwargs can no longer redirect the resource. Added a regression test (`test_environment_specs_create_name_arg_is_authoritative`).

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -280,9 +309,15 @@ def _workspace(self, workspace: str | None) -> str:
     def _agents_url(self, path: str) -> str:
         return self._base_url() + "/apis/agents" + path
 
+    def _default_headers(self) -> dict[str, str] | None:
+        headers = getattr(self._platform, "default_headers", None)
+        if isinstance(headers, dict) and headers:
+            return {str(key): str(value) for key, value in headers.items()}
```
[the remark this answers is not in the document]
benmccown: agent injected reply:

> Reject credentialed HTTP requests.

Not taking this one, deliberately. The agents CLI's documented default base URL is `http://localhost:8080` (see `cli_context.py`), and every `nemo` command already sends its bearer token to that loopback HTTP endpoint in normal local dev. Rejecting credentialed HTTP in the SDK would break that primary local workflow. This PR only *threads* the auth header the CLI already resolves through the SDK — it does not introduce token-over-HTTP, so scoping a transport-security policy here is both a behavior regression for local dev and out of scope for this refactor. If we want to enforce HTTPS-for-credentials, it belongs as a platform-wide client policy (applied to all `nemo` commands), tracked separately.

mmogallapalli (approved):

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -29,11 +29,16 @@
 
     # Environments / specs (the request/fulfill split)
     spec = nemo.agents.environment_specs.create(
-        name="ben", env={"LOG_LEVEL": "debug"},
-        secrets={"GITHUB_PERSONAL_ACCESS_TOKEN": "default/ben-pat"},
+        name="ben",
```
mikeknep, line 32: lol this has surprised me multiple times, I keep thinking "when did I jump into a test file?" before realizing its the example docstring

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -386,21 +422,32 @@ class _EnvironmentSpecResource:
...
             workspace: Target workspace.
-            **spec: EnvironmentSpecInline fields (``env``, ``secrets``, ``mcp``,
-                ``provider``, ``model_provider_override``, ``workspace_path``,
-                ``artifacts_path``, ``connection``, ``metadata``, ``settings``,
-                ...). See :class:`EnvironmentSpecInline`.
+            **spec_kwargs: Back-compat loose EnvironmentSpecInline fields (``env``,
+                ``secrets``, ``mcp``, ``provider``, ``model_provider_override``,
+                ``workspace_path``, ``artifacts_path``, ``connection``,
+                ``metadata``, ``settings``, ...). Merged over ``spec`` when both
+                are given. Prefer the typed ``spec=`` argument.
```
mikeknep: Genuine question: do we want to support this long-term? If we **don't**, I'd argue this is new enough that we could just break it. I lean towards dropping it but I don't feel strongly about it.

Same with `ComputeSpec`. And, while I understand why `Environment` is different and _doesn't_ expose the spec_kwargs argument (it is a higher level "bundling" object), I do kinda feel like that lends a little more support towards dropping spec_kwargs—just feels like it'd be a little more consistent across the board, instead of only _some_ `create` methods supporting it (albeit for legit reasons).

benmccown: I agree with dropping, pushed another commit that made it so.

---
*agent injected reply (Ben's words above):* dropped `**spec_kwargs` from `environment_specs.create` and `compute_specs.create` in `14e9f494c` — every create method now takes a single `spec=` (an `*Inline` model or a plain dict), consistent with `environments`.

mikeknep (approved):

PR #1595, opened 2026-08-27
feat(agents): add missing SDK for execute agent job

mikeknep:
## Summary
Adds SDK methods for working with ExecuteAgentJob

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

---

## Summary by CodeRabbit

* **New Features**
  * Added synchronous and asynchronous SDK support for creating, retrieving, and listing execution job results.
  * Added job access through both synchronous and asynchronous agent resources.
  * Added registration for the asynchronous agents resource.
  * Workspace selection now supports explicit values, client defaults, and platform defaults.
* **Tests**
  * Added coverage for job execution requests, optional fields, result paths, response handling, workspace selection, and resource registration.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
@@ -43,12 +43,21 @@ Usage (once the SDK hub is wired up)::
         session_id="session-entity-id",
         input="Continue",
     )
+
+    # agents.execute jobs
+    job = nemo.agents.jobs.execute.create(spec={"agent": "calculator", "input": "What is 2+2?"})
+    job = nemo.agents.jobs.execute.get(job["name"])
+    results = nemo.agents.jobs.execute.list_results(job["name"])
+
+An async namespace is mounted as ``client.agents`` on ``AsyncNeMoPlatform``.
+It currently exposes ``jobs`` only — agent CRUD, deployments, and ``invoke``
+remain sync-only.
 """
 
 from __future__ import annotations
 
 import re
-from typing import Any, List
+from typing import Any, List, Mapping
 
 import httpx
 from nemo_agents_plugin.session_protocol import SESSION_ID_HEADER
@@ -85,6 +94,7 @@ class AgentsResource:
         self._environments: _EnvironmentResource | None = None
         self._environment_specs: _EnvironmentSpecResource | None = None
         self._compute_specs: _ComputeSpecResource | None = None
+        self._jobs: _JobsResource | None = None
 
     # ------------------------------------------------------------------
     # Agent CRUD
@@ -173,6 +183,13 @@ class AgentsResource:
         if self._compute_specs is None:
             self._compute_specs = _ComputeSpecResource(self)
         return self._compute_specs
+
+    @property
+    def jobs(self) -> "_JobsResource":
+        """Sub-resource for agents job collections."""
+        if self._jobs is None:
+            self._jobs = _JobsResource(self._platform)
+        return self._jobs
 
     # ------------------------------------------------------------------
     # Invocation and evaluation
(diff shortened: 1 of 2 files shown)


mmogallapalli (approved):

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -478,4 +495,147 @@ def delete(self, name: str, workspace: str | None = None) -> None:
...
+    def __init__(self, platform: Any) -> None:
+        self._platform = platform
+
+    def create(
+        self,
+        *,
+        spec: Mapping[str, Any],
+        name: str | None = None,
+        description: str | None = None,
+        workspace: str = _DEFAULT_WORKSPACE,
```
benmccown: **Severity:** P1 (blocking)
**Found by:** opus, sonnet, gpt-sol, nemotron

These job methods hardcode `workspace: str = _DEFAULT_WORKSPACE`, so they always target `default` unless the caller passes `workspace=` on every call. Every sibling resource instead does `workspace: str | None = None` -> `self._workspace(workspace)`, which falls back to the client-level workspace (`getattr(self._platform, "workspace", None)`).

Net effect: a client constructed with a non-default workspace —
```python
nemo = NeMoPlatform(base_url=..., workspace="team-a")
nemo.agents.get("foo")                      # -> /workspaces/team-a/...
nemo.agents.jobs.execute.create(spec=...)   # -> /workspaces/default/...  (silently ignores team-a)
```
The `platform.post/get` pipeline you route through applies auth/base-url/retry but does **not** re-inject the client's `workspace` into a literal path (verified against `enhanced.py` / the Stainless base), so the client-level default is genuinely dropped here.

Root cause: `AgentsResource.jobs` hands `_JobsResource` the raw `self._platform` rather than the `AgentsResource`, so these resources can't reach `_workspace()`. Suggest passing the parent (or its `_workspace`) in and defaulting `workspace: str | None = None`, resolving through the same 3-level fallback the rest of the SDK uses — for `create`, the async mirror, `get`, and `list_results`.

Also worth adding a `test_execute_job_create_uses_client_workspace_by_default` mirroring `test_deployments_create_uses_client_workspace_by_default` (a `SimpleNamespace(workspace="team-a")` platform, **no** `workspace=` arg, assert `team-a` in the path). The current execute-job tests all pass `workspace=` explicitly (or use a fake with no `.workspace`), so none would catch this.

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
```diff
@@ -478,4 +495,147 @@ def delete(self, name: str, workspace: str | None = None) -> None:
...
+    ) -> dict[str, Any]:
+        """Submit an execute-agent job. *spec* is an ``ExecuteAgentJobConfig``."""
+        return self._platform.post(
+            _execute_jobs_base(workspace),
+            body=_execute_job_body(spec, name, description),
+            cast_to=dict[str, Any],
+        )
+
+    def get(self, name: str, *, workspace: str = _DEFAULT_WORKSPACE) -> dict[str, Any]:
```
benmccown: **Severity:** P2 (non-blocking)
**Found by:** opus, sonnet, gpt-sol, nemotron

`get`/`list_results` (and the async mirrors) make `workspace` keyword-only (`def get(self, name, *, workspace=...)`), whereas the sibling `get`/`delete` across the SDK take it positional-or-keyword (`def get(self, name, workspace=None)`). So `execute.get("my-job", "team-a")`, mirroring the sibling call style, raises `TypeError`.

Not opinionated about *which* convention is better here — keyword-only isn't wrong — but we probably want the SDK surface to be internally consistent. Flagging so it's a deliberate choice either way.

benmccown (changes requested):
Independent multi-model review (opus / sonnet / gpt-sol / nemotron, parent-verified against the diff + platform SDK). Nicely scoped and the server-route contract checks out. One blocking item: `jobs.execute.*` silently ignores a client-configured workspace, plus a smaller signature-consistency nit. Details inline.

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
mikeknep: Good catch, thanks. Addressed in a8519a347

plugins/nemo-agents/src/nemo_agents_plugin/sdk.py
mikeknep: Thanks, I went with positional-or-keyword to match existing siblings. Addressed in a8519a347

benmccown (approved):

PR #1596, opened 2026-08-27
feat(agents): accept inline agent definitions on ExecuteAgent jobs

mikeknep:
## Summary
Adds support for passing an inline agent definition to the execution job, as an alternative to referencing an existing agent entity

## Type of Change

- [ ] Documentation only

## Quality Gates

- [s] Tests added or updated for changed behavior

## Verification

---

## Summary by CodeRabbit

- **New Features**
  - Execute agent jobs now support inline agent configurations alongside saved agent references.
  - Added options for execution environments, evaluator prompt templates, LoRA training, deployment, privacy, and data settings.
- **Bug Fixes**
  - Improved validation for inline agents, malformed references, unsupported formats, and incomplete configurations.
  - Preserved compatibility with saved agents and legacy workflow entities.
- **Documentation**
  - Expanded CLI references with schema-derived flag guidance and clearer `--spec`/`--spec-file` limitations.
  - Improved command help for partially supported union-based options.

### DIFF
--- a/docs/cli/reference.mdx
+++ b/docs/cli/reference.mdx
@@ -6871,7 +6871,6 @@ nemo agents execute run [OPTIONS]
 
 **Job Spec:**
 
-* `--agent`: Agent entity name or workspace/name ref.
 * `--input`: Prompt to pass to the agent.
 * `--workdir.base-workdir`: Optional Files reference for the initial working directory.
 * `--timeout-seconds <FLOAT>`: Maximum time to wait for Fabric to return an execution result.
@@ -6897,7 +6896,6 @@ nemo agents execute submit [OPTIONS]
 
 **Job Spec:**
 
-* `--agent`: Agent entity name or workspace/name ref.
 * `--input`: Prompt to pass to the agent.
 * `--workdir.base-workdir`: Optional Files reference for the initial working directory.
 * `--timeout-seconds <FLOAT>`: Maximum time to wait for Fabric to return an execution result.
--- a/docs/cli/reference.mdx
+++ b/docs/cli/reference.mdx
@@ -6573,6 +6573,11 @@ nemo agents analyze run [OPTIONS]
 * `--spec`: Spec as a JSON string. [default: \{}]
 * `--spec-file <PATH>`: Path to a YAML or JSON spec file (used as base; per-flag values override).
 
+Job Spec flags are generated from the AnalyzeBatchConfig Pydantic schema. Precedence: --spec-file (base) → --spec JSON (overlay) → per-flag values (top).
+
+Some spec fields cannot be represented as CLI flags: `format`. Supply them with --spec or --spec-file. Individual flags override values from the supplied spec.
+​
+
 ##### nemo agents analyze submit
 
 Submit to a cluster.
@@ -6606,6 +6611,11 @@ nemo agents analyze submit [OPTIONS]
 * `--cluster`: Configured cluster name to resolve via the NeMo CLI config.
 * `--base-url`: Explicit plugin-service base URL (overrides --cluster and all other submit host resolution).
 * `--workspace`: Workspace scope for the submission. [default: default]
+
+Job Spec flags are generated from the AnalyzeBatchConfig Pydantic schema. Precedence: --spec-file (base) → --spec JSON (overlay) → per-flag values (top).
+
+Some spec fields cannot be represented as CLI flags: `format`. Supply them with --spec or --spec-file. Individual flags override values from the supplied spec.
+​
 
 ##### nemo agents analyze explain
 
@@ -6673,6 +6683,9 @@ nemo agents evaluate run [OPTIONS]
 * `--spec`: Spec as a JSON string. [default: \{}]
 * `--spec-file <PATH>`: Path to a YAML or JSON spec file (used as base; per-flag values override).
 
+Job Spec flags are generated from the EvaluateAgentSpec Pydantic schema. Precedence: --spec-file (base) → --spec JSON (overlay) → per-flag values (top).
+​
+
 ##### nemo agents evaluate submit
 
 Submit to a cluster.
@@ -6707,6 +6720,11 @@ nemo agents evaluate submit [OPTIONS]
 * `--cluster`: Configured cluster name to resolve via the NeMo CLI config.
 * `--base-url`: Explicit plugin-service base URL (overrides --cluster and all other submit host resolution).
 * `--workspace`: Workspace scope for the submission. [default: default]
+
+Job Spec flags are generated from the EvaluateAgentSpec Pydantic schema. Precedence: --spec-file (base) → --spec JSON (overlay) → per-flag values (top).
+
+Some spec fields cannot be represented as CLI flags: `workspace`. Supply them with --spec or --spec-file. Individual flags override values from the supplied spec.
+​
 
 ##### nemo agents evaluate explain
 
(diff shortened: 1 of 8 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -131,7 +133,12 @@
 class ExecuteAgentJobConfig(BaseModel):
     model_config = {"json_schema_mode_override": "validation"}
 
-    agent: str = Field(pattern=ENTITY_REF_PATTERN, description="Agent entity name or workspace/name ref.")
+    agent: str | AgentInline = Field(
```
mmogallapalli, line 136: Changing this field from `str` to `str | AgentInline` removes the generated `--agent` CLI option. The shared `_spec_flags` walker skips mixed scalar/model unions, so the generated flags change from `['--agent', '--input', ...]` to `['--input', ...]`. Existing commands using `--agent calc` will therefore break unless callers switch to `--spec`.

Could we extend the CLI walker to emit the single scalar arm (`str`) as `--agent`, while keeping the `AgentInline` arm available through `--spec`/`--spec-file`, and add a regression test for both input forms?

mikeknep: Hmm, interesting. I took a look at the CLI help output and indeed this field is now gone, plus the `environment: str | AgentEnvironmentInline | None` field that @benmccown added doesn't show up as an option, and also the `workdir: AgentWorkdir | None` exposes the base_workdir string...
```
  --workdir.base-workdir <WORKDIR__BASE_WORKDIR>  Optional Files reference for the initial working directory.
```
...but not the artifact_mounts list. It kind of seems like in practice the spec/spec-file args are going to be the only really useful way to invoke this from the CLI... what do you think? I'm almost inclined to implement a way to _opt out of_ the generated flags entirely 🤔

mikeknep: Chatting with Claude about this, seems like there are three possible directions here.
1. Add full support for all these complex argument types as flags: not just "union between scalar and models" but also nested lists. (This sounds complicated enough to warrant its own PR, if we want to even try taking it on at all.)
2. Add a per-job, all-or-nothing `generate_cli_flags: ClassVar[bool]` attribute to `NemoJob`; default to True (all existing jobs) but set this one to False; force users go entirely through `--spec/--spec-file`
3. Leave per-field generation as-is, but include some sort of messaging when some args are skipped. In the case of this job, for example, we would still generate flags for `--input` and `--timeout-seconds`, but then somehow we'd include a contextual note in the help output like "this job's spec includes required fields not available as flags; use --spec/--spec-file".

I kind of like option 3, because being able to still pass the simple `--input` and `--timeout-seconds` args seems friendly. That said, I'm not sure...
- where/how we would display the contextual note
- how much helpful content we could include in it (names of the fields that aren't supported as flags? descriptions as well?)
- if/how we would avoid "partial" flags, like for `--workdir` where we currently expose only exactly half of that model's fields

mmogallapalli: I like option 3 for discoverability, but I think it would be worth preserving a supported scalar arm when a field changes from a scalar to a scalar/model union.

For example, `agent: str | AgentInline` could continue generating `--agent TEXT` for the `str` arm, while the inline-object arm remains available only through `--spec`/`--spec-file`. The same could apply to the string arm of `environment`. This seems substantially narrower than adding full CLI support for complex models and nested lists, and it avoids breaking existing commands that use `--agent calc`.

Separately, a generated help note could identify the fields or paths that are unavailable as individual flags, for example:

> Some spec fields cannot be represented as CLI flags: `agent` (inline form), `environment` (inline form), and `workdir.artifact_mounts`. Supply them with `--spec` or `--spec-file`. Individual flags override values from the supplied spec.

That would retain useful simple flags and backward compatibility while making partial field generation explicit.

mikeknep: Ah, I misread your initial comment here. I think what you're proposing makes a lot of sense

benmccown (approved):
LGTM, minus the outstanding work in flight for the cli flag

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
mikeknep: Addressed in 7cadb1602

packages/nemo_platform_plugin/tests/test_spec_flags.py
```diff
@@ -95,6 +95,52 @@ class _Custom(BaseModel):
...
+    def test_mixed_scalar_model_union_reports_unavailable(self) -> None:
+        class _Inline(BaseModel):
+            value: str
+
+        class _Mixed(BaseModel):
+            agent: str | _Inline
+
+        unavailable: list[str] = []
+        walk_spec_leaves(_Mixed, unavailable=unavailable)
+        assert unavailable == ["agent (inline form)"]
```
mmogallapalli: I think this needs to be updated to be `agent (other union forms)` same with a few tests below

docs/cli/reference.mdx
```diff
@@ -6871,7 +6871,6 @@ nemo agents execute run [OPTIONS]
 
 **Job Spec:**
 
-* `--agent`: Agent entity name or workspace/name ref.
```
mmogallapalli, line 6874: I think if `make generate-cli-reference-docs` is run these should come back.

mmogallapalli (approved):
two minor comments, otherwise lgtm

packages/nemo_platform_plugin/tests/test_spec_flags.py
mikeknep: 🤦 thanks, updating

docs/cli/reference.mdx
mikeknep: Thanks. Actually: running this revealed that quite a few jobs are affected by this, not just ExecuteAgentJob (which isn't too surprising). I wonder if we want to just go ahead and update all of these with this change, or if we should make it opt-in.

mikeknep: Discussed on Slack, decided to just add them all in. It's more informative and avoids bloating cli-generation code with more opt in/out functionality.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -409,6 +434,61 @@ def _save_fabric_error_results(
...
+        # ``ExecuteAgentJobConfig._validate_agent``; this is the spec check.
+        try:
+            inline_config = AgentConfig.model_validate(agent.config)
+        except Exception as exc:
+            raise ValueError(f"Inline agent config is invalid: {exc}") from exc
+        return _AgentSource(
+            name=inline_config.name,
+            workspace=workspace,
+            config=agent.config,
+            config_format=agent.config_format,
```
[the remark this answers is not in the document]
mikeknep: Started an internal thread to discuss this. We don't have an obvious existing mechanism for this today. For purposes of unblocking this stack I'm punting on this for now, with the understanding that anyone given create-execute-agent-job permissions effectively has the near-equivalent of create-agent permissions. Might revisit in a future PR

PR #1598, opened 2026-08-27
feat(insights): Analyzer as agent execution job

mikeknep:
## Summary
Adds a notion of "execution extensions" to the execute-agent job.
This allows clients to define side-effect behavior that should occur within the job process, after the agent invocation completes.
The implementation (using Python entry-points) effectively limits "who/what" can define and provide extension logic to the installed packages/plugins;
we don't accept completely untrusted code.

Implements the Insights Analyst as an ExecuteAgent job.
- Defines a custom Fabric adapter so that the nooa Agent can be invoked through Fabric
- Defines an extension for the side effect of actually persisting insights

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

---

## Summary by CodeRabbit

* **New Features**
  * Added Insights Analyst analysis runs through the platform API.
  * Added trusted extensions for agent executions, including post-run Insights analysis and report results.
  * Added the Insights Analyst Fabric adapter with model, workspace, and Ethos configuration.
  * Added CLI support for extension kinds in local and cluster runs.
* **Documentation**
  * Added an end-to-end Insights Analyst example with sample scenarios, data seeding, run submission, and report inspection.
* **Bug Fixes**
  * Improved validation for analysis run inputs and error handling for failed executions.


plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -0,0 +1,142 @@
...
+    timeout_seconds: float | None = Field(default=None, gt=0, description="Optional execute job timeout.")
+
+
+class CreateAnalysisRunResponse(BaseModel):
+    """Response returned after creating the backing execute-agent job."""
+
+    job: dict[str, Any]
+
+
+@router.post("/analysis-runs", response_model=CreateAnalysisRunResponse)
```
mikeknep, line 77: This endpoint currently just calls out to the agents service to create an `ExecuteAgentJob` and returns that response (wrapped as lightly as possible). We'll probably want to persist some Insights-owned entity here so that users can list historical analysis runs? I'm not sure yet what kind of patterns we want to establish for "specific ExecuteAgentJob implementations defined by other plugins"; for example, when you get the Job ID back, you can access everything about that job via Agents-plugin endpoints, but if you don't keep track of that Job ID I'm not sure how you'd re-discover it later. I think the _minimum_ would be:
- `AnalysisRun` entity that stores a job ID
- `list_analysis_runs` endpoint to list the runs

That might actually be all we need server-side, and then Insights clients (SDK, CLI) would be implemented such that they know how to take those job IDs and get what users want by _actually_ talking to the Agents and/or Jobs plugins.

mikeknep: Been digging into this. We'll want to create an `AnalysisRun` entity and give it a unique name, e.g. `insights-analysis-{uuid}`. Then when we submit the job, we pass that name as the job name—this ensures we have a link between the Insights-owned domain object and the lower-level Job object, without having to get into failure modes around dual writes.

The Insights service can expose list and get endpoints for analysis runs—in addition to the name / job ID we'd also the high-level, domain-specific details of the request (agent, models, since...)

I'll have to also wire up the job name getting persisted on the created `Insight` records too.

mikeknep: Added ^ in layer 5 of this stack

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
```diff
@@ -0,0 +1,76 @@
...
+            _persist_result(
+                workspace=workspace,
+                agent=agent,
+                result=result,
+                insights_output=config.insights_output,
+                local_only=config.local_only,
+            )
+        )
+
+        report_path = context.ctx.storage.persistent / REPORT_FILE_NAME
```
benmccown: for docker/k8s jobs backend scenarios, ExecuteAgentJob.compile() doesn't declare PERSISTENT_JOB_STORAGE_PATH_ENVVAR so a PVC is never mounted. A job that includes `insights.analysis` as an extension will end up raising an exception every time it runs and the job will fail at the persist step. This won't happen with the subprocess backend, so we didn't catch this. Maybe worth an integration or end to end test to catch once we address? As far as how to address, there's some nuance here that's up to you. We don't need to blindly mount the PVC for every ExecuteAgentJob, only ones that reference an extension that needs persistent storage. My agent mentioned AIRCORE-844 might be relevant (something about a first class 'requires persistent storage' field as a possible feature). This one's blocking since we can't have jobs failing every execution :)

---

**Agent Summary** — *P0, blocking; found by sonnet* — This line reads `ctx.storage.persistent`, but the execute-agent job never requests a persistent volume, so on the Kubernetes/Docker backends this raises `RuntimeError` and fails the job every run; it only survives locally because the subprocess backend always provides the path.

> Relevant code:
> - **Crash site:** `plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py:50` — `report_path = context.ctx.storage.persistent / REPORT_FILE_NAME`.
> - **Missing declaration:** `plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py` `compile()` builds `environment=_secret_environment(step_config.secrets)` only and never declares `PERSISTENT_JOB_STORAGE_PATH_ENVVAR`.
> - **Correct contrast:** `plugins/nemo-insights/src/nemo_insights_plugin/jobs/analyze.py:98-99` — `AnalyzeJob.compile()` DOES declare `EnvironmentVariable(name=PERSISTENT_JOB_STORAGE_PATH_ENVVAR, value=DEFAULT_JOB_STORAGE_PATH)`.
> - **The contract:** `services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py` (~L1085) — "Persistent job storage (PVC mount) is only provisioned when the step explicitly declares NEMO_JOB_PERSISTENT_JOB_STORAGE_PATH"; `packages/nemo_platform_plugin/src/nemo_platform_plugin/job_context.py:68-72` raises `RuntimeError("This job did not request persistent storage…")`. `services/core/jobs/.../backends/subprocess.py:450` always sets it — which is why local dev + the unit test (which hand-builds `persistent`) stay green.
> - **Nuance:** `_persist_result` (the SDK/DB write) runs on the line *above* this one, so insights may already be written to the platform when this line then fails — a half-success that still reports the job as failed.

```
Verify this finding against the current code and only fix it if still present.
In plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py compile(), the
execute-agent step never declares PERSISTENT_JOB_STORAGE_PATH_ENVVAR, so on the
k8s/docker backends no PVC is mounted and the insights.analysis extension's
`ctx.storage.persistent` access (execute_extensions.py:50) raises RuntimeError,
failing the job every run. Do NOT unconditionally declare the env var (a noop
execute-agent job shouldn't be forced a PVC); instead plumb the attached
extension's persistent-storage requirement into compile() so the PVC is
declared only when an extension needs it (see AnalyzeJob.compile() analyze.py:98
for the declaration pattern, and AIRCORE-844 for the proposed first-class field).
Add an integration/e2e test that drives create -> run -> extension -> persist on
a non-subprocess backend so this is caught; the current unit fixture hand-builds
persistent storage and cannot. Run the narrowest useful verification.
```

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -272,6 +292,9 @@ async def to_spec(  # type: ignore[override]
             sdk = cast(AsyncNeMoPlatform, async_sdk)
             workdir = await validate_agent_workdir(request.workdir, sdk.files, default_workspace=workspace)
 
+        extension = request.extension or _make_noop_extension_config()
+        resolve_execute_agent_extension(extension.kind)
```
benmccown: If we're making revisions, seems like an easy addition/win?

---

**Agent Summary** — *P3, non-blocking; found by opus, nemotron-ultra* — At create time this only resolves the extension *kind* and never validates the *contents* of `extension.config`, so a bad/unknown/typo'd config key is accepted, the job runs Fabric, and only then fails deep in `after_invoke` instead of returning a fast create-time 422.

> `resolve_execute_agent_extension(extension.kind)` confirms the kind is installed but does not validate `extension.config` against the extension's own schema. The rest of `to_spec` fails-fast (agent/env/workdir refs → 422 on create), but the extension config body is the one input with no create-time net. Since `InsightsAnalysisExtensionConfig` is `extra="forbid"`, a facade regression in `build_execute_agent_job_config` (e.g. a renamed/typo'd key) — or a direct generic-execute caller — would surface as a mid-job failure after real Fabric work rather than a clean rejection. The envelope IS validated (`ExecuteAgentJobConfig`); only the config body isn't.

```
Verify this finding against the current code and only fix it if still present.
In plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py to_spec (~L296),
extension.config contents are not validated at create time (only the kind is
resolved). Add create-time validation of extension.config against the resolved
extension's config schema so a bad/unknown key returns a 422 on create rather
than failing late in after_invoke. Keep the noop path working. Add a focused
test for a create-time reject of an unknown extension config key, then run the
narrowest useful verification.
```

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
```diff
@@ -0,0 +1,153 @@
...
+            _string_setting(self._settings, "base_url")
+            or os.environ.get("NMP_BASE_URL")
+            or os.environ.get("NEMO_BASE_URL")
+        )
+        result, _backend = await run_analyst_change_set(
+            agent=target_agent,
+            agent_spec=_string_setting(self._settings, "agent_spec"),
+            workspace=workspace,
+            base_url=base_url,
+            client=get_async_task_sdk("insights"),
```
benmccown: seems worth fixing just to ensure we don't leak resources?

---

**Agent Summary** — *P2, non-blocking; found by opus* — The SDK client is constructed as a keyword argument *before* the `model_refs` argument; if building the model refs raises (a misconfigured model), `run_analyst_change_set` is never entered so its `finally` never closes the client, leaking the httpx connection.

> Keyword-argument values evaluate left-to-right, so `client=get_async_task_sdk("insights")` is built before `model_refs=ConfiguredModelRefs(default=_default_model_ref(...), fast=_fast_model_ref(...))`. `_default_model_ref`/`_fast_model_ref` can raise `AnalystAdapterConfigError` ("models.default or harness.settings.default_model is required"). If they do, the client exists but `run_analyst_change_set` (which owns closing it in its `finally`) is never called, so the client is orphaned. `invoke()` catches the error and returns a FAILED result, so nothing crashes — the connection just dangles.

```
Verify this finding against the current code and only fix it if still present.
In plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py _run_analysis
(~L78-90), get_async_task_sdk("insights") is constructed before the model_refs
kwarg; if _default_model_ref/_fast_model_ref raises, run_analyst_change_set is
never entered and never closes the client, leaking it. Restructure so the client
is closed on that error path — e.g. build model_refs before the client, or
acquire the client in a try/finally that closes it if the call is never made.
Add a focused test asserting the client is closed when a model-ref error is
raised, then run the narrowest useful verification.
```

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
```diff
@@ -0,0 +1,76 @@
...
+        report_path.parent.mkdir(parents=True, exist_ok=True)
+        report_path.write_text(report, encoding="utf-8")
+        context.ctx.results.save(REPORT_RESULT_NAME, report_path)
+
+
+def _analyst_result_from_fabric_output(output: Any) -> AnalystResult:
+    if not isinstance(output, dict):
+        raise ValueError("Insights analysis extension expected Fabric output to be a mapping.")
+    payload = output.get("analyst_result", output)
```
benmccown: If I'm tracking correctly, the dict fallback might not make sense. Is there a reason it should default to the output object?

---

**Agent Summary** — *P3, non-blocking; found by sonnet, opus, nemotron-super* — When the Fabric output has no `analyst_result` key, this falls back to validating the *entire* output dict as an `AnalystResult`, producing a confusing field-level Pydantic error instead of a clear "missing analyst_result" message.

> `payload = output.get("analyst_result", output)` — the default is the whole `output` dict. The adapter always emits `analyst_result` today, so this only bites on a shape change or if a non-analyst adapter is wired to this extension; in that case the raw output is validated as `AnalystResult` and the error names missing `AnalystResult` fields rather than the actual problem. The non-dict case one line above is already handled with a clear `ValueError`.

```
Verify this finding against the current code and only fix it if still present.
In plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
_analyst_result_from_fabric_output (~L56-60), replace the
output.get("analyst_result", output) fallback with an explicit check: if
"analyst_result" not in output, raise a descriptive ValueError naming the
missing key. Add a focused test for the missing-key case, then run the narrowest
useful verification.
```

plugins/nemo-insights/tests/test_execute_extensions.py
```diff
@@ -0,0 +1,84 @@
...
+from nemo_insights_plugin.jobs.analyze import REPORT_RESULT_NAME
+from nemo_platform_plugin.job_context import JobContext, StoragePaths
+from nemo_platform_plugin.job_results import LocalJobResults
+
+
+@pytest.fixture
+def ctx(tmp_path: Path) -> JobContext:
+    persistent = tmp_path / "persistent"
+    ephemeral = tmp_path / "ephemeral"
+    persistent.mkdir()
```
benmccown, line 24: **Agent Summary** — *P2, non-blocking; found by gpt-sol, sonnet, nemotron-super, opus* — Unit coverage is solid but two gaps leave real regressions invisible: this fixture hand-builds persistent storage (so the missing-PVC crash can't be caught), and nothing exercises the real entry-point dispatch path end to end.

> - This `ctx` fixture does `persistent = tmp_path / "persistent"; persistent.mkdir()`, manufacturing the persistent storage that the real k8s/docker execute-agent job never mounts — so no test here can catch the persist-step crash.
> - The execute-job tests patch out `run_execute_agent_after_invoke_extension`, and this test constructs `InsightsAnalysisExtension` directly, so nothing drives the real seam: register the `insights.analysis` entry point → resolve by kind → dispatch through `ExecuteAgentJob.run` → persist. A packaging/entry-point-registration regression would leave the whole suite green.
> - An integration test on a non-subprocess backend driving create → run → Fabric → extension → persist would cover both. (Also minor: no negative test for the missing-`analyst_result` fallback.)

```
Verify this finding against the current code. Add integration-level coverage for
the execute-agent extension pipeline that (a) exercises real entry-point
discovery/resolution of insights.analysis through ExecuteAgentJob.run rather than
patching the dispatcher or constructing the extension directly, and (b) runs on a
backend that does NOT auto-provide persistent storage so the missing-PVC path is
observable. Keep the existing unit tests. Run the narrowest useful verification.
```

benmccown (changes requested):
One blocking issue my agent caught. Rest are nits/minor while you're in there.

---

**Agent Summary** — *independent multi-model review; 4 reviewers (opus, sonnet, gpt-sol, nemotron-super); findings parent-verified against the diff at this SHA.*

- **Blocking (1):** the `insights.analysis` extension crashes at the persist step on the Kubernetes/Docker job backends because the execute-agent job never requests persistent storage (inline comment on `execute_extensions.py:50`). It only survives on the subprocess backend, which is why local runs and the unit tests pass.
- **Non-blocking (4):** create-time validation of extension config (`execute.py:296`); an SDK client leak on a model-ref error (`fabric_adapter.py:83`); the implicit `analyst_result` fallback (`execute_extensions.py:59`); a silent `"default"` workspace fallback (`fabric_adapter.py:71`); and a test-coverage gap — no integration/entry-point-dispatch test (`test_execute_extensions.py`).

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
mikeknep: Good catch. I think rather than declaring persistent storage, the simpler fix here is for the extension to save the report in the `ctx.storage.ephermal` directory. I don't think any single-step Job should ever require persistent storage 🤔

mikeknep: Swapped to ephemeral storage in dd1cc7ffa

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
mikeknep: Cleaned this up a little wider than just this raw suggestion in 9bcef697f

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
mikeknep: Tightened this up in daa58c2c7

plugins/nemo-insights/tests/test_execute_extensions.py
mikeknep: I'm working on e2e tests, but will probably land them in layer 5 of the stack because I'll want to use the public SDK methods for them, and the endpoints and interfaces aren't fully fleshed out in this stack layer

plugins/nemo-insights/src/nemo_insights_plugin/analyst/run.py
```diff
@@ -82,6 +84,47 @@ async def run_analyst(
...
+            insights_output=insights_output,
+            local_only=local_only,
+            verbose=verbose,
+            since=since,
+            evaluation_id=evaluation_id,
+            analyst_evaluation=analyst_evaluation,
+            enable_observability=enable_observability,
+            model_refs=model_refs,
+            close_client=False,
+        )
```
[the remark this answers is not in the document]
mikeknep: Things got misaligned after a rebase somewhere; fixed this up in 875ecdf00

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -0,0 +1,142 @@
...
+
+INSIGHTS_ANALYSIS_EXTENSION_KIND = "insights.analysis"
+
+router = APIRouter(tags=["Insights Analysis Runs"])
+
+
+class CreateAnalysisRunRequest(BaseModel):
+    """Client-facing request to run the Insights Analyst through ``agents.execute``."""
+
+    agent: str = Field(description="Agent whose telemetry should be analyzed.")
```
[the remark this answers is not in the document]
mikeknep: Addressed in ed1b420d0

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
```diff
@@ -0,0 +1,161 @@
...
+            or "default"
+        )
+        base_url = (
+            _string_setting(self._settings, "base_url")
+            or os.environ.get("NMP_BASE_URL")
+            or os.environ.get("NEMO_BASE_URL")
+        )
+        # Every settings read can raise, so resolve them before opening the
+        # client: nothing is worth a live SDK handle that no one closes.
+        ethos = _string_setting(self._settings, "ethos")
```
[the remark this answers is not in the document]
mikeknep: Fixed in 60c72438d

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
mikeknep: Yep good idea. I had considered a `validate` method on the extension protocol and rejected it earlier; revisiting this led me to a simpler approach where the protocol implementation simply declares the config Pydantic model as a classvar. Implemented in 808cf56d0

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
benmccown: nice fix, that makes sense!

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
benmccown: nice, seems lightweight

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
benmccown: Resolved by 9bcef697 — thanks.

plugins/nemo-insights/src/nemo_insights_plugin/execute_extensions.py
benmccown: Resolved by daa58c2c — thanks.

plugins/nemo-insights/tests/test_execute_extensions.py
benmccown: SGTM!

benmccown (approved):
All five review threads addressed — the blocking persist-storage issue is fixed (report now written to ephemeral storage) and the non-blocking items (create-time extension-config validation, client cleanup, explicit analyst_result check) are in. Noted the e2e coverage is deferred to a later stack layer, which is reasonable. LGTM — thanks for the quick turnaround!

plugins/nemo-insights/tests/test_analyst_fabric_descriptor.py
```diff
@@ -0,0 +1,146 @@
...
+    assert settings["since"] == "2026-08-01T00:00:00+00:00"
+    assert settings["evaluation_id"] == "eval-123"
+    assert settings["base_url"] == "http://platform:8080"
+    assert settings["enable_observability"] is False
+
+
+def test_fabric_plans_a_config_carrying_an_ethos(tmp_path: Path) -> None:
+    """Planning is where a setting the descriptor omits actually blows up."""
+    if not _installed_descriptor().exists():
+        pytest.skip("Adapter descriptor is not installed; Fabric cannot resolve it.")
```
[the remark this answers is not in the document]
mikeknep: I'll punt this to a separate PR

PR #1603, opened 2026-08-27
refactor(studio): replace virtual model side panel with detail routes (ASTD-505)

aray12:
## Summary

The Virtual Models details side panel opened and closed twice on every interaction. Rather than keep fighting the dialog lifecycle, this moves the details and chat content to dedicated routes, mirroring the existing guardrails detail pages. The URL is now the state because it is a route, so the whole class of bug stops being expressible: there is no dialog, no `open` prop to keep in sync, and no exit animation to outlive. Before, clicking a row flashed a panel open and shut and deep links to off-page models silently failed; after, each model has a real page at a stable URL.

Two separate causes produced the double-open:

- The panel's open state was written both imperatively by the click handlers and reactively by an effect reconciling the URL. Router navigations are transitions, so the two writers committed in different renders and fought each other.
- `SidePanel`'s dialog primitive re-shows itself from a layout effect when a render lands while the native dialog is already closed but the controlled `open` prop is still true. Any latency between the dismissal gesture and `open` flipping produced a visible reopen.

https://github.com/user-attachments/assets/e978fb4f-6759-4b70-9065-99cb129d4122

## Changes

- Delete `VirtualModelDetailsSidePanel` and its tests.
- Add routes at `/workspaces/:workspace/virtual-models/:virtualModelName/details` and `/workspaces/:workspace/virtual-models/:virtualModelName/chat`, backed by a new `VirtualModelDetailRoute` with `VirtualModelDetailsTab` and `VirtualModelChatTab`.
- Fetch the detail subject by name instead of looking it up in the current page of list results, so deep links resolve models that are not on the first page, and searching or paginating no longer closes the view.
- Reduce `VirtualModelsDataView` to a plain table that navigates on row click.
- Add the route constants and `getVirtualModelDetailsRoute` helper, and update the title-change test for the new routes.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs describe the Virtual Models Studio screens; this changes in-app navigation only, with no CLI, API, or config surface change.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test src/routes/virtualModels src/components/dataViews/VirtualModelsDataView src/tests/title-change.test.tsx` — passed (3 files, 85 tests).
- `pnpm --filter nemo-studio-ui typecheck` — passed.
- `pnpm lint` (from `web/`, `--max-warnings 0`) — passed.
- `uv run pre-commit run -a` — all hooks passed except two blocked by the local environment and unrelated to this change:
  - `Helm Docs` — failed because `helm-docs` is not installed locally. This change touches no Helm chart.
  - `Run uv lock with platform uv` — failed because the local uv is 0.9.28 while the hook pins 0.9.14. The companion `Check for uv.lock drift` hook passed, and this change adds no Python dependencies.
  - Both are enforced in CI against the pinned toolchain.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated virtual model detail pages with separate Details and Chat tabs.
  * Virtual model rows now navigate to full-page detail and chat routes.
  * Added model metadata, configured models, middleware pipelines, loading states, and error handling.
  * Added chat inference parameter controls, including a default token limit.

* **Bug Fixes**
  * Improved navigation and route-based state handling for virtual model interactions.


web/packages/studio/src/routes/virtualModels/VirtualModelDetailsTab/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+        <KVPair
+          label="Created"
+          orientation="horizontal"
+          size="medium"
+          value={
+            virtualModel.created_at ? (
+              <RelativeTime datetime={virtualModel.created_at} focusableForTooltip={false} />
+            ) : (
+              '—'
+            )
```
steramae-nvidia: Pretty sure KVPair should handle the undefined state automatically with `value={virtualModel.created_at ? <RelativeTime /> : undefined}`

web/packages/studio/src/routes/virtualModels/VirtualModelDetailsTab/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+          }
+        />
+        <KVPair
+          label="Default model"
+          orientation="horizontal"
+          size="medium"
+          truncate={false}
+          value={virtualModel.default_model_entity || '—'}
```
steramae-nvidia: Same here with `value={virtualModel.default_model_entity}`

web/packages/studio/src/routes/virtualModels/VirtualModelDetailsTab/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  if (virtualModel) {
+    return <VirtualModelDetails virtualModel={virtualModel} />;
+  }
+
+  if (error) {
+    return (
+      <ErrorPanel
+        title={`Failed to load virtual model '${virtualModelName}'`}
+        // getErrorMessage intentionally prefers the error's own message over a fallback, so the
+        // technical detail ("404 Not Found") goes in the body and the context goes in the title.
```
steramae-nvidia: Remove comment?

steramae-nvidia (approved):

PR #1613, opened 2026-08-27
feat(studio): delete guardrail test cases from the Test and Validate tab

aray12:
## Summary

A guardrail check could be created and edited in the **Test and Validate** tab but never removed, so a typo'd or obsolete test case stayed in the list forever and could only be cleaned up outside Studio. Individual messages *within* a check already had a delete action; the gap was at the check level. Each test card header now carries a delete action that confirms through the shared `DeleteConfirmationModal` before removing the check.

The plumbing already existed — `useDeleteGuardrailCheck` was implemented but referenced nowhere in the app — so this is the missing UI affordance plus tests.

## Changes

- `GuardrailTestCard.tsx`: a trash action in the card header (matching the icon-button idiom `GuardrailMessageRow` already uses), a `DeleteConfirmationModal` naming the test it would remove, and a `handleDelete` wired to `useDeleteGuardrailCheck`. The hook is used without an `onError` toast — the modal reports its own outcome, so adding one would double up.
- New required `configId` prop on the card: checks are hard children, so the delete call needs the owning config's entity id. `Entity.parent` is optional in the generated schema, whereas the editor already holds `configId` as a definite string.
- `GuardrailTestCasesEditor.tsx`: passes `configId` through.
- Tests in `GuardrailChecksTab/index.test.tsx` covering confirm-before-delete, delete-only-the-confirmed-check, cancel, and the cache invalidation that makes the card disappear.

The card disappears on its own because `useDeleteGuardrailCheck.onSuccess` calls `invalidateGuardrailChecksCaches`, refetching the config's check list. That invalidation targets the app's module-singleton `queryClient`, which is not the one `TestProviders` renders — so the test asserts the `invalidateQueries` call itself rather than pretending to observe the refetch.

**Results sub-tab:** no matching affordance is needed. Runs are stored on the check entity as `data.runs`, so a deleted check takes its run history with it and the Results table (a read view over the same `checks` array) updates from the same refetch. Bulk delete is left out as the issue marks it optional.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal Studio UI affordance, no documented interface changes

## Verification

Targeted validation:

- `pnpm vitest --run src/routes/guardrails src/components/dataViews/GuardrailChecksDataView src/components/sidePanels/GuardrailCheckDetailSidePanel src/tests` — 21 files, 236 tests passed.
- `pnpm vitest --run src/api/guardrail-checks src/components/dataViews/GuardrailsDataView` — 3 files, 48 tests passed (the remaining suites touching the guardrails mock handlers).
- `pnpm exec eslint packages/studio/src/routes/guardrails/GuardrailChecksTab --max-warnings 0` — clean.
- `pnpm exec prettier --check packages/studio/src/routes/guardrails/GuardrailChecksTab` — clean.
- `pnpm exec tsc -p packages/studio --noEmit` — one error in `packages/common/src/components/FilesetSearchableSelect/index.tsx`, verified present on the unmodified tree as well; no new type errors from this change.
- `uv run pre-commit run -a` was not run over the whole repo; the commit hooks (copyright headers, UI lint-staged, DCO, merge-conflict checks) passed. The Python hooks are no-ops here — this change touches only `web/`.

## Stacked on

Builds on #1611 (`fix(studio): append new guardrail test cases to the end of the list`), which this branch's card and editor changes sit directly on top of. Review that one first.

### DIFF
--- a/web/packages/studio/src/routes/guardrails/GuardrailChecksTab/GuardrailTestCard.tsx
+++ b/web/packages/studio/src/routes/guardrails/GuardrailChecksTab/GuardrailTestCard.tsx
@@ -2,9 +2,21 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { getErrorMessage } from '@nemo/common/src/api/common/utils';
+import { DeleteConfirmationModal } from '@nemo/common/src/components/DeleteConfirmationModal';
 import { useToast } from '@nemo/common/src/providers/toast/useToast';
-import { Button, Flex, Panel, SegmentedControl, Stack, Text } from '@nvidia/foundations-react-core';
-import { useUpdateGuardrailCheck } from '@studio/api/guardrail-checks/hooks';
+import {
+  Button,
+  Flex,
+  Panel,
+  SegmentedControl,
+  Stack,
+  Text,
+  Tooltip,
+} from '@nvidia/foundations-react-core';
+import {
+  useDeleteGuardrailCheck,
+  useUpdateGuardrailCheck,
+} from '@studio/api/guardrail-checks/hooks';
 import type {
   GuardrailCheckEntity,
   GuardrailCheckMessage,
@@ -14,7 +26,7 @@ import {
   type GuardrailMessageFormRow,
   GuardrailMessageRow,
 } from '@studio/routes/guardrails/GuardrailChecksTab/GuardrailMessageRow';
-import { Plus } from 'lucide-react';
+import { Plus, Trash2 } from 'lucide-react';
 import { type FC, type FocusEvent, useCallback, useEffect, useRef, useState } from 'react';
 import { useFieldArray, useForm, useWatch } from 'react-hook-form';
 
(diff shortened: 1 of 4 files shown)


web/packages/studio/src/routes/guardrails/GuardrailChecksTab/GuardrailTestCard.tsx
```diff
@@ -171,6 +187,19 @@ export const GuardrailTestCard: FC<GuardrailTestCardProps> = ({
     persist(form.getValues('messages'));
   };
 
+  // The modal reports the outcome itself, so the hook stays quiet — an onError toast here
+  // would double up on the modal's own error text.
+  const deleteMutation = useDeleteGuardrailCheck();
+
+  const handleDelete = useCallback(async (): Promise<boolean> => {
+    try {
```
walston, line 195: Can we log the error somewhere for debugging? at least in dev?

walston (approved):
Looks good to me, one small request

PR #1622, opened 2026-08-28
fix(cve): drop unreferenced mlflow-skinny cache metadata

tylersbray:
## Reviewer context

Trivial Dockerfile glob fix. `mlflow-skinny` wheels install `mlflow_skinny-<ver>.dist-info`; the RL uv-cache sweep only matched `mlflow-*.dist-info` and could leave scanner-visible archives. Please confirm the extra glob sits next to the existing mlflow/sqlparse/wandb loop.

## CVE / sheet

- Cluster: mlflow / mlflow-skinny leftover cache metadata (post-1609 skinny install)
- Change: add `mlflow_skinny-*.dist-info` to `dist_info_glob` in `docker/rl/Dockerfile.nmp-rl-base`
- Images: nmp-rl-base (and downstream RL GPU images)

## Non-changes

Does not bump mlflow, rebuild GPU images locally, or change 1615 / distroless.

## Verification

Cold `nmp-rl-training-smoke-test` is CI only.

Fixed in source; not re-scanned by Pulse.

### DIFF
--- a/tests/unit/test_docker_workspace_slices.py
+++ b/tests/unit/test_docker_workspace_slices.py
@@ -154,12 +154,6 @@ def test_automodel_cve_layer_does_not_pin_full_mlflow_with_cryptography() -> Non
     assert not FULL_MLFLOW_SPEC_RE.search(text)
 
 
-def test_rl_uv_cache_cleanup_includes_mlflow_skinny_dist_info() -> None:
-    """Skinny metadata uses mlflow_skinny-*.dist-info, which mlflow-*.dist-info does not match."""
-    text = (ROOT / "docker/rl/Dockerfile.nmp-rl-base").read_text(encoding="utf-8")
-    assert "mlflow_skinny-*.dist-info" in text
-
-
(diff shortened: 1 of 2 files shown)


tests/unit/test_docker_workspace_slices.py
```diff
@@ -154,6 +154,12 @@ def test_automodel_cve_layer_does_not_pin_full_mlflow_with_cryptography() -> Non
     assert not FULL_MLFLOW_SPEC_RE.search(text)
 
 
+def test_rl_uv_cache_cleanup_includes_mlflow_skinny_dist_info() -> None:
```
mckornfield: this test (and probably a lot in this file) I think are just str comparisons

mckornfield (approved):

tests/unit/test_docker_workspace_slices.py
tylersbray: Dropped this PR's string-grep in 4d918da5da. Pre-existing checks in that file stay; a real bake is the place to assert image contents.

PR #1633, opened 2026-08-28
feat(insights): Persist AnalysisRun entity, list/get endpoints

mikeknep:
## Summary
Adds an `AnalysisRun` entity, with list and get endpoints, so that users can view analysis jobs separate from other agent execution jobs. The `AnalysisRun.name` is used as the job name to link the two.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

---

## Summary by CodeRabbit

- **New Features**
  - Added on-demand Insights analysis runs for submitting, monitoring, listing, and retrieving agent analyses.
  - Added CLI options for model selection, time and evaluation filters, Ethos input, timeouts, sorting, pagination, JSON output, and polling.
  - Added synchronous and asynchronous SDK support with wait helpers and clear timeout or missing-job errors.
  - Analysis runs retain status and backing-job linkage for later retrieval, including runs without a backing job.

- **Documentation**
  - Added CLI reference, SDK guidance, and updated examples for creating and managing analysis runs.

### DIFF
--- a/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
+++ b/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
@@ -30,7 +30,7 @@ from nemo_insights_plugin.analyst.agent_config import AGENT_CONFIG_FORMAT, build
 from nemo_insights_plugin.authz import scope
 from nemo_insights_plugin.entities import AnalysisRun
 from nemo_insights_plugin.schema import AnalysisRunPage, AnalysisRunResponse, CreateAnalysisRunRequest
-from nemo_platform import APIStatusError, AsyncNeMoPlatform
+from nemo_platform import APIConnectionError, APIStatusError, AsyncNeMoPlatform
 from nemo_platform_plugin.authz import CallerKind, path_rule
 from nemo_platform_plugin.dependencies import get_sdk_client
 from nemo_platform_plugin.entity_client import NemoEntitiesClient, NemoEntityNotFoundError, get_entity_client
@@ -93,6 +93,21 @@ async def create_analysis_run(
         # new name and no resubmit-under-an-existing-name route exists yet.
         logger.warning("Analysis run %r recorded but its job was not created: %s", saved.name, exc)
         raise HTTPException(status_code=exc.status_code, detail=_error_detail(exc)) from exc
+    except APIConnectionError as exc:
+        # Same stranded run, but a sibling of APIStatusError rather than a
+        # subclass, so it needs its own arm — and it is the case most likely to
+        # be worth retrying, since the request may never have reached Jobs.
+        # Log the run name: without it the orphan cannot be found afterwards.
+        logger.warning("Analysis run %r recorded but the Jobs service was unreachable: %s", saved.name, exc)
+        raise HTTPException(
+            status_code=503,
+            # Name the run: it exists, and this is the caller's only chance to
+            # learn what was persisted before the response is all they have.
+            detail={
+                "message": "Could not reach the Jobs service to submit the analysis run.",
+                "run": saved.name,
+            },
+        ) from exc
     return AnalysisRunResponse(run=saved, job=job)
 
 
--- a/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
+++ b/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
@@ -92,7 +92,13 @@ async def create_analysis_run(
         # submitted"; recovering one is a read today, since every create mints a
         # new name and no resubmit-under-an-existing-name route exists yet.
         logger.warning("Analysis run %r recorded but its job was not created: %s", saved.name, exc)
-        raise HTTPException(status_code=exc.status_code, detail=_error_detail(exc)) from exc
+        raise HTTPException(
+            status_code=exc.status_code,
+            # ``error`` keeps the Agents service's own detail — a missing entity
+            # or invalid config reads better from the service that rejected it —
+            # while ``run`` names the record this call stranded.
+            detail={"error": _error_detail(exc), "run": saved.name},
+        ) from exc
     except APIConnectionError as exc:
         # Same stranded run, but a sibling of APIStatusError rather than a
         # subclass, so it needs its own arm — and it is the case most likely to
@@ -101,10 +107,11 @@ async def create_analysis_run(
         logger.warning("Analysis run %r recorded but the Jobs service was unreachable: %s", saved.name, exc)
         raise HTTPException(
             status_code=503,
-            # Name the run: it exists, and this is the caller's only chance to
-            # learn what was persisted before the response is all they have.
+            # Same shape as the status arm above: ``run`` names the record this
+            # call stranded, which is the caller's only chance to learn what was
+            # persisted before the response is all they have.
             detail={
-                "message": "Could not reach the Jobs service to submit the analysis run.",
+                "error": "Could not reach the Jobs service to submit the analysis run.",
                 "run": saved.name,
             },
         ) from exc
--- a/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
+++ b/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
@@ -279,6 +279,13 @@ def _inline_analyst(request: CreateAnalysisRunRequest, *, workspace: str) -> dic
             ethos=request.ethos,
             since=request.since,
             evaluation_id=request.evaluation_id,
+            # The Analyst's self-observability builds its own OTLP exporter from
+            # the operator's local CLI config, which does not exist in a task
+            # pod, so on a cluster it exports unauthenticated at best. Telemetry
+            # for these runs belongs in the agent config's relay endpoints,
+            # where the agents layer already carries auth; until that is wired,
+            # a run must not depend on it.
+            enable_observability=False,
         ),
     }
 
--- a/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
+++ b/plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
@@ -3,100 +3,162 @@
 
 """High-level Insights analysis-run request helpers.
 
-This module is intentionally a thin facade over the generic ``agents.execute``
-job. The current ``AnalyzeJob`` can continue to run side-by-side while this
-path exercises the Analyst-as-Agent implementation.
+This module is a thin facade over the generic ``agents.execute`` job. The
+current ``AnalyzeJob`` can continue to run side-by-side while this path
+exercises the Analyst-as-Agent implementation.
+
+Insights persists one :class:`~nemo_insights_plugin.entities.AnalysisRun` per
+run, holding only what the Jobs layer cannot know: that a job was an analysis
+run, and over which agent and scope. Execution state is *not* copied — it is
+read back from the job on demand.
+
+The run is named before the job is submitted and the job takes that same name,
+so the two are linked by construction rather than by a write-back. That is what
+makes the failure modes distinguishable: a run whose job 404s never landed, and
+a resubmission that conflicts is proof the original did.
 """
 
 from __future__ import annotations
 
-from datetime import datetime
-from typing import Annotated, Any
+import logging
+import uuid
+from typing import Any
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Query
 from nemo_insights_plugin._perms import AnalysisRunPerms
 from nemo_insights_plugin.analyst.agent_config import AGENT_CONFIG_FORMAT, build_analyst_agent_config
 from nemo_insights_plugin.authz import scope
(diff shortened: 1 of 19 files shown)


plugins/nemo-insights/src/nemo_insights_plugin/sdk_resources/insights.py
```diff
@@ -56,13 +56,15 @@ def _build_create_body(
     description: str,
     status: InsightStatus | str,
     trace_refs: list[str] | None,
+    analysis_run: str,
```
mikeknep: I'm souring on this field—it too closely couples the Insight entity with how one particular way that entity can be produced. That said, I think having explicit provenance would be good. I wonder if instead of this `analysis_run: str` field we should have something like:
```python
provenance: Provenance
```
where provenance is a discriminated union, something like
```python
class AnalysisRunProvenance:
    kind: Literal["analysis_run"] = "analysis_run"
    name: str


class UnknownProvenance:
    kind: Literal["unknown"] = "unknown"


# would this be useful for lightweight ad-hoc extensibility, or no?
class MiscProvenance:
    kind: Literal["misc"] = "misc"
    details: dict[str, Any]
```
and we can add more as we care to track them. Anyone have thoughts?

plugins/nemo-insights/src/nemo_insights_plugin/service.py
```diff
@@ -128,6 +128,7 @@ async def create_insight(
             description=body.description,
             status=body.status,
             trace_refs=list(body.trace_refs),
+            analysis_run=body.analysis_run,
```
mmogallapalli: Since `analysis_run` represents provenance, would it make sense to verify that the referenced run exists in the same workspace and belongs to the same agent? It’s currently stored directly from the request, so a stale or mismatched relationship could otherwise be recorded.

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -52,46 +71,145 @@ class CreateAnalysisRunRequest(BaseModel):
...
+    Minted by Insights rather than assigned by the Jobs service so the run can
+    be persisted *before* the job is submitted. Uniqueness comes from the uuid,
+    not from the agent — an earlier version derived the name from the agent and
+    collided on the second run.
+    """
+    return f"{ANALYSIS_RUN_NAME_PREFIX}{uuid.uuid4().hex}"
 
 
-@router.post("/analysis-runs", response_model=CreateAnalysisRunResponse)
+@router.post("/analysis-runs", response_model=AnalysisRunResponse, status_code=201)
```
mmogallapalli, line 58: I’m not sure callers can retry this run under the same name yet. The run is preserved when job creation fails, but the error response doesn’t include its name, and every new POST mints another UUID. After an ambiguous failure, a retry could create another run and duplicate the job. Could we add an idempotency/resubmit mechanism, or return enough information to recover the persisted run?

mikeknep: Yep, good callouts. Maybe a route like this:
```
POST /analysis-runs/{name}/reattempt  # or /retry
```

plugins/nemo-insights/src/nemo_insights_plugin/service.py
mikeknep: I think there are a lot of questions around provenance. I left a separate comment about it on this PR, but also insight entities can be "touched" by multiple executions of the analyst (via analysis-run or otherwise), and I'm not sure if we want to track each of those somehow (e.g. updating a list on insight records) or what.

plugins/nemo-insights/src/nemo_insights_plugin/sdk_resources/insights.py
mikeknep: I'm going to back out of this for now. Today on `main` there is no provenance captured on insight entities, and there is no API for fetching "insights created/updated by Job X"; the only record is the human-readable report produced as a job result. Provenance would be net-new and requires deeper understanding of the product requirements. Going to back out of this for now

plugins/nemo-insights/src/nemo_insights_plugin/service.py
mikeknep: I'm backing out of the `analysis_run` provenance field entirely, see https://github.com/NVIDIA-NeMo/nemo-platform/pull/1633#discussion_r3905918922

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -0,0 +1,213 @@
...
+        # submitted" and can be resubmitted under the same name.
+        logger.warning("Analysis run %r recorded but its job was not created: %s", saved.name, exc)
+        raise HTTPException(status_code=exc.status_code, detail=_error_detail(exc)) from exc
+    return AnalysisRunResponse(run=saved, job=job)
+
+
+@router.get("/analysis-runs", response_model=AnalysisRunPage)
+@scope.read
+@path_rule(callers=[CallerKind.PRINCIPAL], permissions=[AnalysisRunPerms.LIST])
+async def list_analysis_runs(
```
benmccown, line 128: I'd classify this as a nit, but maybe worth addressing.

---

**Agent Summary** — *P2, non-blocking; found by opus, sonnet, gpt-codex, deepseek-flash, gpt-sol* — The `list_analysis_runs` route handler has no direct route-level test, so its filter/sort/pagination assembly is only exercised indirectly through SDK/CLI stubs.

> `create_analysis_run` and `get_analysis_run` each have tests that call the handler directly with a stubbed entity client, but `list_analysis_runs` is never imported or called in any test file (`_StubEntities` has only `create`+`get`, no `list`). The list route does real work — building `filter_obj={"agent": agent}`, passing the `sort` default, and reshaping `result.pagination` via `PaginationData.model_validate(...)` — none of which is asserted against the real handler. It's a thin passthrough so risk is low, but it's the PR's headline feature going untested at the route layer.

```
Add a route-level unit test for list_analysis_runs in
plugins/nemo-insights/tests/test_analysis_runs.py: give the entity-client stub a
`list` method returning a paginated result, call the route handler directly, and
assert the agent filter_obj is built, the sort default is passed through, and the
pagination is reshaped into the response envelope. Mirror the existing
create/get route tests. Run the narrowest useful check.
```

plugins/nemo-insights/src/nemo_insights_plugin/sdk_resources/analysis_runs.py
```diff
@@ -0,0 +1,301 @@
...
+                fast_model=fast_model,
+                since=since,
+                evaluation_id=evaluation_id,
+                timeout_seconds=timeout_seconds,
+            ),
+        )
+        response.raise_for_status()
+        return _run_response_from_response(response.json())
+
+    def list_runs(
```
benmccown, line 160: Another small nit

---

**Agent Summary** — *P3, non-blocking; found by sonnet, deepseek-flash, qwen3-coder* — The list route accepts a `sort` query param, but the SDK `list_runs` and the CLI `list` never forward one — they hardcode the default — so server-side sorting by any other field is unreachable except via raw HTTP.

> The route has `sort: str = Query(default="-created_at")`, but `list_runs` (and its async twin) + the CLI `list` command only pass `page`/`page_size`/`agent`. Harmless today since the default matches, but the route's sort capability is dead from the supported surface. Either expose `--sort` / a `sort=` kwarg, or drop the route param if sort isn't intended to be configurable yet.

```
Either expose sort on the SDK list_runs (sync+async) and the CLI `list` command
(forwarding to the route's sort query param), or remove the sort param from the
list_analysis_runs route if server-side sort isn't meant to be configurable yet.
Keep the -created_at default. Add/adjust a focused test. Run the narrowest check.
```

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -0,0 +1,213 @@
...
+
+    spec = build_execute_agent_job_config(request, workspace=workspace, run_name=saved.name)
+    try:
+        job = await sdk.agents.jobs.execute.create(spec=spec, name=saved.name, workspace=workspace)
+    except APIStatusError as exc:
+        # The run record is deliberately left in place. Deleting it here could
+        # remove the only pointer to a job that was in fact created (a create
+        # that timed out client-side still lands), which is the untracked-job
+        # case this design exists to prevent. A run with no job reads as "never
+        # submitted" and can be resubmitted under the same name.
```
benmccown: is resubmit a feature we'll add? I don't think it's in the PR at present. So the comment might just need updated.

---

**Agent Summary** — *P3, non-blocking; found by gpt-sol* — This comment says a stranded run "can be resubmitted under the same name," but no route/SDK/CLI operation actually accepts an existing run name — every create mints a fresh uuid — so the documented resubmit path isn't reachable today.

> `create_analysis_run` always sets `name=mint_analysis_run_name()` (a fresh uuid); the server exposes only create/list/get, and the SDK/CLI create take no existing-run identifier. So after an ambiguous client-side timeout, a naive retry creates a *new* run + a *new* job while the first job may already be running (duplicate analysis work), and the stranded first run can't be recovered by name. The record-first design and leaving the record in place are sound; it's only the "can be resubmitted" wording that implies a capability that isn't present yet.

```
Either (a) add a resubmit-under-existing-name operation (a create that accepts an
existing AnalysisRun name and submits the backing job under it), or (b) soften the
comment in create_analysis_run's failure handler so it no longer implies a
resubmit capability that isn't exposed. No functional change needed for (b).
```

plugins/nemo-insights/src/nemo_insights_plugin/cli.py
```diff
@@ -493,6 +699,149 @@ async def _analysis_config_command(
...
+) -> tuple[str, bool]:
+    """Poll one run to a terminal job state, reporting status changes on stderr."""
+    response = await client.insights.analysis_runs.wait(
+        workspace=workspace,
+        name=name,
+        timeout=poll_timeout,
+        poll_interval=poll_interval,
+        on_status=_status_reporter(),
+    )
+    return _json(response.model_dump(mode="json")), response.job_status == "completed"
```
benmccown: small nit/nice-to-have

---

**Agent Summary** — *nit, non-blocking; found by sonnet* — `_wait_for_run` compares `job_status` to the string literal `"completed"` instead of `PlatformJobStatus.COMPLETED.value`; identical today, but a silent divergence if the enum value ever changes, and inconsistent with `schema.py`'s `job_is_terminal` which uses the enum set.

> `return _json(...), response.job_status == "completed"` hardcodes the string, while `schema.py`'s `job_is_terminal` uses `{t.value for t in PlatformJobStatus.terminals()}`. Using `PlatformJobStatus.COMPLETED.value` here keeps the CLI in lockstep with the enum.

```
In plugins/nemo-insights/src/nemo_insights_plugin/cli.py:842, replace the
"completed" string literal with PlatformJobStatus.COMPLETED.value so the success
check stays in sync with the enum (matching schema.py job_is_terminal). Run the
narrowest check.
```

benmccown (approved):
LGTM, a few small nits. Feel free to resolve open comment threads, I'll let approval ride

---

**Agent Summary** — *independent multi-model review; 7 reviewers (opus, sonnet, gpt-sol, gpt-codex, qwen3-coder, gemini-flash, deepseek-flash); findings parent-verified against the diff at this SHA.* No blocking issues. Clean, well-tested PR that follows the plugin's existing patterns; the derivable-shared-name design (run name == job name, no write-back) is nicely done. Four non-blocking nits left inline: the list route handler lacks a direct route-level test (5/7 reviewers); the route's `sort` param isn't exposed on the SDK/CLI; the failure-handler comment implies a resubmit capability that isn't present yet; and a hardcoded `"completed"` string vs the enum.

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
mikeknep: Decided to punt on this, made a separate ticket where we can discuss exact requirements, desired UX, etc.

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
mikeknep: Addressed in b1135fd82

plugins/nemo-insights/src/nemo_insights_plugin/sdk_resources/analysis_runs.py
mikeknep: Addressed in 7d1068b99. Note it was already supported on the SDK, but indeed was missing from the CLI

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
mikeknep: Adjusted the comment in 5e20a0643, and made a separate dedicated ticket for resubmit/reattempt

plugins/nemo-insights/src/nemo_insights_plugin/cli.py
mikeknep: Addressed in 5605200a3

PR #1660, opened 2026-08-31
fix: correct ssl handing so the RL docker build works again

crookedstorm:
## Summary
The docker build for the RL base container is breaking right now. It seems to be the change in how python was installed and ssl handling, though there could be another problem with the causal-conv1d wheel. Testing this and making sure this is the right way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTPS compatibility for extensions using the bundled Python runtime.
  * Added installation-time verification of SSL functionality and OpenSSL version detection.
  * Improved build reliability for Transformer Engine components by limiting parallel build activity.


svvarom (approved):

crookedstorm:
This works now, after the second commit.. https://github.com/NVIDIA-NeMo/Platform-Deploy/actions/runs/33439486500/job/99644046864

PR #1667, opened 2026-08-31
fix(agents): remove legacy job verbs

mckornfield:
## Summary
Remove the generated `run`, `submit`, and `explain` verb layer from nemo-agents job commands so users invoke jobs directly, for example `nemo agents evaluate` and `nemo agents optimize`. Preserve the job-specific `nemo agents optimize prepare-fileset` flow by wrapping the direct optimize command with its existing subcommand.

## Changes
- Disable legacy generated verbs for nemo-agents job entry points and the agents-mounted optimize job.
- Preserve `nemo agents optimize prepare-fileset` while keeping `nemo agents optimize` as the direct remote submission command, including generated epilog/help metadata.
- Use job descriptions for flat generated job commands so top-level help remains discoverable.
- Require the optimize config fileset with actionable `prepare-fileset` guidance in the remote submit schema while preserving local scheduler behavior.
- Preserve local optimize workspace context in the canonical job spec.
- Propagate submitted Anthropic base URLs into agent-improvement subprocess environments alongside secret-backed API keys.
- Update docs, examples, plugin skills, generated CLI reference, OpenAPI text, Studio empty-state copy, and agentic-use fixtures to use the direct job command flow, absolute host paths, and platform secret/base-URL flags where direct jobs require them.
- Add CLI, job-environment, submit-schema, optimize-job, and anonymizer flat-job coverage for direct commands, retained job-specific subcommands, schema validation, local execution compatibility, and command help metadata.
- Merge current `origin/main` so CI validates against the same dependency lock state used by the PR merge commit.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [ ] Existing tests cover changed behavior - justification:
- [ ] Tests not applicable - justification:
- [ ] Documentation not applicable - justification:

## Verification

Targeted validation:

- Trusted base refreshed to `origin/main` at `2a1d3c2a9`; PR head `bb1cfa9a72657ee6b6fe0590a7ae8f2c6fea1be6` is mergeable.
- DCO audit over `origin/main..HEAD` - passed for `1044b26afe7dd7e2abf9449c58aaf7f7e6c961ff`, `d4bf24482b2659164bc40241f1c27e7317accc97`, `0ef1e9503a242c7cc67cb540be5c971cb94bf27e`, and `bb1cfa9a72657ee6b6fe0590a7ae8f2c6fea1be6`.
- `flox -q activate --dir /home/mkornfield/wt-np-6557144-agentic-base-image -- uv run pre-commit run -a` - passed after merging current `origin/main`; reran after the final base merge with no `uv.lock` drift.
- `uv run --frozen pytest plugins/nemo-anonymizer/tests/unit/test_cli.py -v` - 4 passed at current head.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_cli.py -v` - 48 passed at current head.
- `make docs-check` - passed after `make docs-deps` installed missing local Fern dependencies.
- `flox -q activate --dir /home/mkornfield/wt-np-6557144-agentic-base-image -- pnpm --dir web --filter @nemo/common typecheck` - passed.
- `flox -q activate --dir /home/mkornfield/wt-np-6557144-agentic-base-image -- pnpm --dir web --filter @nemo/common lint` - passed.
- `uv run _nemo agents evaluate --help` - passed.
- `uv run _nemo agents optimize --help` - passed.
- `uv run _nemo agents optimize prepare-fileset --help` - passed.
- `uv run _nemo agents optimize-skills --help` - passed.
- `uv run _nemo agents evaluate-suite --help` - passed.
- `rg -n "nemo agents (analyze|evaluate|evaluate-suite|execute|optimize|optimize-skills|package-agent) (run|submit|explain)|agents (analyze|evaluate|evaluate-suite|execute|optimize|optimize-skills|package-agent) (run|submit|explain)" README.md docs tests web plugins packages services` - no matches.
- `make refresh-openapi` - passed before the follow-up comment sweep.
- `flox -q activate --dir /home/mkornfield/wt-np-6557144-agentic-base-image -- make generate-cli-reference-docs` - passed before the follow-up comment sweep.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_optimize_prepare_fileset_cli.py plugins/nemo-agents/tests/unit/test_cli.py plugins/nemo-agents/tests/unit/test_improvement_jobs.py plugins/nemo-optimization/tests/test_optimize_job.py packages/nemo_platform_plugin/tests/test_commands.py -v` - 188 passed before the follow-up comment sweep.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit -k evaluate_suite -v` - 10 passed, 1629 deselected.
- `uv run --frozen pytest plugins/nemo-agents/tests/unit -k "optimize_skills or analyze" -v` - 19 passed, 1620 deselected.
- `uv run --frozen pytest packages/nemo_platform_plugin/tests/test_jobs_routes.py -v` - 23 passed.
- `pnpm --filter @nemo/sdk gen:platform` - passed; generated ignored local SDK files needed for web typecheck.
- `pnpm --filter @nemo/sdk gen:platform-zod` - passed; generated ignored local SDK Zod files needed for web typecheck.
- `uv run --frozen ty check packages/nemo_platform_plugin plugins/nemo-agents plugins/nemo-optimization` - still reports 139 existing diagnostics outside the files changed here; no remaining diagnostics matched the touched source or route-test fixtures.
- `git diff --check && git diff --cached --check` - passed before commit and after the base merge.
- `rg -n '^(<<<<<<<|>>>>>>>)' .` - no conflict markers found before the follow-up commit.


plugins/nemo-agents/examples/agent-improver.example.yml
```diff
@@ -20,11 +20,11 @@
 # task.toml (Harbor) or workflow.yml (NAT) plus instruction.md and
 # tests/test_outputs.py. See tests/agentic-use/ in the NeMo Platform repo for a real
 # example.
-evals: ./agent-evals
+evals: /path/to/my-agent/agent-evals
```
mckornfield, line 28: do relative paths not work? I guess this change is to make it in the container, is there a better path for the containers?

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
```diff
@@ -92,7 +95,7 @@ cd /path/to/my-agent
 export ANTHROPIC_API_KEY='<key>'
 export ANTHROPIC_BASE_URL='https://inference-api.nvidia.com'
 
-nemo agents optimize-skills run --spec-file .agent-improver.yml
+nemo agents optimize-skills --spec-file .agent-improver.yml
```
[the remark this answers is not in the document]
mckornfield: I think it's fine for the CLI, just internal paths for the job that are bad

plugins/nemo-agents/src/nemo_agents_plugin/jobs/analyze_batch.py
```diff
@@ -155,7 +152,7 @@ def run(self, config: dict, *, ctx: JobContext | None = None) -> dict:
             ga = asyncio.run(generate_gap_analysis(batch=batch, parser=parser, baselines=baselines))
 
         if cfg.format == "json":
-            return _serialize(ga)  # type: ignore[return-value]
+            return cast(dict[str, Any], _serialize(ga))
```
mckornfield: is there a better way to fix these type issues, here and the ignores above?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/optimize_skills.py
```diff
@@ -76,10 +72,11 @@ class OptimizeSkillsJob(NemoJob):
     name: ClassVar[str] = "optimize-skills"
     description: ClassVar[str] = "Optimize an agent's skills against eval failures via a coding agent (Claude)."
     container: ClassVar[str] = "cpu-tasks"
+    generate_legacy_verbs: ClassVar[bool] = False
     spec_schema: ClassVar[type[BaseModel]] = OptimizeSkillsConfig
 
     @classmethod
-    async def compile(  # type: ignore[override]
+    async def compile(  # type: ignore[override]  # ty: ignore[invalid-method-override]
```
mckornfield: feel like these type casting things are papering over something

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -172,9 +174,53 @@ def update_job_cli(self, job_cls: type[NemoJob], group: typer.Typer) -> None:
...
+            _replace_flat_job_command_with_group(
+                group,
+                job_cls=job_cls,
+                register_subcommands=register_prepare_fileset_command,
+            )
+            return
         register_prepare_fileset_command(group)
 
 
+def _replace_flat_job_command_with_group(
```
mckornfield, line 299: did we need this shim for data-designer and others?

plugins/nemo-optimization/src/nemo_optimization/schemas/optimize.py
```diff
@@ -64,6 +65,32 @@ def _validate_config_location(self) -> "OptimizeSpec":
...
+        "every asset it references (Agent under Test package, dataset, eval.fabric.base_dir tree, "
+        "hooks and MCP configs).  Stage one with `nemo agents optimize prepare-fileset`. Required "
+        "for remote submissions, where the job has no access to the client's filesystem. "
+        "Programmatic local runs may omit it and use an absolute host path for optimize_config.",
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _allow_local_missing_fileset(cls, data: Any, info: ValidationInfo) -> Any:
+        if isinstance(data, dict) and _is_local(info) and "optimize_config_fileset" not in data:
```
mckornfield, line 91: is this "local" thing something that still works?

plugins/nemo-agents/examples/agent-improver.example.yml
mckornfield: [bot] The example template already uses absolute placeholder paths. I updated the checked-in NeMo `.agent-improver.yml` and docs to stop implying relative paths can be submitted as-is: direct platform submissions need absolute host paths for `evals` and `agent`, while bundle-internal optimize paths remain fileset-relative.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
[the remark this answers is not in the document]
mckornfield: [bot] Agreed on separating the cases. I left CLI bundle paths relative where they are meant to be interpreted inside the staged fileset, and clarified/fixed the optimize-skills job docs/config for host paths that the subprocess job reads directly.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/analyze_batch.py
mckornfield: [bot] Reworked this instead of carrying the ignores. `NemoJob.to_spec` and `compile` now accept `Any` at the abstract lifecycle boundary, matching the existing loose `run` contract, so concrete jobs can type their Pydantic models without override ignores. JSON returns now go through `_serialize_dict()` rather than per-call casts.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/optimize_skills.py
mckornfield: [bot] Agreed. I moved the mapping assertion into `_serialize_dict()` and removed the local casts from `analyze_batch` and `optimize_skills`; the helper now validates that serialized dataclass-like results are top-level mappings before returning `dict[str, Any]`.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mckornfield: [bot] The direct-job path already works for data-designer and the other non-legacy jobs. This shim is scoped to `agents optimize` because it is the only direct submit command that also needs a nested `prepare-fileset` subcommand; the wrapper preserves the bare optimize submit callback and adds only that subcommand group.

plugins/nemo-optimization/src/nemo_optimization/schemas/optimize.py
mckornfield: [bot] Yes, the programmatic local path still works through `NemoJobScheduler.run_local()`. I kept the local validation context, changed `OptimizeJob.to_spec()` to return the canonical `OptimizeSpec`, and added coverage for a non-default local workspace with no workspace key in the input spec.

plugins/nemo-optimization/src/nemo_optimization/schemas/optimize.py
```diff
@@ -18,15 +18,15 @@ class OptimizeSpec(BaseModel):
...
+        "(programmatic local runs only) it is an absolute path on the host running the job.",
     )
     optimize_config_fileset: FilesetRef | None = Field(
         default=None,
         description="Fileset holding the optimization bundle: the config named by optimize_config plus "
         "every asset it references (Agent under Test package, dataset, eval.fabric.base_dir tree, "
-        "hooks and MCP configs).  Stage one with `nemo agents optimize prepare-fileset`.  Required "
-        "for `submit`, where the job has no access to the client's filesystem.",
+        "hooks and MCP configs).  Stage one with `nemo agents optimize prepare-fileset`. Required "
+        "for remote submissions, where the job has no access to the client's filesystem.",
```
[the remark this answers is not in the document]
mckornfield: [bot] Addressed by keeping `OptimizeSubmitSpec` as the request schema so remote submissions require `optimize_config_fileset`, while `OptimizeSpec` remains the canonical/local schema. Regenerated `plugins/nemo-agents/openapi/openapi.yaml` now has `OptimizeJobRequest.spec` pointing at `OptimizeSubmitSpec` and stored `OptimizeJob.spec` pointing at `OptimizeSpec`.

plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
```diff
@@ -81,14 +81,29 @@ class OptimizeJob(NemoJob):
...
+        cls,
+        input_spec: OptimizeSubmitSpec,
+        *,
+        workspace: str,
+        entity_client: object,
+        async_sdk: AsyncNeMoPlatform,
+        is_local: bool,
+    ) -> OptimizeSubmitSpec:
+        del workspace, entity_client, async_sdk
+        return OptimizeSubmitSpec.model_validate(input_spec.model_dump(mode="json"), context={"is_local": is_local})
```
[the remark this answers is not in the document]
mckornfield: [bot] Addressed: `OptimizeJob.to_spec()` now stamps the URL or scheduler workspace into the canonical payload before validation, and the new scheduler test covers `workspace="research"` with no input workspace field.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/README.md
```diff
@@ -14,7 +14,7 @@ plugin commands; the canonical NeMo self-improvement config lives at
 `.agent-improver.yml` in the repo root and is invoked as:
 
 ```bash
-nemo agents optimize-skills run --spec-file .agent-improver.yml
+nemo agents optimize-skills --spec-file .agent-improver.yml
```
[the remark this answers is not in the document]
mckornfield: [bot] Addressed: added a prerequisites section before the CLI surface with agents CLI install, platform auth/workspace setup, absolute host path requirement, and `anthropic-api-key` secret creation.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
```diff
@@ -70,7 +70,7 @@ data is also written to `<batch_dir>/<eval-name>__trials.json`.
 ### 3. Inspect the results
 
 ```bash
-nemo agents analyze run --spec '{"batch": "./runs/batch-<timestamp>", "format": "md"}'
+nemo agents analyze --spec '{"batch": "./runs/batch-<timestamp>", "format": "md"}'
```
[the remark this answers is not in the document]
mckornfield: [bot] Addressed and clarified: batch analyze examples now use `$(pwd)` and latest batch directories, the example template uses absolute placeholder paths, and the NeMo self-improve command overrides the checked-in relative `evals` and `agent` fields with absolute paths before direct platform submission.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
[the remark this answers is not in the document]
mckornfield: [bot] Addressed and clarified: direct job submissions need absolute host paths for filesystem inputs, while bundle-internal optimize paths remain relative to the staged fileset. The config comments and docs now make that split explicit.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -278,9 +284,53 @@ def update_job_cli(self, job_cls: type[NemoJob], group: typer.Typer) -> None:
         except ImportError:
             logger.warning("nemo-optimization unavailable; skipping optimize prepare-fileset", exc_info=True)
             return
+
+        if not job_cls.generate_legacy_verbs:
```
tylersbray, line 289: **Severity:** P1 (blocking)
**Found by:** Auto

`update_job_cli` no longer registers `prepare-fileset` unless a flat `optimize` command is already present on the group it is handed, which breaks the existing hook test and that contract.

With `generate_legacy_verbs = False`, `update_job_cli` delegates to `_replace_flat_job_command_with_group`, which searches `app.registered_commands` for a command named `optimize`; when none is found it logs `Cannot extend generated job command %r because it was not registered` and returns **without** calling `register_prepare_fileset_command`.

`plugins/nemo-agents/tests/unit/test_optimize_prepare_fileset_cli.py:173-177` (untouched by this PR) does:

```python
optimize_group = typer.Typer(name="optimize")
AgentsCLI().update_job_cli(OptimizeJob, optimize_group)
assert [command.name for command in optimize_group.registered_commands] == ["prepare-fileset"]
```

That now yields `[]`. Production `add_job_commands` still works because it registers the flat command first, and the targeted pytest list in the PR description does not include this file, so the failure is not covered by the reported validation.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/cli.py` around lines 284-308, make `prepare-fileset` registration reach the group even when no flat `optimize` command was pre-registered, or update the stale expectation in `plugins/nemo-agents/tests/unit/test_optimize_prepare_fileset_cli.py:173-183` so the hook is exercised through `nemo_platform_plugin.commands.add_job_commands`. Preserve both current behaviors: only `optimize` gains the extra subcommand, and `nemo agents optimize` stays a direct remote-submit command. Run `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_optimize_prepare_fileset_cli.py plugins/nemo-agents/tests/unit/test_cli.py -v`.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
```diff
@@ -59,8 +59,9 @@ documented inline.
 cd /path/to/my-agent
 export ANTHROPIC_API_KEY='<key>'
 export ANTHROPIC_BASE_URL='https://inference-api.nvidia.com'
+printf '%s' "$ANTHROPIC_API_KEY" | nemo secrets create anthropic-api-key --from-file -
 
-nemo agents evaluate-suite run --spec-file .agent-improver.yml
+nemo agents evaluate-suite --spec-file .agent-improver.yml
```
tylersbray: **Severity:** P1 (blocking)
**Found by:** GPT-Sol

This quick-start command cannot validate: the submit schema requires `output`, and the referenced config does not set it.

`plugins/nemo-agents/src/nemo_agents_plugin/jobs/evaluate_suite.py:87-100` defines `EvaluateSuiteSubmitConfig` as the job's `input_spec_schema` and narrows both `agent` and `output` to required `str`; `compile` then re-checks with `_require_absolute(spec.output, "output")` and `_require_absolute(spec.agent, "agent")`.

`.agent-improver.yml` at this SHA has `evals`, `agent`, `skills_path`, and the runner/loop keys, but no `output` — and `evals`/`agent` are relative (`tests/agentic-use`, `.`). So `nemo agents evaluate-suite --spec-file .agent-improver.yml` returns 422 rather than producing the `report.md` / `report.csv` / `baselines.json` artifacts described immediately below on lines 67-69.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md` around line 64, make the documented `evaluate-suite` invocation satisfy `EvaluateSuiteSubmitConfig`: supply absolute `output`, `agent`, and `evals` values, either as explicit flags on the command or as fields in the config it references. Preserve the local in-process fallback behavior of the canonical `EvaluateSuiteConfig` (where both stay optional). Add or update a focused test asserting the documented spec shape validates against `EvaluateSuiteSubmitConfig`, then run `uv run --frozen pytest plugins/nemo-agents/tests/unit -k evaluate_suite -v`.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/skills/skills-optimization/SKILL.md
```diff
@@ -107,15 +107,15 @@ reading the first error in any failure is usually enough.
...
+`evaluate-suite` writes `report.md`, `report.csv`, `report.json`, and
 `baselines.json` to the batch directory. Per-eval trial data lands at
 `<batch_dir>/<eval-name>__trials.json` when `repeats > 1`.
 
 ## Common command shapes
 
 ```bash
 # Run a single eval to debug
-nemo agents evaluate-suite run --spec '{
+nemo agents evaluate-suite --spec '{
```
tylersbray: **Severity:** P1 (blocking)
**Found by:** Opus, Auto

The skill's command examples still pass relative host paths and an un-overridden `--spec-file`, all of which the direct submit path rejects, so an agent following this skill will emit failing commands.

At this SHA the examples are:

- `nemo agents evaluate-suite --spec '{ "evals": "tests/agentic-use", "agent": ".", ... }'` (lines 118-123)
- `nemo agents analyze --spec '{ "batch": "./runs/batch-2026-04-30__09-10-42", ... }'` (lines 130-133)
- `nemo agents optimize-skills --spec-file .agent-improver.yml` with no path overrides (lines 43, 127, 137, 140)

`evaluate_suite.py:159-171`, `analyze_batch.py:79`, and `optimize_skills.py:108-111` all call `_require_absolute` on those fields, so each of these 422s. This also contradicts the same PR: `.agent-improver.yml:22-23` and `GETTING_STARTED.md:136-140` instruct users to override with `--evals "$(pwd)/tests/agentic-use" --agent "$(pwd)"`.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/skills/skills-optimization/SKILL.md` around lines 32-48 and 114-141, update the evaluate-suite, analyze, and optimize-skills examples to use absolute host paths and the same `--evals` / `--agent` overrides used elsewhere in this change set, so every documented command passes `_require_absolute` at submit time. Preserve the skill's structure and the set of commands it teaches. Verify each revised command against the current submit schemas in `plugins/nemo-agents/src/nemo_agents_plugin/jobs/`.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
tylersbray: **Severity:** P1 (blocking)
**Found by:** Opus, GPT-Sol

The guide has users create the `anthropic-api-key` secret but never requests its injection for `optimize-skills` (or `evaluate-suite`), so the dispatched subprocess runs without a credential.

Line 62 adds `printf '%s' "$ANTHROPIC_API_KEY" | nemo secrets create anthropic-api-key --from-file -`, but this command and the one at line 64 pass no `--anthropic-api-key-secret`, and `.agent-improver.yml` sets no `anthropic_api_key_secret`.

`optimize_skills.py:130-140` appends `ANTHROPIC_API_KEY` to the step environment **only** when `spec.anthropic_api_key_secret` is set, and the field description states the value is injected "When set on a platform submission". The remote loop then runs `check_anthropic_api` / the Claude analyzer preflight with no key. `evaluate_suite.py:76-84` has the same conditional. The PR gets this right for `analyze` (line 75 adds `"anthropic_api_key_secret": "anthropic-api-key"`), just not for the two commands that drive the eval and loop passes.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md` around lines 62-64 and 95-98, make the documented `optimize-skills` and `evaluate-suite` submissions request the secret they just created, either via `--anthropic-api-key-secret anthropic-api-key` on the command or an `anthropic_api_key_secret` field in the referenced config (and mirror it in `.agent-improver.yml` / `examples/agent-improver.example.yml` if that is the intended source). Preserve the documented local in-process behavior of reading `ANTHROPIC_API_KEY` from the calling shell. Keep the `analyze` example consistent with whichever form you choose.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/jobs/optimize_skills.py
```diff
@@ -62,7 +58,7 @@ class OptimizeSkillsConfig(BaseModel):
     anthropic_api_key_secret: str | None = Field(
         default=None,
         description=(
-            "Name of a platform Secret holding the Anthropic API key.  When set on a submit, "
+            "Name of a platform Secret holding the Anthropic API key.  When set on a platform submission, "
             "the value is injected as ``ANTHROPIC_API_KEY`` into the dispatched subprocess so "
```
tylersbray, line 62: **Severity:** P1 (blocking)
**Found by:** GPT-Sol

Remote submissions receive the Anthropic API key from a platform Secret but not the `ANTHROPIC_BASE_URL` this change set documents, so the configured NVIDIA proxy endpoint is silently dropped.

`compile` (lines 127-140) appends only `ANTHROPIC_API_KEY` to the step environment. The consumer, `plugins/nemo-agents/src/nemo_agents_plugin/improvement/analysis/llm.py:286`, does `base_url = os.environ.get("ANTHROPIC_BASE_URL")` and otherwise falls back to Anthropic's default endpoint.

Meanwhile the docs updated in this PR export `ANTHROPIC_BASE_URL='https://inference-api.nvidia.com'` (`GETTING_STARTED.md:61,96`; `skills/skills-optimization/SKILL.md:42`), and the subprocess backend's sanitized environment does not inherit the platform process env. `analyze_batch.py:94-98` has the same shape. Since direct submit is now the only CLI path for these jobs, the local-shell escape hatch that used to cover this is gone.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/jobs/optimize_skills.py` around lines 127-140 and `plugins/nemo-agents/src/nemo_agents_plugin/jobs/analyze_batch.py` around lines 94-98, propagate the Anthropic base URL into the dispatched step environment alongside the API key (an explicit spec field is preferable to reading submitter env implicitly), so `improvement/analysis/llm.py` sees the configured endpoint. Preserve the existing secret-based injection of `ANTHROPIC_API_KEY` and never log the credential. Add a unit test asserting the compiled `PlatformJobSpec` step environment contains the base-URL variable, then run `uv run --frozen pytest plugins/nemo-agents/tests/unit -k "optimize_skills or analyze" -v`.

</details>

docs/cli/reference.mdx
```diff
@@ -5830,13 +5830,13 @@ nemo agents [OPTIONS] [COMMAND] [ARGS]...
 
 *Jobs:*
 
-* `analyze`: Analyze a batch of eval-suite results (clusters,...
-* `evaluate`: Evaluate an agent workflow against a dataset as a...
-* `evaluate-suite`: Run a directory of containerized eval tasks (Harbor or...
-* `execute`: Execute an agent to completion as a scheduled platform job.
+* `analyze`: Submit to a cluster.
```
tylersbray: **Severity:** P2 (non-blocking)
**Found by:** Opus, Auto

Collapsing the job groups into flat commands replaces every job's description in `nemo agents --help` with the generic submit help string, losing top-level discoverability.

This regenerated panel now lists `analyze`, `evaluate`, `evaluate-suite`, `execute`, `optimize-skills`, and `package-agent` all as "Submit to a cluster.", where previously each showed its own description (for example "Analyze a batch of eval-suite results..." and "Evaluate an agent workflow against a dataset...").

`optimize` is the one job that keeps its real description — precisely because `_replace_flat_job_command_with_group` re-wraps it in a Typer group with `help=job_cls.description` (`plugins/nemo-agents/src/nemo_agents_plugin/cli.py:313-317`). So the fix pattern already exists in this PR; it is just not applied to the flat commands.

<details>

Verify this finding against the current code and only fix it if it is still present.

Make the flat direct job commands use `job_cls.description` as their command help instead of the hardcoded submit help string, so `nemo agents --help` describes each job again. Prefer changing the generated-command path in `nemo_platform_plugin` (where `help_text = "Submit to a cluster."` is set) so all jobs benefit rather than special-casing the agents plugin. Preserve the existing per-command epilog and submission help text. Regenerate the CLI reference with `make generate-cli-reference-docs` and confirm `docs/cli/reference.mdx` shows distinct descriptions.

</details>

plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
```diff
@@ -104,8 +121,8 @@ async def compile(  # ty: ignore[invalid-method-override]
             PERSISTENT_JOB_STORAGE_PATH_ENVVAR,
         )
 
-        # ``compile`` is the submit path only — ``NemoJobScheduler.run_local`` goes straight to
-        # ``run`` — so requiring the fileset here is exactly the "submit is remote-safe" rule.
+        # ``compile`` is the remote submission path only — ``NemoJobScheduler.run_local`` goes
+        # straight to ``run`` — so requiring the fileset here keeps platform execution remote-safe.
         if spec.optimize_config_fileset is None:
```
tylersbray, line 118: **Severity:** P2 (non-blocking)
**Found by:** Opus

Requiring the fileset on `OptimizeSubmitSpec` makes this actionable error unreachable, so users now get a generic required-field message with no `prepare-fileset` guidance.

`schemas/optimize.py:87-91` raises `optimize_config_fileset is required for remote submissions.` during input-schema validation, which runs before `to_spec` / `compile`. The local path (`run_local`) goes straight to `run` and never calls `compile` either. So the multi-sentence `_FILESET_REQUIRED` text at lines 68-74 — the one that names `nemo agents optimize prepare-fileset --source <dir> --optimize-config <file> --fileset <name>` — is dead on both paths.

`plugins/nemo-optimization/tests/test_optimize_job.py::test_compile_requires_a_staged_fileset` still passes because it calls `compile` directly, which masks the regression. The hermes README troubleshooting row updated in this PR also points users at guidance they will no longer see.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py` around lines 68-74 and 126-127, and `plugins/nemo-optimization/src/nemo_optimization/schemas/optimize.py` around lines 87-91, surface the actionable `prepare-fileset` guidance on the path users actually hit — the remote input-schema validation — rather than only in `compile`. Preserve the local behavior where `context={"is_local": True}` permits omitting the fileset with an absolute host path. Update the test so the user-facing message is asserted on the submit-schema path, then run `uv run --frozen pytest plugins/nemo-optimization/tests/test_optimize_job.py -v`.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -278,9 +284,53 @@ def update_job_cli(self, job_cls: type[NemoJob], group: typer.Typer) -> None:
...
+    """Keep ``nemo agents <job>`` as remote submit while allowing job-specific subcommands."""
+    command_info = next((command for command in app.registered_commands if command.name == job_cls.name), None)
+    if command_info is None or command_info.callback is None:
+        logger.warning("Cannot extend generated job command %r because it was not registered", job_cls.name)
+        return
+
+    original = command_info.callback
+    app.registered_commands = [command for command in app.registered_commands if command is not command_info]
+
+    job_group = typer.Typer(
```
tylersbray, line 317: **Severity:** P2 (non-blocking)
**Found by:** Opus

The re-wrapped `optimize` group drops the generated epilog, so `nemo agents optimize --help` loses the spec-precedence guidance that every other job command still shows.

The wrapper reads `command_info.callback`, copies `__doc__` and `__signature__`, but never reads `command_info.epilog`, and constructs `typer.Typer(name=..., help=..., no_args_is_help=False)` with no `epilog=`. The generated submit command sets one via `build_epilog(...)`.

Visible in the regenerated reference: `analyze`, `evaluate`, `evaluate-suite`, `execute`, `optimize-skills`, and `package-agent` all retain "Job Spec flags are generated from the ... Pydantic schema. Precedence: --spec-file (base) -> --spec JSON (overlay) -> per-flag values (top)." plus the "Some spec fields cannot be represented as CLI flags" note, while the `optimize` section has neither.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/src/nemo_agents_plugin/cli.py` around lines 298-336, carry the generated command's `epilog` onto the replacement Typer group (or its callback) so `nemo agents optimize --help` keeps the spec-precedence and unrepresentable-fields notes shown by the other job commands. Preserve the existing early-return-on-subcommand behavior and the copied `__signature__`. Regenerate the CLI reference and confirm the `optimize` section renders the epilog like its siblings.

</details>

plugins/nemo-agents/tests/unit/test_cli.py
```diff
@@ -228,6 +227,65 @@ def _submit_remote(_self, job_cls, spec, **kwargs):
...
+
+    assert result.exit_code == 0, result.output
+    assert "--source" in result.output
+    assert "--fileset" in result.output
+
+    top_level_result = CliRunner().invoke(app, ["prepare-fileset", "--help"])
+    assert top_level_result.exit_code != 0
+
+
+def test_agent_jobs_do_not_register_legacy_run_submit_verbs() -> None:
```
tylersbray, line 254: **Severity:** P2 (non-blocking)
**Found by:** Opus

This test guards its real assertions behind an `isinstance(..., click.Group)` check, so for six of the seven jobs it only restates a class attribute.

```python
for job_cls in jobs.values():
    assert job_cls.generate_legacy_verbs is False
    job_command = command.commands[job_cls.name]
    if isinstance(job_command, click.Group):
        assert "run" not in job_command.commands
        assert "submit" not in job_command.commands
```

After this PR only `optimize` is a `click.Group`; the other six are flat `click.Command`s, so the loop body degenerates to `assert job_cls.generate_legacy_verbs is False` — a restatement of the source line it is meant to protect. The nearby `assert legacy_result.exit_code != 0` is also loose: it passes on any non-zero exit, including an unrelated crash.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-agents/tests/unit/test_cli.py` around lines 253-291, make the test pin the intended CLI surface rather than the class attribute: assert the six flat jobs are not Click groups (or that invoking them with `run` / `submit` fails as unexpected arguments), and assert `optimize`'s subcommands are exactly `{"prepare-fileset"}`. Tighten the legacy-verb exit assertion so it distinguishes a usage error from an unrelated failure. Then run `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_cli.py -v`.

</details>

packages/nemo_platform_plugin/src/nemo_platform_plugin/job.py
```diff
@@ -237,13 +232,13 @@ def run(self, *args: Any, **kwargs: Any) -> dict:
     @classmethod
     async def to_spec(
         cls,
-        input_spec: "BaseModel",
+        input_spec: Any,
```
tylersbray: **Severity:** P3 (non-blocking)
**Found by:** Opus

Widening the base `to_spec` / `compile` parameter types to `Any` to delete override suppressions disables type checking of those arguments for every plugin in the repo.

`input_spec: "BaseModel"` becomes `input_spec: Any` here, and `spec: "BaseModel"` becomes `spec: Any` at line 281, paired with the removal of `# type: ignore[override]` from six agents job overrides and `# ty: ignore[invalid-method-override]` in nemo-optimization.

That trades seven contained, explicit suppressions for an `Any` hole in a shared framework base class: it propagates to all current and future job implementations and to the `to_spec` / `compile` adapters in `jobs/routes.py`, where a genuinely wrong spec type would now go unflagged. Narrowing a parameter in an override is a real Liskov violation, which is what those ignore comments were documenting.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `packages/nemo_platform_plugin/src/nemo_platform_plugin/job.py` at lines 235 and 281, restore a precise annotation for the `input_spec` and `spec` parameters (`BaseModel`, or a bounded `TypeVar` if subclass-specific overrides need to type-check cleanly) instead of `Any`, reinstating narrowly scoped ignore comments at the override sites if still required. Preserve the current runtime behavior and the identity-function default for `to_spec`. Verify with `uv run --frozen ty check packages/nemo_platform_plugin plugins/nemo-agents plugins/nemo-optimization`.

</details>

.agent-improver.yml
```diff
@@ -19,23 +19,28 @@
...
 # ---------------------------------------------------------------------------
 
 # The Harbor agentic-use eval suite — discovered as task.toml-marked dirs.
+# Override with --evals "$(pwd)/tests/agentic-use" when submitting this checked-in
+# config through the direct platform command.
 evals: tests/agentic-use
 
-# Agent root: this repo. The optimize-skills strategy is allowed to write
+# Agent root: this repo. Override with --agent "$(pwd)" when submitting this
+# checked-in config through the direct platform command. The optimize-skills
```
tylersbray, line 46: **Severity:** nit (non-blocking)
**Found by:** Opus

Mechanical-edit artifact: the inserted override sentence breaks the surrounding comment mid-clause, leaving a stranded line.

```
# Agent root: this repo. Override with --agent "$(pwd)" when submitting this
# checked-in config through the direct platform command. The optimize-skills
# strategy is allowed to write
# only under skills_path; the loop's post-edit guard reverts any change
```

Related: the Prerequisites block above (lines 16-17) still says `ANTHROPIC_API_KEY` is exported "for the LLM analyzer pass", which for a platform submission is now the dispatched subprocess reading a secret.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `.agent-improver.yml` around lines 40-46, rewrap the `agent:` comment so "The optimize-skills strategy is allowed to write only under skills_path" reads as one sentence, and refresh the Prerequisites wording at lines 16-17 to match the platform-secret flow this change set documents. Comment-only change; do not alter any YAML values.

</details>

tylersbray:
**Severity:** P1 (blocking)
**Found by:** Opus

**Scope:** `docs/agents/optimization.mdx`, `docs/agents/index.mdx`, `README.md` (not in this PR's change set)

The published agents documentation and the root README still teach the removed verbs, so the canonical docs now describe commands that no longer exist.

At the pinned head SHA these files are unchanged and still contain, among others:

- `docs/agents/optimization.mdx:27,198,276,305,317,387,453,491` — `nemo agents optimize run`, `nemo agents optimize submit`, `nemo agents optimize-skills run --spec-file .agent-improver.yml`, and prose such as "`nemo agents optimize run` executes in your shell. `submit` hands the study to..."
- `docs/agents/optimization.mdx:328` — documents the skill as "Runs `nemo agents optimize run` (or `submit` for platform jobs)", which this PR changed in `skills/agents-optimize/SKILL.md`
- `docs/agents/index.mdx:56,65,219` — `nemo agents optimize run --optimize-config <path>`, `nemo agents evaluate run --eval-config <path> --agent <name>`
- `README.md:169,178` — `nemo agents evaluate run`, `nemo agents optimize run`

The PR description states "Update docs, examples, plugin skills, generated CLI reference, and OpenAPI text to use the direct job command flow", so this looks like an incomplete sweep rather than a deliberate exclusion.

<details>

Verify this finding against the current code and only fix it if it is still present.

Replace the removed `run` / `submit` / `explain` verb forms in `docs/agents/optimization.mdx` (lines 27, 198, 276, 305, 317, 328, 387, 453, 491), `docs/agents/index.mdx` (lines 56, 65, 219), and `README.md` (lines 169, 178) with the direct job commands introduced by this change (`nemo agents optimize`, `nemo agents evaluate`, `nemo agents optimize-skills`, plus `nemo agents optimize prepare-fileset` where a staged bundle is needed). Preserve each page's existing structure and intent; do not invent new workflows. Confirm no `nemo agents <job> (run|submit|explain)` occurrences remain in `docs/` or `README.md`.

</details>

tylersbray:
**Severity:** P2 (non-blocking)
**Found by:** Opus

**Scope:** `web/packages/common/src/components/EntityEmptyState/registry.ts:262` (not in this PR's change set)

Studio's empty-state card still offers a command that no longer parses. At the pinned head SHA:

```ts
cliCommand: 'nemo agents optimize-skills run --spec-file <spec>.yml',
```

After this PR `optimize-skills` is a flat command, so `run` is parsed as an unexpected argument and the copy-pasteable command shown in the UI fails.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `web/packages/common/src/components/EntityEmptyState/registry.ts` around line 262, update `cliCommand` from `nemo agents optimize-skills run --spec-file <spec>.yml` to the direct form `nemo agents optimize-skills --spec-file <spec>.yml`. Preserve the surrounding registry entry shape. Check the same file for any other `nemo agents <job> run|submit` strings.

</details>

tylersbray:
**Severity:** P3 (non-blocking)
**Found by:** Opus

**Scope:** `tests/agentic-use/README.md`, `tests/agentic-use/aut-eval-diagnostic.yml`, `tests/agentic-use/aut-optimize.yml` (not in this PR's change set)

The agentic eval fixtures still document and exercise the removed verbs:

- `tests/agentic-use/README.md:36,291` and `aut-eval-diagnostic.yml:8` — `uv run _nemo agents evaluate run --help`
- `tests/agentic-use/README.md:84` and `aut-optimize.yml:14` — `nemo agents optimize run --optimize-config ...`
- `tests/agentic-use/README.md:286` — "`agents evaluate run` and `agents optimize run` are provided by the..."
- `tests/agentic-use/README.md:98` — `uv run _nemo --base-url ... agents evaluate run --spec '{...}'`

These are the evals the `optimize-skills` loop grades itself against (and the target of the repo's `.agent-improver.yml`), so stale commands both teach the wrong surface and risk a diagnostic task failing on a `--help` invocation that no longer resolves.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `tests/agentic-use/README.md` (lines 36, 84, 98, 286, 291), `tests/agentic-use/aut-eval-diagnostic.yml` (lines 8, 15), and `tests/agentic-use/aut-optimize.yml` (line 14), replace the removed `run` verb with the direct job commands from this change (`agents evaluate`, `agents optimize`, staging via `agents optimize prepare-fileset` where a fileset is now required). Preserve the local-diagnostic guidance about passing `--base-url`. Confirm any `--help` invocations in the eval instructions still resolve against the new CLI surface.

</details>

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mckornfield: [bot] Addressed in 1044b26af. `_replace_flat_job_command_with_group` now registers `prepare-fileset` directly when the hook receives the `optimize` Typer group without a generated flat command, while the root app still avoids a stray top-level command. Covered by `test_optimize_prepare_fileset_cli.py` and `test_cli.py`.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
mckornfield: [bot] Addressed in 1044b26af. The quick-start `evaluate-suite` invocation now supplies absolute `--evals`, `--agent`, and `--output` values plus the platform secret/base-url flags, and `test_evaluate_suite_documented_submit_shape_validates` pins that submit shape. `uv run --frozen pytest plugins/nemo-agents/tests/unit -k evaluate_suite -v` passed.

plugins/nemo-agents/src/nemo_agents_plugin/skills/skills-optimization/SKILL.md
mckornfield: [bot] Addressed in 1044b26af. The skill examples now use absolute path overrides for `evaluate-suite`, `analyze`, and `optimize-skills`, include explicit output paths for submitted eval batches, and pass the Anthropic secret/base-url fields needed by the direct submit schemas.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/optimize_skills.py
mckornfield: [bot] Addressed in 1044b26af. Added an explicit `anthropic_base_url` spec field and propagate it as `ANTHROPIC_BASE_URL` in the compiled subprocess environment for `analyze`, `evaluate-suite`, and `optimize-skills`, alongside the existing secret-backed API key. Unit coverage asserts the compiled env includes the URL, and `uv run --frozen pytest plugins/nemo-agents/tests/unit -k "optimize_skills or analyze" -v` passed.

plugins/nemo-agents/src/nemo_agents_plugin/improvement/GETTING_STARTED.md
mckornfield: [bot] Addressed in 1044b26af. The guide, root config comments, template, and improvement README now request `--anthropic-api-key-secret anthropic-api-key` for submitted `optimize-skills` and `evaluate-suite` flows, and the `analyze` example stays consistent with the same secret field.

plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
mckornfield: [bot] Addressed in 1044b26af. The actionable `prepare-fileset` guidance now lives in the `OptimizeSubmitSpec` remote validation path and is reused by the compile guard. Local scheduler validation with `context={"is_local": True}` still permits omitting the fileset for absolute host-path configs. `plugins/nemo-optimization/tests/test_optimize_job.py` now asserts the submit-schema message, and the focused test file passed.

docs/cli/reference.mdx
mckornfield: [bot] Addressed in 1044b26af. `_add_submit_command` now uses `job_cls.description` for flat direct job commands while preserving the generic `Submit to a cluster.` help for legacy submit verbs. Regenerated `docs/cli/reference.mdx`; the `agents` job list now shows distinct descriptions.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mckornfield: [bot] Addressed in 1044b26af. The replacement `optimize` Typer group now carries the generated command epilog, preserving the spec-precedence and unavailable-field notes. Regenerated `docs/cli/reference.mdx` and added a CLI assertion that `nemo agents optimize --help` includes `Precedence:`.

.agent-improver.yml
mckornfield: [bot] Addressed in 1044b26af. Rewrapped the `agent:` comment so the `skills_path` sentence is intact, refreshed the prerequisites wording to describe the platform-secret flow, and did not change any YAML values.

plugins/nemo-agents/tests/unit/test_cli.py
mckornfield: [bot] Addressed in 1044b26af. The CLI test now asserts the six flat jobs are not Click groups, legacy `run` and `submit` are usage errors, `optimize` exposes exactly `prepare-fileset`, and the `optimize submit` check expects the specific no-such-command failure. `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_cli.py -v` passed as part of the focused bundle.

packages/nemo_platform_plugin/src/nemo_platform_plugin/job.py
mckornfield: [bot] Addressed in 1044b26af. Restored `BaseModel` annotations on the shared `NemoJob.to_spec` and `NemoJob.compile` parameters, added scoped `ty` suppressions only where concrete job overrides intentionally narrow schema types, and cast the route/scheduler adapters at the dynamic boundaries. Focused `uv run --frozen ty check <touched source and route-test files>` passed. The broad package command still has 139 existing diagnostics outside the files changed here.

mckornfield:
Addressed the follow-up items from the open top-level comments in `d4bf24482` and refreshed the branch with current `main` through signed merge commits, latest head `bb1cfa9a7`.

Covered:
- Updated published docs and root README to use direct agent job commands, including `optimize prepare-fileset` where staging is required.
- Updated Studio empty-state copy from `nemo agents optimize-skills run` to the flat `nemo agents optimize-skills` form.
- Updated agentic-use README and fixtures to stop exercising removed `evaluate run` / `optimize run` verbs.
- Fixed the anonymizer flat-job test expectation to assert the generated job help content now shown by `anonymizer run --help`.

Validation is recorded in the PR body. The relevant CI gates are now green on the current head: `Check uv lock`, `Lint all`, `Python unit tests`, Web checks, DCO, CodeQL, and the wheel build/test matrix.

tylersbray (approved):
Thanks dude.

PR #1700, opened 2026-09-01
feat(evaluator): Support running custom Gym evals and native-v1 environment format

JashG:
## Summary
Adds end-to-end support for running custom Gym environments in agent evaluations. Environment packages can provide their own agent and resources-server components, and these evaluations now run correctly through both subprocess and Kubernetes OpenSandbox execution.

### Environment Packages
- Add support for `native-v1` packages alongside `wheels-v1`.
- Discover agent and resources-server components from package configuration. Reject customer-provided models and conflicting component names.
- Use package components when available and built-in components as fallbacks.
- Allow `agent_config` to be omitted when the package provides the agent.

### Agent Evaluation
- Stage custom environment FileSets before running evaluations.
- Support container and subprocess execution profiles.
- Keep each rollout result linked to the input task that produced it.

### Runtime Image
- Add a dedicated Gym orchestration dependency group specifically so Evaluator can invoke OpenSandbox when invoked in the gym-task image.
- Include the Evaluator plugin, `sandboxed-gym`, and OpenSandbox SDK in `nmp-gym-tasks`.
- Add smoke checks for required runtime packages.

### Fixes
- Start the OpenSandbox Gym host instead of its default keepalive process.
- Propagate the platform protocol so in-cluster readiness checks use HTTP.
- Run OpenSandbox lifecycle operations on one event loop to prevent cleanup failures.
- Include the health-check URL in readiness timeout errors.
- Correct subprocess Gym commands to use the host Python runtime.

### Validation
- Add and update tests for package validation and composition.
- Add tests for agent-evaluation compilation and subprocess execution.
- Add tests for OpenSandbox entrypoints, protocol propagation, and lifecycle cleanup.
- I verified a custom-environment evaluation end-to-end on Kubernetes OpenSandbox.

## Summary by CodeRabbit

* **New Features**
  * Added support for native-v1 Gym environments alongside wheels-v1 packages.
  * Environment packages can provide agent and resource-server components with automatic discovery and fallback configuration.
  * Agent configuration is optional when supplied by the environment package.
  * Added configurable OpenSandbox protocols, subprocess execution, and enhanced execution-profile handling.

* **Bug Fixes**
  * Improved validation for duplicate, conflicting, unsupported, or invalid component declarations.
  * Improved sandbox lifecycle reliability and readiness error details.
  * Expanded validation and support for custom environments and storage configurations.


docker/Dockerfile.nmp-gym-tasks
```diff
@@ -13,10 +13,20 @@ ARG NMP_PYTHON_BASE=nmp-python-base
 FROM ${NMP_CPU_TASKS_BASE} AS cpu-tasks
 
 FROM ${NMP_PYTHON_BASE} AS gym-builder
-COPY --from=cpu-tasks /app/.venv /app/.venv
+COPY --from=nmp-workspace . .
 ENV VIRTUAL_ENV=/app/.venv \
     PATH="/app/.venv/bin:$PATH"
 
+# Create a fresh /app/.venv containing the CPU task dependencies, Evaluator
```
JashG, line 20: Bug I ran into: Evaluator couldn't invoke OpenSandbox due to the dependency missing from the virtualenv.

I don't think we need this image anymore, though - agent-evaluate jobs should be able to use the default `cpu-tasks` image, since evaluator no longer talks to Gym directly. I don't think this should block this PR though - can handle this in a follow-up PR.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -343,23 +344,34 @@ async def compile(
...
-                    profile=profile or "default",
-                )
-                require_fileset_sandbox_storage_identity(
-                    canonical_spec.target,
-                    evaluator_config,
-                    execution_profile=execution_profile,
-                )
-            except SandboxUnavailableError as exc:
-                raise PlatformJobCompilationError(str(exc)) from exc
+        execution_profile: BaseExecutionProfile | None = None
```
JashG, line 347: Context: Jobs using a subprocess execution profile need a host command rather than a container specification. Otherwise staging/evaluation could run with the wrong command or interpreter.

packages/sandboxed_gym/src/sandboxed_gym/host/opensandbox.py
```diff
@@ -131,14 +132,18 @@ def _to_sandbox_spec(self, spec: GymHostSpec):
                     resource_kwargs["memory_mib"] = int(memory_s)
             resources = SandboxResources.from_mapping(resource_kwargs)
 
+        # OpenSandbox treats an omitted entrypoint as a request for its persistent-sandbox
+        # keepalive (`tail -f /dev/null`), rather than preserving the image CMD. Start the standard
+        # Gym host explicitly while still allowing runtime images to configure their own command.
+        entrypoint = spec.entrypoint or _DEFAULT_RUNTIME_ENTRYPOINT
```
JashG, line 138: Context: If OpenSandbox entrypoint is `None`, it starts a keepalive command instead of using the image's CMD. So, I'm explicitly passing the entrypoint command via `_DEFAULT_RUNTIME_ENTRYPOINT`, which I got from the Gym Host's CMD.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_compiler.py
```diff
@@ -111,29 +116,60 @@ def _environment(spec: AgentEvalSpec) -> list[EnvironmentVariable]:
     environment = build_task_environment(_secret_refs(spec))
     if not isinstance(spec.target, GymRunnerTarget):
         return environment
-    plan = resolve_sandbox_plan(config, spec.target)
+    plan = resolve_sandbox_plan(
+        config,
+        spec.target,
+        sandbox_server_protocol=platform_config.sandbox_server_protocol,
```
JashG, line 122: Bug I ran into: The shared `sandbox_server_protocol` setting was not propagated into the Evaluator sandbox plan. The readiness probe consequently timed out despite a running host because it used HTTPS against an HTTP endpoint.

So, needed to pass `sandbox_server_protocol` here.

packages/sandboxed_gym/src/sandboxed_gym/orchestrator.py
```diff
@@ -49,6 +50,38 @@ def _run_coro_sync(coro: Coroutine[Any, Any, T]) -> T:
         return pool.submit(asyncio.run, coro).result()  # ty: ignore[invalid-return-type]
 
 
+class _SessionAsyncRunner:
```
JashG, line 53: Bug I ran into: Separate `asyncio.run()` calls created and destroyed the SDK client on different loops. Cleanup then failed with `RuntimeError: Event loop is closed, potentially leaking the sandbox pod.`

The fix that worked was for each host session to use one event loop for create, readiness, and destroy operations.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -370,19 +382,42 @@ async def compile(
...
+        """Resolve the profile that Jobs will use for this submission."""
         if async_sdk is None:
             return None
         try:
             profiles = (await client_from_platform(async_sdk, AsyncJobsClient).get_execution_profiles()).data()
-        except (NemoTransportError, NemoResponseValidationError, InternalServerError):
-            return None
-        return next((item for item in profiles if item.profile == profile), None)
+        except (NemoTransportError, NemoResponseValidationError, InternalServerError) as exc:
+            raise _profile_dependency_unavailable(profile) from exc
```
JashG, line 397: Context: Temporary Jobs-service failures returned `None`, causing FileSet/OpenSandbox storage identity validation to be silently bypassed.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -343,23 +344,34 @@ async def compile(
...
-                require_fileset_sandbox_storage_identity(
-                    canonical_spec.target,
-                    evaluator_config,
-                    execution_profile=execution_profile,
-                )
-            except SandboxUnavailableError as exc:
-                raise PlatformJobCompilationError(str(exc)) from exc
+        execution_profile: BaseExecutionProfile | None = None
+        if isinstance(canonical_spec.target, GymRunnerTarget):
+            evaluator_config = get_config() if canonical_spec.target.environment is not None else None
```
JashG, line 349: Bug I ran into: Constructing `EvaluatorConfig()` directly could miss deployment-specific configuration. We should simply read the active configuration via `get_config()`.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_compiler.py
```diff
@@ -111,29 +116,60 @@ def _environment(spec: AgentEvalSpec) -> list[EnvironmentVariable]:
...
+            profile=resolved_profile,
+            provider="subprocess",
+            command=[*AGENT_EVAL_ENTRYPOINT, *command],
+        )
+
+    return CPUExecutionProviderSpec(
+        profile=resolved_profile,
+        provider="cpu",
+        container=ContainerSpec(image=image, entrypoint=entrypoint, command=command),
+    )
```
SandyChapman, line 175: huh, I totally thought the backend selected was a global config and not a spec that gets provided.

SandyChapman (approved):

plugins/nemo-evaluator/src/nemo_evaluator/config.py
```diff
@@ -116,3 +116,4 @@ def get_config() -> EvaluatorConfig:
 
 
 config = get_config()
+platform_config = get_platform_config()
```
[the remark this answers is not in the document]
JashG: Practically, I think the concern isn't a big issue. Platform config is a service-level singleton; if it's malformed or invalid, we'd likely see a failure much earlier (probably during service initialization). Accessing the platform config in this way is a common pattern in the repo, so I'd suggest we make no change here.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/gym_sandbox.py
```diff
@@ -192,14 +209,21 @@ def resolve_sandbox_plan(config: EvaluatorConfig, target: GymRunnerTarget) -> Sa
     # `require_sandbox_available` has just established that neither is empty.
     assert config.sandbox_runtime_image is not None
     assert config.sandbox_job_storage_pvc_claim is not None
+    host_provider_options = dict(config.sandbox_host_provider_options)
+    # In-cluster OpenSandbox speaks http. If we leave this unset, the host probes the
+    # sandbox over https and the readiness wait times out.
+    if config.sandbox_host_provider == "opensandbox" and sandbox_server_protocol:
+        connection = dict(host_provider_options.get("connection") or {})
+        connection["protocol"] = sandbox_server_protocol
```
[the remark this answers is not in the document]
JashG: The OpenSandbox API key is usually supplied through the `OPEN_SANDBOX_API_KEY` environment variable, not `connection["api_key"]`. Therefore, checking this dictionary would miss the normal authenticated connection.

Also, our current Kubernetes setup intentionally connects to OpenSandbox over an internal HTTP service. Requiring HTTPS here would break that supported setup. So in this PR, this would be a breaking fix without TLS/encrypted transport support for the OpenSandbox deployment.

PR #1704, opened 2026-09-01
feat(agents): add deployment-backed sessions chat functionality

mmogallapalli:
## Summary

Adds an interactive multi-turn CLI for persisted sessions on deployed
Fabric-backed agents. Users can create, discover, resume, inspect, and close
sessions while the deployment runtime—not the CLI—retains conversation context.

This PR is stacked on `sessions-cli-1/mmogallapalli`, which extracts the reusable
chat TUI. The TUI refactor is therefore excluded from this PR's functional scope.

## Related Issue

[AIRCORE-951: Add interactive multi-turn CLI for deployed Fabric-backed agents](https://linear.app/nvidia/issue/AIRCORE-951/add-interactive-multi-turn-cli-for-deployed-fabric-backed-agents)

## Changes

- Add `nemo agents chat` with two mutually exclusive flows:
  - `--agent-deployment/-d` creates a new persisted session.
  - `--session/-s` resumes an existing session by name.
- Support `--session-name` for explicit naming and preserve API-generated names
  when it is omitted.
- Support an optional first turn through `--input/-i` while keeping chat
  interactive; reject non-interactive terminals and direct scripted callers to
  `nemo agents invoke`.
- Add `nemo agents sessions list|get|close`:
  - list sessions newest-first, optionally filtered by deployment;
  - reuse table, JSON, YAML, CSV, Markdown, and raw output formats;
  - close with confirmation unless `--yes/-y` is supplied;
  - omit deletion and user-facing pagination.
- Validate new-session deployments as Fabric-backed and routable before creating
  the session with the deployment entity ID.
- Resolve resumed sessions by name, reject closed, expired, or lost sessions, and
  find their immutable deployment binding through the existing paginated
  deployment-list API.
- Stream each turn through the deployment's `/v1/chat/completions` gateway route
  with `X-Nemo-Session-Id` and the existing authentication headers.
- Send only the current user turn; persisted Fabric runtime state owns the
  conversation instead of a replayed CLI transcript.
- Show deployment, session, status, and expiration metadata in the shared TUI.
- Print a reusable resume command after session creation and explain that resumed
  runtime context does not redisplay earlier messages.
- Treat Ctrl+C and EOF as detach operations that leave the persisted session
  active, including interruptions during the optional initial streamed turn.
- Synchronize healthy subprocess backend state to `running` so controller
  reconciliation does not falsely mark active sessions as `lost`.
- Add CLI regression coverage for the command contract, management operations,
  creation, resumption, deployment resolution, streaming, authentication,
  current-turn-only requests, terminal session errors, missing sessions, and
  detach behavior.

## Design Decisions

- **Use the deployment gateway.** Session chat continues through the deployment's
  existing OpenAI-compatible gateway route and identifies the persisted session
  with `X-Nemo-Session-Id`; no session-specific gateway is introduced.
- **Keep runtime context server-side.** The CLI sends only the current user turn.
  Fabric owns multi-turn state, so resuming does not require local transcript
  persistence or replay.
- **Resolve the immutable deployment binding through existing APIs.** Resumption
  reads `AgentSession.deployment_id` and internally walks the deployment list in
  pages of 100 until it finds the matching entity. No deployment-by-ID endpoint
  or user-facing pagination options are added.
- **Separate detach from close.** Leaving the TUI never closes the session.
  Lifecycle termination remains an explicit `nemo agents sessions close` action.
- **Keep management intentionally small.** The CLI exposes list, get, and close;
  permanent deletion remains API-only and outside this ticket.
- **Defer user documentation.** General session-chat documentation will be added
  separately rather than attaching the workflow to an unrelated example in this
  PR.

## Type of Change

- [x] Code change (feature)
- [ ] Documentation only

## Quality Gates

- [x] Existing model-chat and Agents CLI behavior remains covered
- [x] Documentation deferred to separate follow-up work

## Verification

Targeted validation:

- `.venv/bin/pytest plugins/nemo-agents/tests/unit/test_cli_sessions.py plugins/nemo-agents/tests/unit/test_cli_session_chat.py -q` — 42 passed
- `.venv/bin/pytest plugins/nemo-agents/tests/unit plugins/nemo-agents/tests/test_authz.py -q` — 1,546 passed
- Focused and full-source Ruff linting and formatting passed.
- Changed-file `ty` type checking passed.
- `git diff --check` passed.

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to chat with deployed agents and resume persisted sessions.
  * Added session management commands to list, view, and close sessions.
  * Added support for detaching from interactive chat sessions.
  * Added CLI reference documentation for agent chat and session management.

* **Bug Fixes**
  * Deployment status now correctly reports as running after a successful health check.
  * Improved validation of deployment and session information when managing sessions.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/chat_tui.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/chat_tui.py
@@ -9,8 +9,9 @@ import json
 import logging
 import re
 import sys
+from enum import Enum
 from types import TracebackType
-from typing import Any, Callable, Iterator, Literal, Protocol
+from typing import Any, Callable, Iterator, Protocol
 
 import click
 from nemo_platform._streaming import SSEDecoder
@@ -22,6 +23,18 @@ from rich.panel import Panel
 from rich.prompt import Prompt
 
 from nemo_platform_ext.cli.core.help_formatter import _get_terminal_width
+
+
+class ExitAction(Enum):
+    """Presentation text for leaving a chat TUI."""
+
+    EXIT = ("exit", "Exit the chat", "Chat session ended")
+    DETACH = ("detach", "Detach from the session", "Session detached")
+
+    def __init__(self, verb: str, help_text: str, successful_exit_text: str) -> None:
+        self.verb = verb
+        self.help_text = help_text
+        self.successful_exit_text = successful_exit_text
 
 
 class StreamingBody(Protocol):
@@ -123,7 +136,7 @@ def run_chat_tui(
     system_message: str | None = None,
     initial_message: str | None = None,
     record_assistant_message: RecordAssistantMessage | None = None,
-    exit_action: Literal["exit", "detach"] = "exit",
+    exit_action: ExitAction = ExitAction.EXIT,
 ) -> None:
     """Run the shared interactive chat UI against a caller-provided transport."""
     root_logger = logging.getLogger()
@@ -309,10 +322,9 @@ def _clear_prompt_line() -> None:
     console.file.flush()
 
 
-def _exit_gracefully(exit_action: Literal["exit", "detach"]) -> None:
+def _exit_gracefully(exit_action: ExitAction) -> None:
     console.print("\n")
-    message = "Session detached" if exit_action == "detach" else "Chat session ended"
-    console.print(Panel.fit(f"[bold]{message}[/bold]", border_style="dim", padding=(0, 2)))
+    console.print(Panel.fit(f"[bold]{exit_action.successful_exit_text}[/bold]", border_style="dim", padding=(0, 2)))
     raise click.exceptions.Exit(0)
 
 
@@ -320,7 +332,7 @@ def _handle_special_command(
     command: str,
     last_thinking: str,
     thinking_displayed: bool,
-    exit_action: Literal["exit", "detach"],
+    exit_action: ExitAction,
 ) -> bool | None:
     if command in {"/thinking", "/t"}:
         if not last_thinking:
@@ -342,12 +354,11 @@ def _handle_special_command(
         return True
 
     if command in {"/help", "/h"}:
-        exit_help = "Detach from the session" if exit_action == "detach" else "Exit the chat"
         console.print(
             Panel(
                 "[cyan]/thinking[/cyan] - Show/hide model reasoning from last response\n"
                 "[cyan]/help[/cyan] - Show this help message\n"
-                f"[cyan]Ctrl+C[/cyan] - {exit_help}",
+                f"[cyan]Ctrl+C[/cyan] - {exit_action.help_text}",
                 title="[bold]Available Commands[/bold]",
                 border_style="blue",
                 padding=(0, 1),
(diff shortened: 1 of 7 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/chat_tui.py
```diff
@@ -334,11 +342,12 @@ def _handle_special_command(command: str, last_thinking: str, thinking_displayed
         return True
 
     if command in {"/help", "/h"}:
+        exit_help = "Detach from the session" if exit_action == "detach" else "Exit the chat"
```
mikeknep: This thought is surely overkill, but whatever I'll share it anyways. If there was just _one more_ presentation text switch based on exit_action, I would advocate for some sort of polymorphic presenter approach, so that we could just call things like `exit_action.help_text`, `exit_action.successful_exit_text`, etc. But that'd be silly for just these two, right?!

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
```diff
@@ -1949,11 +2115,303 @@ def _platform_invoke(
...
+            base_url=base_url,
+            workspace=workspace,
+            deployment=deployment,
+            session=resolved_session,
+            session_id=session_id,
+            input=input,
+            timeout=timeout,
+        )
+        return
+    typer.echo("Error: Provide exactly one of --agent-deployment or --session.", err=True)
```
mikeknep: Do we need a check at the top of this function that the user hasn't supplied both these flags? We'll only reach this line if they provide **neither**.

mmogallapalli: The command calls `_validate_session_chat_options()` before reaching this helper. Its `(agent_deployment is None) == (session is None)` check rejects both “both provided” and “neither provided,” and both cases have regression coverage.

Since `_platform_session_chat()` is private and only called from that validated path, an additional check here would duplicate the command-boundary validation. The final branch is just a defensive fallback for a missing selector.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/chat_tui.py
mmogallapalli: It may be "overkill" but I think centralizing the text instead of spreading conditionals everywhere readsbetter so I went ahead and added it.

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mikeknep: Ah, nice, I didn't catch that. Makes sense. Does that also mean this typer.echo call at the end of this private function is functionally useless as well? (Not a blocker)

packages/nemo_platform_ext/src/nemo_platform_ext/cli/chat_tui.py
mikeknep: Sweet!

mikeknep (approved):

plugins/nemo-agents/src/nemo_agents_plugin/cli.py
mmogallapalli: Fair enough, since its not reachable in a practical sense I'll remove it

PR #1715, opened 2026-09-02
feat(studio): save an experiment's column layout

rrhyne:
## Summary

The evaluations table's column order lived in `localStorage`, so it followed the browser rather than the experiment it describes — which of an experiment's evaluators are worth a column, and in what order they read, is a property of the experiment, not of the machine looking at it. Column visibility was not persisted at all. Both now live on the experiment: a **Save columns** button appears at the toolbar's leading edge once the live layout differs from the saved one, and disappears again once it matches.

## Changes

- `ExperimentGroup` gains `column_layout` (`order` + `hidden`). The entity store is schema-on-read, so **no migration is required**. Update follows the existing `pareto` convention: an omitted `column_layout` preserves what is saved rather than clearing it.
- The field is **nullable** so "never saved" stays distinct from "saved with nothing hidden". The client applies its own default hidden columns only in the null case, so a deliberate "show everything" survives a reload instead of being re-hidden — and the Save button does not appear on a freshly loaded page.
- Visibility is stored as the **hidden ids** rather than a map over every column, so a column that appears later (a new evaluator, a new metadata key) shows up by default instead of being invisible because a layout saved before it existed never mentioned it.
- Studio seeds column order and visibility from the experiment and persists them through the existing `PUT /experiments/{name}`.
- `StudioDataViewToolbar` gains a `slotStart` (surfaced as `toolbarSlotStart` on `StudioDataView`), rendered before the search bar. This is additive and optional; existing callers are unaffected.
- Regenerated `openapi/` and the web SDK.

### Behavior change worth calling out

This replaces the `localStorage` column-order persistence (`nemo-studio:experiment-columns:<id>`). Anyone holding a local layout will see it reset once to the experiment's saved (initially absent) layout. Keeping both stores seemed worse than picking one.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no documented surface changes; the field is described in the OpenAPI schema it ships with.

## Verification

Targeted validation, all run after rebasing onto `origin/main`:

- `uv run pre-commit run --files <changed files>` — passed (ruff, ruff format, **ty**, copyright headers, UI lint-staged, merge-conflict check). The full `-a` run was **not** performed locally, so that box is left unchecked; CI runs the complete gate.
- `uv run --frozen pytest services/intake/tests/integration/test_experiments_crud.py -k "column_layout or pareto"` — 3 passed. Covers the null default, the PUT round-trip, preservation on an update that omits the field, and that an empty layout stays distinguishable from null.
- `pnpm --filter="...[origin/main]" run --parallel --if-present test:ci` — `@nemo/common` 1518 passed (122 files), `studio` 3358 passed (346 files). Includes 11 new `useColumnLayout` cases (load-time clean state, hide/show, reorder, hidden-compared-as-a-set).
- `pnpm --filter="...[origin/main]" run --parallel --if-present typecheck`, `pnpm lint`, `pnpm format` — clean.
- Regenerated `openapi/` and the web SDK on the rebased base and confirmed the only remaining diff is this change; re-running the generator is stable.
- Manually exercised against a local platform: reordering and hiding columns reveals the button, saving persists, and a reload comes back on the saved layout. The stored layout round-trips through the API.

Not done: the Stainless **Python** SDK was not regenerated — `make update-sdk` requires `STAINLESS_API_KEY`, which this environment does not have. The OpenAPI spec and web SDK are committed.

## Summary by CodeRabbit

* **New Features**
  * Experiment evaluation-table column order and hidden columns can now be saved and restored.
  * Added a toolbar action to save unsaved column-layout changes, with saving status and success/error feedback.
  * Saved layouts persist through experiment creation and updates and appear in experiment details.
  * New experiments start without a saved layout; an explicitly saved empty layout is preserved separately.
  * Added support for custom content at the start of data-view toolbars.
* **Tests**
  * Added coverage for layout defaults, persistence, visibility changes, ordering, and unsaved-state detection.

### DIFF
--- a/services/intake/src/nmp/intake/api/v2/experiments/schemas.py
+++ b/services/intake/src/nmp/intake/api/v2/experiments/schemas.py
@@ -13,7 +13,7 @@ from datetime import datetime
 from typing import Annotated, Self
 
 from nmp.common.entities.values import DatetimeFilter, Filter, NumberFilter, map_entity_field
-from nmp.intake.entities.experiments import Experiment, ExperimentGroup, ParetoConfig
+from nmp.intake.entities.experiments import ColumnLayout, Experiment, ExperimentGroup, ParetoConfig
 from nmp.intake.repository.evaluation_session import EvaluationSessionRow
 from nmp.intake.spans.domain import (
     INTAKE_PREVIEW_PAYLOAD_CHAR_LIMIT,
@@ -55,6 +55,13 @@ class ExperimentRequest(BaseModel):
         description=(
             "Default X/Y metrics for the experiment's Pareto view. Omit to preserve the existing value on "
             "update; on create, defaults to cost vs. latency."
+        ),
+    )
+    column_layout: ColumnLayout | None = Field(
+        default=None,
+        description=(
+            "Saved column order and hidden columns for the experiment's evaluations table. Omit to preserve "
+            "the existing layout on update; on create, defaults to an empty layout."
         ),
     )
     is_favorite: bool = Field(
@@ -185,6 +192,13 @@ class ExperimentResponse(BaseModel):
     metadata: dict[str, str] | None = None
     default_sort: str
     pareto: ParetoConfig = Field(default_factory=ParetoConfig)
+    column_layout: ColumnLayout | None = Field(
+        default=None,
+        description=(
+            "Saved column order and hidden columns for the experiment's evaluations table. Null when "
+            "none has been saved."
+        ),
+    )
     is_favorite: bool = Field(default=False, description="Whether this Experiment is marked as a favorite.")
     show_evaluations_over_time: bool = Field(
         default=False,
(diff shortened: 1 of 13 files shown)


services/intake/src/nmp/intake/api/v2/experiments/schemas.py
```diff
@@ -57,6 +57,13 @@ class ExperimentRequest(BaseModel):
             "update; on create, defaults to cost vs. latency."
         ),
     )
+    column_layout: ColumnLayout | None = Field(
+        default=None,
+        description=(
+            "Saved column order and hidden columns for the experiment's evaluations table. Omit to preserve "
```
rrhyne:
```suggestion
            "Saved column order and column visiblity for the experiment's evaluations table. "
```

PR #1717, opened 2026-09-02
feat(evaluator): prefer OTLP over ATIF when reading agent traces

SandyChapman:
## Summary

ATIF's turn-based shape cannot represent span timing, concurrency, or per-call detail, so the evaluator now reads **OTLP first**. A metric can ask for the trace view its question needs (`evidence.trace(format="otlp")`) and get a narrowed handle back; a Harbor trial that emits both encodings exposes both. The `trace_format` runner config is deleted — the primary trace is OTLP when the agent emits one and ATIF otherwise.

Before: one config-selected trace view, and a configured format that matched no artifact left the trial with **no trace at all**. After: both views are always reachable, and the choice of primary is a property of what the trial contains.

Publishing OTLP to Intake is deliberately **not** in this PR — see below.

## Related Issue

Part of AALGO-569 (Linear). Not a GitHub issue, so no `Fixes` keyword.

## Changes

- **`CandidateEvidence.trace()`** takes a keyword-only `format` with `@overload` declarations, so `format="atif"` narrows to `ATIFTraceHandle` and `format="otlp"` to `OTLPTraceHandle`. `trace()` with no `format` is unchanged and returns the union, so no existing caller needed a port.
- **Handle cache keyed on the resolved descriptor key**, not `name`. Keyed on `name` alone, asking for OTLP after ATIF returned the cached ATIF handle — the static type promised `OTLPTraceHandle` and the runtime raised `AttributeError` far from the cause.
- **A candidate is accepted on its declared `descriptor.format`, never on its key suffix**, so a mis-filed descriptor cannot falsify the caller's narrowed return type.
- **Harbor registers `trace:atif` and `trace:otlp`** alongside the standard `trace` key.
- **`trace_format` deleted end-to-end** (3 `harbor_runtime` signatures, `_validate_trace_format`, 2 `harbor_trial_adapter` signatures, and the `runner_info()` key), along with the "configured format matched no trace artifact" warning branch and its tests.
- **`output_text` now comes from the OTLP trace**, falling back to ATIF. Root span preferred, else the latest-ending span carrying an output attribute.
- **New `values/otlp.py`** — OTLP/JSON decoding moved out of `OTLPTraceHandle`, plus protobuf parsing via `opentelemetry-proto` and final-answer extraction.
- **`SkillUsedMetric`** resolves ATIF first, falling back to OTLP, so its answer is independent of which encoding a runner made primary while still working for agents that emit only spans.
- **Generated bundle manifests re-synced** (`make vendor`). The `nemo-evaluator-sdk` extra is mirrored into `packages/nemo_platform`, `packages/nemo_platform_plugin`, and `sdk/python/nemo-platform`, so adding a dependency to the SDK leaves all three stale. Second commit; no hand edits.
- **`atif_steps_from_trial`** reads the ATIF view by name rather than the primary trace, keeping *absent* distinct from *unreadable*.

### Design notes

**The output-attribute key list is deliberately narrower than Intake's.** Intake's `OTLP_OUTPUT_PAYLOAD_ATTRIBUTE_KEYS` answers "what payload did this span emit" for storage and admits `gen_ai.tool.call.result` / `tool_response`. A tool result is not the agent's answer, and `output_text` is compared against a reference by `exact_match`/`bleu`/`rouge`, so admitting one would score a tool's output as the model's. Duplicated rather than imported because the SDK ships independently of the services.

**Text search reads decoded JSON, not the parsed protobuf.** `ParseDict` is strict, so one span with a malformed unrelated field would fail the whole batch and hide every other span's attributes — a false negative in `SkillUsedMetric`, which corrupts a skill A/B result.

**Scope is the evaluator only.** The Experimentalist has its own complete Harbor adapter and never passed `trace_format` to the SDK runner, so its config field is untouched. Verified: its suite passes unmodified against this change.

**Deliberately left out:** publishing OTLP to Intake. That changes how spans are identified in a `ReplacingMergeTree` keyed on `(workspace, session_id, start_time, id)`, where a wrong session id republishes as duplicates instead of replacing — a different risk class that deserves isolated review and a live round-trip test. It follows as a stacked PR.

## Type of Change

- [ ] Documentation only

## Quality Gates

`docs/evaluator/agent-eval/writing-metrics.mdx` gained an "Asking for one trace format" section; its handle table and the description of what `trace()` returns were corrected.

## Verification

Targeted validation:

| Command | Result |
| -- | -- |
| `pytest packages/nemo_evaluator_sdk/tests plugins/nemo-evaluator/tests` | **2687 passed, 49 skipped** |
| `pytest plugins/nemo-experimentalist/tests` | **1015 passed, 52 skipped** (unmodified by this PR) |
| `pytest plugins/nemo-eval-author/tests plugins/nemo-optimization/tests` | **235 passed, 1 skipped** |
| `tools/lint/lint-python-types.sh` (CI's type gate) | All checks passed |
| `uv run ruff check packages plugins` | All checks passed |
| `uv run ruff format --check packages plugins` | 2192 files already formatted |
| `docs/fern` `validate-mdx` | 217 files parsed cleanly |

### `pre-commit run -a` — two hooks did not pass, neither caused by this change

**`ty` — 9 diagnostics, all pre-existing.** Six are in `test_harbor_runtime.py:2159-2188`, below this PR's last hunk at `+2189`; three are `unused-ignore-comment` warnings on `# ty: ignore[unresolved-import]` directives at `harbor_runtime.py:806-808`, which no hunk touches. All are verbatim on `main`. The local hook runs bare `ty check`, while CI's `tools/lint/lint-python-types.sh` passes `--ignore unresolved-attribute --ignore unused-ignore-comment`, which is why the CI gate is green and the local hook is not.

**`uv-lock` — environment, and the artifact is verified correct.** The hook requires uv 0.9.14; this shell has 0.9.30. I re-ran `uv lock` under the repo's pinned toolchain (`flox activate --dir tools/python`, uv 0.9.14) and the result is **byte-identical** to the committed `uv.lock`, so the lock is right even though the hook could not confirm it here.

Every other hook passed, including copyright headers, config-reference docs, helm-docs, uv-lock drift, and the plugin/`nmp-common` boundary check.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting ATIF or OTLP trace evidence by format.
  * Added OTLP trace parsing, validation, and agent output extraction.
  * Added support for JSONL trace input and resilient nested attribute handling.

* **Improvements**
  * Harbor evaluations prefer OTLP traces and fall back to ATIF.
  * Skill-use detection supports both trace formats and tolerates unreadable data.
  * Missing requested formats now return clear errors.

* **Documentation**
  * Clarified trace formats, defaults, precedence, and selection behavior.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
```diff
@@ -0,0 +1,300 @@
...
+    Args:
+        request: Parsed OTLP export request.
+
+    Returns:
+        The answer text, or ``None`` when no span carries one.
+    """
+    spans = [span for rs in request.resource_spans for ss in rs.scope_spans for span in ss.spans]
+    by_recency = sorted(spans, key=lambda span: span.end_time_unix_nano, reverse=True)
+    roots = [span for span in by_recency if not span.parent_span_id]
+    for span in roots + by_recency:
```
ngoncharenko, line 167: **[P1] Exclude tool spans from final-answer fallback**

When the root has no output, this scans every span. The repository itself stores TOOL results in `output.value` in [`otlp_build.py`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/d9aae64b9e7adb3bfac1740a091c3c07504229c6/plugins/nemo-insights/evaluation/otlp_build.py#L247), so the latest tool result can become `AgentOutput.output_text` and be scored as the agent answer. Restrict fallback candidates to agent/chain/LLM semantics and explicitly exclude tool and evaluator spans.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
```diff
@@ -0,0 +1,300 @@
...
+        contents: list[str] = []
+        for part in parts:
+            if not isinstance(part, dict) or part.get("type") != "text":
+                continue
+            content = part.get("content")
+            if isinstance(content, str):
+                contents.append(content)
+        if contents:
+            return "".join(contents)
+    return raw
```
ngoncharenko: **[P1] Let ATIF handle empty message envelopes**

For a parsed `gen_ai.output.messages` payload with no assistant text—such as `[]` or user-only messages—returning the raw JSON produces a non-answer. `_trial_output_text` treats that string as truthy and never falls back to the valid ATIF answer, causing content metrics to score serialized telemetry. Return `None` for recognized payloads without assistant text so the existing ATIF fallback can run.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_trial_adapter.py
```diff
@@ -407,24 +428,16 @@ def _harbor_extension_evidence(
...
+    atif_trace = collected_traces[EVIDENCE_FORMAT_ATIF]
+    if atif_trace is None:
+        atif_trace = legacy_atif
+    otlp_trace = collected_traces[EVIDENCE_FORMAT_OTLP]
+    for evidence_format, descriptor in ((EVIDENCE_FORMAT_ATIF, atif_trace), (EVIDENCE_FORMAT_OTLP, otlp_trace)):
+        if descriptor is not None:
+            descriptors[f"{EVIDENCE_TRACE}:{evidence_format}"] = descriptor
+
+    # OTLP carries span timing and concurrency that ATIF's turn shape cannot represent.
+    return descriptors, otlp_trace if otlp_trace is not None else atif_trace
```
ngoncharenko: **[P2] Do not promote unreadable JSONL over valid ATIF**

Any `.jsonl` below a `traces` directory becomes `otlp_trace` based only on its path. With malformed JSONL and valid ATIF, output extraction falls back to ATIF, but the primary `evidence.trace()` remains the unreadable OTLP handle; after removing the selector, default trace consumers lose the usable trace. Validate the OTLP request before promoting it, falling back to ATIF while retaining the malformed file under its extension key.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/metrics.py
```diff
@@ -182,6 +189,23 @@ async def _any_skill_used(self, candidate: CandidateOutput, provenances: list[Ma
...
 
+async def _skill_trace_handle(evidence: CandidateEvidence, name: str) -> TraceHandle:
+    """The ATIF view of a trace when the trial has one, otherwise the OTLP view.
+
+    ATIF answers a content question against typed messages and tool calls, so it wins when
+    both exist and the score stays independent of which encoding a runner made primary.
+    """
+    try:
+        return await evidence.trace(name, format=EVIDENCE_FORMAT_ATIF)
+    except KeyError:
```
ngoncharenko: **[P2] Retry OTLP when ATIF cannot be materialized**

The helper selects ATIF from descriptor existence alone, but validation occurs afterward. If Harbor discovers a malformed `*.atif.json` alongside valid OTLP, the later read exception immediately records `skill_used=False`, ignoring usable evidence. Retry the OTLP view when ATIF reading or validation fails, while retaining ATIF precedence when it parses successfully.

docs/evaluator/agent-eval/writing-metrics.mdx
```diff
@@ -136,14 +137,34 @@ async def read_trace(input: MetricInput) -> None:
...
 It's a responsibility of consumer to parse `resourceSpans` tree specific to OTLP exporter to extract the searched data.
 
+### Asking for one trace format
+
+A runner may record the same trial in more than one encoding. Harbor registers both under
+`trace:atif` and `trace:otlp` when the agent under test emits both. Pass `format=` to choose one
+and get a narrowed handle back, with no runtime branch:
+
+```python
+otlp = await evidence.trace(format="otlp")  # OTLPTraceHandle
```
ngoncharenko: **[P2] Make the async snippet independently checkable**

`make docs-check-python-snippets DOCS_PATH=docs/evaluator/agent-eval/writing-metrics.mdx` fails because this fence has top-level `await` and an undefined `evidence`. Wrap the example in an async function that accepts or initializes `CandidateEvidence`. The applicable documentation guidance requires this focused check: [`docs/AGENTS.md`](https://github.com/NVIDIA-NeMo/nemo-platform/blob/d9aae64b9e7adb3bfac1740a091c3c07504229c6/docs/AGENTS.md#L18).

ngoncharenko (commented):
Posting the five verified findings from the requested evaluator trace-format review.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
```diff
@@ -0,0 +1,300 @@
...
+# Deliberately narrower than Intake's ``OTLP_OUTPUT_PAYLOAD_ATTRIBUTE_KEYS``
+# (``services/intake/src/nmp/intake/spans/ingest/otlp.py``), which answers "what payload did
+# this span emit" for storage and so also admits ``gen_ai.tool.call.result`` and
+# ``tool_response``. A tool result is not the agent's answer, and this value is compared
+# against a reference by the content metrics, so admitting one would score a tool's output
+# as the model's. The list is duplicated rather than imported because the SDK ships
+# independently of the services and must not depend on one.
+FINAL_OUTPUT_ATTRIBUTE_KEYS = (
+    "output.value",
+    "gen_ai.output.messages",
```
ngoncharenko: should we replace this with `_MESSAGE_ATTRIBUTE_KEY` defined below? They are identical

docs/evaluator/agent-eval/writing-metrics.mdx
```diff
@@ -136,14 +137,34 @@ async def read_trace(input: MetricInput) -> None:
...
+calls = await atif.tool_calls()
+```
+
+`trace()` with no `format` still returns whichever encoding the runner made primary. For Harbor
+that is the OTLP trace when the agent emits one and the ATIF trace otherwise — there is no setting
+to choose between them. A `format=` request raises `KeyError` when the trial carries no trace in
+that encoding, so a metric that depends on one specific view fails loudly instead of silently
+scoring nothing.
+
 The [Score by Component](/documentation/evaluate-models/agent-eval/score-by-component) guide has a
```
albcui, line 171: we should prolly also update [score-by-component.mdx:76](https://github.com/NVIDIA-NeMo/nemo-platform/blob/main/docs/evaluator/agent-eval/score-by-component.mdx?plain=1#L253-L255) to leverage the new `format=` key, since we default to `otlp` now.

```python
handle = await evidence.trace(EVIDENCE_TRACE, format="atif")
# can also drop the raise since we statically type to ATIFTraceHandle now
```

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
```diff
@@ -0,0 +1,300 @@
...
+            except json.JSONDecodeError as exc:
+                raise ValueError(f"invalid OTLP JSON on line {line_number}") from exc
+
+    resource_spans: list[dict[str, Any]] = []
+    for request in requests:
+        resource_spans.extend(resource_spans_from_request(request))
+    return resource_spans
+
+
+def export_request_from_resource_spans(resource_spans: list[dict[str, Any]]) -> ExportTraceServiceRequest:
```
ngoncharenko, line 123: Experimentalist does something similar: plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/otlp.py > spans_to_protobuf()

I wonder if we should have a shared utility for this?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
```diff
@@ -578,20 +529,64 @@ async def filesystem(self, name: str) -> LocalFilesystemEvidence:
...
+        A candidate is accepted on its declared ``format``, never on its key suffix,
+        so a mis-filed descriptor cannot falsify the caller's narrowed return type.
+        """
+        if evidence_format is None:
+            return name
+        for key in (name, f"{name}:{evidence_format}"):
+            descriptor = self.get(key)
+            if descriptor is not None and (descriptor.format or EVIDENCE_FORMAT_ATIF) == evidence_format:
+                return key
+        raise KeyError(f"missing evidence descriptor {name!r} in format {evidence_format!r}")
```
ngoncharenko, line 548: I wonder if should raise here vs. return None - basically have a dict look up pattern instead

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
SandyChapman: Only the hex→base64 id bit overlaps — `spans_to_protobuf()` chunks Intake rows for upload, this parses one trace into one request. I think this is outside the scope of the current work as the functions differ a bit too much.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/evidence.py
SandyChapman: `CandidateEvidence` already has both halves of that pattern — `get()` returns `None` like `dict.get`, `require()`/`trace()` raise like `d[k]`. Keeping the raise here since `None` can't distinguish "no trace" from "trace exists in the other encoding", and it would cost the narrowed return type.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
```diff
@@ -0,0 +1,300 @@
...
+    Raises:
+        ValueError: The payload does not match the OTLP schema.
+    """
+    try:
+        return json_format.ParseDict(
+            {"resourceSpans": [_base64_ids(resource_span) for resource_span in resource_spans]},
+            ExportTraceServiceRequest(),
+            # A producer on a newer OTLP schema stays readable rather than failing wholesale.
+            ignore_unknown_fields=True,
+        )
```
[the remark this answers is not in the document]
SandyChapman: Done — normalized at both boundaries: `export_request_from_resource_spans` now converts it to `ValueError` (`deepcopy` recurses too), and `_any_value_strings` walks an explicit stack; regressions cover depth 400 and 2000.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
SandyChapman: Done — only `AGENT`/`CHAIN`/`LLM` spans are eligible now, applied to the root as well, so a tool-only trace yields no answer rather than a misattributed one.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
SandyChapman: Done — `_assistant_text` returns `None` for a recognized payload with no assistant text, so the key loop continues and the ATIF fallback runs.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_trial_adapter.py
SandyChapman: Done — the OTLP trace is read before anything is promoted, so it only becomes primary if it parses; the malformed file stays reachable under its extension key.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/metrics.py
SandyChapman: Done — each view is now resolved *and* read before the next is considered, OTLP-first per this ticket's default, scoring `False` only when neither reads and logging which failed.

docs/evaluator/agent-eval/writing-metrics.mdx
SandyChapman: Done — wrapped in `async def read_both_views(evidence: CandidateEvidence)`; the check passes 5/5 and now proves the narrowing, since flipping to `format="atif"` fails on `.resource_spans()`.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/otlp.py
SandyChapman: Done — `_MESSAGE_ATTRIBUTE_KEY` moved above the tuple and substituted for the literal, so the `key ==` comparison can't drift from the list.

docs/evaluator/agent-eval/writing-metrics.mdx
SandyChapman: Done — both occurrences now pass `format="atif"` and drop the raise; a trial with no ATIF view scores `0.0` instead of raising, and the surrounding prose was updated to say so.

PR #1680, opened 2026-09-01
fix(intake): declare the OTLP trace-ingest protobuf body

SandyChapman:
## Summary

The Intake OTLP trace-ingest endpoint streams its body off the raw FastAPI `Request`, so FastAPI inferred no schema for it and emitted **no `requestBody`** in the OpenAPI spec. Generated clients consequently had no way to send a payload — the SDK's `create()` accepted only `workspace` and the `extra_*` passthroughs, and the CLI command posted an empty body. The endpoint itself was always fine; producers reach it over raw HTTP, which is why this went unnoticed.

After this change the spec declares the protobuf body, and `client.intake.ingest.otlp.v1.traces.create(body=..., workspace=...)` works. `nemo-insights` and `nemo-experimentalist` currently hand-build the URL and `Content-Type` header over `httpx`; they can be simplified onto the SDK in follow-ups.

## Related Issue

Relates to AALGO-569 (Linear). No GitHub issue.

## Changes

- `services/intake/.../ingest/otlp.py`: declare the body via `openapi_extra` (`application/x-protobuf`, `type: string` / `format: binary`), matching the fileset upload endpoint. Drop the now-redundant `content_type` Header parameter so Content-Type is described once — by the requestBody — instead of twice by two mechanisms that can disagree; the value is read from `request.headers` as the fileset endpoint does. **The handler's behavior is unchanged**: it already read the body correctly and already rejected non-protobuf media types.
- `openapi/`: regenerated (3 specs).
- `sdk/python/nemo-platform/`: regenerated via `sdk/stainless.sh sync`. `create()` gains `body: FileContent | BinaryTypes`, sets `Content-Type: application/x-protobuf`, and passes `content=`.
- `cli_config.yaml`: skip CLI generation for `[intake, ingest, otlp]`, removing `nemo intake ingest otlp`. The body is binary protobuf, which cannot be passed as a shell argument; the generated positional made the command impossible to invoke correctly rather than merely useless. No docs, skills, or `.md`/`.mdx` referenced it. This matches the effective outcome for the other binary endpoint (fileset upload has no CLI command).
- `services/intake/tests/.../test_otlp_ingest_media_type.py`: new. Covers the 415 media-type path (previously untested), asserts the spec declares the body, and exercises `create()` end-to-end through both the sync and async SDK.

### Unrelated drift included

[...]

PR #1725, opened 2026-09-02
feat(evaluator): publish the OTLP trace to Intake when a trial has one

SandyChapman:
## Summary

A trial's OTLP trace is now what gets published to Intake, carrying the span tree, per-call detail and timing that ATIF's turn shape flattens. Runners whose agent emits no OTLP keep publishing ATIF exactly as before — this adds a path rather than replacing one.

**Stacked on #1717.** Review that one first; this PR's diff is only the publish work.

## Related Issue

AALGO-569 (Linear), "PR 4" in its sequencing table. Not a GitHub issue, so no `Fixes` keyword.

## Changes

- **`mapping.otlp_ingest_for_trial`** reads the trial's OTLP trace as a typed `ExportTraceServiceRequest`, stamps identity and trial totals, and returns the serialized payload plus its root span id. Returns `None` — falling back to ATIF — when the trial has no readable OTLP trace or no single scorable root span.
- **`publish.py`** sends that payload via `intake.ingest.otlp.v1.traces.create(body=…)` (landed in #1680) and scores against the locally-read span id.
- **`values/otlp.py`** gains `set_span_attributes`, `set_root_span_attributes`, `set_root_span_error`, `fill_missing_start_times`, and `root_span_id`.
- **`OTLPTraceHandle.export_request()`** returns, now that it has a caller.

### Design calls

**Identity goes on span attributes, not resource attributes.** Intake merges the layers as `{**resource, **span}`, so an agent recording its own `gen_ai.conversation.id` would win from the resource layer and take the session id with it. That id is part of the key Intake's `ReplacingMergeTree` replaces on, so losing it turns a re-publish into duplicate rows.

**`_resolve_root_span_id` stays, for the ATIF path only.** ATIF span ids are minted inside Intake via `stable_id(workspace, session_id, *identity, "trajectory", prefix="span")` and cannot be known locally without replicating that scheme. Deleting it outright would have broken score attachment for exactly the ATIF-only runners this PR keeps supporting. Note this corrects AALGO-569, which justifies removing the round trip with "OTLP span IDs are producer-chosen, so Evaluator already knows the id" — untrue for Harbor, where the *agent under test* writes the trace. The real reason is that the root span id is readable from the payload we are about to publish.

**Four things the OTLP path must carry that the agent's spans do not**, each found by review rather than by a failing test:

1. Trial token and cost totals, which ATIF sent as `final_metrics`. On the **root span alone**, so a rollup summing across a trace cannot count them once per span.
2. The trial's error, as **span status** — Intake reads status, not attributes, to decide a span failed. Recorded only as `exception.type`, a failed trial read as successful.
3. A start time for any span lacking one. Intake stores such a span against its own ingest clock (`_nanos_to_datetime(...) or ingested_at`), and start time is in the replace key, so re-publish would insert instead of replacing.
4. A check of the per-span error list ingest returns **with its 200**. A dropped span is otherwise invisible — including when the dropped span is the one about to be scored.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification:

`publish_to_intake` is an internal publish path with no user-facing configuration or CLI surface; which encoding a trial publishes is a property of what the runner produced, not something a user selects. The user-facing trace-format documentation was updated in #1717.

## Verification

Targeted validation:

| Command | Result |
| -- | -- |
| `pytest plugins/nemo-evaluator/tests/integration/test_publish_to_intake.py` (real ClickHouse + platform) | **6 passed** |
| `pytest packages/nemo_evaluator_sdk/tests plugins/nemo-evaluator/tests` | **2706 passed, 51 skipped** |
| `tools/lint/lint-python-types.sh` (CI's type gate) | All checks passed |
| `uv run ruff check packages plugins` | All checks passed |
| `uv run ruff format --check packages plugins` | 2192 files already formatted |

### The integration test was run, not just written

The idempotency claim is the one unit doubles cannot reach, so it is covered end to end against real ClickHouse: publish twice, assert **one** trace and **one** span. To confirm the assertion has teeth, `session_id_for` was mutated to append a per-call uuid — the exact "unstable session id" failure the `ReplacingMergeTree` key is vulnerable to — and the test went red. Code restored and re-verified.

The four ATIF integration tests pass alongside the two new OTLP ones, so the fallback path is unregressed.

### `pre-commit run -a` — one hook blocked, unrelated to this change

**`uv-lock`** requires uv 0.9.14; this shell has 0.9.30. No dependency changed in this PR, and the sibling **`uv-lock-check` (drift) passed**, confirming `uv.lock` is untouched. Every other hook passed, including `ty`, ruff, copyright headers, config-reference docs, helm-docs, and the plugin/`nmp-common` boundary check.

## Known gap, not addressed here

Intake's ATIF ingest endpoint calls `validate_evaluation_context`; its OTLP endpoint has no equivalent. So a nonexistent evaluation name now publishes successfully where the ATIF path would have failed. This PR is the first evaluator caller to exercise that endpoint and so the first to expose the gap, but the fix belongs in Intake alongside its ATIF sibling — a client-side check here would duplicate a server responsibility and need removing later. Worth its own ticket.

## Summary by CodeRabbit

* **New Features**
  * Added support for publishing OTLP traces from evaluation trials.
  * Preserved trace and span identities, timestamps, usage totals, errors, and evaluation metadata.
  * Added validation for root spans and malformed trace data.
  * Added fallback to the existing ATIF publishing path when OTLP evidence is unavailable.
  * Re-publishing results now avoids creating duplicate traces or spans.

* **Bug Fixes**
  * Unknown evaluations are rejected before trial data is written.
  * Trial and per-span ingestion errors are surfaced during publication.

### DIFF
--- a/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
@@ -73,15 +73,6 @@ _SESSION_ID_ATTRIBUTE = "gen_ai.conversation.id"
 _EVALUATION_NAME_ATTRIBUTE = "nemo.evaluation.name"
 _TEST_CASE_NAME_ATTRIBUTE = "nemo.test_case.name"
 
-# Trial-level totals, which Intake reads as numbers. ATIF carried these as ``final_metrics``;
-# under OTLP they go on the root span, which is the span that stands for the whole trial.
-_MEASUREMENT_ATTRIBUTES = {
-    "prompt_tokens": "gen_ai.usage.input_tokens",
-    "completion_tokens": "gen_ai.usage.output_tokens",
-    "cache_read_tokens": "gen_ai.usage.cached_tokens",
-    "cost_usd": "gen_ai.usage.cost",
-}
-
 
 def session_id_for(run_id: str, trial_id: str) -> str:
     """Return the stable, adapter-minted session id for a trial.
@@ -201,9 +192,15 @@ async def otlp_ingest_for_trial(
 
 def _trial_totals(trial: AgentEvalTrial, measurements: TrialMeasurements) -> dict[str, str | int | float]:
     """The trial-level totals and error that ATIF conveyed outside the step list."""
+    # Read as numbers by Intake. ATIF carried these as ``final_metrics``; under OTLP they go on
+    # the root span, which is the span that stands for the whole trial.
     totals: dict[str, str | int | float] = {}
-    for field, attribute in _MEASUREMENT_ATTRIBUTES.items():
-        value = getattr(measurements, field)
+    for value, attribute in (
+        (measurements.prompt_tokens, "gen_ai.usage.input_tokens"),
+        (measurements.completion_tokens, "gen_ai.usage.output_tokens"),
+        (measurements.cache_read_tokens, "gen_ai.usage.cached_tokens"),
+        (measurements.cost_usd, "gen_ai.usage.cost"),
+    ):
         if value is not None:
             totals[attribute] = value
     if trial.error is not None:
--- a/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
+++ b/plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
@@ -36,9 +36,17 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Literal, cast
 
+from nemo_evaluator_sdk.agent_eval.metrics import TrialMeasurements
 from nemo_evaluator_sdk.agent_eval.scores import AgentEvalScoreStatus, AgentEvalTaskScore
 from nemo_evaluator_sdk.agent_eval.trials import AgentEvalTrial
-from nemo_evaluator_sdk.values.evidence import EVIDENCE_FORMAT_ATIF, EVIDENCE_TRACE
+from nemo_evaluator_sdk.values.evidence import EVIDENCE_FORMAT_ATIF, EVIDENCE_FORMAT_OTLP, EVIDENCE_TRACE
+from nemo_evaluator_sdk.values.otlp import (
+    fill_missing_start_times,
+    root_span_id,
+    set_root_span_attributes,
+    set_root_span_error,
+    set_span_attributes,
+)
 from nemo_platform.types.intake.evaluation_context_param import EvaluationContextParam
 from nemo_platform.types.intake.evaluator_result_create_params import EvaluatorResultCreateParams
 from nemo_platform.types.intake.evaluator_result_data_type import EvaluatorResultDataType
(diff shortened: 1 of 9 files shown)


plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
```diff
@@ -119,6 +142,77 @@ async def atif_steps_from_trial(trial: AgentEvalTrial, *, started_at: datetime)
...
+        set_root_span_error(request, message=trial.error.message)
+    fill_missing_start_times(request, start_time_unix_nano=int(started_at.timestamp() * 1_000_000_000))
+    return request.SerializeToString(), span_id
+
+
+def _trial_totals(trial: AgentEvalTrial, measurements: TrialMeasurements) -> dict[str, str | int | float]:
+    """The trial-level totals and error that ATIF conveyed outside the step list."""
+    totals: dict[str, str | int | float] = {}
+    for field, attribute in _MEASUREMENT_ATTRIBUTES.items():
+        value = getattr(measurements, field)
```
ngoncharenko: not sure I'd use here `getattr(measurements, field)`. Looks concise but what if _MEASUREMENT_ATTRIBUTES and TrialMeasurements diverge in the future?

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
ngoncharenko: function is quite long. Can we put atif and oltp related logic into separate functions and invoke them here?

ngoncharenko (approved):

plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
SandyChapman: Dropped the name table — the four totals are read as typed attributes now, so a rename on `TrialMeasurements` fails `ty` here rather than only reddening this plugin's tests.

plugins/nemo-evaluator/src/nemo_evaluator/intake/publish.py
SandyChapman: Split into `_publish_otlp` and `_publish_atif`; `_publish_trial` is 8 lines that just pick the path, with `ended_at` and `build_body` moved into the ATIF helper that was their only caller.

plugins/nemo-evaluator/src/nemo_evaluator/intake/mapping.py
```diff
@@ -119,6 +142,77 @@ async def atif_steps_from_trial(trial: AgentEvalTrial, *, started_at: datetime)
...
+            "Ignoring OTLP trace for trial %s: it has no single root span with a usable id to score.", trial.id
+        )
+        return None
+
+    set_span_attributes(
+        request,
+        {
+            _SESSION_ID_ATTRIBUTE: session_id,
+            _EVALUATION_NAME_ATTRIBUTE: evaluation_name,
+            _TEST_CASE_NAME_ATTRIBUTE: trial.task_id,
```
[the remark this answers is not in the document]
SandyChapman: Confirmed and fixed — `validate_evaluation_context` is called from `atif.py` and `chat_completions.py` only, so OTLP ingest really did skip it. `publish_to_intake` now resolves the evaluation once up front and stops with a `PublishError`. Note it has to be `evaluations.retrieve`, not `experiments.retrieve`: `/experiments/{name}` resolves an `ExperimentGroup`, while ingest validates the `Experiment` behind `/evaluations/{name}`. This is only a client-side guard — enforcing it in `ingest_otlp_traces` alongside ATIF is the real fix, but that is Intake's call since it would start 400ing existing OTLP producers.

PR #1728, opened 2026-09-02
fix(entities): hide pending-deletion workspaces from the list API

tylersbray:
## Summary

- `GET /apis/entities/v2/workspaces` now omits workspaces with a non-null `deletion_stage` (`deletion_stage IS NULL`) so list matches GET/DELETE-by-name (those already 404 while cleanup is in progress). Fixes the Studio contradiction in NVBug 6709050.
- Integration tests cover: pending hidden after DELETE, live sibling still listed, DELETING/FAILED omitted, and the cleanup repo query with `EQ pending` still finds candidates.
- Guardrail: `--service-group all` still starts **zero** controllers.

## Docs vs runner (intentional)

We **did not** change the platform runner so that `--service-group all` auto-starts controllers. Helm API pods pass `--service-group=all` and must keep running with no controllers.

Instead, local/dev docs (`AGENTS.md`, `CONTRIBUTING.md`, `packages/nmp_platform/config/local.yaml`) now tell you to add **`--controller-group all`** when you actually want controllers (including WorkspaceCleanup):

```
nemo services run --service-group all --controller-group all
```

A runner config test locks the Helm-safe behavior: `--service-group all` ⇒ empty controller set.

## Test plan

- [x] Entities workspace-deletion integration tests
- [x] Runner config test: `--service-group all` does not start controllers
- [x] Pod smoke (nemo-dev-blue / tbray-dev): after DELETE, workspace is gone from list immediately, GET 404, live sibling remains, logs show `Processing workspace deletion` / `Entities controller started`
- [ ] Full row delete in that pod still 404s listing deployments (inference/deployments plugin not in this smoke service set). That is an env gap, not the list-filter bug.

## Summary by CodeRabbit

- **Bug Fixes**
  - Workspace listings now omit workspaces pending deletion, being deleted, or marked as failed.
  - Deleted workspaces remain unavailable through direct retrieval, returning a not-found response.
  - Active sibling workspaces continue to appear normally.

- **Documentation**
  - Local development instructions now show how to start all service and controller groups together.
  - Workspace API documentation clarifies that deleting workspaces are excluded from listings.

### DIFF
--- a/packages/nmp_platform/config/local.yaml
+++ b/packages/nmp_platform/config/local.yaml
@@ -7,7 +7,7 @@
 #   export NMP_BASE_URL=http://127.0.0.1:8080
 #   uv run nemo services run --host 127.0.0.1 --port 8080
 #
-# Use default service set (omit --services) or --service-group all. Then:
+# Use default service set (omit --services) or --service-group all --controller-group all. Then:
 #   nemo auth login --unsigned-token
 #   uv run nemo-platform run task --task nmp.platform_seed
 
--- a/services/core/entities/src/nmp/core/entities/api/v2/workspaces/endpoints.py
+++ b/services/core/entities/src/nmp/core/entities/api/v2/workspaces/endpoints.py
@@ -25,7 +25,7 @@ from datetime import datetime, timezone
 
 from fastapi import APIRouter, HTTPException, Query, status
 from nmp.common.api.common import DeleteResponse, GenericSortField, Page, PaginationData
-from nmp.common.api.filter import ComparisonOperation, FilterOperator
+from nmp.common.api.filter import ComparisonOperation, FilterOperation, FilterOperator, LogicalOperation
 from nmp.common.auth.models import Principal
 from nmp.core.entities.api.dependencies import AuthClientDep, EntityRepository, WorkspaceRepository
 from nmp.core.entities.api.v2.utils import (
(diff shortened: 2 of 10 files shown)


packages/nmp_platform/config/local.yaml
```diff
@@ -7,7 +7,7 @@
 #   export NMP_BASE_URL=http://127.0.0.1:8080
 #   uv run nemo services run --host 127.0.0.1 --port 8080
 #
-# Use default service set (omit --services) or --service-group all. Then:
+# Use default service set (omit --services) or --service-group all --controller-group all. Then:
```
mckornfield, line 10: lol did you or the agent hit this

services/core/entities/src/nmp/core/entities/api/v2/workspaces/endpoints.py
```diff
@@ -50,6 +50,22 @@
 API_TAG = "Entity Store"
 logger = logging.getLogger(__name__)
 
+_ACTIVE_WORKSPACE_FILTER = ComparisonOperation(
```
mckornfield, line 53: does this need to be a constant if it's only in this fn?

mckornfield (approved):
should this target release/0.5

packages/nmp_platform/config/local.yaml
tylersbray: Agent, but it also said it updated the docs but shied away from updating the runner itself.  I had it call that out in the PR description in case we didn't want to load up on too many flags.

services/core/entities/src/nmp/core/entities/api/v2/workspaces/endpoints.py
tylersbray: It's part of my anti-slop cleanup that I have it do.  It uses that filter more than once so I'm okay with having it once at the top as a "magic constant" filter, helps readability imho.

PR #1731, opened 2026-09-02
fix(studio): stop side panels closing twice after react-router v8

aray12:
## Summary

Side panels close twice: the panel animates out, snaps back open, then animates out again — visible as a flash. This regressed with the react-router v6 → v8 upgrade (a6904bd83). Passing `flushSync: true` on the navigations that close these panels restores the pre-upgrade commit ordering, and switching `RouterProvider` to the `react-router/dom` entry point makes that option actually take effect.

### Before

https://github.com/user-attachments/assets/2a043b16-b602-4393-9558-583adcc779d0

### After

https://github.com/user-attachments/assets/b4d260c2-ea7c-478e-8933-69fef900e05e

## Changes

**Root cause.** Two deferrals stack on the close path:

1. KUI's `SidePanel` passes `animateOutDuration = 200` into `useDialog`, so `onOpenChange(false)` — and therefore the consumer's `navigate()` — fires from inside a `setTimeout` after the exit animation.
2. `react-router@8.3.0` wraps every router state update in `React.startTransition` by default (`lib/components.js:348`). v6 was synchronous.

React flushes the urgent `setIsClosing(false)` first, producing an intermediate commit where the `open` prop is still `true` but the native `<dialog>` is already closed. `useDialog`'s layout effect takes its `if (isOpen)` branch and calls `showNativeDialog()`, re-opening the panel; when the transition finally lands, it closes a second time.

**Fixes**

- `App.tsx` — import `RouterProvider` from `react-router/dom`, keeping `createBrowserRouter` on `react-router`. Only the DOM variant injects `ReactDOM.flushSync`. `a6904bd83` moved both onto the bare entry point during the v8 upgrade, which silently turned the `flushSync` navigate option into a no-op that only logs a warning. This is a prerequisite for everything below.
- `DeploymentsListRoute` — `flushSync: true` on the details-panel close and on the post-delete close.
- `DeploymentsListRoute` — `flushSync: true` on the create-panel deep-link param cleanup. Same shape via `setSearchParams`: a deferred param clear let the `createPrefill` effect observe stale params and reopen the panel.
- `FilesetListRoute/PanelManagement` — `flushSync: true` on the dataset-panel close, file-panel close, and file-panel outside-click. The existing mirrored open flags are kept; the comment explaining them is corrected to say the two mechanisms are complementary.
- `MetricRunSidePanel` — same treatment on its post-submit navigation. This component currently has no consumer outside its own directory and stories, so nothing exercises it; hardened so it is correct when wired up.
- `tests/util/render.tsx` — shared harness uses the `react-router/dom` `RouterProvider`, so tests can exercise the flushed path. Note this only applies to the `createMemoryRouter` branch; the declarative `MemoryRouter` fallback ignores `flushSync` entirely.
- `FilesetNewRoute` — `flushSync: true` on the close navigation. This is the create-fileset panel, and the most pronounced instance: the route *is* the panel and is mounted with a literal `open`, so the prop can never go false and the navigation unmounting the route is the only thing that closes it. The re-show is therefore guaranteed rather than merely likely, and the panel flashes back up until the transition lands.
- New regression tests at `routes/DeploymentsListRoute/index.test.tsx` and in `routes/FilesetNewRoute/index.test.tsx`.

Navigations that move *between* panel states rather than closing to a panel-less route were deliberately left alone, as were panels whose `open` is plain local state.

The `flushSync: true` options carry no explainer comments — the rationale lives here and in the commit history, with the regression tests as the executable record. One line is kept in the test harness: that the declarative `MemoryRouter` fallback silently ignores `flushSync` is not inferable from the code, and a test written against that path would pass vacuously.

**Sweep coverage.** The first pass looked for `open={...}` expressions derived from route params, which does not match a bare `open` attribute — that is how `FilesetNewRoute` was missed initially. Re-swept for the bare-`open` form: three other panels use it (`SafeSynthesizerFilesetPreview/FilePreview`, `customizer/CustomizationFilesetSelect`, `PromptTuningForm/InContextLearningSection`), but all three close via local `setState`, which is urgent and leaves no window for the re-show.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal bug fix with no user-facing documented behavior change; the reasoning is captured in code comments at each call site.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pnpm --filter nemo-studio-ui typecheck` | pass |
| `pnpm --filter nemo-studio-ui lint` (`--max-warnings 0`) | pass |
| `pnpm --filter nemo-studio-ui format` | pass |
| `pnpm --filter nemo-studio-ui test` (full suite) | 3 pre-existing flakes, see below |
| `pnpm --filter nemo-studio-ui test src/routes/DeploymentsListRoute src/routes/FilesetListRoute` | pass — 7 files, 62 tests |
| `pnpm --filter nemo-studio-ui test src/routes/FilesetNewRoute` | pass — 23 tests |
| `uv run pre-commit run -a` | 2 environment-only failures, see below |

**Full-suite flakes.** Two runs of the full suite on an identical tree produced *different* failure sets (4 files / 49 skipped, then 2 files / 0 skipped), and an earlier run of the same tree was fully green. The affected tests — `components/Layouts/GlobalNav` and `routes/SafeSynthesizerNewRoute` — pass in isolation, take ~11.5s when they fail (timeouts under load), build their own routers directly from `react-router`, and do not touch the shared test harness or any route changed here. Treating these as pre-existing load flakiness rather than a regression; CI is the arbiter.

**Red/green check on the regression tests.** With `flushSync: true` reverted from `handleCloseDetailsPanel`, the deployment test fails with `expected "showModal" to not be called at all, but actually been called 1 times` — exactly the predicted spurious re-show. It passes with the fix restored. The `FilesetNewRoute` test behaves identically — one spurious re-show without the fix, none with it. An earlier `MutationObserver`-based version of these tests passed against *both* the fixed and broken code; the shared test setup (`packages/testing/src/react/setup.ts:50`) stubs `MutationObserver` to a no-op suite-wide, so the test now spies on `HTMLDialogElement.prototype.showModal` instead. Worth knowing more broadly: that stub also makes KUI's own top-layer-heal `MutationObserver` inert under test.

**`uv run pre-commit run -a` — two failures, both environment-only and unrelated to this change:**

- `Helm Docs` — `helm-docs` is not installed locally. This PR changes no Helm files.
- `Run uv lock with platform uv` — local uv is 0.9.28, the hook requires 0.9.14. This PR changes no `pyproject.toml`; the separate `Check for uv.lock drift` hook passed.

Both hooks reported `(no files to check) Skipped` on the actual commit, confirming they do not apply here. Every other hook passed.

**Not verified.** The `FilesetNewRoute` flash was reported from manual browser use and the fix is verified by test, not yet re-confirmed in a browser. Regression tests cover the deployment details panel's close button and the create-fileset panel's close button; Escape, outside click, the fileset list panels, the create-deployment deep link, and the focus-restoration behavior remain unverified. `MetricRunSidePanel` is unverifiable by construction since it has no consumer.

## Bug Fixes

- Fixed deployment, dataset, file, and fileset panels reopening or remaining visible after being closed.
- Improved panel dismissal during animated transitions and deep-linked navigation.
- Synchronized URL updates with panel visibility when closing panels or clearing deployment creation links.
- Fixed navigation to metric evaluation results when closing the creation panel.
- Prevented duplicate navigation after dismissing the fileset creation panel.


marcusds (approved):
Image

PR #1735, opened 2026-09-02
chore(customizer): expose param to configure full results table in W&B

anubhutivyas:
## Summary
Adds `training.log_nemo_gym_full_result_tables` (default false) to GRPO jobs. When enabled, NeMo-RL logs each rollout's full NeMo-Gym result payload to the user's own W&B as a Table under `<agent>/full_result`.

Today there is no way to read a rollout's generated text. The per-step train_data_step*.jsonl is written by default but carries generations as token ids with an empty content field — nemo_gym.py sets "content": "" on both the user and assistant messages. The W&B full-result table is the only surface where NeMo-RL exposes decoded text (prompt_str / generation_str) alongside every reward component the environment returned.

NeMo-RL already supports the setting end to end: it reads it off logger.wandb, gates the tables on wandb_enabled AND flag, and pops the key before wandb.init(). The platform already passes a W&B config block through. The switch just wasn't exposed.

## Related Issue
https://linear.app/nvidia/issue/AALGO-555/support-rollout-viewer-in-studio

## Changes
- `GRPOTraining.log_nemo_gym_full_result_tables` -> `GRPOConfig` -> `compiler` -> injected into `cfg["logger"]["wandb"]` in `grpo_config.py`.
- Placed on `GRPOTraining` rather than the shared `IntegrationsSpec.wandb`, which is also used by automodel SFT where a Gym flag is meaningless.
- Layered on after the shared DPO/GRPO `_build_logger_config`, since it's GRPO-only.
- Validation: setting the flag without `integrations.wandb` is rejected at submit. NeMo-RL skips the tables entirely when W&B is off, so without this the flag is a silent no-op and the user goes looking for rollouts that were never logged.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Added an optional GRPO training setting to log complete NeMo-Gym rollout results as W&B Tables.
  - The setting is disabled by default and can be enabled for runs using W&B logging.

- **Bug Fixes**
  - Added validation to prevent enabling full result-table logging without a configured W&B integration.

- **Documentation**
  - Documented the new GRPO hyperparameter and its W&B integration requirement.


anubhutivyas:
closing because changes included in https://github.com/NVIDIA-NeMo/nemo-platform/pull/2093

PR #1742, opened 2026-09-02
feat(auth): enable deployment workload token exchange

ironcommit:
## TL;DR

Adds managed workload token exchange for deployments, so Docker and Kubernetes deployment workloads can call platform APIs using delegated creator context. This brings deployment-backed agents and model deployments onto the same workload identity path that already exists for managed jobs.

## Details

This PR generalizes workload delegation records beyond jobs with workload kind, workload id, workload generation, and optional logical claim metadata while preserving the legacy job delegation names. The shared delegation and workload identity helpers move into plugin-safe `nemo_platform_plugin.auth` modules, with compatibility re-exports from `nmp.common.auth` for existing callers.

Deployment configs now accept a `workloadIdentity` block, and deployment creation captures the request auth context from the runtime or trusted principal headers. Backends only activate managed workload identity when workload token exchange is enabled and an auth context is available; otherwise they fail the deployment with an explicit status message. User-supplied `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` values are rejected when platform-managed workload identity is enabled.

The Docker deployment backend provisions per-container opaque proof tokens into managed volumes, injects the workload token file environment, refreshes active delegations, and revokes or removes delegation state on startup failures, one-shot recreation, terminal statuses, missing containers, and delete. The Docker executor also gains configurable shared volume mounts and a pinned/configurable token-writer image for writing the managed token file.

The Kubernetes deployment and job-backed deployment paths project service-account tokens, select the effective service account from workload identity or backend config, and reconcile Pod UID-bound delegations only for verified live workload pods. Reconciliation skips unsafe cleanup when pod or ReplicaSet listing is unavailable, revokes stale pod delegations after confirmed live-pod changes, and revokes owned delegations during delete.

Agent deployments and model deployments now propagate captured auth context into the deployments plugin and request workload identity when platform workload token exchange is enabled. Existing managed job backends are refactored onto the same shared helpers while retaining their current token exchange behavior.

Authentik compose and Kubernetes fixtures now include the deployments service/controller surfaces and advertise managed workload deployment capabilities. The contract coverage adds a live deployment workload OBO flow, plus unit and static coverage for the new schemas, lifecycle cleanup, Kubernetes reconciliation, Authentik runtime configuration, and token exchange behavior.

No breaking changes or migrations are expected. Workload identity remains opt-in through `workloadIdentity` and the existing `auth.oidc.workload_token_exchange_enabled` platform switch.


plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -409,21 +476,39 @@ async def create_deployment(
...
-                name=sidecar_name,
-                labels={**base_labels, CONTAINER_ROLE_LABEL: sidecar.name},
-                host_ports={},
-                gpu_ids=[],
-                network=f"container:{c_name}",
-            )
+            sidecar_identity: tuple[str, str] | None = None
+            sidecar_labels = {**base_labels, CONTAINER_ROLE_LABEL: sidecar.name}
             try:
+                sidecar_identity = await self._prepare_workload_identity_for_container(
```
tylersbray: **Severity:** P1 (blocking)
**Found by:** Opus

Docker calls `_prepare_workload_identity_for_container` for every sidecar, including the platform auth-proxy that `build_docker_plan` appends in `backends/docker/containers.py`. Kubernetes compiles that sidecar with `include_workload_identity=False`. On the default agent path both `auth_proxy_sidecar` and `workloadIdentity` are set, so the proxy gets a second, unrelated creator OBO proof token.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py` around the sidecar loop (~477-506), do not mint or mount workload identity for the platform auth-proxy sidecar. Match Kubernetes `backends/k8s/compiler.py` (`include_workload_identity=False`). Keep identity for user sidecars. Add a Docker unit test that auth-proxy is in `plan.sidecars` but gets no volume/delegation when both auth-proxy and workload identity are enabled.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
```diff
@@ -160,6 +162,19 @@ class Affinity(BaseModel):
...
+class WorkloadIdentitySpec(BaseModel):
+    """Workload token-exchange settings for deployment backends."""
+
+    enabled: bool = False
+    workload_kind: str | None = Field(default=None, alias="workloadKind")
+    workload_id: str | None = Field(default=None, alias="workloadId")
+    token_audience: str | None = Field(default=None, alias="tokenAudience")
+    service_account_name: str | None = Field(default=None, alias="serviceAccountName")
+    token_expiration_seconds: int = Field(default=3600, ge=600, alias="tokenExpirationSeconds")
+
```
tylersbray, line 182: **Severity:** P1 (blocking)
**Found by:** Opus

`tokenExpirationSeconds` is `ge=600` with no ceiling. Docker uses this as `ttl_seconds_active` for the opaque proof token (the whole credential lifetime). Managed jobs use an operator-controlled execution-profile TTL instead. A DeploymentConfig author can set an unbounded TTL for an identity supplied later at Deployment create.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `WorkloadIdentitySpec.token_expiration_seconds` (`plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py` around 174), add a documented upper bound, or ignore user TTL for Docker and use an operator-controlled value like jobs. Keep `ge=600`. Add schema and Docker TTL-source tests.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/workload_identity.py
```diff
@@ -0,0 +1,83 @@
...
+DEFAULT_WORKLOAD_KIND = "deployment"
+
+
+def workload_identity_requested(config: DeploymentConfig) -> bool:
+    spec = config.workload_identity
+    return spec is not None and spec.enabled
+
+
+def workload_identity_activation_error(
+    *,
```
tylersbray, line 28: **Severity:** P1 (blocking)
**Found by:** Opus

`workload_identity_activation_error` is the fail-closed gate advertised in the PR, but it has no tests. New backend tests patch `is_workload_identity_token_exchange_enabled` to `True`, so neither the disabled-exchange branch nor missing `auth_context` is asserted, nor that create returns `FAILED` instead of proceeding.

<details>

Verify this finding against the current code and only fix it if it is still present.

Add unit tests that `workload_identity_activation_error` in `plugins/nemo-deployments/src/nemo_deployments_plugin/backends/workload_identity.py` returns the disabled-exchange and missing-auth_context messages. Assert Docker and Kubernetes create return `FAILED` when the gate fires. Do not patch the exchange flag to `True` for those cases.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/deployments.py
```diff
@@ -462,12 +586,34 @@ def _read() -> Any:
...
+                clients,
+                namespace=namespace,
+                resource_name=resource_name,
+                deployment=deployment,
+                config=config,
+                k8s_config=k8s_config,
+                pods=pods,
+            )
+            await reconcile_pod_uid_delegations(
+                workload_delegation_store,
```
tylersbray, line 621: **Severity:** P1 (blocking)
**Found by:** GPT-Sol

Distinct from CodeRabbit’s pod-label spoof comment: this path only requires `CONFIG_NAME_LABEL` before `reconcile_pod_uid_delegations`. Full `resource_labels_match` runs later in `status_from_deployment`. Job create/status compute `status_from_job` first but still reconcile even when that update is `FAILED` (foreign resource). A namespaced object occupying the expected name can receive creator OBO before identity-label failure.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `read_deployment_status` / `create_deployment` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/deployments.py`) and job create/status (`backends/k8s/jobs.py`), run full identity-label matching (`resource_labels_match` / `status_from_*`) before `reconcile_pod_uid_delegations`. Skip reconcile when the resource is foreign or identity-FAILED. Add a test: `CONFIG_NAME_LABEL` present, other identity labels wrong → no `store.register`.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
```diff
@@ -0,0 +1,278 @@
...
+    for pod in pods:
+        pod_uid = _pod_uid(pod)
+        if pod_uid is not None and _pod_is_live(pod):
+            live_pod_uid_set.add(pod_uid)
+    live_pod_uids = sorted(live_pod_uid_set)
+
+    for pod_uid in live_pod_uids:
+        current = existing_by_pod_uid.get(pod_uid)
+        if current is not None and current.is_active():
+            continue
```
tylersbray, line 132: **Severity:** P1 (blocking)
**Found by:** GPT-Sol

Reconciliation `continue`s every still-active Pod UID delegation and only registers a replacement after `is_active()` is false. Docker refreshes at half TTL. Long-running Kubernetes workloads therefore lose exchange for `tokenExpirationSeconds + 300s` until the next reconcile after expiry.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `reconcile_pod_uid_delegations` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py` around 106-120), refresh or replace still-active rows before expiry (mirror Docker half-TTL refresh with versioned `update`). Add a test that an active row near expiry is extended without waiting for `is_active()` to become false.

</details>

tylersbray (commented):
Triple-review (Opus / GPT-Sol / Auto) on `7c4fa8db18e8aeb4ef67e5068b09f44ea6868761`.

Skipped CodeRabbit duplicates: ReplicaSet→Deployment UID / name-prefix ownership (marked addressed), reserved `WORKLOAD_IDENTITY_VOLUME_NAME`, Docker/K8s revoke when config load returns `None` or custom `workload_kind` on the None path, Docker refresh writing every poll, and contract-test cleanup timeout / config-delete assertions.

Inline threads below are remaining distinct findings. Each includes an agent prompt.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
```diff
@@ -278,19 +304,81 @@ def _collect_pvc_mounts(config: DeploymentConfig) -> list[VolumeMount]:
...
+    spec = config.workload_identity
+    return spec is not None and spec.enabled
+
+
+def build_workload_identity_volume(config: DeploymentConfig) -> V1Volume:
+    if config.workload_identity is None:
+        raise DeploymentConfigError("workload_identity is required to build a workload identity volume")
+    k8s = k8s_client_module()
+    audience = config.workload_identity.token_audience or get_workload_identity_token_audience()
+    return k8s.client.V1Volume(
```
tylersbray, line 327: **Severity:** P1 (blocking)
**Found by:** Opus, GPT-Sol, Auto

`tokenAudience` is user-writable and projected as the SA token audience. TokenReview always uses platform `workload_client_id`/`client_id`, so a custom value cannot exchange. Kubernetes will still issue a bound SA JWT for that arbitrary audience, which other audience-trusting consumers may accept. Agent configs snapshot the platform audience, so the default path works.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `build_workload_identity_volume` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py` around 317) and `WorkloadIdentitySpec.token_audience`, either drop the user field, validate it equals `get_workload_identity_token_audience()` at config time, or pass the same audience into TokenReview. Fail closed at compile/validate. Add a mismatched-audience test.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
```diff
@@ -0,0 +1,278 @@
...
+    metadata = pod.metadata
+    owner_references = metadata.owner_references if metadata is not None else None
+    for owner_reference in owner_references or []:
+        if owner_reference.controller is not True:
+            continue
+        if owner_reference.kind != controller_kind:
+            continue
+        owner_name = owner_reference.name
+        if owner_name in controller_names:
+            return True
```
tylersbray, line 264: **Severity:** P1 (blocking)
**Found by:** Auto, Opus

Distinct from CodeRabbit’s ReplicaSet→Deployment UID / name-prefix issue (already addressed): `_pod_has_controller_owner` still matches controller **name** only, not UID. A Pod in a shared namespace can copy the live ReplicaSet/Job name, expected SA, and labels. Require `owner_reference.uid` to match the live controller UID. Apply to both Deployment (Pod→ReplicaSet) and Job (Pod→Job) filters.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `_pod_has_controller_owner` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py` around 217-232), require `owner_reference.uid` to equal the live controller UID (ReplicaSet UID set / Job UID). Keep the existing Deployment→ReplicaSet UID filter. Add tests for matching name + wrong UID on both Deployment and Job paths.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
```diff
@@ -0,0 +1,278 @@
...
+    *,
+    config: DeploymentConfig | None,
+    workspace: str,
+    deployment_name: str,
+) -> None:
+    """Revoke all delegation rows owned by one physical deployment workload."""
+    if store is None:
+        return
+    if config is not None and not workload_identity_requested(config):
+        return
```
tylersbray: **Severity:** P2 (non-blocking)
**Found by:** Opus

Distinct from CodeRabbit’s config=None / custom-kind cleanup: `revoke_workload_delegations` returns immediately when the **current** config no longer requests workload identity (`enabled=false` or block removed). Docker `_revoke_workload_delegations_for_config` does the same. Rows minted under a prior revision survive until user-controlled `expires_at`. Changing `workloadKind` has the same miss because scope is derived from the current config.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `revoke_workload_delegations` and Docker `_revoke_workload_delegations_for_config`, use the broad `config=None` lookup (no `workload_kind` filter) on delete and whenever the current config no longer requests identity, so prior kinds/rows are still revoked. Add tests for disable-WI and kind-change cleanup.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
```diff
@@ -473,6 +558,323 @@ async def _allocate_host_ports(
...
+        try:
+            now = datetime.now(UTC)
+            refresh_threshold_seconds = config.workload_identity.token_expiration_seconds / 2
+            delegations = await self._workload_delegations.list_by_workload(
+                workload_delegation_scope(workspace=workspace, deployment_name=name, config=config)
+            )
+            for delegation in delegations:
+                if delegation.revoked_at is not None or not delegation.opaque_subject_token_hash:
+                    continue
+                remaining_seconds = (delegation.expires_at - now).total_seconds()
```
tylersbray: **Severity:** P2 (non-blocking)
**Found by:** Opus, Auto

Two issues in this refresh (distinct from CodeRabbit’s already-addressed “write every poll”): (1) `expires_at - now` is not normalized with `_as_aware_utc`; a naive store round-trip raises `TypeError` swallowed by `except Exception`, stopping refresh. (2) refresh only slides `expires_at`; the opaque proof bytes and `opaque_subject_token_hash` are not rotated, so a stolen `nmp_obo_v1` token stays exchangeable for as long as the deployment stays non-terminal.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `_refresh_workload_delegations_for_config` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py` around 708-731): normalize `expires_at` with `_as_aware_utc` (or equivalent); do not let one bad row abort the loop; rotate the opaque proof and rewrite the volume (and hash) on refresh rather than only extending expiry. Add tests for naive datetime and for hash/token rotation.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/deployments.py
```diff
@@ -445,6 +565,10 @@ async def read_deployment_status(
     try:
         k8s_config = resolve_k8s_deployment_config(backend_config)
         namespace = resolve_deployment_namespace(default_namespace=default_namespace, k8s_config=k8s_config)
+        if config is not None:
+            identity_error = workload_identity_activation_error(config=config, auth_context=auth_context)
+            if identity_error is not None:
```
tylersbray: **Severity:** P2 (non-blocking)
**Found by:** Opus

Status-read applies `workload_identity_activation_error` and returns `FAILED` if a running Deployment’s config requests WI but the entity has no `auth_context` (pre-change deployments, or any path that skipped `with_auth_context`). Create-time fail-closed is correct; this turns a config edit into an outage. Same shape in `backends/k8s/jobs.py` ~395-398.

<details>

Verify this finding against the current code and only fix it if it is still present.

Keep fail-closed on **create**. On status-read in `read_deployment_status` / `read_job_status`, do not flip READY workloads to FAILED solely for missing `auth_context` when WI was newly enabled; skip identity reconcile or use a non-fatal message. Add a regression test for a READY deployment with WI enabled and `auth_context=None`.

</details>

packages/nemo_platform_plugin/src/nemo_platform_plugin/auth/__init__.py
```diff
@@ -13,10 +13,146 @@
...
+            principal_on_behalf_of_email=principal.on_behalf_of_email,
+        )
+
+    @classmethod
+    def from_runtime_or_headers(cls, headers: Mapping[str, str]) -> AuthContext | None:
+        """Create from the resolved auth runtime, falling back to trusted principal headers."""
+        runtime_context = current_auth_context()
+        if runtime_context is not None:
+            return runtime_context
+        return cls.from_headers(headers)
```
tylersbray: **Severity:** P2 (non-blocking)
**Found by:** Opus

`from_runtime_or_headers` falls back to raw `x-nmp-principal-*` headers. Deployments/agents APIs use this for delegation-minting input. Jobs use `AuthContext.from_principal(auth_client.principal)`. In-platform middleware usually sets runtime context first; the fallback still widens the trust boundary for a minting input.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `AuthContext.from_runtime_or_headers` and the deployments API call sites (`plugins/nemo-deployments/.../api/v2/deployments.py`, `plugins/nemo-agents/.../api/v2/deployments.py`), mint only from `current_auth_context()` / `from_principal`. Do not build `AuthContext` from raw headers when runtime context is missing. Preserve existing header stripping at the gateway.

</details>

packages/nemo_platform_plugin/src/nemo_platform_plugin/auth/workload_identity.py
```diff
@@ -0,0 +1,240 @@
...
+
+def get_workload_delegation_audience() -> str:
+    """Return the RFC 8693 audience used when exchanging workload subject tokens."""
+    try:
+        from nmp.common.config import get_auth_config
+
+        oidc = get_auth_config().oidc
+        return oidc.workload_audience or oidc.audience or DEFAULT_WORKLOAD_AUDIENCE
+    except Exception:
+        logger.debug("Could not resolve auth config for workload delegation audience", exc_info=True)
```
tylersbray, line 87: **Severity:** P2 (non-blocking)
**Found by:** Opus

`get_workload_delegation_audience` catches bare `Exception` and logs at debug (jobs previously warned on `(RuntimeError, ValidationError)`). Silent default audience mints unredeemable rows. Config-resolution errors also surface through the activation gate as “token exchange disabled.”

<details>

Verify this finding against the current code and only fix it if it is still present.

In `get_workload_delegation_audience` (`packages/nemo_platform_plugin/src/nemo_platform_plugin/auth/workload_identity.py` around 67-76), restore a narrow catch and a warning log. Do not map config-resolution failure to the “token exchange disabled” activation message without a distinct error. Add a test for the fallback path.

</details>

packages/nemo_platform/pyproject.toml
```diff
@@ -392,8 +392,8 @@ nemo-guardrails-plugin = [
 nemo-insights-plugin = [
   "fastapi>=0.115",
   "httpx",
-  "nemo-fabric-adapter-contract==0.3.0b1",
-  "nemo-fabric-adapters-common==0.3.0b1",
+  "nemo-fabric-adapter-contract>=0.3.0b1",
```
tylersbray, line 395: **Severity:** P2 (non-blocking)
**Found by:** Opus

Unrelated to workload token exchange: `nemo-fabric-adapter-contract` and `nemo-fabric-adapters-common` move from `==0.3.0b1` to `>=0.3.0b1` (pre-release admits later majors). Sibling `nemo-fabric-adapters-deepagents` stays pinned.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `packages/nemo_platform/pyproject.toml` nemo-insights extras, revert the fabric-adapter `>=0.3.0b1` unpin unless it is required and justified in a separate change. Keep `uv.lock` consistent with the revert.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
```diff
@@ -174,11 +189,22 @@ def _validate_port_names(config: DeploymentConfig) -> None:
             seen_ports.add(port_key)
 
 
+def _validate_workload_identity_volume_mount_names(config: DeploymentConfig) -> None:
+    if not workload_identity_enabled(config):
+        return
+    for mount in _collect_pvc_mounts(config):
+        if mount.name == WORKLOAD_IDENTITY_VOLUME_NAME:
+            raise DeploymentConfigError(
```
tylersbray, line 197: **Severity:** P3 (non-blocking)
**Found by:** Opus

Distinct from CodeRabbit’s reserved **volume name** (already addressed): `_validate_workload_identity_volume_mount_names` still only compares `mount.name` to `WORKLOAD_IDENTITY_VOLUME_NAME`. A user mount with another name and `mountPath: /var/run/secrets/nemo-platform/workload` collides with the projected token mount and fails as an opaque Kubernetes validation error.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `_validate_workload_identity_volume_mount_names` (`plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py` around 193-200), also reject mounts whose `mountPath` equals `WORKLOAD_IDENTITY_VOLUME_PATH`. Raise `DeploymentConfigError`. Add a unit test for path collision with a different volume name.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
```diff
@@ -295,6 +296,7 @@ async def create_deployment(
         config_name: str,
         labels: dict[str, str],
         backend_config: dict[str, Any],
+        auth_context: AuthContext | None = None,
```
tylersbray, line 303: **Severity:** P2 (non-blocking)
**Found by:** Auto

OpenShell `create_deployment` accepts `auth_context` for the backend contract but never calls `workload_identity_requested` / `workload_identity_activation_error`. A WI-enabled config can go READY with no subject token and no explicit FAILED status, unlike Docker/K8s.

<details>

Verify this finding against the current code and only fix it if it is still present.

In `plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py` `create_deployment`, if workload identity is requested, fail closed with `workload_identity_activation_error` (same messages as Docker/K8s). Do not silently ignore `auth_context`. Add a unit test that WI-enabled OpenShell create returns FAILED.

</details>

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
ironcommit: Resolved in current PR head `20af7966`.

Docker sidecars now skip workload-identity setup when `is_auth_proxy_container(sidecar)` is true. Coverage: `test_create_deployment_does_not_mount_workload_identity_on_auth_proxy_sidecar`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/deployments.py
ironcommit: Resolved in current PR head `20af7966`.

K8s deployment/job create and status now compute the status update and check full `resource_labels_match(...)` before pod-UID reconciliation. Foreign identity-label paths return before listing pods or registering delegations. Coverage includes `test_read_deployment_status_skips_workload_delegation_for_foreign_identity_labels` and `test_read_job_status_skips_workload_delegation_for_foreign_identity_labels`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/entities.py
ironcommit: Resolved in current PR head `20af7966`.

`WorkloadIdentitySpec.token_expiration_seconds` now has the pydantic upper bound `le=WORKLOAD_IDENTITY_TOKEN_MAX_EXPIRATION_SECONDS`. Coverage: `test_workload_identity_spec_caps_token_expiration_seconds`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/workload_identity.py
ironcommit: Resolved in current PR head `20af7966`.

The activation gate now has unit coverage for disabled exchange and missing auth context: `test_workload_identity_activation_error_when_exchange_disabled` and `test_workload_identity_activation_error_when_auth_context_missing`. Backend create paths fail closed before workload setup; K8s deployment/job coverage asserts FAILED on missing auth context. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
ironcommit: Resolved in current PR head `20af7966`.

K8s pod-UID reconciliation now refreshes active delegations before expiry when remaining lifetime is at or below half the configured TTL, using a versioned store update. Coverage: `test_read_deployment_status_refreshes_active_pod_uid_before_expiry`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
ironcommit: Resolved in current PR head `20af7966`.

`_pod_has_controller_owner` now requires the Pod controller owner reference UID to match the live controller UID for both Deployment ReplicaSets and Jobs. Coverage includes `test_create_deployment_ignores_pod_with_matching_replicaset_name_wrong_uid` and `test_read_job_status_ignores_pod_with_matching_job_name_wrong_uid`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
ironcommit: Resolved in current PR head `20af7966`.

`build_workload_identity_volume` now resolves the platform projected-token audience and rejects a configured `workload_identity.tokenAudience` that differs from it. Coverage: `test_compile_workload_rejects_mismatched_workload_identity_token_audience`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/deployments.py
ironcommit: Resolved in current PR head `20af7966`.

K8s status reads no longer mark an existing workload FAILED just because the status read lacks `auth_context`; they skip workload-identity reconciliation when activation is not allowed and preserve the Kubernetes-derived status. Coverage: `test_read_deployment_status_with_missing_auth_context_keeps_ready_status` and `test_read_job_status_with_missing_auth_context_keeps_running_status`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/docker/backend.py
ironcommit: Resolved in current PR head `20af7966`.

Docker refresh now normalizes `expires_at` with `as_aware_utc`, isolates per-row refresh failures, and rotates the opaque proof token plus `opaque_subject_token_hash` before writing the replacement token. Coverage: `test_refresh_workload_identity_delegation_extends_near_expiry` and `test_refresh_workload_identity_delegation_handles_naive_expiry`. CI is green at `20af7966`.

packages/nemo_platform_plugin/src/nemo_platform_plugin/auth/__init__.py
ironcommit: Resolved in current PR head `20af7966`.

`AuthContext.from_runtime_or_headers` now ignores raw principal headers and only returns the platform runtime auth context. Coverage: `test_auth_context_from_runtime_or_headers_ignores_raw_headers` and `test_auth_context_from_runtime_or_headers_uses_platform_runtime_context`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/workload_identity.py
ironcommit: Resolved in current PR head `20af7966`.

K8s revocation now ignores the current config shape and uses the deployment-wide `config=None` workload scope, so disabling workload identity or changing kind still revokes prior rows. Coverage: `test_revoke_workload_delegations_uses_broad_scope_when_disabled` and `test_revoke_workload_delegations_uses_broad_scope_after_workload_kind_change`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/k8s/compiler.py
ironcommit: Resolved in current PR head `20af7966`.

K8s workload config validation now rejects PVC volume mounts using the reserved workload-identity mount path even when the volume name differs. Coverage: `test_compile_workload_rejects_reserved_workload_identity_volume_mount_path`. CI is green at `20af7966`.

packages/nemo_platform_plugin/src/nemo_platform_plugin/auth/workload_identity.py
ironcommit: Resolved in current PR head `20af7966`.

`get_workload_delegation_audience` now catches only `ImportError`, `RuntimeError`, and `ValidationError`, with config-resolution failures logged at warning level. The activation gate also surfaces `WorkloadIdentityConfigError` distinctly for token-exchange config resolution failures. Coverage: `test_get_workload_delegation_audience_warns_and_defaults_on_config_error` and `test_workload_identity_token_exchange_enabled_raises_distinct_config_error`. CI is green at `20af7966`.

plugins/nemo-deployments/src/nemo_deployments_plugin/backends/openshell/backend.py
ironcommit: Resolved in current PR head `20af7966`.

OpenShell create now fails closed when workload identity is requested, returning FAILED before sandbox creation because the backend does not support injecting the subject token. Coverage: `test_create_fails_when_workload_identity_requested`. CI is green at `20af7966`.

tylersbray (approved):
🤌

PR #1744, opened 2026-09-02
fix(agents): prevent unbounded fabric runtime for stateless invocations

mmogallapalli:
## Summary

Prevents unbounded Fabric runtime-session growth for clients that invoke a
deployed agent without `X-Nemo-Session-Id`. Headerless requests now use an
ephemeral one-shot runtime that is stopped when the response or response stream
completes, while requests with a session header retain the existing registered
session lifecycle.

This covers evaluation and security-testing clients that send independent
headerless requests.

## Related Issue

[AIRCORE-1108](https://linear.app/nvidia/issue/AIRCORE-1108)

Related prior approach: [NVIDIA-NeMo/nemo-platform#1240](https://github.com/NVIDIA-NeMo/nemo-platform/pull/1240)

## Changes

- Route requests without `X-Nemo-Session-Id` through one-shot Fabric invocation
  instead of creating a registered runtime session.
- Keep requests with a session header on the existing lazy session creation,
  reuse, idle expiration, explicit closure, and shutdown paths.
- Add non-streaming and streaming one-shot helpers that start an ephemeral
  runtime and guarantee cleanup after completion, failure, timeout, cancellation,
  or client disconnect.
- Keep streaming runtimes alive until response delivery finishes and close the
  stream when delivery ends early.
- Shield streaming cleanup from request cancellation and bound the response's
  cleanup wait without limiting invocation or stream duration.
- Classify runtime context-entry failures as startup errors while preserving
  cleanup error classification for failures after startup.
- Apply the existing invocation concurrency limit to both stateful and one-shot
  work, including one-shot runtime construction and cleanup.
- Preserve the current OpenAI-compatible request, response, streaming, and error
  contracts, including the existing Studio chat behavior.
- Document the session-header lifecycle contract for direct API clients.
- Add regression coverage for routing, runtime isolation, concurrency, streaming
  lifetime, disconnects, direct cancellation, delayed cleanup failures, error
  mapping, and exact-once cleanup.

## Design Decisions

- **Use header presence as the lifecycle boundary.** A missing session header
  means the caller did not request persistent runtime state, so the server does
  not generate or register an implicit session. A supplied header continues to
  identify a Platform-managed runtime session.
- **Reuse Fabric's one-shot lifecycle.** Stateless requests use the existing
  runtime start/invoke/stop primitives rather than introducing a second runtime
  registry or manual harness-state reset behavior.
- **Tie streaming cleanup to response delivery.** The ephemeral runtime remains
  available while the response iterator is active and is closed on normal
  completion, error, cancellation, or client disconnect. Response teardown waits
  up to five seconds for cleanup; if cleanup takes longer, it continues in a
  shielded background task without shortening the invocation itself.
- **Preserve execution timeout semantics.** Runtime startup, context entry, and
  cleanup do not consume the caller's invocation or result-waiting budget.
- **Share admission control.** Stateful and one-shot invocations use the same
  concurrency limit so headerless traffic cannot bypass the deployment's
  existing capacity bound.
- **Keep deliberate sessions lifecycle-managed.** This PR does not add a global
  or per-deployment session cap; explicitly created Platform sessions remain
  governed by the existing session lifecycle.
- **Preserve Studio compatibility.** This PR does not change transcript shaping
  or Studio session ownership. Removing transcript replay remains scoped to a
  follow-up PR.

## Type of Change

- [ ] Code change (refactor)
- [ ] Documentation only

## Quality Gates

- [x] Stateful session behavior remains covered

## Verification

- [x] Every existing commit includes an appropriate `Signed-off-by:` trailer

Targeted validation:

- Agents plugin unit suite — 1,618 passed
- Focused Fabric runtime, server, and session-manager suite — 95 passed
- `ty` on all changed Python source and test files — passed
- Ruff lint and formatting checks on all changed Python files — passed
- `git diff --check` — passed

## Summary by CodeRabbit

* **New Features**
  * Added support for stateless one-shot and streaming requests without a session ID.
  * Session-backed requests continue reusing registered runtimes according to the session lifecycle.
  * Streaming responses now clean up resources when completed, interrupted, disconnected, or cancelled.

* **Bug Fixes**
  * Runtime startup no longer reduces the available invocation timeout.
  * Runtime-start failures now return HTTP 503 responses.
  * Improved handling and logging of streaming cleanup failures without masking client disconnects.

* **Documentation**
  * Documented stateless and session-backed runtime behavior.

### DIFF
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
@@ -15,12 +15,13 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import AsyncIterator, Mapping, Sequence
+from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
 # CI type-checks this plugin via ty extra-paths without installing nemo-agents deps.
-from nemo_fabric import (  # ty: ignore[unresolved-import]
+from nemo_fabric import (
     Fabric,
     FabricConfig,
     FabricError,
@@ -200,6 +201,37 @@ async def run_fabric_agent_once(
         ) from error
 
     return _normalize_fabric_run_result(result)
+
+
+@asynccontextmanager
+async def stream_fabric_agent_once(
+    request: FabricOneShotRequest,
+    *,
+    fabric: Any | None = None,
+) -> AsyncIterator[FabricRuntimeStream]:
+    """Start an ephemeral Fabric runtime and keep it alive for one stream."""
+    fabric_client = fabric or Fabric()
+
+    try:
+        async with await fabric_client.start_runtime(
+            request.fabric_config,
+            base_dir=request.base_dir,
+            overrides=request.overrides,
+            streaming=True,
+        ) as runtime:
+            yield stream_fabric_runtime(
+                runtime,
+                FabricInvocationRequest(
+                    input=request.input,
+                    request_id=request.request_id,
+                    caller_context=request.caller_context,
+                    timeout_seconds=request.timeout_seconds,
+                ),
+            )
+    except FabricError as error:
+        raise FabricRuntimeExecutionError(
+            f"Fabric runtime streaming failed: {error}",
+        ) from error
 
 
 async def _invoke_fabric_agent_once(
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
@@ -189,15 +189,26 @@ async def run_fabric_agent_once(
 ) -> FabricRuntimeResult:
     """Start an ephemeral Fabric runtime, invoke it once, and stop it."""
     fabric_client = fabric or Fabric()
-
-    runtime = await _start_one_shot_runtime(request, fabric=fabric_client)
+    deadline = None if request.timeout_seconds is None else asyncio.get_running_loop().time() + request.timeout_seconds
 
     try:
+        runtime = await asyncio.wait_for(
+            _start_one_shot_runtime(request, fabric=fabric_client),
+            timeout=_remaining_timeout(deadline),
+        )
+    except TimeoutError as error:
+        raise FabricRuntimeTimeoutError(
+            _timeout_error_message(request.timeout_seconds),
+        ) from error
+
+    runtime_entered = False
+    try:
         async with runtime:
+            runtime_entered = True
             try:
                 result = await asyncio.wait_for(
                     runtime.invoke(request=_with_platform_invocation_context(request)),
-                    timeout=request.timeout_seconds,
+                    timeout=_remaining_timeout(deadline),
                 )
             except TimeoutError as error:
                 raise FabricRuntimeTimeoutError(
@@ -208,9 +219,15 @@ async def run_fabric_agent_once(
                     f"Fabric runtime invocation failed: {error}",
                 ) from error
     except FabricError as error:
-        raise FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}") from error
+        raise _runtime_context_error(error, entered=runtime_entered) from error
 
     return _normalize_fabric_run_result(result)
+
+
+def _remaining_timeout(deadline: float | None) -> float | None:
+    if deadline is None:
+        return None
+    return max(0.0, deadline - asyncio.get_running_loop().time())
 
 
 @asynccontextmanager
@@ -224,8 +241,10 @@ async def stream_fabric_agent_once(
 
     runtime = await _start_one_shot_runtime(request, fabric=fabric_client, streaming=True)
 
+    runtime_entered = False
     try:
         async with runtime:
+            runtime_entered = True
             yield stream_fabric_runtime(
                 runtime,
                 FabricInvocationRequest(
@@ -236,7 +255,13 @@ async def stream_fabric_agent_once(
                 ),
             )
     except FabricError as error:
-        raise FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}") from error
+        raise _runtime_context_error(error, entered=runtime_entered) from error
+
+
+def _runtime_context_error(error: FabricError, *, entered: bool) -> FabricRuntimeExecutionError:
+    if entered:
+        return FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}")
+    return FabricRuntimeStartError(f"Fabric runtime startup failed: {error}")
 
 
 async def _start_one_shot_runtime(
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import AsyncIterator, Mapping, Sequence
-from contextlib import asynccontextmanager
+from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
@@ -189,27 +189,20 @@ async def run_fabric_agent_once(
 ) -> FabricRuntimeResult:
     """Start an ephemeral Fabric runtime, invoke it once, and stop it."""
     fabric_client = fabric or Fabric()
-    runtime = await _start_one_shot_runtime(request, fabric=fabric_client)
-
-    runtime_entered = False
-    try:
-        async with runtime:
-            runtime_entered = True
-            try:
-                result = await asyncio.wait_for(
-                    runtime.invoke(request=_with_platform_invocation_context(request)),
-                    timeout=request.timeout_seconds,
-                )
-            except TimeoutError as error:
-                raise FabricRuntimeTimeoutError(
-                    _timeout_error_message(request.timeout_seconds),
-                ) from error
-            except FabricError as error:
-                raise FabricRuntimeExecutionError(
-                    f"Fabric runtime invocation failed: {error}",
-                ) from error
-    except FabricError as error:
-        raise _runtime_context_error(error, entered=runtime_entered) from error
+    async with _one_shot_runtime(request, fabric=fabric_client) as runtime:
+        try:
+            result = await asyncio.wait_for(
+                runtime.invoke(request=_with_platform_invocation_context(request)),
+                timeout=request.timeout_seconds,
+            )
+        except TimeoutError as error:
+            raise FabricRuntimeTimeoutError(
+                _timeout_error_message(request.timeout_seconds),
+            ) from error
+        except FabricError as error:
+            raise FabricRuntimeExecutionError(
+                f"Fabric runtime invocation failed: {error}",
+            ) from error
 
     return _normalize_fabric_run_result(result)
 
@@ -222,47 +215,40 @@ async def stream_fabric_agent_once(
 ) -> AsyncIterator[FabricRuntimeStream]:
     """Start an ephemeral Fabric runtime and keep it alive for one stream."""
     fabric_client = fabric or Fabric()
-
-    runtime = await _start_one_shot_runtime(request, fabric=fabric_client, streaming=True)
-
-    runtime_entered = False
-    try:
-        async with runtime:
-            runtime_entered = True
-            yield stream_fabric_runtime(
-                runtime,
-                FabricInvocationRequest(
-                    input=request.input,
-                    request_id=request.request_id,
-                    caller_context=request.caller_context,
-                    timeout_seconds=request.timeout_seconds,
-                ),
-            )
-    except FabricError as error:
-        raise _runtime_context_error(error, entered=runtime_entered) from error
+    async with _one_shot_runtime(request, fabric=fabric_client, streaming=True) as runtime:
+        yield stream_fabric_runtime(
+            runtime,
+            FabricInvocationRequest(
+                input=request.input,
+                request_id=request.request_id,
+                caller_context=request.caller_context,
+                timeout_seconds=request.timeout_seconds,
+            ),
+        )
 
 
-def _runtime_context_error(error: FabricError, *, entered: bool) -> FabricRuntimeExecutionError:
-    if entered:
-        return FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}")
-    return FabricRuntimeStartError(f"Fabric runtime startup failed: {error}")
-
-
-async def _start_one_shot_runtime(
+@asynccontextmanager
+async def _one_shot_runtime(
     request: FabricOneShotRequest,
     *,
     fabric: Any,
     streaming: bool = False,
-) -> Runtime:
+) -> AsyncIterator[Runtime]:
     try:
-        return await fabric.start_runtime(
-            request.fabric_config,
-            base_dir=request.base_dir,
-            overrides=request.overrides,
-            streaming=streaming,
-        )
+        async with AsyncExitStack() as stack:
+            try:
+                runtime = await fabric.start_runtime(
+                    request.fabric_config,
+                    base_dir=request.base_dir,
+                    overrides=request.overrides,
+                    streaming=streaming,
+                )
+                runtime = await stack.enter_async_context(runtime)
+            except FabricError as error:
+                raise FabricRuntimeStartError(f"Fabric runtime startup failed: {error}") from error
+            yield runtime
     except FabricError as error:
-        raise FabricRuntimeStartError(f"Fabric runtime startup failed: {error}") from error
+        raise FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}") from error
 
 
 def _with_platform_invocation_context(request: FabricInvocationRequest | FabricOneShotRequest) -> RunRequest:
--- a/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
+++ b/plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
@@ -132,6 +132,10 @@ class FabricRuntimeExecutionError(RuntimeError):
     """Raised when Fabric cannot return a normalized runtime result."""
 
 
+class FabricRuntimeStartError(FabricRuntimeExecutionError):
+    """Raised when Fabric cannot start a runtime."""
+
+
 class FabricRuntimeTimeoutError(FabricRuntimeExecutionError):
     """Raised when a Fabric runtime invocation times out."""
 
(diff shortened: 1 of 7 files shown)


plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
```diff
@@ -184,35 +189,96 @@ async def run_fabric_agent_once(
...
+            timeout=_remaining_timeout(deadline),
         )
     except TimeoutError as error:
         raise FabricRuntimeTimeoutError(
             _timeout_error_message(request.timeout_seconds),
         ) from error
+
+    runtime_entered = False
+    try:
+        async with runtime:
```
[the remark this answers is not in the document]
mmogallapalli: Reviewed this but will not be adopting a single deadline across runtime startup, context entry, and execution.

For Jobs, `timeout_seconds` represents the invocation/result-waiting budget. Deducting runtime setup time from that budget would be a behavioral regression—for example, five minutes of startup would leave only 55 minutes of a configured one-hour execution timeout.

The one-shot helpers therefore apply the full configured timeout to `invoke()` or the streaming result, while runtime startup, context entry, and cleanup remain outside that budget. This does not introduce a universal runtime lifetime limit, and cleanup is still guaranteed by the runtime context. Streaming response cleanup is handled separately with cancellation shielding and a bounded five-second teardown wait.

Added non-streaming and streaming regression tests confirming that delayed startup and context entry do not reduce the configured execution budget.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
```diff
@@ -184,35 +189,80 @@ async def run_fabric_agent_once(
...
+) -> AsyncIterator[FabricRuntimeStream]:
+    """Start an ephemeral Fabric runtime and keep it alive for one stream."""
+    fabric_client = fabric or Fabric()
+
+    runtime = await _start_one_shot_runtime(request, fabric=fabric_client, streaming=True)
+
+    runtime_entered = False
+    try:
+        async with runtime:
+            runtime_entered = True
```
mikeknep: style nit: I think this could read a little friendlier if you made `_start_one_shot_runtime` an asynccontextmanager? So that this call site would look something like:
```python
async with _one_shot_runtime(request, fabric=fabric_client, streaming=True) as runtime:
    yield stream_fabric_runtime(...)
```
And same in `run_fabric_agent_once`. Mainly gets rid of the boilerplate noise of `runtime_entered`, and also I think the `try` blocks become a little simpler to follow... or maybe just different, your call. The helper would look something like this (sketched with Claude)
```python
  @asynccontextmanager
  async def _one_shot_runtime(
      request: FabricOneShotRequest,
      *,
      fabric: Any,
      streaming: bool = False,
  ) -> AsyncIterator[Runtime]:
      try:
          async with AsyncExitStack() as stack:
              try:
                  runtime = await fabric.start_runtime(
                      request.fabric_config,
                      base_dir=request.base_dir,
                      overrides=request.overrides,
                      streaming=streaming,
                  )
                  await stack.enter_async_context(runtime)
              except FabricError as error:
                  raise FabricRuntimeStartError(f"Fabric runtime startup failed: {error}") from error
              yield runtime
      except FabricError as error:
          raise FabricRuntimeExecutionError(f"Fabric runtime cleanup failed: {error}") from error
```

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
```diff
@@ -305,67 +379,81 @@ async def chat_completions(
...
+            try:
+                session = await app.state.session_manager.resolve_session(session_id)
+            except FabricSessionNotFoundError as error:
+                raise HTTPException(status_code=404, detail=str(error)) from error
+            except FabricSessionStartError as error:
+                raise HTTPException(status_code=503, detail=str(error)) from error
+            invocation_request = _to_fabric_invocation_request(request, session_id=session.session_id)
+            response_headers = _session_headers(session.session_id)
+            invoke = partial(app.state.session_manager.invoke_session, session)
+            stream = partial(app.state.session_manager.stream_session, session)
```
mikeknep, line 397: I'm torn on this; part of me feels like the `session_manager` module should continue exposing only a single pair of public methods (invoke, stream) and those methods should accept an optional session and understand how to handle it being None, as opposed to this change where `session_manager` exposes two pairs and the caller has to use the correct one. Maybe because it seems weird for a module named `session_manager` to expose a function like `invoke_once` that doesn't do anything with a session at all?

I don't feel strongly about it, but figured I'd share the thought—you can take it or leave it.

mikeknep (approved):
Two non-blocking style thoughts

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
```diff
@@ -243,6 +252,38 @@ async def _stop_session(self, session: FabricRuntimeSession) -> None:
             except FabricError as error:
                 raise FabricSessionStopError(f"Fabric runtime shutdown failed: {error}") from error
 
+    async def _materialize_fabric_config(self, *, streaming: bool) -> FabricConfig:
+        try:
+            fabric_config = translate_agent_config(self._agent_config)
+        except FabricTranslationError as error:
+            raise FabricRuntimeStartError(f"Fabric config translation failed: {error}") from error
+
+        await asyncio.to_thread(ensure_local_workspace_dir, self._agent_config, self._base_dir)
```
benmccown, line 261: Would it be tricky to manage ephemeral subdirs keyed on session ID? Could be a straight forward fix? Probs worth capturing as a follow up issue, not in scope for this PR.

---

**Concurrent one-shot runtimes share one workspace dir**
*Priority: P3 · non-blocking · `session_manager.py:261` · found by deepseek-flash*

#### Finding: Agent Summary

Every headerless one-shot invocation builds its ephemeral runtime against the *same* shared `base_dir`/workspace. With the concurrency limit at its default of 8, that means up to eight one-shot runtimes can be reading and writing the same workspace directory at the same time.

This isn't a new class of problem — registered sessions already share `base_dir` the same way — but one-shot invocations add more concurrent writers to that shared space.

- **What:** all one-shot runtimes share one workspace directory, no per-invocation isolation
- **Where:** `_materialize_fabric_config` / `_to_one_shot_request` in `session_manager.py`
- **When:** two or more headerless requests run concurrently and touch the same workspace paths
- **Why it matters:** concurrent writes to the same files could collide under load, depending on whether Fabric namespaces per-runtime writes

#### Finding: Agent Breakdown

`_materialize_fabric_config` re-runs `ensure_local_workspace_dir(self._agent_config, self._base_dir)` for every one-shot, and `_to_one_shot_request` builds `FabricOneShotRequest(base_dir=self._base_dir, ...)` — one shared directory for all of them.

Worth noting this is a *non-destructive* `mkdir(parents=True, exist_ok=True)`, **not** `invocation.py`'s `_reset_directory` (which `rmtree`s the workspace). So there's no clobber-on-create; the concern is purely concurrent file *writes* into the same tree stepping on each other.

Whether that's actually safe comes down to one thing outside this diff: does Fabric namespace each runtime's writes (e.g. per-runtime subdirs)? If yes, this is a non-issue. If no, one-shot concurrency under load could corrupt workspace files — though that latent risk already existed for registered sessions.

Feels like a follow-up item (per-session-id ephemeral subdirs) rather than in-scope for this PR.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
```diff
@@ -184,35 +189,80 @@ async def run_fabric_agent_once(
 ) -> FabricRuntimeResult:
     """Start an ephemeral Fabric runtime, invoke it once, and stop it."""
     fabric_client = fabric or Fabric()
+    runtime = await _start_one_shot_runtime(request, fabric=fabric_client)
```
benmccown: potential follow up item to flag. If I'm tracking correctly, the least common denominator will just be the http timeouts between the client and server, and those timeouts will still result in proper resource cleanup, but if we ever did change the `timeout=None` fabric default that we aren't overriding, then the streaming path won't properly honor it either. If we ever do make a change and introduce a timeout value to the streaming logic branch, this other mentioned latent bug/finding will manifest.

---

**No server-side invocation timeout, and streaming wouldn't honor one if added**
*Priority: P2 · non-blocking · `runtime.py:192` · found by gpt-sol, gpt-codex, sonnet*

#### Finding: Agent Summary

The chat path never sets `timeout_seconds`, so there is no server-side runtime deadline today — a hung invocation waits until the client disconnects. That's fine as-is (HTTP timeouts bound it, and disconnect cleanup still runs).

The catch is latent: if a timeout is ever wired in, the streaming path wouldn't fully honor it, and a streaming start-timeout would return the wrong HTTP status.

- **What:** no invocation timeout is set; streaming timeout coverage is incomplete
- **Where:** `_to_fabric_invocation_request` (`server.py:78`) + `stream_fabric_agent_once` / `FabricRuntimeStream.records()` in `runtime.py`
- **When:** only matters the day someone overrides the `timeout=None` default
- **Why it matters:** a future timeout would silently not apply to streaming startup / mid-stream, and a start-timeout would map to 502 instead of 504

#### Finding: Agent Breakdown

`_to_fabric_invocation_request` builds the request without `timeout_seconds`, so it's `None` end-to-end for both stateful and one-shot chat. `asyncio.wait_for(..., timeout=None)` is a no-op, so nothing bounds the invocation server-side.

That's not a leak today: the client/server HTTP timeouts are the effective bound, and disconnect cleanup tears the runtime down properly — exactly as you noted.

Two things would bite if a real timeout is introduced later:

1. **Startup / mid-stream aren't bounded.** `run_fabric_agent_once` and `stream_fabric_agent_once` both `await _start_one_shot_runtime(...)` with no `wait_for`, and `FabricRuntimeStream.records()` iterates with a plain `async for`. Only the terminal `invoke`/`result()` applies the deadline — so a hung startup or a stalled stream would never time out.

2. **Wrong status on a streaming start-timeout.** The streaming-start `except` block (`server.py:402-412`) has no `FabricRuntimeTimeoutError` branch. Since that error subclasses `FabricRuntimeExecutionError`, it falls into the 502 branch — whereas the non-streaming path explicitly returns 504 (`server.py:441`).

Both are latent (dormant while `timeout` is `None`) and largely pre-existing, so this reads as a follow-up flag rather than a blocker.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
```diff
@@ -184,35 +189,80 @@ async def run_fabric_agent_once(
...
+            try:
+                result = await asyncio.wait_for(
+                    runtime.invoke(request=_with_platform_invocation_context(request)),
+                    timeout=request.timeout_seconds,
+                )
+            except TimeoutError as error:
+                raise FabricRuntimeTimeoutError(
+                    _timeout_error_message(request.timeout_seconds),
+                ) from error
+            except FabricError as error:
```
benmccown: Unlikely edge case but in the event it happens, the cryptic error without helpful details would be kind of a pain.

---

**Unmapped errors return a bare 500 with no detail**
*Priority: P3 · non-blocking · `runtime.py:207` · found by deepseek-flash*

#### Finding: Agent Summary

Some startup/invoke failures skip the error-mapping entirely and reach the client as a blank `500 Internal Server Error` with nothing useful in the body.

The real error message only lands in the server logs, so an operator hitting this has to go digging rather than reading the response.

- **What:** a non-`FabricError`/non-`TimeoutError` exception isn't translated into a proper HTTP status + detail
- **Where:** `run_fabric_agent_once` in `runtime.py`, surfacing through `chat_completions`
- **When:** e.g. an `OSError` while creating the workspace, or a plain `RuntimeError` from the adapter
- **Why it matters:** the client sees a cryptic 500 with no message; diagnosing it means digging through server logs

#### Finding: Agent Breakdown

The one-shot helpers only catch two exception types — `TimeoutError` and `FabricError` (in `run_fabric_agent_once` and `invoke_fabric_runtime`).

But two things on that path can raise *other* types. `_materialize_fabric_config` → `ensure_local_workspace_dir` can raise a plain `OSError` (disk full, permission denied), and `runtime.invoke()` could raise a non-Fabric `RuntimeError` from the adapter.

Those aren't `FabricError`, so they slip past every `HTTPException` mapping and propagate all the way up to FastAPI.

Because the app registers no custom exception handler and runs with no debug mode, Starlette falls back to a bare `Internal Server Error` **text body** — the actual message (e.g. `[Errno 28] No space left on device`) goes *only* to the server logs, never the response. Compare the mapped path, which returns a helpful `{"detail": "..."}`.

No resource leak here (nothing was started, or the `async with` still stops the runtime) — it's purely an observability gap, on an uncommon path, and it's pre-existing (the stateful path shares it).

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
```diff
@@ -179,19 +188,70 @@ async def __anext__(self) -> str:
...
+            self._close_deadline = asyncio.get_running_loop().time() + _FABRIC_STREAM_CLEANUP_TIMEOUT_SECONDS
+        return self._close_task
+
+    def _observe_background_cleanup(self, close_task: asyncio.Task[None]) -> None:
+        if self._close_observer_added:
+            return
+        self._close_observer_added = True
+        close_task.add_done_callback(self._log_background_cleanup_result)
+        logger.warning(
+            "Timed out waiting for Fabric stream cleanup after %gs; cleanup will continue in the background.",
```
benmccown, line 227: The shutdown race isn't super concerning, only undesirable side effect would likely be some noise in the logs that would be 'false positives' as far as exceptions occurring.

---

**Shutdown can delete the workspace out from under a still-stopping one-shot runtime**
*Priority: P3 · non-blocking · `server.py:227` · found by gpt-sol, gpt-codex, deepseek-flash*

#### Finding: Agent Summary

When streaming cleanup overruns its 5s bound it finishes in a background task. At shutdown the server only waits for *registered* sessions before deleting the staged workspace dir — and one-shot runtimes are never registered.

So a still-stopping one-shot runtime can have its scratch dir removed out from under it during shutdown. In practice this is low-impact (see below) — mostly some log noise.

- **What:** shutdown drains only registered sessions, not detached one-shot cleanup, before removing the staged dir
- **Where:** the lifespan shutdown block in `server.py` (`close_all_sessions()` → `release_runtime_base_dir`)
- **When:** a streaming one-shot's stop overruns 5s right as the process is shutting down
- **Why it matters:** the still-running cleanup hits a vanished directory — spurious exception noise in shutdown logs (not data loss)

#### Finding: Agent Breakdown

At shutdown the server runs `close_all_sessions()` then `release_runtime_base_dir`. But `close_all_sessions()` only drains the session registry — and one-shot runtimes are deliberately never registered (that's the leak fix). So a detached one-shot cleanup task isn't waited on.

The practical harm is small, and it's worth being explicit about why:

- `release_runtime_base_dir` is a no-op unless the dir was `staged`.
- A staged dir is a per-process `tempfile.mkdtemp` — "belongs to one process lifecycle," not shared across servers.
- The process is terminating anyway (SIGTERM → SIGKILL), and in k8s the whole pod's filesystem goes with it.

So the realistic outcome is exactly what you said: a background stop hitting a "file gone" error and logging a false-positive exception on a process that's about to exit — not data loss.

*(Separate but related: the background stop itself has no timeout — see the sibling comment on this line.)*

benmccown (approved):
Caught some nits/minor things, passing approval along and will let you address as you see fit. Test suite looks pretty robust, nice work.

One open question: this drops the old "call without a header and the server mints a session for you" behavior — `resolve_session` no longer auto-creates on `None`, so the client is now the sole source of session IDs (headerless = ephemeral one-shot, nothing registered, no ID returned). Fine for Studio (always sends an ID) and the eval/security harnesses (want ephemeral), and the README documents "send a stable ID to retain context." Just want to confirm no other client relied on server-minted session IDs — such a client would silently switch to ephemeral (lose cross-turn context) rather than error.

(Reviewed at 6ce736b9. Two earlier findings — the unguarded stream `_cleanup` and the unobserved background-cleanup task — look already handled by the latest `aclose`/observer rework, so they're dropped.)

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
benmccown, line 227: I'm not sure it's worth a bunch of effort to make the background stop more robust, could we just set a naive timeout or something? Not sure the best path forward, and it's just minor/nit level, so do whatever you think is best.

---

**Background stream cleanup has no timeout — a hung stop leaks a concurrency slot**
*Priority: nit · non-blocking · `server.py:227` · found by deepseek-flash*

#### Finding: Agent Summary

The 5-second bound only applies to the *foreground* wait; once cleanup detaches into the background task, the actual `runtime.stop()` it awaits has no timeout of its own.

If a stop hangs (rather than shutting down cleanly), that background task never completes — and because the concurrency slot is held until the stream context exits, the slot is never released. Repeated wedged stops would slowly erode the concurrency budget. This is independent of shutdown.

- **What:** the detached background cleanup awaits `runtime.stop()` with no timeout
- **Where:** `_cleanup` in `server.py` (the `__aexit__` call), reached via the background `_close_task`
- **When:** `runtime.stop()` hangs instead of returning — normal operation, not just shutdown
- **Why it matters:** a hung stop holds a shared concurrency permit forever; enough of them exhaust the budget (default 8)

#### Finding: Agent Breakdown

`stream_once` holds the `_invocation_slot` semaphore until the stream context exits. For the overrun case, that exit happens inside the detached `_close_task`.

The foreground `asyncio.wait_for(asyncio.shield(close_task), timeout=remaining)` (`server.py:205`) only bounds how long the *request handler* waits — after that it logs and lets cleanup continue in the background. But the background `_cleanup` then does `await self._stream_context.__aexit__(...)` (i.e. `runtime.stop()`) with no timeout of its own (`server.py:253`).

So a genuinely wedged stop never releases its permit. It's a slow, edge-case slot leak — not the common path, and it needs an actually-hanging `runtime.stop()`, not just a slow one.

Agreeing with your instinct: a naive bounded timeout on the background stop (wrap that `__aexit__` in a `wait_for`, log + move on if it fires) is a proportionate fix — no need for elaborate robustness. Genuinely minor/nit.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
mmogallapalli: I agree with the consolidation you suggested, made that change

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
mmogallapalli: I like keeping them separate because I want to guard against callers forgetting to provide a session and accidentally keeping a request as one shot when they wanted stateful.

plugins/nemo-agents/src/nemo_agents_plugin/fabric/session_manager.py
mmogallapalli: This is valid, I'm adding this to followup [issue](https://linear.app/nvidia/issue/AIRCORE-1118/harden-fabric-one-shot-runtime-isolation-timeouts-and-cleanup)

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
mmogallapalli: Agreed, tracking in followup [issue](https://linear.app/nvidia/issue/AIRCORE-1118/harden-fabric-one-shot-runtime-isolation-timeouts-and-cleanup)

plugins/nemo-agents/src/nemo_agents_plugin/fabric/server.py
mmogallapalli: Tracking in followup [issue](https://linear.app/nvidia/issue/AIRCORE-1118/harden-fabric-one-shot-runtime-isolation-timeouts-and-cleanup)

mmogallapalli: Tracking in followup [issue](https://linear.app/nvidia/issue/AIRCORE-1118/harden-fabric-one-shot-runtime-isolation-timeouts-and-cleanup)

plugins/nemo-agents/src/nemo_agents_plugin/fabric/runtime.py
mmogallapalli: A bit hesitant to expose any filesystem level errors that will detail filesystem paths, will have to look into a sanitized 500 response, tracking in [issue](https://linear.app/nvidia/issue/AIRCORE-1118/harden-fabric-one-shot-runtime-isolation-timeouts-and-cleanup)

PR #1781, opened 2026-09-03
fix(studio): don't apply the deployment grace period to endpoint-backed models (ASTD-554)

aray12:
## Summary

`getModelEntityChatStatus` returned `'pending'` for any model entity created in the last five minutes, and that check ran ahead of the one that recognises a live `api_endpoint`. A model autodiscovered from an external inference provider has its endpoint populated at creation and is serveable immediately, so the timer imposed a false five-minute wait and the Chat Playground reported **"Model Deployment in Progress"** for a deployment that does not exist and never will. After this change, an entity that already carries `api_endpoint.url` skips the grace period entirely and is evaluated on its own merits.

Observed on a local instance: 83 models autodiscovered from a newly registered provider, every one with a live `api_endpoint`, all blocked for five minutes.

## Changes

- `web/packages/common/src/utils/models.ts`
  - Hoist `hasApiEndpoint` above the grace-period check and gate the timer on `!hasApiEndpoint`.
  - `deploymentLoading` still short-circuits first: a model whose provider/deployment chain is still resolving is genuinely unknown regardless of endpoint.
  - Document the exemption on `isWithinDeploymentGracePeriod` and in the `getModelEntityChatStatus` evaluation-order list.
- `web/packages/common/src/utils/models.test.ts` — three cases: exempt when an endpoint is present, timer still applies without one, and `deploymentLoading` still wins over an endpoint.

### Behavioural note

Adapters and entities with `base_model` continue to require a READY deployment. They now reach that check immediately rather than sitting at `'pending'` for five minutes first, which does not change what the user sees: that check returns `'pending'` on its own while a deployment is non-terminal.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)
- [ ] Code change with documentation updates
- [ ] Documentation only
- [ ] Contributor tooling or automation
- [ ] CI, build, or test infrastructure

## Quality Gates

- [x] Tests added or updated for changed behavior
- [ ] Existing tests cover changed behavior — justification:
- [ ] Tests not applicable — justification:
- [ ] Documentation updated for user-visible behavior
- [x] Documentation not applicable — justification: no published documentation describes the chat-availability grace period; the rationale is captured in code comments alongside the change.

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
[...]

PR #1765, opened 2026-09-03
feat(studio): let the deployment wizard choose an inference engine (ASTD-553)

aray12:
## Summary

The create-deployment wizard hardcoded `engine: nim` for every source, so a deployment could not be created with vLLM even though that is what the LoRA customization docs use. `lora_enabled` was collected by the form but sent only on the NGC path, so the Workspace and HuggingFace paths — the two a user reaches a fine-tuned checkpoint through — silently dropped it. The wizard can now pick an engine and always sends `lora_enabled`.

| HuggingFace | Workspace |
| --- | --- |
| image | image |

## Changes

- Add an engine picker (`EngineFields`) to the HuggingFace and Workspace sources, defaulting to vLLM. NGC keeps `nim` and its required image fields, since that source is a NIM container by definition.
- Send `lora_enabled` on all three sources.
- Require an image when the engine is `nim` or `generic`. A blank image with `nim` silently resolves to `nvcr.io/nim/meta/llama-3.1-8b-instruct`, so an arbitrary checkpoint would be served by a Llama-specific NIM; `generic` has no compiler and cannot run without one.
- Omit blank image fields entirely so vLLM resolves its own model-agnostic default.
- Correct the side panel copy, which told the user that Workspace and HuggingFace deployments use the multi-LLM NIM. They did not: no image was sent, and the platform default is Llama-specific.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Documentation not applicable — justification: the deployments route is still behind `VITE_FF_DEPLOYMENTS_ENABLED` and has no published user documentation yet.

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

Targeted validation:

- `pnpm --filter nemo-studio-ui test src/routes/DeploymentsListRoute` — 38 passed across 3 files
- `pnpm --filter nemo-studio-ui typecheck` — clean
- `pnpm lint` — clean
- `uv run pre-commit run -a` was not run; the commit-time hooks (copyright headers, UI lint-staged, DCO, merge-conflict) passed.

Verified against a live platform: a vLLM deployment with `lora_enabled` created through this wizard reaches `READY` and serves both its base model and a LoRA adapter.

PR #1764, opened 2026-09-03
fix(studio): send the backend's own model name when chatting with an adapter (ASTD-552)

aray12:
## Summary

Chatting with a LoRA adapter sent `model: <adapter-name>` through the provider proxy. The proxy forwards the body to the inference backend unrewritten, and the backend knows an adapter by the name the sidecar registered it under (`{adapter_workspace}--{adapter_name}`), so a bare adapter name never resolved. The adapter's backend name is now resolved from the provider that serves it, so adapter chat reaches the adapter instead of failing.

https://github.com/user-attachments/assets/566ae0a0-15ab-4bad-a909-a32bac353569

## Changes

- Add `useServedModel`, which walks `model.model_providers` and returns the `served_models` entry matching a given model entity id, along with the provider serving it.
- `useModelIsServed` delegates to it, so there is a single path for fetching providers.
- Add `toInferenceModelEntityId`, which builds the composite `{base}&adapters/{ws}/{name}` id the provider reconciler records for an adapter.
- `ModelPanel` sends the resolved `served_model_name` and keeps adapters on the provider proxy. The model-entity and OpenAI gateway routes both require a VirtualModel, and the reconciler skips any served model whose id contains `&adapters/`, so neither can address an adapter today.
- When no provider serves the adapter, render an empty state rather than falling back to the base model. A silent fallback answers from the base weights and makes a fine-tune look like it had no effect.

## Type of Change

- [x] Code change (feature, bug fix, or refactor)

## Quality Gates

- [x] Tests added or updated for changed behavior
- [x] Documentation not applicable — justification: no user-facing documentation covers the model chat panel's request construction.

## Verification

- [x] Pull request title follows the repository's Conventional Commit format
- [x] Every commit includes an appropriate `Signed-off-by:` trailer
- [ ] `uv run pre-commit run -a` passes, or any blocked checks are identified below
- [x] Targeted tests pass, or tests are marked not applicable above
- [x] No secrets, API keys, or credentials are included

Targeted validation:

- `pnpm --filter @nemo/common test src/utils/models.test.ts` — 21 passed
- `pnpm --filter nemo-studio-ui test src/hooks src/components/sidePanels` — 110 passed across 14 files
- `pnpm --filter @nemo/common typecheck` — clean
- `pnpm --filter nemo-studio-ui typecheck` — clean
- `pnpm lint` — clean
- `uv run pre-commit run -a` was not run; the commit-time hooks (copyright headers, UI lint-staged, DCO, merge-conflict) passed.

[...]

PR #1782, opened 2026-09-03
fix(models): link deployment-backed providers to their Model Entity (ASTD-555)

aray12:
## Summary

A deployment-backed `ModelProvider` never wrote itself into its base Model Entity's `model_providers` list. The model therefore became fully routable for inference — provider `READY`, passthrough VirtualModel created, listed by IGW `/v1/models` — while remaining invisible in every Studio surface that gates on `hasModelProvider`. In practice, every model deployed through the Studio deployment wizard deployed successfully and then could not be selected in the UI that created it.

After this change the reconciler stages the back-reference for each served entity, so `model_providers` is populated for deployment-backed providers exactly as it already was for external ones.

## Changes

- `services/core/models/src/nmp/core/models/controllers/provider_reconciler.py` — stage `stage_provider_link` for every served model entity inside the existing `served_models` loop that creates passthrough VirtualModels.
- `services/core/models/tests/unit/controllers/test_provider_reconciler.py` — three tests: the base entity is linked, the link is idempotent across ticks, and LoRA composite ids are not linked. Section header amended, since "no autocreate" still holds but "no linking" no longer does.

### Root cause

`_reconcile_single_provider` dispatches on provider kind:

```python
if provider.model_deployment_id is None:
    await self._ensure_external_entities(...)   # reaches stage_provider_link
    served_models = self._generate_external_served_model_mappings(...)
else:
    served_models = self._generate_deployment_served_model_mappings(...)   # generator only
```

The only two `stage_provider_link` call sites live in `_ensure_model_entity_for_provider`, reachable solely from `_ensure_external_entities`, so the back-reference was never staged for a deployment-backed provider.

The omission traces to the "No entity autocreation happens here, by design" note on `_generate_deployment_served_model_mappings` — sound for *creation*, but silently extended to *linking*. An entity that already exists still needs the provider reference. The asymmetry is the giveaway: the deployment reconciler already stages an **unlink** on teardown, so the invariant was maintained on delete but never established on create.

### Why this location

The chosen loop already runs for both branches, already applies the `&adapters/` skip for LoRA composite ids that have no standalone Model Entity, and already has `provider_id` in scope — so one statement covers both provider kinds with no new control flow, and `_generate_deployment_served_model_mappings` stays a pure generator.

Staging a link never populates `create_kwargs`, so the deployment-backed no-autocreation contract is preserved and the existing `test_deployment_backed_never_calls_ensure_model_entity` passes unmodified. An entity that does not exist yet (a prompt-tuned variant still being registered by its training job) has its link dropped by `ModelEntityCache._create` and re-staged on the next tick, which is the intended self-healing behaviour.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal controller reconciliation behaviour restoring a documented data-model invariant; no published documentation changes.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pytest services/core/models/tests/unit/controllers/test_provider_reconciler.py` | 130 passed |
| `pytest services/core/models/tests/unit/controllers/` | 379 passed |
| `uv run ruff check services/core/models` | All checks passed |
| `uv run ruff format --check services/core/models` | 127 files already formatted |

The new tests were confirmed non-vacuous: reverting only the source change while keeping the tests makes `test_deployment_backed_links_base_entity_to_provider` and `test_deployment_backed_does_not_link_lora_composite_ids` fail, and restoring it makes them pass. `test_deployment_backed_link_is_idempotent` passes either way by construction — it asserts that no write occurs — so it guards against duplicate links rather than detecting the original defect.

`uv run --frozen ty check services/core/models` reports 780 diagnostics, which is the pre-existing baseline: the count is identical with this change stashed, and none of the diagnostics fall on the added lines. The `ty` pre-commit hook passed on the staged files.

`uv run pre-commit run -a` was not run across the whole repository. The pre-commit hook suite ran on the staged files at commit time and passed, including ruff, ruff-format, `ty`, copyright headers, and DCO.

### Not covered here

End-to-end verification against a running platform (restart the models controller, confirm `model_providers` is populated without a manual `PATCH`, and confirm the model becomes selectable in the Chat Playground) has not been performed in this PR.

## Summary by CodeRabbit

- **Improvements**
  - Deployment-backed providers now maintain associations with the base models they serve.
  - Prompt-tuned model registrations can be handled when their entities become available.
  - Existing associations are preserved without duplicate updates.
  - LoRA composite mappings continue to support served-model routing without altering base-model associations.


aray12:

- #1764
    - #1765
        - #1781
            - #1782 ◀

benmccown (approved):
Sorry this slipped by, thank you for fixing!

PR #1790, opened 2026-09-04
fix(setup): address sorting for models resolve nvbug-6676750

soluwalana:
## Summary

`nemo setup --auto` was persisting the first alphabetically sorted chat-looking model from provider discovery as both the default and fast model. NVIDIA Build's `GET /v1/models` advertises the full catalog, not the models the current account can serve, so setup saved unusable defaults (QA: `ai21labs/jamba-1.5-large-instruct` returned upstream 404). Setup now probes candidates with a short chat request, prefers NVIDIA models, and saves the largest usable model as default and the smallest usable model as fast. If none answer, it saves no default.

## Related Issue

NVBug 6676750

## Changes

- Filter out non-chat names (`embed`, `vision`, `guard`, `rerank`, `reward`, `translate`, …).
- Order remaining candidates NVIDIA-first, then by parameter count parsed from the entity name.
- Probe each candidate through the inference gateway (`POST v1/chat/completions`, 16 tokens) before persisting it; skip non-2xx (including IGW-wrapped upstream 404).
- Cap probes at 8 attempts per slot so gated large models at the head of the list do not block selection.
- Honor `NEMO_DEFAULT_MODEL` / `NEMO_FAST_MODEL` without probing.
- Interactive setup uses the same probed pair as picker suggestions.
- If the gateway is unreachable or no candidate answers, save nothing rather than falling back to name sort.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: setup already documents `NEMO_DEFAULT_MODEL` / `NEMO_FAST_MODEL`; the change is automatic selection plus CLI warnings when nothing is usable.

## Verification

Targeted validation:

```
uv run --no-sync pytest packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestPickDefaultChatEntity \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestModelSizeOrdering \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestModelProbe \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestUsableModelPairSelection \
  packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestAutoModelPairSelection -q
```

33 passed.

Live catalog check against `https://integrate.api.nvidia.com` (not in CI): old sort would still pick Jamba; probed selection picked `nvidia/nemotron-3-super-120b-a12b` (default) and `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning` (fast) for an entitled nvapi key.

`uv run pre-commit run -a` was not run for this PR.

## Summary by CodeRabbit

* **New Features**
  * Automatic setup recognizes slash-qualified NVIDIA model IDs.
  * Model probing retries route-readiness issues independently for each candidate.
  * Account-entitlement 404 responses are skipped without additional retries.
  * Automatic selection prioritizes responsive, compatible model pairs and waits for newly available routes.
* **Bug Fixes**
  * Interactive model selection no longer chooses unverified defaults when probing fails.
  * Interactive setup requires explicit model choices when verification is unavailable.
  * Updated guidance clarifies the choices required during interactive setup.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
@@ -27,7 +27,7 @@ from urllib.parse import urlparse
 import httpx
 import typer
 import yaml as _yaml
-from nemo_platform import APIConnectionError, APIStatusError, NeMoPlatform
+from nemo_platform import APIConnectionError, APIStatusError, APITimeoutError, NeMoPlatform
 from nemo_platform_plugin.capabilities import probe_docker
 from nemo_platform_plugin.client.adapter import client_from_platform
 from nemo_platform_plugin.secrets.client import SecretsClient
@@ -249,11 +249,14 @@ _KEY_REJECTED_STATUS_CODES = (401, 403)
 _TERMINAL_PROBE_STATUS_CODES = (404, 405, 410)
 _KEY_REJECTED_MESSAGE = "API key validation failed. The provider rejected the credentials."
 
-# Catalog listing is not entitlement-scoped, so a candidate is only trusted
-# after a real chat request. The largest names are often gated, so the budget
-# has to absorb a run of 404s at the head of the list.
+# Catalog listing is not entitlement-scoped; only a successful chat request counts.
 _MODEL_PROBE_TIMEOUT = 20.0
 _MODEL_PROBE_MAX_ATTEMPTS = 8
+_MODEL_PROBE_BUDGET_SECONDS = 90.0
+# Gateway 404s until the model's VirtualModel exists; retries are shared and capped.
+_MODEL_ROUTE_READY_SECONDS = 10.0
+_MODEL_ROUTE_RETRY_INTERVAL = 1.0
+_MODEL_ROUTE_MAX_RETRIES = 10
 
 _MODEL_DISCOVERY_ROUND_SECONDS = 30
 _MODEL_DISCOVERY_MAX_ROUNDS = 2
@@ -859,32 +862,47 @@ def _order_candidates_by_size(entity_ids: list[str], *, largest_first: bool) -> 
     return sorted(entity_ids, key=sort_key)
 
 
-def _probe_model_entity(client: NeMoPlatform, workspace: str, entity_id: str) -> bool:
+def _probe_model_entity(
+    client: NeMoPlatform, workspace: str, entity_id: str, route_ready_deadline: float = 0.0
+) -> bool:
     """Return True when a short chat request against *entity_id* succeeds.
 
-    Non-2xx responses (including wrapped upstream 404s) skip the candidate.
-    Connection failures are re-raised so the caller can stop probing.
+    Retries 404s until *route_ready_deadline*; timeouts skip; connection errors raise.
     """
-    try:
-        client.inference.gateway.openai.post(
-            "v1/chat/completions",
-            workspace=workspace,
-            body={
-                "model": entity_id,
-                "messages": [{"role": "user", "content": "Respond with 'OK'"}],
-                "max_tokens": 16,
-            },
-        )
-    except APIConnectionError:
-        raise
-    except APIStatusError as exc:
-        console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} (HTTP {exc.status_code})")
-        return False
-    except Exception as exc:
-        logger.debug("Model probe for '%s' failed", entity_id, exc_info=True)
-        console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} ({exc})")
-        return False
-    return True
+    retries = 0
+    while True:
+        try:
+            client.inference.gateway.openai.post(
+                "v1/chat/completions",
+                workspace=workspace,
+                body={
+                    "model": entity_id,
+                    "messages": [{"role": "user", "content": "Respond with 'OK'"}],
+                    "max_tokens": 16,
+                },
+            )
+        except APITimeoutError:
+            logger.debug("Model probe for '%s' timed out", entity_id, exc_info=True)
+            console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} (timed out)")
+            return False
+        except APIConnectionError:
+            raise
+        except APIStatusError as exc:
+            if (
+                exc.status_code == 404
+                and retries < _MODEL_ROUTE_MAX_RETRIES
+                and time.monotonic() < route_ready_deadline
+            ):
+                retries += 1
+                _pause(_MODEL_ROUTE_RETRY_INTERVAL)
+                continue
+            console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} (HTTP {exc.status_code})")
+            return False
+        except Exception as exc:
+            logger.debug("Model probe for '%s' failed", entity_id, exc_info=True)
+            console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} ({exc})")
+            return False
+        return True
 
 
 def _select_usable_model_pair(client: NeMoPlatform, workspace: str, entity_ids: list[str]) -> ModelPair | None:
@@ -899,15 +917,18 @@ def _select_usable_model_pair(client: NeMoPlatform, workspace: str, entity_ids: 
 
     usable: dict[str, bool] = {}
     probe_client = client.with_options(max_retries=0, timeout=_MODEL_PROBE_TIMEOUT)
+    started = time.monotonic()
+    deadline = started + _MODEL_PROBE_BUDGET_SECONDS
+    route_ready_deadline = started + _MODEL_ROUTE_READY_SECONDS
 
     def first_usable(ordered: list[str]) -> str | None:
         attempts = 0
         for entity_id in ordered:
             if entity_id not in usable:
-                if attempts >= _MODEL_PROBE_MAX_ATTEMPTS:
+                if attempts >= _MODEL_PROBE_MAX_ATTEMPTS or time.monotonic() >= deadline:
                     return None
                 attempts += 1
-                usable[entity_id] = _probe_model_entity(probe_client, workspace, entity_id)
+                usable[entity_id] = _probe_model_entity(probe_client, workspace, entity_id, route_ready_deadline)
             if usable[entity_id]:
                 return entity_id
         return None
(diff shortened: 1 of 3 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -816,6 +832,98 @@ def _pick_default_chat_entity(entity_ids: list[str]) -> str | None:
...
+            "v1/chat/completions",
+            workspace=workspace,
+            body={
+                "model": entity_id,
+                "messages": [{"role": "user", "content": "Respond with 'OK'"}],
+                "max_tokens": 16,
+            },
+        )
+    except APIConnectionError:
+        raise
```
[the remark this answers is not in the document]
soluwalana: APITimeoutError is caught before APIConnectionError. A timeout is logged, printed as a skip, and treated as a failed candidate (False) so later models still get probed:


setup.py
Ln 884–889
except APITimeoutError:
    logger.debug("Model probe for '%s' timed out", entity_id, exc_info=True)
    console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} (timed out)")
    return False
except APIConnectionError:
    raise
The two eight-attempt passes share one wall-clock deadline (_MODEL_PROBE_BUDGET_SECONDS = 90.0). Each probe still uses the 20-second client timeout, but selection stops once the 90-second budget is gone rather than running up to 8 × 2 × 20s.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -816,6 +835,116 @@ def _pick_default_chat_entity(entity_ids: list[str]) -> str | None:
...
+    """
+    candidates = [entity_id for entity_id in entity_ids if _is_usable_chat_model_entity(entity_id)]
+    if not candidates:
+        return None
+
+    usable: dict[str, bool] = {}
+    probe_client = client.with_options(max_retries=0, timeout=_MODEL_PROBE_TIMEOUT)
+    started = time.monotonic()
+    deadline = started + _MODEL_PROBE_BUDGET_SECONDS
+    route_ready_deadline = started + _MODEL_ROUTE_READY_SECONDS
```
tylersbray: > **agent** (`pr-review`): Per-candidate 404 retries, not one shared 10s window

`route_ready_deadline` is computed once for the whole pair selection, and `_probe_model_entity` retries **every** HTTP 404 until that deadline.

On a cold start the largest NVIDIA name is probed first. If that 404 is either "VirtualModel not published yet" **or** a terminal missing-function 404, the loop can spend the full 10s here. Later candidates then get a single 404 and are treated as unusable even if their route appears a moment later. The late-route e2e only has one chat candidate, so it does not cover this. 502-wrapped entitlement misses skip immediately; raw gateway 404s do not.

Give each candidate its own retry window (or only retry 404s that look like a missing route).

```python
# Agent fix: per-candidate route window, not one shared deadline.
# In `_select_usable_model_pair`, stop passing `started + _MODEL_ROUTE_READY_SECONDS`
# into every probe. Compute it inside `_probe_model_entity` (or at the call site
# per entity) so each candidate gets a fresh `_MODEL_ROUTE_READY_SECONDS` budget.
#
# Optional: do not retry entitlement 404s. If `exc.body` / `exc.message` contains
# "Function not found for account" (or similar), return False immediately.

def _probe_model_entity(
    client: NeMoPlatform, workspace: str, entity_id: str, *, wait_for_route: bool = True
) -> bool:
    route_ready_deadline = (
        time.monotonic() + _MODEL_ROUTE_READY_SECONDS if wait_for_route else 0.0
    )
    retries = 0
    while True:
        try:
            client.inference.gateway.openai.post(
                "v1/chat/completions",
                workspace=workspace,
                body={
                    "model": entity_id,
                    "messages": [{"role": "user", "content": "Respond with 'OK'"}],
                    "max_tokens": 16,
                },
            )
        except APITimeoutError:
            ...
        except APIConnectionError:
            raise
        except APIStatusError as exc:
            detail = str(getattr(exc, "body", None) or exc)
            entitlement_miss = "not found for account" in detail.lower()
            if (
                exc.status_code == 404
                and not entitlement_miss
                and retries < _MODEL_ROUTE_MAX_RETRIES
                and time.monotonic() < route_ready_deadline
            ):
                retries += 1
                _pause(_MODEL_ROUTE_RETRY_INTERVAL)
                continue
            console.print(f"  {WARN} Skipping {_display_model_name(entity_id)} (HTTP {exc.status_code})")
            return False
        ...

# Tests to add:
# 1. First candidate 404s for the full route window, second candidate 404s once then 200
#    after its own window starts -> pair uses the second model.
# 2. 404 body "Function not found for account" is not retried.
```

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -816,6 +835,116 @@ def _pick_default_chat_entity(entity_ids: list[str]) -> str | None:
...
+    """Return the largest ``Nb`` parameter count parsed from an entity name."""
+    name = _display_model_name(entity_id).lower().replace("_", "-")
+    sizes = [float(match) for match in _MODEL_SIZE_PATTERN.findall(name)]
+    return max(sizes) if sizes else None
+
+
+def _is_preferred_vendor(entity_id: str) -> bool:
+    """Return whether the entity name identifies an NVIDIA-published model."""
+    name = _display_model_name(entity_id).lower().replace("_", "-")
+    return name.startswith(("nvidia-", "nv-")) or "nemotron" in name
```
tylersbray: > **agent** (`pr-review`): Treat `nvidia/` as an NVIDIA-published model

`_display_model_name` only strips the first `/`, so `default/nvidia/nemotron-3-super-120b-a12b` becomes `nvidia/nemotron-3-super-…`. `startswith(("nvidia-", "nv-"))` is false for slash IDs; they only count as preferred when `"nemotron"` is in the name.

The live catalog check on this PR used slash IDs. Unit tests only cover hyphenated `nvidia-nemotron-…`. Non-Nemotron NVIDIA models (`nvidia/mistral-…`, `nvidia/llama-…` without `nemotron`) sort with Meta/Jamba and can consume the 8-attempt budget first.

```python
# Agent fix: prefer nvidia/ as well as nvidia- / nv- / nemotron.

def _is_preferred_vendor(entity_id: str) -> bool:
    """Return whether the entity name identifies an NVIDIA-published model."""
    name = _display_model_name(entity_id).lower().replace("_", "-")
    return name.startswith(("nvidia-", "nvidia/", "nv-", "nv/")) or "nemotron" in name

# Add to TestModelSizeOrdering.test_orders_nvidia_first_then_by_size (or a new test):
#   "default/nvidia/mistral-nemo-minitron-8b-instruct" should sort ahead of
#   "default/meta/llama-3.3-70b-instruct" even though it is not a Nemotron name.
# Also assert _is_preferred_vendor("default/nvidia/llama-3.1-8b-instruct") is True.
```

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -1973,21 +2102,25 @@ def _select_model_pair(
...
 
+    suggested = _select_usable_model_pair(client, workspace, entity_ids)
+    if suggested is None:
+        console.print(f"  {WARN} None of the discovered models served a test request; suggestions may not work.")
+
     default_model = prompt_select(
         "Choose your default model (used for quality-critical agent work):",
         choices=display_models,
-        default=suggested,
+        default=suggested.default if suggested else _pick_default_chat_entity(entity_ids),
```
tylersbray: > **agent** (`pr-review`): Do not default the picker to alphabetical first-chat after a failed probe

Auto mode correctly saves nothing when `_select_usable_model_pair` returns `None`. Interactive setup still does `default=_pick_default_chat_entity(entity_ids)` in that case. Choices are sorted, so Enter can persist `ai21labs/jamba-…` — the same NVBug 6676750 failure — with only a warning.

If probing found nothing usable, do not suggest a name that was not verified.

```python
# Agent fix: only pre-select a probed model. If probing failed, leave the picker
# without a verified default (first choice is fine as a cursor position only if
# prompt_select requires one; prefer None / no default if the helper allows it).

suggested = _select_usable_model_pair(client, workspace, entity_ids)
if suggested is None:
    console.print(
        f"  {WARN} None of the discovered models served a test request; "
        "do not accept an unverified default."
    )

default_model = prompt_select(
    "Choose your default model (used for quality-critical agent work):",
    choices=display_models,
    default=suggested.default if suggested else None,
    hint="Press Enter to accept the default." if suggested else "Pick a model you can actually call.",
)
fast = prompt_select(
    "Choose your fast model (used for latency-sensitive agent work):",
    choices=display_models,
    default=suggested.fast if suggested else default_model,
    hint="Press Enter to reuse the default model.",
)

# If prompt_select cannot take default=None, pass no default / require an explicit
# selection rather than _pick_default_chat_entity (alphabetical first chat model).
#
# Test: when _select_usable_model_pair returns None, prompt_select's default kwarg
# must not be the Jamba (or any other unverified) entity from _pick_default_chat_entity.
```

tylersbray (commented):
> **agent** (`pr-review`): Three P2s on automatic model selection

Inline notes: shared 404 retry window, `nvidia/` vendor detection, and interactive Enter-to-accept fallback after a failed probe.

tylersbray (approved):
Looks good to me.  Agentic sweep found some non-blocking P2's, posted them with agent-friendly snippets for quick fixes.  The alphabetical always felt a little crappy, thanks for making this UX improvement.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
soluwalana: Implemented

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
soluwalana: Implemented.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
soluwalana: Implemented

PR #1798, opened 2026-09-04
refactor(studio): tighten agent and evaluation surfaces

rrhyne:
Screenshot 2026-09-04 at 10 01 50 AM
Screenshot 2026-09-04 at 9 26 03 AM
Screenshot 2026-09-04 at 9 25 40 AM

## Summary

Assorted Studio cleanup across the agent detail page, the new-agent modal, and the evaluation wizard. The only functional change is the agent overview's number formatting; everything else removes a control, renames a tab, or shortens copy.

## Changes

- **Agent overview tiles compact large magnitudes.** Avg latency and Avg token count now read `2.5K ms` and `1.1M` instead of `2,500 ms` and `1,125,000`, via `formatNumericValue` from `@nemo/common/src/components/charts/format` — the formatter the chart components already use, so no new dependency and no second convention. Latency below 1000 ms keeps its existing precision rather than gaining decimals.
- **Removed the Open traces button** from the agent header. The Traces nav item and the overview panel's own View traces control already lead there. This also retires the now-unused `useNavigate`, `getIntakeTracesRoute`, `INTAKE_ENABLED` and `ListTree` references in that route.
- **Renamed the new-agent modal's first tab** from "Coding agent prompt" to "Integrate agent", naming the goal rather than the artifact, and moved the agent name field above the file picker on the Upload tab so it is filled before the upload it labels.
- **Shortened copy** on the evaluation wizard's start options, the evaluation-name helper, the evaluation picker's helper, and the two experiment flag hints.
- **Renamed the agents list action to Register Agent**, carried through the modal's title, instruction and tab-list label. "Instrument" describes what the user does to their own code; this modal only records the agent with the platform.
- **Removed the wizard's step-blocker paragraph.** It restated what the field directly above it already reported (`An experiment named X already exists`, `Checking name…`), so field validation now carries it alone. The underlying blocker still disables **Next**, so a step cannot be skipped into a state submit would silently reject.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: Studio UI copy and layout; no documented surface describes these controls.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test` — 333 files, 3263 tests passed (exit 0). Updated assertions that pinned the old behaviour: the agent trace-statistics tiles now expect `1.5K` / `2.5K ms`, the agent detail route asserts the Open traces button is *absent* and drops the test that navigated with it, and the new-agent modal tests follow the tab rename.
- `pnpm --filter nemo-studio-ui typecheck`, `pnpm --filter @nemo/common typecheck`, `pnpm lint`, `pnpm format` — all clean.
- `uv run pre-commit run --files <the 12 changed files>` — `Fix copyright headers`, `Run UI lint-staged` and `check for merge conflicts` passed; Python and config hooks reported `(no files to check)` since this change is entirely under `web/`. **`uv run pre-commit run -a` was not run**, as it would also act on unrelated untracked work in the checkout; flagging rather than claiming the repo-wide gate passed.
- Copy reviewed across every string the branch touches: fixed `NeMo platform` -> `NeMo Platform` casing against the title beside it, and the evaluation picker's helper (a past run is what you evaluate a change *with*, not *by*).
- One full-suite run exited non-zero without reporting a failing test; two subsequent runs passed with exit 0 and identical counts, so it is being treated as a flake rather than a regression.


aray12 (approved):

walston:
The prompt text is really confusing to me, and doesn't read like you're registering an agent, but rather that you're creating a new evaluation/experiment group. Specifically it seems odd to put the AGENT_NAME down in an environment variables with PLATFORM & WORKSPACE and not as a question the agent should ask you along where it's running. If we're integrating an agent where is it, wouldn't that need to be an environment variable too?

PR #1814, opened 2026-09-04
feat(evaluator): give Gym trials an OTLP trace projected from ng_trajectory

SandyChapman:
Part 1 of 2. Stacked follow-up: #1815 (the rename).

## What

Gym reports each rollout in its own `ng_trajectory` shape, so a metric could not read a Gym trial the way it reads a Harbor one. This projects each rollout onto OTLP resource spans, so both runners feed the same typed trace evidence.

- **`nemo_evaluator_sdk.ng_trajectory_otlp`** — the projection. Imports nothing from the rest of the SDK and depends only on `opentelemetry-proto`, so it can be lifted out if Gym or anyone else wants it. Liftable, deliberately not lifted.
- **Per-call spans.** Gym only records timing and per-call token usage when observability is on, so the runtime asserts `observability_enabled` and a `model_call_capture_dir` and reads the captures back: one LLM span per model call, with `gen_ai.usage.*` and TTFT. It overrides a caller's `hydra_params` on purpose — turning it off degrades every trace silently.
- **`ng_trajectory` evidence** is kept in its own right, by reference. A projection is not a reason to lose what it was projected from.
- **One representation.** The trace is `list[ResourceSpans]` throughout. `ExportTraceServiceRequest` is the Export RPC envelope and is now built only at the wire boundary in `intake/mapping.py`. `resource_spans_json` is gone.

## Commits

1. `feat` — the projection and the runtime wiring.
2. `test` — a coverage pass over paths the first round missed, kept separate so the findings are readable on their own.
3. `merge` — `origin/main`, to pick up the CI fix below.
4. `chore` — **an unrelated repo-wide CI fix; see "One out-of-scope commit" at the bottom.**

## Validation

Real `gym eval run` output, not fixtures: `mcqa` (1 LLM span, 35.06ms) and `workplace_assistant` for the TOOL path (2 LLM spans, 26.96ms / 3.00ms). Descriptors on a real trial: `['ng_trajectory', 'rollouts', 'trace', 'trace:otlp']`. `final_output_text` returns the agent's answer rather than a tool's output.

**2805 passed, 38 skipped.** CI is fully green.

Coverage on the new and rewritten modules, measured across the SDK *and* plugin suites together (SDK-only under-reports, since Intake exercises several of these functions):

| module | before the test commit | after |
|---|---|---|
| `ng_trajectory_otlp/convert.py` | 89.6% | **94.6%** |
| `values/otlp.py` | 94.3% | **95.5%** |

Every new test was verified by breaking the code and confirming it goes red. The one worth calling out: nothing covered the `RecursionError` guard in `parse_resource_spans`. A payload nested past protobuf's limit raises an error that is neither a `ParseError` nor a `ValueError`, so without that guard it escapes every caller and takes a run down over one unreadable trace.

## Known gaps

- **Sandboxed Gym runs get no captures** ([AALGO-593](https://linear.app/nvidia/issue/AALGO-593/enable-gym-model-call-capture-on-the-sandboxed-host-so-sandboxed-runs)). The host doesn't enable Gym's capture, so captures never leave the sandbox and those traces carry no per-call timing. Not a regression — that path had no trace at all before — and marked at the call site.
- **`convert.py`'s `_any_value` bool branch is unreachable** and is the only uncovered line left. The comment on it defends a real trap (bool is an int subclass, so an int check ordered first writes `True` as `1`), but every caller excludes bool upstream. Kept for whoever lifts the package out; flagging it so it isn't mistaken for tested code. Happy to delete it instead.
- **The gym package still imports underscore-prefixed names across modules.** Real smell, fixed in #1815 rather than mixed in here.

## One out-of-scope commit

`chore: re-vendor nemo-platform extras after nemo-insights loosened its pins` has nothing to do with the OTLP work. It is here because `main` was red and this PR inherited the failure.

`plugins/nemo-insights/pyproject.toml` moved two nemo-fabric adapter deps from `==0.3.0b1` to `>=0.3.0b1`, but `packages/nemo_platform/pyproject.toml` is **generated** from the plugin pyprojects by `make vendor` and was never regenerated. So the declared extras — and `uv.lock`, derived from them — still carried the old pins.

That made two CI jobs demand opposite things: `Check uv lock` compared the lock against the stale pyproject and rejected `>=`, while `lint-sdk-vendored` / `lint-pre-commit-all` regenerated the pyproject and rejected `==`. Both were right; the input was stale, not the output. `make vendor && uv lock` (uv 0.9.14) fixes both, and this is the first commit on which they pass together.

**This also fixes `main`**, which is still red for everyone else. Say the word and I will pull it into its own PR to main instead, so the right reviewers see it and it unblocks other branches sooner.

## Summary by CodeRabbit

- **New Features**
  - Added conversion of Gym rollout records into OTLP traces with agent, LLM, and tool spans.
  - Included model-call captures, prompts, outputs, token usage, timing, errors, and deterministic identifiers.
  - Gym trial evidence now includes per-rollout files, captures, and projected OTLP traces.
  - OTLP evidence now exposes parsed, typed resource spans.
  - Retry attempts receive distinct trial and trace identifiers.

- **Bug Fixes**
  - Improved handling of missing, unreadable, malformed, or unparsable traces without discarding trials.

- **Documentation**
  - Added OTLP conversion usage, limitations, and evidence-processing guidance.

### DIFF
--- a/docs/evaluator/agent-eval/writing-metrics.mdx
+++ b/docs/evaluator/agent-eval/writing-metrics.mdx
@@ -142,8 +142,10 @@ async def read_trace(input: MetricInput) -> None:
 | **filesystem** (`EVIDENCE_FINAL_STATE`, `EVIDENCE_INITIAL_STATE`) | `await evidence.filesystem(name)` | `.run_verifier(command)`, `.diff(other)` — run a check or diff two snapshots |
 | **logs** | `await evidence.logs(name)` | `.read_text(file)`, `.tail(file)` |
 
-OTLP exposes the unflattened `resourceSpans` tree; it does not provide a generic `tool_calls()` method.
-It's a responsibility of consumer to parse `resourceSpans` tree specific to OTLP exporter to extract the searched data.
+OTLP exposes the unflattened span tree as parsed `ResourceSpans` protobuf messages — the same type
+`opentelemetry-proto` defines and Intake ingests — so a reader walks `.scope_spans[].spans[]` and
+typed attributes rather than decoded JSON. It does not provide a generic `tool_calls()` method: what
+counts as a tool call is exporter-specific, so extracting them is the consumer's job.
 
 ### Asking for one trace format
 
--- a/docs/evaluator/agent-eval/writing-metrics.mdx
+++ b/docs/evaluator/agent-eval/writing-metrics.mdx
@@ -145,7 +145,9 @@ async def read_trace(input: MetricInput) -> None:
 OTLP exposes the unflattened span tree as parsed `ResourceSpans` protobuf messages — the same type
 `opentelemetry-proto` defines and Intake ingests — so a reader walks `.scope_spans[].spans[]` and
 typed attributes rather than decoded JSON. It does not provide a generic `tool_calls()` method: what
-counts as a tool call is exporter-specific, so extracting them is the consumer's job.
+counts as a tool call depends on whatever produced the spans — the Gym projection, or a harness like
+Fabric or Codex — each of which names and nests them differently, so extracting them is the
+consumer's job.
 
 ### Asking for one trace format
 
--- a/packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
+++ b/packages/nemo_evaluator_sdk/examples/run_agent_eval/example_metrics.py
@@ -22,6 +22,7 @@ from typing import Any
 
 from nemo_evaluator_sdk.agent_eval.trials import EVIDENCE_FINAL_STATE, EVIDENCE_INITIAL_STATE, EVIDENCE_TRACE
 from nemo_evaluator_sdk.metrics.protocol import MetricInput, MetricOutput, MetricOutputSpec, MetricResult
+from opentelemetry.proto.trace.v1.trace_pb2 import ResourceSpans, Span
 
 
 class TestsPassMetric:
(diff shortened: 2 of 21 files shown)


docs/evaluator/agent-eval/writing-metrics.mdx
```diff
@@ -142,8 +142,10 @@ async def read_trace(input: MetricInput) -> None:
 | **filesystem** (`EVIDENCE_FINAL_STATE`, `EVIDENCE_INITIAL_STATE`) | `await evidence.filesystem(name)` | `.run_verifier(command)`, `.diff(other)` — run a check or diff two snapshots |
 | **logs** | `await evidence.logs(name)` | `.read_text(file)`, `.tail(file)` |
 
-OTLP exposes the unflattened `resourceSpans` tree; it does not provide a generic `tool_calls()` method.
-It's a responsibility of consumer to parse `resourceSpans` tree specific to OTLP exporter to extract the searched data.
+OTLP exposes the unflattened span tree as parsed `ResourceSpans` protobuf messages — the same type
+`opentelemetry-proto` defines and Intake ingests — so a reader walks `.scope_spans[].spans[]` and
+typed attributes rather than decoded JSON. It does not provide a generic `tool_calls()` method: what
+counts as a tool call is exporter-specific, so extracting them is the consumer's job.
```
arpitsardhana: exporter in this agent harness like fabric, codex?

JashG (approved):

packages/nemo_platform/pyproject.toml
```diff
@@ -392,8 +392,8 @@ nemo-guardrails-plugin = [
 nemo-insights-plugin = [
   "fastapi>=0.115",
   "httpx",
-  "nemo-fabric-adapter-contract==0.3.0b1",
-  "nemo-fabric-adapters-common==0.3.0b1",
+  "nemo-fabric-adapter-contract>=0.3.0b1",
+  "nemo-fabric-adapters-common>=0.3.0b1",
```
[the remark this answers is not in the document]
SandyChapman: Not this PR — the vendoring commit landed on main separately as #1824, so this branch no longer touches `pyproject.toml`, `uv.lock`, or `third_party/`; the snapshot is stale on main and is a license/OSV scan input rather than an install source, so I'll regenerate it there.

docs/evaluator/agent-eval/writing-metrics.mdx
SandyChapman: Yes — "exporter" here means whatever emitted the spans (the Gym projection, or a harness like Fabric or Codex), since each names and nests tool calls differently; reworded the line to say that rather than leaving "exporter" ambiguous.

PR #1818, opened 2026-09-04
feat(studio): default the deployment name from the Hugging Face repo (ASTD-556)

aray12:
## Summary

The HuggingFace tab in the deployment wizard names every resource from `values.name`, which defaults to `generateDefaultName()` — a random `adjective-colour-animal` string. Deploying `Qwen/Qwen2.5-7B-Instruct` produced a model entity called `human-yellow-elephant`, and the repo ID appeared nowhere in Models, Deployments, or the chat playground picker. A model could deploy successfully, serve inference, and still be unfindable by the identifier the user typed.

This derives the default as `<org>-<repo>` — `Qwen/Qwen2.5-7B-Instruct` becomes `qwen-qwen2.5-7b-instruct`. The field stays editable, and derivation stops permanently once the user types their own name. Only the HuggingFace source is affected; NGC and Workspace keep the random default.

https://github.com/user-attachments/assets/31bba8cf-08cd-412f-b38d-c9498ae1198b

## Changes

- **`huggingFaceDeploymentArtifacts.ts`** — new `huggingFaceRepoIdToBaseName()` slug helper and `HF_DERIVED_BASE_NAME_MAX_LEN`.
- **`useHuggingFaceNameDefault.ts`** (new) — watches the Repo ID and writes the derived name until the user edits the field.
- **`CreateDeploymentSidePanel/index.tsx`** — calls the hook.
- **`useCreateDeploymentBySource.ts`** — pre-flight model-name lookup so a taken name is reported before anything is created.
- Tests for all of the above.

### Meeting the name contract

The derived slug must satisfy the deployment-name regex (`modelsCreateDeployment.ts:23`):

```
^[a-z](?!.*--)[a-z0-9\-@.+_]{1,62}(?<!-)$
```

So the helper collapses consecutive hyphens, trims trailing hyphens *after* truncation (the cut itself can leave one), and prefixes a repo that would otherwise start with a digit rather than truncating it — `01-ai/Yi-Large` becomes `m-01-ai-yi-large`, matching the `m-` prefix the platform already applies to autodiscovered entities. Dots are inside the allowed class, so version numbers stay readable (`qwen2.5`). Every slug case in the tests is asserted against the imported regex, so the tests fail if that contract moves.

### Why generated names are truncated to 45, not 52

`schema.ts` budgets the base name against the deployment suffix (`63 - 11 = 52`), but the longest name derived from it is the fileset's `<base>-deployment-hf-src` (18 characters). A 52-character base yields a 70-character fileset name. Random three-word names never came close; HF repo IDs plausibly do.

`wizardBaseNameSchema` is deliberately **not** tightened — it also governs NGC and Workspace, which create no fileset. Only auto-derived names are held to the stricter bound. A manually typed 46–52 character name still hits the pre-existing gap; see Follow-ups.

### Why the pre-flight name check is included

`createHuggingFaceDeployment` creates the fileset **before** the model. With random names a collision was effectively impossible; with deterministic names, redeploying the same repo is an ordinary action. Without a guard the first call to fail would be `filesCreateFileset`, so the user would be told a *fileset* already exists rather than a model — and that fileset would be stranded, because the chain has no rollback. The pre-flight `modelsGetModel` reports the conflict against the model and creates nothing. A transport failure is treated as "not taken", since the create call is still authoritative and returns 409.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no published documentation describes the wizard's naming behaviour; the rationale is captured in code comments beside the change.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pnpm --filter nemo-studio-ui test src/routes/DeploymentsListRoute` | **68 passed (6 files)** — was 39 across 4 |
| `pnpm --filter nemo-studio-ui test .../huggingFaceDeploymentArtifacts.test.ts` | 21 passed |
| `pnpm --filter nemo-studio-ui test .../useHuggingFaceNameDefault.test.tsx` | 6 passed |
| `pnpm --filter nemo-studio-ui typecheck` | clean |
| `pnpm --filter nemo-studio-ui lint` | clean |

The collision tests were confirmed non-vacuous: reverting only `useCreateDeploymentBySource.ts` while keeping the tests makes `fails before creating anything when the model name is taken` fail, and restoring it passes.

`uv run pre-commit run -a` was not run across the whole repository. The pre-commit hook suite ran on the staged files at commit time and passed, including copyright headers, UI lint-staged, and DCO.

Import order in `huggingFaceDeploymentArtifacts.ts` was resolved with `pnpm lint:fix` per the repo's `web/AGENTS.md` guidance rather than hand-sorted.

## Not covered here

End-to-end verification against a running platform — deploying `Qwen/Qwen2.5-7B-Instruct` with the Name field untouched and confirming the created resources are `qwen-qwen2.5-7b-instruct{,-deployment,-config,-deployment-hf-src}` — has not been performed in this PR.

## Follow-ups

- `wizardBaseNameSchema` still budgets against `-deployment` rather than `-deployment-hf-src`, so a manually typed 46–52 character name overflows the fileset name.
- `createHuggingFaceDeployment` has no rollback, so a failure at the config or deployment step still leaves the fileset and model behind.
- Storing the repo ID in `custom_fields` and rendering it as a column would make models findable regardless of their name — complementary to this change.

## Summary by CodeRabbit

- **New Features**
  - Deployment names are automatically suggested from Hugging Face repository IDs and normalized to valid names.
  - Users can edit the suggested name, which then remains unchanged.
  - Existing deployment names are checked before resource creation to prevent conflicts.
  - Customization pages now warn when the base model is not deployed and provide a direct deployment link when available.

- **Bug Fixes**
  - Invalid repository IDs no longer produce unsuitable deployment names.
  - Deployment creation stops and reports lookup errors instead of continuing when name availability cannot be verified.


web/packages/studio/src/routes/DeploymentsListRoute/CreateDeploymentSidePanel/useHuggingFaceNameDefault.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  useEffect(() => {
+    if (source !== SOURCE_HF || nameEdited) return;
+
+    const derived = huggingFaceRepoIdToBaseName(repoId ?? '');
+    // Leave the existing default in place rather than blanking the field while the
+    // user is still part-way through typing a repo ID.
+    if (!derived) return;
+
+    setValue('name', derived, { shouldDirty: false, shouldValidate: true });
```
steramae-nvidia: Could this be written as 
```
if(huggingFaceRepoIdToBaseName(repoId ?? '') {
    setValue('name', derived, { shouldDirty: false, shouldValidate: true });
}
```

steramae-nvidia (approved):

PR #1822, opened 2026-09-04
chore: Improve logging in run_task, execute agent, analyst agent

mikeknep:
## Summary
Improves job logging at three layers.

### run_task

This is a generic function that several (but not all) plugins use in/as their job task entrypoints. We now configure a standardized logging setup here for consistent formatting, and make sure we at least log an error message right before exit-1'ing a failed job.

This is configured to pick up nmp.common observability settings **without** carrying a direct dependency on nmp.common, following the existing pattern used for SDK clients.

### ExecuteAgentJob

Add several log statements so that these jobs aren't silent as the grave

### Analyst agent (fabric adapter)

Configure logging inside the fabric subprocess, and include an exception log when `_run_analysis` raises

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

## Summary by CodeRabbit

- **New Features**
  - Task processes now initialize logging with platform-configured levels and formats.
  - Structured logging providers are supported, with a fallback format when configuration is unavailable.
  - Task and agent logs now include progress, timing, identifiers, and captured error output.
  - Error output is safely bounded to prevent excessive log volume.
  - Logging settings are propagated to job runtimes.

- **Bug Fixes**
  - Failed jobs and analyst runs now record detailed failure information, including tracebacks and result details.
  - Existing logging configuration is preserved, and repeated initialization is handled safely.

### DIFF
--- a/packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/dispatcher.py
+++ b/packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/dispatcher.py
@@ -57,6 +57,7 @@ from nemo_platform_plugin.jobs.constants import (
     PERSISTENT_JOB_STORAGE_PATH_ENVVAR,
 )
 from nemo_platform_plugin.run_dependencies import LocalRunError, resolve_run_kwargs
+from nemo_platform_plugin.tasks.logging_setup import configure_task_logging
 
 logger = logging.getLogger(__name__)
 
@@ -80,6 +81,11 @@ def run_task(
     :class:`~nemo_platform_plugin.job_results.PlatformJobResults` as ``ctx.results``,
     so *sdk* is required when *ctx* is not supplied.
 
+    Root logging is bootstrapped via
+    :func:`~nemo_platform_plugin.tasks.logging_setup.configure_task_logging`
+    before anything else, since a task container configures none of its own.
+    A caller that already configured logging keeps it.
+
     Args:
         job_cls: The :class:`~nemo_platform_plugin.job.NemoJob` subclass to run.
         sdk: :class:`~nemo_platform.NeMoPlatform` handle, typically built via
@@ -93,6 +99,11 @@ def run_task(
     Returns:
         Process exit code suitable for :func:`sys.exit`.
     """
+    # First thing in the process: a task container configures nothing itself,
+    # so without this every message below INFO+lastResort is lost - including
+    # the failure reporting further down. Must precede the first log call.
+    configure_task_logging()
+
     try:
         config = _read_step_config()
     except Exception:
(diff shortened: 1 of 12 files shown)


packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/dispatcher.py
```diff
@@ -155,6 +166,9 @@ def _exit_code_for(result: Any) -> int:
     if not isinstance(result, dict):
         return 0
     if result.get("status") == "failed":
+        # A job that reports failure through its return value never raises, so
+        # nothing else in this process says why the exit code is non-zero.
+        logger.error("Job reported status=failed; result: %s", result)
         return 1
     exit_code = result.get("exit_code")
```
mmogallapalli, line 173: **[P1] Log non-zero exit codes at ERROR**

This documented failure shape still returns `1` without emitting an error-level record. The result is logged earlier at `INFO`, but that message is suppressed when the supported `LOG_LEVEL=ERROR` configuration is used, leaving the failed task with no diagnostic log.

Could we log the result before returning here?

```suggestion
    if isinstance(exit_code, int) and exit_code != 0:
        logger.error("Job reported non-zero exit_code=%s; result: %s", exit_code, result)
        return 1

File: packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/logging_setup.py [hunk 2]
```diff
@@ -0,0 +1,204 @@
...
+
+    ``CommonServiceConfig`` is owned by this package and reads the same
+    ``LOG_LEVEL`` / ``LOG_FORMAT`` environment the services use, so a task
+    honours whatever the deployment sets. Falls back to the field defaults
+    rather than failing: a task with misconfigured settings should still log.
+    """
+    try:
+        from nemo_platform_plugin.config import CommonServiceConfig
+
+        service_config = CommonServiceConfig()
```
mmogallapalli: **[P1] Ensure tasks receive the deployed logging configuration**

Instantiating `CommonServiceConfig()` directly only reads environment variables and defaults; it bypasses the YAML-backed configuration loaded through `Configuration.get_service_config()` / `get_common_service_config()`. The task backends also do not currently propagate `LOG_FORMAT` or `LOG_LEVEL`.

As a result, a Helm deployment configured with `service.log_format: json` can still start these tasks with the default `plain` format. Could we either propagate the resolved logging settings into the task environment or make the same platform configuration available to this bootstrap? Please add coverage for YAML-backed logging configuration as well.

packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/dispatcher.py
mikeknep: Good call, added in dff96e614

packages/nemo_platform_plugin/src/nemo_platform_plugin/tasks/logging_setup.py
mikeknep: Great catches here, see 49a41df33

mmogallapalli (approved):

PR #1847, opened 2026-09-05
refactor(sdk): route platform clients through owned endpoints

ironcommit:
## TL;DR

This refactors platform SDK and `NemoClient` construction so generated and plugin clients use endpoint-owned, immutable HTTP clients with service-aware routing instead of shared mutable clients or module-level test fallbacks.

## What changed

- Platform endpoint resolution now carries service routes from configuration, environment overrides, and configured local services. SDK-owned transports route `/apis/<service>/...` requests to the matching TCP or UDS service endpoint while leaving non-API paths and already-service-specific URLs alone.
- SDK factory construction now uses routed `NeMoPlatform` subclasses and preserves routing through `copy()` / `with_options()`, including typed plugin client adapters, instead of attaching request routing to individual SDK instances.
- SDK-owned HTTP clients are frozen after construction, blocking mutation of base URL, headers, cookies, params, timeouts, hooks, and trust settings so derived SDK handles cannot leak request or session state into each other.
- Test harnesses and affected service call sites no longer depend on module-level `_test_http_client` fallbacks. In-process clients are passed explicitly, request-scoped SDKs are reused for files/jobs flows, and jobs log filtering normalizes `job_attempt` to a string.

## Why

The previous lifecycle mixed per-request SDK state with reusable HTTP client state, which made routing and auth propagation harder to reason about across production, service-discovery, UDS, plugin adapter, and in-process test contexts. Owning routing at the endpoint/SDK layer keeps transport configuration stable while allowing per-request headers and scoped SDK options to stay on the SDK instances that need them.

## Summary by CodeRabbit

- **New Features**
  - Added endpoint-aware routing across configured, environment-based, TCP, and Unix-socket service endpoints.
  - Added protected HTTP clients that prevent unintended changes to connection settings, headers, cookies, and event hooks.
  - SDK clients now support custom HTTP client injection and managed synchronous and asynchronous lifecycles.

- **Improvements**
  - Requests retain service routing when SDK instances are copied or reused.
  - Jobs, Files, and result downloads now use consistent SDK handling and delegated authentication.
  - Improved routing diagnostics while keeping sensitive request details redacted.


packages/nmp_common/src/nmp/common/immutable_http_client.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def update(self, headers: HeaderTypes | None = None) -> None:
+        _raise_immutable_client_mutation_error()
+
+
+class _ImmutableCookies(httpx.Cookies):
+    def extract_cookies(self, response: httpx.Response) -> None:
+        # httpx normally persists response cookies on the client. SDK clients
+        # should not carry request/session state between derived SDK handles.
+        pass
```
mckornfield, line 80: would we not want to raise here?

services/core/inference-gateway/src/nmp/core/inference_gateway/testing/fixtures.py
```diff
@@ -66,23 +66,16 @@ def _build_app_context(
...
-       config snapshot (captured at first import), so a ``service_configs``
-       override is too late. Patch the snapshot field to 0 *before*
-       entering ``create_test_client`` and ``on_startup`` never schedules
-       the loop.
-    2. The shared SDK HTTP client's ``aclose``. Plugins like
-       ``nemo-guardrails`` call ``await sdk.close()`` in ``on_shutdown``,
-       which would close the shared client for every later test in the
-       module. Patch ``aclose`` to a no-op for the module's lifetime;
-       ``ASGITransport`` is in-process so nothing actually leaks.
+    One module-scope hazard is neutralised here:
```
mckornfield, line 69: lol this phrasing

mckornfield (approved):

packages/nmp_common/src/nmp/common/immutable_http_client.py
ironcommit: this is a normal path way, adding an exception would break normally functioning code I think.  leaving as is.

PR #1856, opened 2026-09-07
fix(customizer): vendor nemo-gym in the RL image so native-v1 can install

anubhutivyas:
## Summary
For every Gym server, Gym checks whether it is an editable install in the working directory. If it is not, it installs nemo-gym from a wheel pinned to the parent venv's version: https://github.com/soluwalana/Gym/blob/nmp/customizer/nemo_gym/cli/setup_command.py#L80

A staged environment FileSet is never an editable install, so every native-v1 server takes that path and asks an index for `nemo-gym==0.5.0rc0`. That is our fork's version and it is not published anywhere, PyPI has 0.5.0 and 0.5.1, no rc0. So the server venv never builds and the job fails at spin-up
```
    No solution found when resolving dependencies:
    Because there is no version of nemo-gym==0.5.0rc0 ...
```
This affects every native-v1 environment.

## Changes
- Build the wheel from the Gym submodule already in the image and set `UV_FIND_LINKS=/opt/gym-wheels`. Gym's install then finds it locally, and everything else still comes from the index as before.
- Letting it fall through to PyPI's `0.5.0` is not an option, the server venv would run upstream Gym while the actor runs our fork.
- `UV_FIND_LINKS` is additive and takes multiple paths, so it does not change how any other dependency resolves. It holds one wheel: nemo-gym.
- Gym relies on setuptools-scm's git file-finder to put configs in the wheel, and there is no git in this build. The build script declares that data explicitly and fails if it is missing, so we cannot ship a wheel that installs cleanly but has no YAML in it.

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

* **New Features**
  * NeMo-Gym is now packaged and included in the RL base image, allowing installations to use the bundled wheel.
  * Required Gym configuration and documentation resources are included with the packaged component.

* **Bug Fixes**
  * Improved package validation helps ensure the bundled wheel is complete, correctly versioned, and contains required files.

* **Chores**
  * Updated the NeMo-RL base image to a newer pinned source revision.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -128,7 +128,7 @@ variable "NEMO_RL_REPO" {
 # RL pins Gym as a git submodule (-> soluwalana/Gym over https), so Gym rides in with the RL git ADD
 # - no separate Gym pin needed.
 variable "NEMO_RL_REF" {
-  default = "07090a9875e8535a185e925e5f1ebd137e7f4d6b" # soluwalana/RL nmp/customizer
+  default = "9932dc8aa63a55fd431670d1b7c9d0bf3b2d2373" # soluwalana/RL nmp/customizer
 }
 variable "RL_BASE_CONTEXT" {
   default = ""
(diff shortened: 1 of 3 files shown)


gabwow (approved):
nit: I think some of the agent comments are hurting readability rather than helping (particularly in the nmp-rl-base). If you agree feel free to remove the ones you fine unneeded. However, this is non-blocking and if you find them helpful, leave them in.

PR #1862, opened 2026-09-08
feat(agents): strategy-based `nemo agents optimize`

stefan-kickoff:
Introduces a `nemo-agent-optimization` plugin that owns `nemo agents optimize` and
dispatches to optimization strategies discovered from installed plugins, plus the
first four strategy implementations.

62 files, +4,226/−77, across 24 commits.

## What changed

**New plugin: `nemo-agent-optimization`.** Owns the `nemo agents optimize` job and
CLI. Discovers strategies through a new entry-point group,
`nemo.optimization-strategy`, and dispatches on `--strategy`. A plugin may register
multiple strategies; each takes an agent plus a strategy-specific config, and the
config format is the strategy's own business.

```
nemo agents optimization-strategies list
  experimentalist  nat  prompt-master  switchyard
```

**Four strategies implemented:**

| Strategy | Plugin | Behaviour |
|---|---|---|
| `nat` | `nemo-optimization` | Existing numeric HPO, wrapped behind the strategy contract |
| `prompt-master` | `prompt-master` | Runs the bundled Prompt Master skill via a deepagents Fabric agent |
| `switchyard` | `nemo-switchyard` | Rewrites a Fabric agent's model to a Switchyard-routed virtual model |
| `experimentalist` | `nemo-experimentalist` | Stub — raises and points at `nemo agents experimentalist` |

All four expose byte-identical `run` / `validate_config` signatures.

**Renamed** `nemo-prompt-master` → `prompt-master` (distribution `prompt-master-plugin`,
module `prompt_master_plugin`). `NOTICE` and `.copyrightignore` repointed; the vendored
MIT skill files are byte-identical to upstream.

**`nemo-agents` and `nemo-optimization` are near-unchanged.** 16 files between them,
+210/−59, limited to: repointing the `agents.optimize` entry point and HTTP route,
registering each plugin's strategy, the regenerated OpenAPI spec, and `--strategy`
added to docs and examples. No behavioural change to the HPO code itself.

## Testing

- 1,101 passed / 51 skipped across the five touched plugins in a single pytest
  invocation (exercises multi-directory collection)
- 1,673 passed in `plugins/nemo-agents/tests`
- `ruff check`, `ruff format --check`, and `ty check` clean on all touched files
- `tools/lint/lint-openapi.sh` exits 0
- `docs/cli/reference.mdx` and `cli-summary.mdx` regenerate byte-identical

## Known limitations

- **Strategy execution is verified in-process only.** No live-platform run happened
  (no local platform available during development). The Switchyard payload is the
  best-covered unverified path — it's validated against the real vendored
  `RandomRoutingFactory` rather than a stub.
- **`prepare-fileset` only accepts NAT-shaped configs.** Its preflight requires an
  `optimizer:` section, so non-`nat` strategies have no first-class staging path and
  must stage via `nemo files filesets create` + `upload` (as documented in the
  prompt-master README). The `FILESET_REQUIRED` message and the
  `optimize_config_fileset` field help both still point at `prepare-fileset`. Worth a
  follow-up issue; fixing it means touching `nemo-optimization`, which this PR
  deliberately leaves alone.
- **Strategies return artifact references, not new agent entities.** Registering the
  result as a new agent is deferred. Note that the primary artifact's schema currently
  depends on the input form — a local `agent.yaml` yields `nemo-agents-spec-v1`, a
  platform agent ref yields the translated `fabric.agent/v1alpha1` package — which that
  later work will need to reconcile.
- `uv.lock` drops one `unsloth` sdist line (same version, same wheel hash, wheel entry
  untouched). Unrelated churn, benign.

### DIFF
--- a/plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
+++ b/plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""OptimizeJob — Agents numeric HPO (``nemo agents optimize``).
+"""OptimizeJob — strategy-based Fabric agent optimization (``nemo agents optimize``).
 
 Implementation lives in ``nemo_optimization``; registration and HTTP mounting
 are owned by the agents plugin (``agents.optimize``).
@@ -51,7 +51,8 @@ from pydantic import BaseModel
 from nemo_optimization.agents import resolve_agent_config
 from nemo_optimization.preflight import preflight_validate_llm_models
 from nemo_optimization.router import OptimizeRouter
-from nemo_optimization.schemas.optimize import FILESET_REQUIRED, OptimizeSpec, OptimizeSubmitSpec
+from nemo_optimization.schemas.optimize import FILESET_REQUIRED, OptimizeSpec, OptimizeStrategy, OptimizeSubmitSpec
+from nemo_optimization.strategies import PRIMARY_ARTIFACT_KEY, discover_optimization_strategies
 
 logger = logging.getLogger(__name__)
 
@@ -67,10 +68,10 @@ OPTIMIZE_TASK_IMAGE = "nmp-cpu-tasks"
 
 
 class OptimizeJob(NemoJob):
-    """Run a Fabric-native numeric optimize study via the Agents optimize job."""
+    """Run the selected Fabric agent optimization strategy."""
 
     name: ClassVar[str] = "optimize"
-    description: ClassVar[str] = "Optimize a Fabric agent workflow (numeric HPO)."
+    description: ClassVar[str] = "Optimize a Fabric agent with HPO or Prompt Master."
     container: ClassVar[str] = "cpu-tasks"
     job_collection_path: ClassVar[str | None] = None
     generate_legacy_verbs: ClassVar[bool] = False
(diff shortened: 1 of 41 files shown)


yamini:
Good design! Didn't spot any bugs or issues, just a couple of product questions: 
- Sweep support: you mentioned we could eventually pick a few strategies and run them in parallel. `RunStrategyJob` right now only produces one job record with no child submission and no polling, so sweep isn't matter of adding a new parameter. It'll need a different execution model. Are you thinking to extend this job to fan out, or build a separate orchestration layer that calls run-strategy multiple times? That'll also influence whether "router" is still the right name for this feature.
- Experiments API: does `nemo-optimization` report results into the Experiments API today? If not, can we figure out that contract before adding more strategies? I think the pattern should be to surface all the results via Experiments.
- Naming: `nemo-agents`, `nemo-agent-optimization`, and `nemo-optimization` are three packages doing three different jobs. You already flagged that nemo-optimization needs a rename. Do you have a timeline on that?
- API compatibility: the old `/apis/agents/.../jobs/optimize` route is gone and there's no deprecation path. Confirming that's intentional and external callers are covered.

yamini (approved):

plugins/nemo-agent-optimization/README.md
```diff
@@ -0,0 +1,109 @@
...
+## The strategy contract
+
+A strategy is an ordinary `nemo.jobs` entry whose class declares one extra
+class variable:
+
+```python
+class OptimizeJob(NemoJob):
+    nemo_agent_optimization_strategy: ClassVar[OptimizationStrategy] = OptimizationStrategy(
+        name="nat",
+        description="Numeric hyperparameter search (Optuna) over a Fabric agent workflow.",
```
gabwow: This will now be Optuna and GA prompt optimization

plugins/nemo-optimization/src/nemo_optimization/jobs/optimize.py
```diff
@@ -69,6 +72,16 @@ class OptimizeJob(NemoJob):
...
     name: ClassVar[str] = "optimize"
+    #: Marks this job as an agent optimization strategy, names it for
+    #: ``nemo agents optimize run-strategy --strategy``, and says what it optimizes for
+    #: ``list-strategies``. Declaring the variable is the whole contract — nothing to
+    #: subclass, and no strategy-specific entry-point group to join. The description is the
+    #: strategy's, not the job's: ``description`` below introduces the job to CLI users,
+    #: while this one tells a caller choosing a ``--strategy`` what this one does.
+    nemo_agent_optimization_strategy: ClassVar[OptimizationStrategy] = OptimizationStrategy(
+        name="nat",
+        description="Numeric hyperparameter search (Optuna) over a Fabric agent workflow.",
```
gabwow: This will now be Optuna and GA prompt optimization

plugins/nemo-agent-optimization/src/nemo_agent_optimization_plugin/jobs/run_strategy.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def _strategy_payload(spec: RunStrategySpec) -> dict[str, Any]:
+    """The router's fields minus ``strategy``, which the strategy job already knows."""
+    payload = spec.model_dump(mode="json")
+    payload.pop("strategy", None)
+    return payload
+
+
+def _strategy_spec(target: type[NemoJob], spec: RunStrategySpec, *, error: type[Exception]) -> Any:
```
gabwow: Never validates target.input_spec_schema or calls target.to_spec(...). Any ordinary NemoJob that resolves references, applies remote-specific validation, or transforms its submitted model in to_spec will receive the wrong input.

plugins/nemo-optimization/pyproject.toml
```diff
@@ -20,6 +26,21 @@ dependencies = [
...
 
+# Discovered by `nemo agents optimize` as the `nat` strategy, via the
+# `nemo_agent_optimization_strategy` class variable on OptimizeJob. The key sits under the
+# dispatcher's `agent-optimization` plugin namespace rather than minting a new one: plugin
+# allowlists and `nemo plugins list` group `nemo.jobs` entries by that prefix, so a deployment
+# that allows the router also gets its strategies. Nothing mounts the key itself — the
+# agent-optimization service adds routes explicitly and has no `nemo.cli` group — so the study
+# stays discoverable, not separately submittable; it runs as a step of the router job's record.
+[project.entry-points."nemo.jobs"]
+"agent-optimization.optimize" = "nemo_optimization.jobs.optimize:OptimizeJob"
```
gabwow: Question: I'm not quite sure what the intended behavior is: do you want strategies to load all or nothing or the user to specify individual strategies? Are all strategies supposed to be listed as agent-optimization?

plugins/nemo-optimization/pyproject.toml
```diff
@@ -20,6 +26,21 @@ dependencies = [
...
+
+# Hangs `prepare-fileset` off the shared `nemo agents optimize` group. The value is a plain
+# function taking that group; joining this group is independent of the strategy job above.
+[project.entry-points."nemo.cli.agents.optimize"]
+prepare-fileset = "nemo_optimization.optimize_cli:register_prepare_fileset_command"
```
gabwow: I don't think this is going to list because I think 
NEMO_PLUGIN_ALLOWLIST gets set to agents,agent-optimization,optimization
with the current logic.

gabwow (commented):
A few changes requested

docs/agents/optimization.mdx
```diff
@@ -277,9 +277,12 @@ nemo files list nemo-agent-telemetry
 
 ## Run Prompt and Parameter Tuning
 
-The `nemo agents optimize` command submits Fabric-backed numeric
-optimization through `agents.optimize` (implementation in
-`nemo-optimization`). Input must be a Fabric-native agent package
+The `nemo agents optimize run-strategy` command submits Fabric-backed numeric
+optimization. `nemo agents optimize` is a shared command group, one per
+installed optimization strategy; `--strategy nat` selects the Fabric-backed
```
gabwow: wide reaching comment that I'll only mention here: "nat" as a strategy probably will mean nothing to a lot of users. "legacy" won't tell them about the expected algorithm but would at least have semantic meaning

stefan-kickoff:
> Good design! Didn't spot any bugs or issues, just a couple of product questions:
> 
> * Sweep support: you mentioned we could eventually pick a few strategies and run them in parallel. `RunStrategyJob` right now only produces one job record with no child submission and no polling, so sweep isn't matter of adding a new parameter. It'll need a different execution model. Are you thinking to extend this job to fan out, or build a separate orchestration layer that calls run-strategy multiple times? That'll also influence whether "router" is still the right name for this feature.

It will be one job per strategy.

> * Experiments API: does `nemo-optimization` report results into the Experiments API today? If not, can we figure out that contract before adding more strategies? I think the pattern should be to surface all the results via Experiments.

That's the plan, but right now Experiments are tied to Evaluations, while an Optimization doesn't necessarily need to be evaluated, so it requires some thinking.

> * Naming: `nemo-agents`, `nemo-agent-optimization`, and `nemo-optimization` are three packages doing three different jobs. You already flagged that nemo-optimization needs a rename. Do you have a timeline on that?

Next week?

> * API compatibility: the old `/apis/agents/.../jobs/optimize` route is gone and there's no deprecation path. Confirming that's intentional and external callers are covered.

Yes that's intentional and Studio (the only caller) has been updated.

docs/agents/optimization.mdx
stefan-kickoff: Thoughts on `nat_legacy`?

plugins/nemo-optimization/pyproject.toml
stefan-kickoff: I'm not sure I follow. I want all pip installed strategies to be loaded, and they're all listed by `list-strategies`. For example (and I'm going to pretty-print this later), this is the current output:
```
% nemo agents optimize list-strategies
Targeting http://localhost:8080/
nat
```

plugins/nemo-optimization/pyproject.toml
stefan-kickoff: Where is that set?

PR #1863, opened 2026-09-08
feat: Auto wire agent telemetry

mikeknep:
## Summary
Auto-configure telemetry to intake for agent execute jobs and deployments

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

## Summary by CodeRabbit

- **New Features**
  - Execute-agent jobs can automatically export agent trajectories to Intake when supported.
  - The `auto-telemetry` option configures missing Intake destinations, authentication, and agent metadata.
  - Telemetry can be enabled, disabled, or left unspecified; explicit destinations and opt-outs are preserved.
  - Fabric deployments and Insights analyst runs support Relay-based telemetry with scoped configuration.

- **Bug Fixes**
  - Unsupported configurations no longer receive automatic telemetry wiring.
  - Relay settings are restored after each analyst run, preventing cross-run configuration leakage.

- **Tests**
  - Added coverage for telemetry setup, span recording, token handling, and configuration isolation.


plugins/nemo-agents/src/nemo_agents_plugin/agent_config.py
```diff
@@ -70,7 +70,11 @@ class RuntimeConfig(BaseModel):
 class TelemetryConfig(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
-    enabled: bool = False
+    # Tri-state so a config can decline telemetry without being mistaken for one
+    # that never mentioned it: unset lets the backend wire an export for this
+    # deployment context, False opts out, True turns it on and still lets the
+    # backend fill in whatever the config left out.
+    enabled: bool | None = None
```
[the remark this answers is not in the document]
mikeknep: This suggestion misreads the tri-state. `None` means "the backend may wire an export," not "telemetry is on." The backends set `enabled=True` explicitly when they wire; `_apply_telemetry` treating `None` as off is the correct fallback for "nobody wired it" — which is exactly what should happen for an adapter that doesn't support Relay, or when `NMP_BASE_URL` is missing.

The proposed guard here enables Relay with no exporters at all for every config that omits telemetry. That's instrumentation overhead with nowhere to send anything, applied to agents that never asked for telemetry. The current guard is right; if anything the fix is to make `_apply_telemetry` reject an `enabled=True` config with no exporters configured.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -636,6 +650,47 @@ def _validate_agent_config_format(config_format: str) -> None:
...
+    one.
+    """
+    base_url = os.environ.get(NMP_BASE_URL_ENVVAR)
+    if not base_url:
+        logger.warning("%s is not set; the agent will run untraced.", NMP_BASE_URL_ENVVAR)
+        return
+
+    headers = get_forwarding_headers(sdk) if sdk is not None else {}
+    for name, value in headers.items():
+        os.environ[_header_envvar(name)] = value
```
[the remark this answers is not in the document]
mikeknep: This concern is real in kind, but the marginal exposure is approximately zero and the proposed alternatives are worse.

The task process already has `NMP_PRINCIPAL` in its environment — the jobs backend puts it there by design, and `get_forwarding_headers` derives our headers from it. An agent that can read env vars can already reconstruct the same identity. We're not adding a credential; we're adding a rendering of one that's already present.

"Keep the headers in a trusted exporter path" isn't available: `header_env` is Relay's contract — it resolves headers from named environment variables at export time. The only way to avoid env entirely is inline headers in the config, which we deliberately rejected because Fabric writes the resolved config into run artifacts that get uploaded as a downloadable job result. That trades env exposure (to a process already holding the principal) for artifact exposure (to anyone who can download job results). Env is clearly the better side of that trade.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -783,6 +802,72 @@ def _validate_agent_config_format(config_format: str) -> None:
...
+    Fabric writes the resolved agent config into the run's artifacts, and those
+    are uploaded as a job result, so an inline header would be a downloadable
+    one.
+    """
+    base_url = os.environ.get(NMP_BASE_URL_ENVVAR)
+    if not base_url:
+        logger.warning("%s is not set; the agent will run untraced.", NMP_BASE_URL_ENVVAR)
+        return
+
+    headers = get_forwarding_headers(sdk) if sdk is not None else {}
```
mmogallapalli, line 839: Workload-identity jobs will not authenticate this Relay request correctly. `get_forwarding_headers()` only copies the SDK's configured headers, while the workload-identity bearer token is added dynamically by the SDK's auth handler. As a result, this returns only `X-NMP-Internal`, and Relay's raw request to Intake will not include `Authorization`.

Could we provide Relay an equivalent workload-identity authentication path and add a test covering telemetry export with `NMP_WORKLOAD_IDENTITY_TOKEN_FILE` set?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -396,6 +409,12 @@ def run(self, config: dict, *, ctx: JobContext, sdk: NeMoPlatform | None = None)
 
         fabric_dirs = FabricDirectories.create(agent_config, ctx.storage.ephemeral)
 
+        if step_config.request.telemetry and _adapter_supports_relay(agent_config, fabric_dirs.base):
```
mmogallapalli: The request field documents `telemetry: false` as “run untraced,” but this condition only disables automatic wiring. If the agent config already contains enabled telemetry, it is passed unchanged to Fabric and Relay still exports the trajectory.

Could we explicitly disable telemetry in the effective agent config when the request-level option is `false`, or change the documented contract if this is intended to mean only “disable auto-wiring”?

plugins/nemo-agents/src/nemo_agents_plugin/telemetry/intake_export.py
```diff
@@ -0,0 +1,103 @@
...
+        return False
+
+    storage: dict[str, object] = {
+        "type": "http",
+        "endpoint": f"{base_url.rstrip('/')}{INTAKE_ATIF_INGEST_PATH.format(workspace=workspace)}",
+    }
+    if header_env:
+        storage["header_env"] = dict(header_env)
+
+    atif = dict(telemetry.atif or {})
```
mmogallapalli, line 136: More of a nit but my agent called this out, this overrides an explicit ATIF opt-out. For example:

```yaml
telemetry:
  enabled: true
  atif:
    enabled: false
```

is rewritten with atif.enabled: true and an Intake destination. Since the helper is intended to preserve explicitly declared telemetry configuration, could we return without wiring when telemetry.atif.enabled is explicitly false?

```suggestion
    atif = dict(telemetry.atif or {})
    if atif.get("enabled") is False:
        return False
    atif["enabled"] = True
```

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -668,6 +669,14 @@ async def create_deployment(
...
 
         if is_fabric:
+            # Wire the trajectory export before the rebase below, so the
+            # endpoint it produces is rewritten for reachability alongside the
+            # inference URLs. No header_env here: the auth-proxy sidecar stamps
+            # identity on the way out, which is why the workload needs none.
+            # config is the caller's deployment entity; rewrite_fabric_config_base_urls
+            # deep-copies for the same reason, and wiring runs before it.
+            config = copy.deepcopy(config)
+            configure_intake_atif_export(config, workspace=workspace, base_url=rewrite_target)
```
mmogallapalli: This auto-wires Relay into every Fabric deployment without checking whether the selected adapter supports Relay and ATIF. Fabric rejects that telemetry configuration for adapters whose descriptors do not advertise those capabilities, so an otherwise valid custom adapter can become undeployable by default.

Could we apply the same capability guard used by the execute-job path here, including verifying that `providers.relay.outputs` contains `atif`, before modifying the configuration?

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
```diff
@@ -16,9 +20,12 @@
 from nemo_platform_plugin.nooa_model_client import ConfiguredModelRefs
 from nemo_platform_plugin.sdk_provider import get_async_task_sdk
 from nemo_platform_plugin.tasks.logging_setup import configure_task_logging
+from nemo_relay import plugin as relay_plugin
```
mmogallapalli, line 23: This introduces a direct runtime import of `nemo_relay`, but `nemo-insights-plugin` does not declare `nemo-relay` as a dependency. It is currently available only transitively in full platform environments through `nemo-agents-plugin`, so installing Insights independently can fail while importing this adapter.

Could we add the Relay package to `plugins/nemo-insights/pyproject.toml`, or make this an optional/lazy import if Relay support is intended to remain optional?

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
```diff
@@ -783,6 +802,72 @@ def _validate_agent_config_format(config_format: str) -> None:
...
+    """
+    try:
+        plan = Fabric().plan(translate_agent_config(agent_config), base_dir=base_dir)
+        descriptor = plan.to_dict().get("adapter_descriptor") or {}
+        providers = descriptor.get("descriptor", descriptor).get("telemetry", {}).get("providers", {})
+        supported = "relay" in providers
+    except Exception:
+        # Planning failures are the invocation's to report, with its own
+        # diagnostics; here they only mean we cannot know, so do not wire.
+        logger.warning("Could not read adapter telemetry support; the agent will run untraced.", exc_info=True)
```
maxdubrinsky: had a few agents flag this as potentially problematic since this `except Exception` can be hit by any plan validation errors

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
mikeknep: Nice catch, addressed in 60d7b0403. Worth calling out: there is no mechanism to refresh the authorization token, so if a job runs longer than the duration that token is valid, it will start/continue exporting with a stale token. I'll make a ticket to address this.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
mikeknep: Renamed the request field to `auto_telemetry: bool`. False = do exactly what the agent config says and nothing more; True = respect the agent config, but fill in the gaps (atif, atif storage configuration)

plugins/nemo-agents/src/nemo_agents_plugin/telemetry/intake_export.py
mikeknep: Improved in the latest batch of commits

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
mikeknep: Now properly verifying atif support

plugins/nemo-insights/src/nemo_insights_plugin/fabric_adapter.py
mikeknep: Added as a standard dep

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute.py
mikeknep: This is marked outdated on GitHub, but only because this function was moved to `intake_export.py`. That said, the broad exception here is OK because we already have true validation error handling at a higher level. Within this function we just want to check if atif is supported, which shouldn't produce a job-stopping crash—we'd rather the job instead run without the telemetry export.

packages/nemo_platform/pyproject.toml
```diff
@@ -395,6 +395,7 @@ nemo-insights-plugin = [
   "nemo-fabric-adapter-contract>=0.3.0b1",
   "nemo-fabric-adapters-common>=0.3.0b1",
```
benmccown, line 396: If I'm tracking correctly, this is a user facing problem that will indeed cause impact.

---

**Wrapper wheel omits the Insights Fabric adapter descriptor**
*Priority: P1 · non-blocking · `packages/nemo_platform/pyproject.toml` (bundle entry, line 867) · found by agent review (gpt-sol)*

#### Finding: Agent Summary

The standalone `nemo-insights` wheel ships `insights-analyst.fabric-adapter.json` as `shared-data` (→ `share/nemo-fabric/adapters/…`), but the wrapper `nemo-platform` wheel's bundle entry doesn't carry it forward, so a packaged install can't discover `nvidia.fabric.insights-analyst`.

- **What:** the Insights adapter descriptor isn't included in the bundled `nemo-platform` wheel.
- **Where:** `packages/nemo_platform/pyproject.toml:867` (the `nemo-insights-plugin` bundle entry) vs `plugins/nemo-insights/pyproject.toml:57-58`.
- **When:** only in a *packaged wheel* install — editable/dev installs have the descriptor, so it passes locally and in the existing descriptor test.
- **Why it matters:** Fabric can't resolve the adapter, so the new Insights Relay path (and possibly the analyst adapter itself) fails or silently falls back to untraced in the shipped artifact.

#### Finding: Agent Breakdown

The bundle entry at line 867 declares only `source` + entry-point `inherit` — no `force_include` for the descriptor. The build hook (`packages/nmp_build_tools/src/nmp_build_tools/hatch.py:80-102`) copies each bundle's source module plus explicit `force_include` entries only; it doesn't read the standalone plugin's `[tool.hatch.build.targets.wheel.shared-data]`. For contrast, the `nemo-agents-plugin` entry (line 856) `force_include`s its vendored adapters explicitly — the same mechanism this needs.

The only descriptor test (`plugins/nemo-insights/tests/test_analyst_fabric_descriptor.py`) checks the standalone editable install, not the wrapper-wheel layout, which is why nothing flags it today.

Likely fix: add a `force_include` for `insights-analyst.fabric-adapter.json → share/nemo-fabric/adapters/insights-analyst/…` on the wrapper's `nemo-insights-plugin` bundle entry, and add a wrapper-wheel-layout test.

```
Verify this finding against the current code and only fix it if still present. In packages/nemo_platform/pyproject.toml, the [tool.bundle-package] entry `nemo-insights-plugin` (~line 867) does not force_include the insights-analyst.fabric-adapter.json that the standalone plugin installs as shared-data. Confirm whether the bundled nemo-platform wheel ships the descriptor to share/nemo-fabric/adapters/insights-analyst/; if not, add a force_include mapping mirroring how nemo-agents-plugin force_includes its vendored adapters, and add a test asserting the descriptor is present in the wrapper wheel layout. Preserve Fabric adapter discovery. Run the narrowest packaging/descriptor test.
```

benmccown (approved):
Passing along approval since I'm OOO tomorrow, I trust Manjesh and Max will close out their more thorough reviews with you, I just wanted to pass along the one finding my agent review caught.

---

**Agent Summary**

Fresh multi-model agent review at `daf95298` (after the round of fixes for the deployment guard, ATIF tri-state, `auto_telemetry` rename, and token exchange — all verified correct + tested). One finding worth surfacing: the wrapper `nemo-platform` wheel likely doesn't bundle the Insights Fabric adapter descriptor. The rest were test-coverage completeness (an `atif.enabled:false` regression test, a deployments-side negative-guard test) and minor notes.

packages/nemo_platform/pyproject.toml
mikeknep: Hmm. If this finding is correct, I'm surprised it's not causing problems in the e2e tests and/or in the dev instance. I'll investigate, and might punt to a separate PR

mmogallapalli (approved):

packages/nemo_platform/pyproject.toml
mikeknep: I'm going to land this PR now, and pick up this comment as a separate effort.

PR #1875, opened 2026-09-08
chore(rl): point the episode broker venv at the sandboxed-gym wheel

SandyChapman:
## Summary

NeMo-RL is moving its sandboxed gym onto the `sandboxed-gym` wheel ([AALGO-583](https://linear.app/nvidia/issue/AALGO-583/share-the-sandboxed-gym-implementation-as-a-sandboxed-gym-wheel)), which deletes `nemo_rl.environments.sandbox.broker_actor`. The broker does **not** go away — NeMo-RL keeps running it in a Ray actor of its own — but the actor is now the wheel's, so its FQN changes and Ray derives a different venv path from it. This repo names the old FQN in two places and would fail the build on it.

Before: the image prefetches a venv for `nemo_rl.environments.sandbox.broker_actor.SandboxEpisodeBrokerActor`. After: the same venv, named for `sandboxed_gym.ray.broker_actor.SandboxEpisodeBrokerActor`. Still eight filters, nine venvs.

## ⚠️ Blocked on two things — do not merge first

1. **The NeMo-RL change.** `test_venv_contents` asserts `sandboxed_gym` is in the `SandboxedGymActor` venv, true only once NeMo-RL adds the dependency.
2. **#1898** (`feat(sandboxed-gym): let a caller choose where the episode broker runs`), which is what lets NeMo-RL keep a Ray-hosted broker at all.

## Changes

- `docker/rl/Dockerfile.nmp-rl-base` — the existence check moves to the wheel's FQN. The prefetch **filter** is unchanged: it is a substring match, so `broker_actor.SandboxEpisodeBrokerActor` matches the new FQN just as well.
- `tests/smoke_gpu/test_rl_training.py` — the broker's per-actor venv entry is rekeyed to the wheel's FQN; `sandboxed_gym` is added to the `SandboxedGymActor` venv's expectations; `test_sandboxed_gym_driver_imports` imports `sandboxed_gym.host.models` rather than the deleted `nemo_rl.environments.sandbox.host.models`.
- `services/rl/.../grpo_driver.py` — comment now names `sandboxed_gym.runtime.gym_host_runtime`.

### Why the broker keeps its own process

The wheel's orchestrator hosts the broker in-process by default, which is right for Evaluator — an ordinary evaluation opens no episodes at all. NeMo-RL is the opposite case: a SWE rollout crosses the broker on every episode `exec`, while the same process postprocesses rollout results (tokenizer-bound). The NeMo-RL fork gave the broker its own Ray actor for exactly that reason, and #1898 restores the ability to choose. Ray supplies placement and lifecycle only — the HTTP server runs on a background thread either way, so nothing serializes through actor dispatch.

Prefetching it is deliberate: an unprefetched venv is built on the node at first use, turning a build-time error into a slow job. That is the same reasoning the existence check below the prefetch already encodes for every other actor.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Existing tests cover changed behavior — justification: `tests/smoke_gpu/test_rl_training.py` is the test for image contents, and is updated here to match. No new behaviour is introduced.
- [x] Documentation not applicable — justification: no user-facing surface changes; the reasoning lives in the Dockerfile comment beside the prefetch list.

## Verification

Targeted validation — built the image from an earlier revision of this branch against a NeMo-RL checkout carrying the wheel change (`--build-context nemo-rl=<checkout>`), 0 errors:

- the `SandboxedGymActor` venv is present with `sandboxed_gym` installed
- a **real Gym host boots** from the packaged `gym_host.sh` + `gym_host_runtime.py` in ~90s and serves `/rollouts/run`; a two-rollout batch round-trips and joins back to its prompts
- hooks on the changed files: `ruff`, `ruff format`, `ty`, copyright, merge-conflict — all pass

**Not yet rebuilt since the broker venv was renamed back in.** That earlier build was of the delete-the-broker revision, so it confirms the `SandboxedGymActor` half but not that a venv appears at the new broker FQN — which additionally needs the NeMo-RL registry entry. `tests/smoke_gpu` itself needs GPUs and a published image and was not run; its assertions were checked by hand against the local build.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable sandboxed Gym versions.
  * Expanded compatibility with additional Nemotron model layer layouts.
  * Added support for newer checkpoint configuration options.

* **Bug Fixes**
  * Improved asynchronous checkpoint saving and finalization reliability.
  * Ensured checkpoint status updates occur only after successful completion.
  * Updated sandboxed Gym runtime and broker integration for the current module layout.


anubhutivyas (commented):
lgtm, be sure to update the RL ref after merging https://github.com/soluwalana/RL/pull/25

PR #1910, opened 2026-09-09
feat(evaluator): capture Relay's OTLP export so Fabric trials carry an OTLP trace

SandyChapman:
Implements [AALGO-601](https://linear.app/nvidia/issue/AALGO-601/decide-how-fabric-emits-otlp-relays-exporter-is-endpoint-only-so-there).

## Summary

Fabric was the only runner family with no OTLP trace at all — both its runtimes registered ATIF — so a metric written against the OTLP path that [AALGO-569](https://linear.app/nvidia/issue/AALGO-569/switch-evaluators-primary-trajectory-format-from-atif-to-otlp) makes primary got nothing from a Fabric trial. Relay cannot write OTLP to disk, so the runner now listens for it: a loopback receiver per task, capturing each export to `<evidence>/traces/relay.jsonl`.

Before: a Fabric trial's `trace` evidence was always the ATIF file. After: it is the OTLP capture when one was recorded, with ATIF still addressable at `trace:atif`.

## Changes

- **`fabric/otlp_receiver.py`** (new) — the endpoint Relay exports to: loopback `HTTPServer` on an OS-assigned port, accepting `POST /v1/traces` as `application/x-protobuf` and storing each export verbatim. 404 on another path, 415 on another media type, 400 on a malformed or negative `Content-Length`, 413 past the size ceiling. The request target is parsed rather than compared raw, so an absolute-form target or a query string still resolves. **Standard library only, deliberately** — the container runtime (#1915) seeds this same file into a sandbox that carries Fabric and nothing of this package, so there is one endpoint with two ways of running it rather than an implementation per runtime.
- **`fabric/otlp_writer.py`** (new) — everything that needs protobuf, and so cannot live in the receiver: folding stored exports into the OTLP/JSON the trace reader already reads, registering trace evidence, and the Relay endpoint settings both runtimes share.
- **`fabric/runtime.py`** — one receiver per task, started inside the guarded block, stopped and folded in `finally` before evidence is read (the fold guarded, since a raise from `finally` escapes the per-task handlers and would abort the whole `gather`); its address threaded into `RelayOpenTelemetryConfig`; the result registered as `trace:otlp` and as the primary `trace` key. ATIF now also gets `trace:atif`.
- **Docs** — the metrics guide described dual-format traces as Harbor-only; that is no longer true.

## Why a receiver rather than config

Verified against `NeMo-Relay@9c68f58` and the shipped `_native.abi3.so`, not inferred: `OpenTelemetrySectionConfig` is `{enabled, endpoints[]}` with no sink; the exporter is a two-arm match on `HttpBinary`/`Grpc`, both networked; `resolve_http_trace_endpoint` rejects any scheme but http/https; only `atif.rs` and `atof.rs` have file sinks; and the binary embeds the OTLP-endpoint env vars but not `OTEL_TRACES_EXPORTER`. Fabric only narrows the config it forwards, so it cannot add a sink Relay lacks.

**A file sink on Relay's OTel section would delete this module** and make Fabric's case the config line AALGO-601 originally assumed. Worth asking the Relay owners before anyone extends this.

## Type of Change

## Quality Gates

## Verification

Targeted validation:

- `uv run pre-commit run -a` — **15 hooks, 0 failures**
- `uv run --frozen pytest packages/nemo_evaluator_sdk plugins/nemo-evaluator` — **2850 passed, 38 skipped**
- `uv run --frozen ty check` — clean on every changed file (the 6 `unused-ignore-comment` warnings in `fabric/runtime.py` are present on `origin/main` unchanged; verified by stashing)
- `make vendor` and `make generate-cli-reference-docs` — no drift
- `lint-openapi` / `lint-sdk-vendored` need `mapfile` (bash 4+) and cannot run on macOS bash 3.2; both invariants checked directly instead, as above

**Live evidence.** `test_fabric_codex_live_eval_captures_atif_trajectory` runs real codex + real Relay (it self-skips without codex, nemo-relay and the adapter on PATH). It passes on this branch's HEAD. The captured trace carries what ATIF cannot — from one such run:

```
codex-turn                15919.93ms   root
  ├─ openai.responses       3434.04ms
  ├─ openai.responses       3108.79ms
  ├─ mcp__dev_brain__bootstrap  87.83ms
  └─ openai.responses       2801.13ms
```

ATIF for the same run: 5 steps, one `timestamp` each, no durations, and step order not matching the wall clock.

Span counts vary run to run with what the agent actually does — a later run captured 3 spans across 2 export flushes rather than 6 across 1. What is invariant, and what the tests pin, is that spans carry durations and parent links and that repeated flushes append rather than replace.

## Design notes

**Storing exports, then folding.** The receiver writes each export to disk verbatim and the fold happens once, at the end. That keeps the receiver free of protobuf so the container can run it, and means a trace never has to fit in memory. Folding streams through a temporary file and moves it into place, so a fold that dies partway cannot leave a half-written trace where a complete one goes; the raw exports are dropped once folded, since the result is a lossless re-encoding of them. It is additive in both directions — repeating a fold changes nothing, and a fold that finds new exports copies the existing trace into the rewrite before appending, since the atomic replace would otherwise drop the earlier spans. Every export a fold reads is consumed, spanless flushes included, so a later pass cannot re-read the same nothing.

**ATIF and OTLP are admitted on different terms.** ATIF is registered on a path alone — Fabric declares it in the run's artifact manifest and `ATIFTraceHandle` validates it on read. OTLP is one we wrote, so an empty file means the exporter never flushed and is not a trace.

## Known limitation

**Local harnesses only, in this PR.** The container runtime is handled in #1915, which reuses this receiver rather than adding a second one.

**A failed trial carries no trace.** `build_failed_trial` files only `error`, so a timed-out or errored run loses the partial OTLP capture even though the fold has already written it. That predates this PR — failed trials have never carried an ATIF trace either — but OTLP makes it more worth fixing, since spans up to the timeout are exactly what you want when diagnosing one. Left for a follow-up rather than restructuring the try/finally here.

## Summary by CodeRabbit

- **New Features**
  - Fabric task runs can capture OTLP traces emitted by Relay.
  - Captured OTLP traces are automatically converted into trajectory evidence.
  - ATIF and OTLP evidence remain available when both are produced.

- **Behavior Changes**
  - OTLP is used as the primary trace when available; ATIF is used as a fallback.
  - Trace-processing failures no longer prevent successful task results from completing.
  - Existing ATIF and ATOF exports remain supported.

- **Documentation**
  - Clarified trace handling for Harbor and Fabric.
  - Documented that no setting controls the default trace format.

### DIFF
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
@@ -1,0 +1,135 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""The OTLP endpoint Relay exports to, for both Fabric runtimes.
+
+Relay exports OTLP only over the network, so capturing a trace means listening for one. The local
+runtime imports this and runs it in a thread; the container runtime seeds this same file into the
+sandbox and runs it as a script beside the harness, where it writes into the directory that is
+downloaded when the task ends.
+
+Standard library only, and it stays that way: the sandbox image ships Fabric and its harnesses, not
+this package, so an import of anything else would have to be installed there first. That is also
+why each export is stored verbatim rather than as OTLP/JSON -- decoding needs protobuf, which the
+image does not guarantee, so :func:`nemo_evaluator_sdk...otlp_writer.fold_exports` does it later.
+"""
+
+from __future__ import annotations
+
+import sys
+import threading
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from pathlib import Path
+from types import TracebackType
+
+TRACES_PATH = "/v1/traces"
+PROTOBUF_MEDIA_TYPE = "application/x-protobuf"
+MAX_EXPORT_BYTES = 256 * 1024 * 1024
+REQUEST_TIMEOUT_S = 30.0
+#: Written once the socket is bound, so the caller can start Relay without racing the listener.
+READY_FILENAME = "receiver.ready"
+#: Read back in export order, which `sorted()` gives while the counter is zero-padded.
+EXPORT_SUFFIX = ".otlp.pb"
+
+
+class _Handler(BaseHTTPRequestHandler):
+    server: OTLPReceiver  # type: ignore[assignment]
+    timeout = REQUEST_TIMEOUT_S
+
+    def do_POST(self) -> None:
+        if self.path.split("?")[0] != TRACES_PATH:
+            self.send_response(404)
+            self.end_headers()
+            return
+        if self.headers.get("Content-Type", "").split(";")[0].strip() != PROTOBUF_MEDIA_TYPE:
+            self.send_response(415)
+            self.end_headers()
+            return
+        try:
+            length = int(self.headers.get("Content-Length", "0"))
+        except ValueError:
+            self.send_response(400)
+            self.end_headers()
+            return
+        if length > MAX_EXPORT_BYTES:
+            self.send_response(413)
+            self.end_headers()
+            return
+        body = self.rfile.read(length)
+        try:
+            self.server.exports += 1
+            target = self.server.directory / f"{self.server.exports:06d}{EXPORT_SUFFIX}"
+            target.write_bytes(body)
+        except OSError:
+            # The agent is still running and its result stands on its own.
+            self.send_response(500)
+            self.end_headers()
+            return
+        self.send_response(200)
+        self.send_header("Content-Type", PROTOBUF_MEDIA_TYPE)
+        self.end_headers()
+        self.wfile.write(b"")
+
+    def log_message(self, format: str, *args: object) -> None:
+        return
+
+
+class OTLPReceiver(HTTPServer):
+    """Loopback OTLP endpoint that stores each export verbatim under ``directory``.
+
+    Used two ways: as a context manager, which serves on its own thread, and as a script via
+    :func:`main`, which serves on the main one.
+    """
+
+    def __init__(self, directory: Path, port: int = 0) -> None:
+        # Port 0 lets the OS pick, which is what the local runtime wants so parallel trials cannot
+        # collide. The container passes a fixed one because it has a sandbox to itself.
+        super().__init__(("127.0.0.1", port), _Handler)
+        directory.mkdir(parents=True, exist_ok=True)
+        self.directory = directory
+        self.exports = 0
+        self._thread: threading.Thread | None = None
+
+    @property
+    def endpoint(self) -> str:
+        return f"http://{self.server_address[0]}:{self.server_address[1]}{TRACES_PATH}"
+
+    def signal_ready(self) -> None:
+        """Announce that the socket is bound, for a caller that has to wait out of process."""
+        (self.directory / READY_FILENAME).write_text("ready\n", encoding="utf-8")
+
+    def __enter__(self) -> OTLPReceiver:
+        self._thread = threading.Thread(target=self.serve_forever, name="otlp-receiver", daemon=True)
+        self._thread.start()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        self.shutdown()
+        if self._thread is not None:
+            self._thread.join(timeout=10.0)
+        self.server_close()
+
+
+def main(argv: list[str]) -> int:
+    if len(argv) != 3:
+        print(f"usage: {argv[0] if argv else 'receiver'} <directory> <port>", file=sys.stderr)
+        return 2
+    directory, port = Path(argv[1]), int(argv[2])
+    receiver = OTLPReceiver(directory, port)
+    receiver.signal_ready()
+    try:
+        receiver.serve_forever()
+    except KeyboardInterrupt:
+        pass
+    finally:
+        receiver.server_close()
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main(sys.argv))
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
@@ -21,6 +21,7 @@ import threading
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
 from types import TracebackType
+from urllib.parse import urlparse
 
 TRACES_PATH = "/v1/traces"
 PROTOBUF_MEDIA_TYPE = "application/x-protobuf"
@@ -37,7 +38,7 @@ class _Handler(BaseHTTPRequestHandler):
     timeout = REQUEST_TIMEOUT_S
 
     def do_POST(self) -> None:
-        if self.path.split("?")[0] != TRACES_PATH:
+        if urlparse(self.path).path != TRACES_PATH:
             self.send_response(404)
             self.end_headers()
             return
@@ -48,6 +49,13 @@ class _Handler(BaseHTTPRequestHandler):
         try:
             length = int(self.headers.get("Content-Length", "0"))
         except ValueError:
+            self.send_response(400)
+            self.end_headers()
+            return
+        # Bounded below as well as above: ``rfile.read(-1)`` reads to EOF, so a negative length would
+        # hold this single-threaded receiver open for as long as the client kept the connection, and
+        # the agent sharing this loopback is untrusted code.
+        if length < 0:
             self.send_response(400)
             self.end_headers()
             return
--- a/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
+++ b/packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import logging
 import os
+import shutil
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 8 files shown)


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -369,6 +381,12 @@ async def _run_task(
                 task, evidence_dir, exc, extra_metadata=self._failed_metadata(skill_provenances, evidence_dir)
             )
         finally:
+            if trace_receiver is not None:
+                # Stopped before the fold, and both before `_to_trial` reads the evidence: the
+                # spans arrive on the exporter's own schedule, and a trace folded before the last
+                # flush looks complete.
+                trace_receiver.__exit__(None, None, None)
```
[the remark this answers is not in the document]
SandyChapman: Correct on the mechanics — the `return` expression is evaluated before `finally`, so the trial is built before the fold. Not actioning here though: `build_failed_trial` files only `error`, so failed trials have never carried a trace in any format, ATIF included. This is an enhancement (a partial trace from a timed-out run is genuinely useful) rather than a regression this PR introduces, and it needs the try/finally restructured. Tracking separately.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
```diff
@@ -0,0 +1,142 @@
...
+#: Read back in export order, which `sorted()` gives while the counter is zero-padded.
+EXPORT_SUFFIX = ".otlp.pb"
+
+
+class _Handler(BaseHTTPRequestHandler):
+    server: OTLPReceiver  # type: ignore[assignment]
+    timeout = REQUEST_TIMEOUT_S
+
+    def do_POST(self) -> None:
+        if self.path.split("?")[0] != TRACES_PATH:
```
arpitsardhana: nit: such kind of split & reference first element could be error prone leading to Exceptions
Does it make sense to make sense to have some sort of Regex logic which can survive None assignments or empty assignments

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
```diff
@@ -0,0 +1,142 @@
...
+        self._thread = threading.Thread(target=self.serve_forever, name="otlp-receiver", daemon=True)
+        self._thread.start()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
```
arpitsardhana, line 119: nit: No cleanup of created directory required right?

arpitsardhana (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
```diff
@@ -0,0 +1,142 @@
...
+PROTOBUF_MEDIA_TYPE = "application/x-protobuf"
+MAX_EXPORT_BYTES = 256 * 1024 * 1024
+REQUEST_TIMEOUT_S = 30.0
+#: Written once the socket is bound, so the caller can start Relay without racing the listener.
+READY_FILENAME = "receiver.ready"
+#: Read back in export order, which `sorted()` gives while the counter is zero-padded.
+EXPORT_SUFFIX = ".otlp.pb"
+
+
+class _Handler(BaseHTTPRequestHandler):
```
ngoncharenko, line 36: I wonder why there's such a discrepancy between OTLP and ATIF in Relay. Why they drop ATIF file but don't do the same for OTLP? This seems like a workaround for now. Should we request a parity with ATIF in terms dropping the file?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
```diff
@@ -0,0 +1,139 @@
...
+                try:
+                    line = export_to_json_line(export.read_bytes())
+                except (OSError, DecodeError):
+                    logger.warning("Discarding an unreadable OTLP export at %s.", export)
+                    continue
+                if line is None:
+                    continue
+                handle.write(f"{line}\n")
+                folded += 1
+    except BaseException:
```
ngoncharenko, line 105: why not to use finally here or it's not a good fit?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
```diff
@@ -383,6 +395,17 @@ async def _run_task(
...
+                # Guarded because this runs in `finally`, where a raise escapes the handlers above
+                # and aborts the whole gather instead of failing this one task.
+                try:
+                    fold_exports(traces_dir(evidence_dir))
+                except Exception as exc:  # noqa: BLE001 - any fold failure costs the trace, not the trial
+                    logger.warning("Could not fold the OTLP trace for task %s: %s", task.id, exc)
```
ngoncharenko, line 408: I wonder is there a more explicit fail we could make, so it ends up with status.PARTIAL considering a trace is an important asset for eval run

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
SandyChapman: Yes, I'm hoping to start a conversation with them on this, but this is definitely the case that ATIF gets written to a file and OTLP requires an endpoint to send the data to.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
SandyChapman: Correct — deliberately none: that directory is the trial's evidence dir, so the fold writes `relay.jsonl` into it and the runtime reads it after the receiver is gone.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
SandyChapman: Not a good fit here — on the success path the partial has to survive so `os.replace` can move it into place, so a `finally` would need a success flag to say what `except … raise` says directly.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/runtime.py
SandyChapman: Agreed the warning is quiet for something this important, but `PARTIAL` means the *agent* only partly completed (`architecture.md`), and a fold failure is a bug on our side rather than a degraded run — so it would tell a metric the agent underperformed when it did not.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
```diff
@@ -0,0 +1,139 @@
...
+        # Whatever went wrong, a half-written trace must not be left where a complete one goes.
+        partial.unlink(missing_ok=True)
+        raise
+    if not folded:
+        partial.unlink(missing_ok=True)
+        return 0
```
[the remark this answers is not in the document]
SandyChapman: Half actioned. Cleanup is now best-effort and warns, naming the duplicate risk, so a commit that lands can never be undone by a failure to consume its inputs. Not adding transaction state: removing an input needs the same directory permission `os.replace` did, so the ordinary failure (unwritable dir) fails the commit first and correctly preserves the inputs — I could only reproduce commit-then-cleanup-failure with `chflags uchg` on an individual file, and rename-out-of-the-glob fails there too, so a journal would be the only remedy for a case that also needs a second fold of the same directory to bite. Both runtimes fold once per task, after the receiver stops.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
```diff
@@ -0,0 +1,148 @@
...
+    except BaseException:
+        # Whatever went wrong, a half-written trace must not be left where a complete one goes.
+        partial.unlink(missing_ok=True)
+        raise
+    if not folded:
+        partial.unlink(missing_ok=True)
+        _drop(exports)
+        return 0
+    os.replace(partial, trace)
+    _drop(exports)
```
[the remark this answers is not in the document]
SandyChapman: Half actioned — and note `unlink` no longer raises, so a commit that lands can no longer be undone by a failure to consume its inputs; cleanup is best-effort and warns, naming the duplicate risk explicitly.

Not adding transaction state. Removing an input needs the same directory permission `os.replace` did, so the ordinary failure — an unwritable evidence dir — fails the commit first and correctly preserves the inputs. I could only reproduce commit-then-cleanup-failure by making an individual file immutable (`chflags uchg`), and renaming the survivor out of the glob fails under exactly the same conditions, so a journal really would be the only remedy. It would have to live in the directory that just proved hostile, and the case additionally needs a second fold of the same directory to bite — both runtimes fold exactly once per task, after the receiver has stopped.

Worth revisiting if a mid-run flush ever lands, since that is what would make the second fold real.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_writer.py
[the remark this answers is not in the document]
SandyChapman: Correction: my earlier reply here was meant for the transaction-state finding on the thread below — this one (spanless exports) was fully actioned, not half.

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/fabric/otlp_receiver.py
SandyChapman: Switched to `urlparse(self.path).path` — `split("?")[0]` cannot actually raise (`str.split` never returns an empty list), but parsing buys something real: an RFC 7230 absolute-form target like `POST http://host/v1/traces` would otherwise 404 and silently cost the trial its trace, so there is a test for that now.

PR #1912, opened 2026-09-09
fix(setup): Don't allow platform to setup if python version is incorrect

soluwalana:
## Summary

`nemo setup` currently succeeds on unsupported Python versions (including 3.14, which `uv` can select automatically). Controllers cannot run on that interpreter, so the platform then fails at startup with little or no useful signal. This change rejects unsupported interpreters at setup time and tells the user how to reinstall against Python 3.12–3.13.

Siemens hit this while debugging on Sep 8: setup completed on Python 3.14, then services failed silently at startup.

Live evaluator sandbox tests were also hitting Docker Hub (`docker.io/library/busybox:latest`) during `docker build` / Compose, which failed when Hub auth returned 500. Those tests now pull busybox from NVIDIA's Docker Hub mirror (`dockerhub.nvidia.com/busybox:latest`), matching Platform-Deploy.

## Changes

- Fail `nemo setup` immediately when `sys.version_info` is outside Python 3.12–3.13 (inclusive).
- Print a clear error and the `uv tool install --python 3.13 --reinstall "nemo-platform[all]"` recovery command.
- Unit tests cover accepted versions (3.12/3.13), rejected versions (3.11/3.14), and that setup exits before starting services.
- Troubleshooting docs note the early exit and add `--reinstall` to the install command.
- Point live Compose/Docker sandbox tests at `dockerhub.nvidia.com/busybox:latest` instead of unqualified `busybox:latest` (which resolves to docker.io).
- Narrow `stdout` in the Docker sandbox live test so `ty` accepts `.strip()` (`str | None`).

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

```bash
uv run pytest packages/nemo_platform_ext/tests/cli/commands/test_setup.py::TestRequireSupportedPython -q
```

Live sandbox tests (when Docker is available) now pin busybox to `dockerhub.nvidia.com` so they do not depend on docker.io auth.

## Summary by CodeRabbit

* **Tests**
  * Updated Docker-based smoke tests to use NVIDIA’s BusyBox image mirror.
  * Improved uploaded-file validation in Docker provider tests.
  * Updated Docker smoke test documentation.

### DIFF
--- a/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
+++ b/packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
@@ -15,6 +15,7 @@ import logging
 import os
 import re
 import subprocess
+import sys
 import time
 from dataclasses import dataclass
 from enum import StrEnum
@@ -78,6 +79,10 @@ from nemo_platform_ext.ui.prompts import (
 
 logger = logging.getLogger(__name__)
 console = Console(stderr=True)
+
+# Supported Python versions (inclusive).
+_SUPPORTED_PYTHON_MIN = (3, 12)
+_SUPPORTED_PYTHON_MAX = (3, 13)
 
 CHECK = "[green]✓[/green]"
 CROSS = "[red]✗[/red]"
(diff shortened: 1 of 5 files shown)


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
```diff
@@ -79,6 +80,10 @@
 logger = logging.getLogger(__name__)
 console = Console(stderr=True)
 
+# Supported Python versions (inclusive).
```
tylersbray, line 83: Just wondering how/when we update this?  Like do we wait for it to break again or is there some better way you have in mind?

tylersbray (approved):
👍

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/setup.py
soluwalana: probably just wait for it to break ...?

PR #1922, opened 2026-09-09
refactor(jobs): move SDK and CLI off Stainless

ironcommit:
## TL;DR

Moves Jobs out of Stainless generation and makes the Jobs SDK/CLI source-owned by `nemo_platform_plugin.jobs`. The public `client.jobs` and `nemo jobs` surfaces remain available, but they now route through the plugin Jobs client and the SDK tooling excludes `/apis/jobs/v2` from Stainless.

## Details

This removes the Stainless `jobs` resource, job-only generated types, and generated jobs resource tests from the vendored Python SDK. A plugin SDK entry point now contributes sync and async compatibility resources for `client.jobs`, including job create/list/get/delete, lifecycle actions, logs, execution profiles, results, steps, and tasks. The compatibility layer preserves SDK paging behavior, request options such as extra headers/query/body and timeouts, binary result downloads, and `tail` support on job logs.

The CLI follows the same migration: `nemo jobs` is registered as a source-owned Typer command group instead of generated API commands. The command group keeps the existing job, result, step, task, wait/watch, tail, table output, input-file/input-data, and code-output workflows while using the source-owned `JobsClient` underneath. The CLI generator now skips jobs so future generation runs do not recreate the deleted Stainless command tree.

The SDK tooling now has a source-owned resource exclusion registry. When a top-level resource such as jobs is absent from Stainless config, the mapper skips matching endpoints and jobs-only schemas while preserving schemas shared with other generated resources. The Customizer SDK and throughput tutorial imports were adjusted to use the plugin-owned jobs schemas/types that remain available after the generated job types are removed.

## Reviewer Notes

The expected import path for jobs-specific schemas and types is now `nemo_platform_plugin.jobs.*` instead of `nemo_platform.types.jobs` or job-only entries under `nemo_platform.types.shared`. Shared generated models that are still used outside jobs, such as job logs, stay in the Stainless SDK.

## Summary by CodeRabbit

- **New Features**
  - Added a top-level `jobs` CLI for managing jobs, results, steps, tasks, logs, and lifecycle actions.
  - Added synchronous and asynchronous jobs access through the platform client.
  - Added optional log tailing and broader filtering support in jobs APIs.
  - Preserved compatibility for existing jobs SDK workflows.

- **Breaking Changes**
  - Removed the legacy `api jobs` command group; use the top-level `jobs` commands instead.
  - Job-related CLI commands now require explicit job and step identifiers where applicable.

- **Improvements**
  - Updated models, projects, and workspaces sort options with a consistent ordering.


maxdubrinsky (approved):

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/adapter.py
```diff
@@ -17,81 +17,94 @@ def make_sync_resource(platform: NeMoPlatform) -> NemoClient:
...
+    _client: httpx.Client | httpx.AsyncClient
+
+    def _prepare_url(self, url: str) -> httpx.URL: ...
+
+
 @overload
-def client_from_platform(platform: NeMoPlatform, client_cls: type[SyncT]) -> SyncT: ...
+def client_from_platform(platform: object, client_cls: type[SyncT]) -> SyncT: ...
 @overload
-def client_from_platform(platform: AsyncNeMoPlatform, client_cls: type[AsyncT]) -> AsyncT: ...
```
mckornfield, line 34: Object seems like a cop out no?

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
```diff
@@ -23,17 +23,274 @@
...
+            yield page
+            if not page.has_next_page():
+                return
+            get_next_page = page._sync_get_next_page
+            if get_next_page is None:
+                raise TypeError("sync pagination returned an async next page")
+            page = get_next_page()
+
+
+class AsyncLegacyPaginatedResponse(Generic[LegacyItemT]):
```
mckornfield, line 192: Jeez these are some generics

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
```diff
@@ -23,17 +23,274 @@
...
+
+    async def iter_pages(self) -> AsyncIterator[LegacyPage[LegacyItemT]]:
+        page = await self._page()
+        while True:
+            yield page
+            if not page.has_next_page():
+                return
+            get_next_page = page._get_next_page
+            if get_next_page is None:
+                raise TypeError("async pagination returned a sync next page")
```
mckornfield: Is this error right?

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
```diff
@@ -122,6 +379,119 @@ class _JobsMethods:
...
+    ) -> LegacyPaginatedResponse[PlatformJobLog]:
+        query_params = _job_logs_query_params(
+            attempt_id=attempt_id,
+            limit=limit,
+            page_cursor=page_cursor,
+            step_id=step_id,
+            tail=tail,
+            task_id=task_id,
+        )
+        return LegacyPaginatedResponse(self.list_job_logs(workspace=workspace, name=name, query_params=query_params))
```
mckornfield, line 500: Are we naming these things all legacy with the intent to remove them in a follow up?

tools/nemo-platform-sdk-tools/tests/sdk/openapi_stainless_mapper_test.py
mckornfield, line 1: Is this file named correctly

mckornfield (approved):
Those _test.py files at the end, is that naming intentional?

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
ironcommit: yes. that's the idea.  This avoids needing to update the callers, which would make this PR even bigger.

packages/nemo_platform_plugin/src/nemo_platform_plugin/client/adapter.py
ironcommit: definitely a cop out.   Fixing.

tools/nemo-platform-sdk-tools/tests/sdk/openapi_stainless_mapper_test.py
ironcommit: renamed

packages/nemo_platform_plugin/src/nemo_platform_plugin/jobs/client.py
ironcommit: fixed.

PR #1957, opened 2026-09-10
feat(files): serve filesets from GitHub and refresh a tracked revision [ASTD-592]

marcusds:
## Summary

Adds a GitHub storage backend for filesets, so an agent's files can be served straight from a repository instead of uploaded, and a way to move a fileset forward when the ref it tracks does.

Fileset creation pins a mutable ref to an immutable id so contents cannot shift under a deployment, and `PATCH /filesets/{name}` does not accept `storage`. Picking up a new commit therefore meant creating another fileset, and another agent. `POST /filesets/{name}/refresh` re-resolves the ref the fileset was created from and repoints it at whatever that names now.

Deployments stage the fileset when they are created, so a refresh deliberately leaves running deployments on the revision they started with. `AgentDeployment` now records that revision, which is what makes the difference visible.

## Changes

- **GitHub storage backend** (`backends/github.py`): lists via the git-trees API, streams file contents, maps GitHub status codes onto the storage exception hierarchy, and pins the revision to a commit SHA at create time. `api.github.com` added to the default allowed external hosts.
- **`POST /filesets/{name}/refresh`**: re-resolves the tracked ref. Repository and directory are untouched. A fileset created from an immutable id has nothing to track and is refused with 409.
- **`tracked_revision` / `config_at_tracked_revision()`** on `StorageImpl`, beside `resolve_config` whose counterpart they are. Implemented for GitHub and HuggingFace; every other backend inherits "tracks nothing".
- **`AgentDeployment.spec_revision` / `spec_tracked_revision`**, captured at deployment create from the Ethos fileset. An absent or unreadable fileset records nothing rather than failing the deployment.
- OpenAPI regenerated for both new surfaces.

The shared `nemo_platform_plugin` package is deliberately untouched: the tracked-ref concept lives in the files service that owns the endpoint.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs cover fileset storage backends yet; the endpoint is described in the generated OpenAPI spec.

## Verification

Targeted validation:

- `pytest services/core/files/tests plugins/nemo-agents/tests/unit/test_deployments_api.py` — 666 passed, 51 skipped
- `ruff check` / `ruff format --check` on the changed files — clean
- `ty check` on the changed files — one pre-existing diagnostic in `backends/base.py` (`return self.config` in the original `resolve_config`)
- `uv run pre-commit run -a` — not run. The pre-push hooks fail in this environment for unrelated reasons: `helm-docs` is not installed, and `uv-lock` wants uv 0.9.14 against 0.9.30 locally. Pushed with `--no-verify`.

Manual verification against a local platform: created a fileset from `owner/repo@main`, pushed a commit, refreshed, and confirmed the fileset moved to the new SHA while an existing deployment stayed on the old one.

## Notes for review

`fix(files): do not treat a self-resolving commit as a tracked ref` fixes a bug the endpoint tests caught: `resolve_config` records `original_revision` unconditionally, so a fileset created from a SHA reported itself as tracking a ref and the 409 was unreachable for GitHub and HuggingFace.

The refresh endpoint has no way to check for a newer revision without applying it. A read-only `GET /filesets/{name}/tracked-revision` would let a client show "update available" first; not included here.

Part of ASTD-519. Closes ASTD-592. The Studio half is stacked on this branch in a separate PR.

## Summary by CodeRabbit

- **New Features**
  - Added GitHub repositories as a fileset storage option, including private repositories, revisions, subdirectories, and GitHub Enterprise URLs.
  - Added API and CLI support for refreshing filesets to the latest tracked mutable revision.
  - Added GitHub storage labels and repository path display in Studio.
- **Documentation**
  - Updated configuration and CLI references for GitHub storage and fileset refresh.
  - Clarified local file removal behavior when deleting filesets.
- **Configuration**
  - Added GitHub’s API host to the default allowed external hosts.

### DIFF
--- a/docs/set-up/config-reference.mdx
+++ b/docs/set-up/config-reference.mdx
@@ -245,8 +245,8 @@ files:
     path: /data/files_storage
     # How many bytes to buffer before flushing to disk | default: 16777216
     write_buffer_size: 16777216
-  # Comma-separated list of external hosts the Files service is allowed to access. | default: 'https://api.ngc.nvidia.com,https://huggingface.co'
-  allowed_external_hosts: https://api.ngc.nvidia.com,https://huggingface.co
+  # Comma-separated list of external hosts the Files service is allowed to access. | default: 'https://api.ngc.nvidia.com,https://huggingface.co,https://api.github.com'
+  allowed_external_hosts: https://api.ngc.nvidia.com,https://huggingface.co,https://api.github.com
   # Allow users to explicitly create filesets with local storage config. Security-sensitive: enable only in trusted deployments. | default: False
   allow_user_local_storage: false
   # TTL for file locks in seconds (default 5 minutes) | default: 300
--- a/openapi/ga/individual/platform.openapi.yaml
+++ b/openapi/ga/individual/platform.openapi.yaml
@@ -2290,6 +2290,62 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPValidationError'
+  /apis/files/v2/workspaces/{workspace}/filesets/{name}/refresh:
+    post:
+      tags:
+      - Files
+      summary: Refresh Fileset Revision
+      description: 'Re-resolve a fileset''s tracked revision against its source.
+
+
+        A fileset created from a mutable ref is pinned to an immutable id so its
+
+        contents cannot shift under a deployment. This re-resolves that same ref and
+
+        repoints the fileset at whatever it names now. Everything else about the
+
+        storage config, including the repository and directory, is left alone.
+
+
+        Deployments stage the fileset when they are created, so existing deployments
+
+        keep serving the revision they were staged from.'
+      operationId: refresh_fileset_apis_files_v2_workspaces__workspace__filesets__name__refresh_post
+      parameters:
+      - name: workspace
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Workspace
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          title: Name
+      responses:
+        '200':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FilesetOutput'
+        '409':
+          description: Fileset does not track a mutable revision
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  detail:
+                    type: string
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
   /apis/guardrails/v2/workspaces/{workspace}/checks:
     post:
       tags:
--- a/openapi/ga/individual/platform.openapi.yaml
+++ b/openapi/ga/individual/platform.openapi.yaml
@@ -2307,9 +2307,10 @@ paths:
         storage config, including the repository and directory, is left alone.
 
 
-        Deployments stage the fileset when they are created, so existing deployments
+        A running deployment keeps serving the revision it staged. It moves to this
+        one
 
-        keep serving the revision they were staged from.'
+        when the runner next stages the fileset, and records the revision it staged.'
       operationId: refresh_fileset_apis_files_v2_workspaces__workspace__filesets__name__refresh_post
       parameters:
       - name: workspace
@@ -2331,8 +2332,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/FilesetOutput'
+        '400':
+          description: Storage configuration, secret, or source rejected the refresh
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  detail:
+                    type: string
+        '404':
+          description: Fileset not found
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  detail:
+                    type: string
         '409':
           description: Fileset does not track a mutable revision
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  detail:
+                    type: string
+        '502':
+          description: Storage backend unavailable
           content:
             application/json:
               schema:
--- a/openapi/ga/individual/platform.openapi.yaml
+++ b/openapi/ga/individual/platform.openapi.yaml
@@ -10266,6 +10266,7 @@ components:
           - $ref: '#/components/schemas/NGCStorageConfig'
           - $ref: '#/components/schemas/HuggingfaceStorageConfig'
           - $ref: '#/components/schemas/S3StorageConfig'
+          - $ref: '#/components/schemas/GithubStorageConfig'
           title: Storage
           description: The storage configuration for the fileset. If not provided,
             uses default storage.
@@ -12616,6 +12617,7 @@ components:
           - $ref: '#/components/schemas/NGCStorageConfig'
           - $ref: '#/components/schemas/HuggingfaceStorageConfig'
           - $ref: '#/components/schemas/S3StorageConfig'
+          - $ref: '#/components/schemas/GithubStorageConfig'
           title: Storage
         metadata:
           $ref: '#/components/schemas/FilesetMetadata'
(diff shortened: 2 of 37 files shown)


openapi/ga/individual/platform.openapi.yaml
```diff
@@ -2290,6 +2290,62 @@ paths:
...
+                type: object
+                properties:
+                  detail:
+                    type: string
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
```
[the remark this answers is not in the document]
marcusds: Claude says: Done in `c678c141e8`. The refresh route now declares `400`, `404`, and `502` alongside `409`, all using `_HTTP_EXCEPTION_DETAIL`, and the OpenAPI specs, CLI, and Python SDK were regenerated.

Noting for the record that no sibling fileset endpoint declares anything beyond the default, so this makes `refresh` the best-documented route in the file rather than a consistent one. Worth doing the rest separately.

openapi/openapi.yaml
```diff
@@ -13122,6 +13180,60 @@ components:
...
+        api_base_url:
+          type: string
+          title: Api Base Url
+          description: GitHub API base URL. Use for GitHub Enterprise instances.
+          default: https://api.github.com
+      type: object
+      required:
+      - owner
+      - repo
+      title: GithubStorageConfig
```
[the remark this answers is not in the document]
marcusds: Claude says: Done in `c678c141e8`. `GithubStorageConfig.api_base_url` now rejects anything that is not `https://` at validation time, so it fails when the fileset is created rather than on first request:

```
raise ValueError(f"api_base_url must use https, got {v!r}")
```

Your reasoning was right: `validate_external_host` compares scheme + netloc, so an allowlisted `http://` host would have been accepted and `_headers()` attaches the token to every request. GitHub Enterprise over HTTPS still works.

services/core/files/src/nmp/core/files/app/backends/github.py
```diff
@@ -0,0 +1,220 @@
...
+    """Map a GitHub response status onto the storage exception hierarchy."""
+    if status < 400:
+        return
+
+    # A private repository is indistinguishable from a missing one without the right
+    # token, so 404 is reported as a config error rather than an access error.
+    if status == 404:
+        raise GithubConfigError(f"GitHub has no {subject}, or the token cannot see it")
+    if status in (401, 403):
+        if headers and headers.get("x-ratelimit-remaining") == "0":
```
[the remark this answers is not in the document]
marcusds: Claude says: Confirmed and fixed in `c678c141e8` — this one was live, not theoretical.

```
CI proxy lookup : 0
dict() lookup   : None
dict() keys     : ["X-RateLimit-Remaining"]
```

aiohttp speaks HTTP/1.1, where GitHub capitalizes the header, so the lowercase `.get()` never matched and an exhausted rate limit was reported as `GithubAccessError` instead of `GithubUnavailableError`. Both call sites now pass `response.headers` through unconverted, and lookup goes via a `_header()` helper that compares case-insensitively.

The existing test only passed because it supplied lowercase keys; it is now parametrized over both spellings.

services/core/files/src/nmp/core/files/app/backends/github.py
```diff
@@ -0,0 +1,220 @@
...
+        """Address the configured directory directly, so the listing is scoped to it."""
+        return f"{self.config.revision}:{self.config.path}" if self.config.path else self.config.revision
+
+    def _repo_path(self, path: str) -> str:
+        return f"{self.config.path}/{path}" if self.config.path else path
+
+    async def _get_json(self, url: str, subject: str) -> Any:
+        session = get_http_session()
+        try:
+            async with session.get(url, headers=self._headers(JSON_MEDIA_TYPE)) as response:
```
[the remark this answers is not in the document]
marcusds: Claude says: Same finding as the thread on `openapi/openapi.yaml` — fixed in `c678c141e8`. HTTPS is enforced on `api_base_url` by a field validator, before any request can attach the token.

plugins/nemo-agents/src/nemo_agents_plugin/runner/deployments_backend.py
```diff
@@ -707,6 +710,9 @@ async def create_deployment(
             except FabricArtifactStagingError as exc:
                 logger.error("Refusing to deploy Fabric agent %r: %s", name, exc)
                 return DeploymentInfo(name=name, status="failed", error=str(exc))
+            staged_spec = await read_spec_revision(
+                client_from_platform(sdk, AsyncFilesClient), workspace=workspace, agent_name=agent
+            )
```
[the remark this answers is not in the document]
marcusds: Claude says: Valid finding, fixed — but the code has moved out of this PR.

This PR was split so the files service and the agents plugin could be reviewed separately. `deployments_backend.py` and the rest of the staging path now live in #2009 (ASTD-604), stacked on this one.

The fix there: staging reads the spec fileset revision either side of the download and stages again when it moved, so a refresh landing mid-stage cannot be recorded against content it did not produce. Three tests cover it, including the refresh-between-staging-and-lookup case you asked for.

Fully closing it needs the fileset download path to accept or report a revision, so staging can pin what it downloads rather than observe it. That is a files-service change and is tracked separately rather than stretched into either PR.

services/core/files/src/nmp/core/files/app/backends/github.py
```diff
@@ -0,0 +1,228 @@
...
+            url = f"{url}/{path}"
+        return f"{url}?{urlencode(query)}" if query else url
+
+    def _subject_at_revision(self) -> str:
+        scope = f"{self._repo_slug}/{self.config.path}" if self.config.path else self._repo_slug
+        return f"{scope} at {self.config.revision}"
+
+    def _tree_ref(self) -> str:
+        """Address the configured directory directly, so the listing is scoped to it."""
+        return f"{self.config.revision}:{self.config.path}" if self.config.path else self.config.revision
```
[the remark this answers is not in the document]
marcusds: Claude says: Declining this one — `git/trees/{tree_sha}` does accept `revision:path`, and the requests below are against this repository on the live API.

| Request | Result |
| --- | --- |
| `git/trees/main:docs` | `200`, 38 entries, sha `8fd4edf0…` ≠ root sha `c138f1bc…` |
| `git/trees/main:definitely-not-a-dir` | `404` — the suffix is parsed, not ignored |
| `contents/docs?ref=main` | identical listing, so it resolves to the right subtree |
| `git/trees/main:docs?recursive=1` | 471 entries — the exact call `list_files` makes |

The 404 is the important one: if the `:path` suffix were being dropped, a nonexistent path would have returned the root tree rather than failing.

Walking the parent tree to find the directory's SHA would also cost an extra request per call and reintroduce the truncation problem this scoping exists to avoid — a large repository truncates its root tree, which is exactly why the request is scoped rather than filtered client-side.

Worth noting the listing looks wrong at a glance: `git/trees/main:docs` returns `.code-critic.yaml`, `.coderabbit.yaml`, `.cursorignore`… which read as root files. The contents API confirms `docs/` genuinely holds those dotfiles.

docs/set-up/config-reference.mdx
```diff
@@ -245,8 +245,8 @@ files:
     path: /data/files_storage
     # How many bytes to buffer before flushing to disk | default: 16777216
     write_buffer_size: 16777216
-  # Comma-separated list of external hosts the Files service is allowed to access. | default: 'https://api.ngc.nvidia.com,https://huggingface.co'
-  allowed_external_hosts: https://api.ngc.nvidia.com,https://huggingface.co
+  # Comma-separated list of external hosts the Files service is allowed to access. | default: 'https://api.ngc.nvidia.com,https://huggingface.co,https://api.github.com'
```
mckornfield, line 248: ooo storing files in gh now, just like the gretel-blueprints days 🆗

sdk/python/nemo-platform/tests/api_resources/files/test_filesets.py
mckornfield, line 1: curious, is there an __init__.py shim for these like for the other things in the sdk or no

sdk/stainless.yaml
```diff
@@ -240,6 +240,7 @@ resources:
       fileset_file: FilesetFileOutput
       fileset_outputs_page: FilesetOutputsPage
       fileset_purpose: FilesetPurpose
+      github_storage_config: GithubStorageConfig
```
mckornfield, line 243: hissss

services/core/files/src/nmp/core/files/api/v2/filesets/endpoints.py
```diff
@@ -582,6 +582,108 @@ async def update_fileset_metadata(
...
+        raise HTTPException(HTTP_400_BAD_REQUEST, f"Invalid URL for external host: {exc}") from exc
+    except (SecretNotFoundError, SecretAccessDeniedError) as exc:
+        logger.warning(
+            "Secret unavailable while refreshing %s/%s: %s",
+            _sanitize_for_log(workspace),
+            _sanitize_for_log(name),
+            exc,
+        )
+        raise HTTPException(HTTP_400_BAD_REQUEST, f"Secret unavailable: {exc}") from exc
+    except StorageAccessError as exc:
```
mckornfield, line 668: lol when the exceptions are longer than the code

mckornfield (approved):

sdk/stainless.yaml
marcusds: animated-gif-f5738a6f

sdk/python/nemo-platform/tests/api_resources/files/test_filesets.py
marcusds: Yeah, there's one: tests/api_resources/files/__init__.py

docs/set-up/config-reference.mdx
marcusds: Its what product wants!

PR #1965, opened 2026-09-11
refactor(experimentalist): migrate to typed platform clients

ironcommit:
## TL;DR

Migrates Experimentalist off the Stainless-generated platform SDK and onto the typed `nemo_platform_plugin` clients. The branch adds the typed Intake and Insights surfaces Experimentalist needs, updates Experimentalist call sites and tests to use those clients, and tightens typed-client transport ownership so shared sync/async transports are not accidentally closed by child clients or option clones.

## What changed

Experimentalist now constructs and passes `AsyncNemoClient` instances instead of `AsyncNeMoPlatform` instances. Remote authenticated platform URLs still use the active NeMo Platform config and OIDC token refresh path, while unauthenticated loopback/direct-mode behavior is preserved for local services.

The remote authenticated client path now rejects non-loopback plaintext HTTP before attaching credentials. Loopback HTTP remains supported for local unauthenticated services, and authenticated remote URLs must use HTTPS.

The Intake typed client now covers the APIs that Experimentalist uses to persist and inspect trials: trace retrieval, span listing, span grouping, experiment create/update, evaluation create/update, ATIF ingest, OTLP ingest, and evaluator-result creation. The wire models were expanded to represent experiments, evaluations, spans, span groups, and the query parameters needed by those flows.

Experimentalist's backend, experiment mirror, trace upload, insight fetch, examples, and tests now call typed resource clients directly. The native experiment projection path still treats Intake projection failures as best-effort, keeps existing source-link preservation behavior, and continues to stamp evaluation/trial identity onto uploaded traces.

The Tau3 NOOA example now declares `nemo-platform-plugin` as an editable local dependency and updates its standalone `uv.lock`, so its typed-client imports resolve when the example is run outside the monorepo workspace.

Fileset-backed task template staging now validates listed paths before writing downloaded content. It rejects traversal outside the staging destination and rejects paths outside a requested Fileset fragment prefix.

Typed client lifecycle support was added for sync and async clients. Root clients own their underlying HTTP transport; resource clients and `with_options`/`with_headers` clones share that transport without owning it, so closing a child client does not close the parent.

## Reviewer notes

The main compatibility surface is the client boundary: code that needs a generated SDK still uses the generated SDK, while Experimentalist-specific platform calls now use typed clients. Sync and async paths are intentionally separate throughout the migrated code.

Nooa model clients now preserve refreshable Platform authentication for completion calls by resolving auth at the Nooa call boundary, rather than relying only on static default headers.

The test harness can now yield an `AsyncNemoClient` backed by the same in-process ASGI transport used by existing SDK tests, and `ClientContext` exposes that typed async client directly.

## Summary by CodeRabbit

- **New Features**
  - Added Intake support for traces, spans, span groups, experiments, and evaluations.
  - Added synchronous and asynchronous Insights retrieval.
  - Added configurable fast-model support with fallback behavior.
  - Added client lifecycle management, including explicit closing and context-manager support.
- **Security**
  - Added authentication handling for model requests and HTTPS enforcement for authenticated remote connections.
- **Improvements**
  - Improved dataset staging with recursive fileset downloads and path-traversal protection.
  - Added clearer validation errors for missing Experimentalist configuration.

### DIFF
--- a/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/experimentalist_backend.py
+++ b/plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/experimentalist_backend.py
@@ -24,7 +24,6 @@ from pathlib import Path
 from typing import Any, TypeVar, cast
 from urllib.parse import urlparse
 
-import httpx
 from nemo_experimentalist_plugin.config import CandidateStorageConfig
 from nemo_experimentalist_plugin.entities import (
     Candidate,
@@ -48,7 +47,16 @@ from nemo_experimentalist_plugin.experimentalist.experiment_mirror import Experi
 from nemo_experimentalist_plugin.experimentalist.otlp import jsonl_to_protobuf, read_trace_id, spans_to_protobuf
 from nemo_experimentalist_plugin.experimentalist.result import ExperimentalistResult
 from nemo_insights_plugin.entities import Insight
-from nemo_platform import AsyncNeMoPlatform
+from nemo_insights_plugin.typed_client import AsyncInsightsClient
+from nemo_platform_plugin.client.client import AsyncNemoClient
+from nemo_platform_plugin.client.errors import NemoHTTPError, NotFoundError
+from nemo_platform_plugin.intake.client import AsyncIntakeClient
+from nemo_platform_plugin.intake.types import (
+    AtifCreateRequest,
+    EvaluatorResultCreateRequest,
+    ListSpansQueryParams,
+    Trace,
+)
 from pydantic import BaseModel
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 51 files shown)


plugins/nemo-experimentalist/src/nemo_experimentalist_plugin/experimentalist/experimentalist_backend.py
```diff
@@ -457,12 +465,12 @@ async def get_insight(self, *, workspace: str, insight_id: str) -> Insight:
                 "client is available to fetch it from the platform."
             )
         try:
-            return await self.client.insights.insights.get(workspace=workspace, insight_id=insight_id)
-        except httpx.HTTPStatusError as exc:
-            # Surface as ValueError so the CLI's clean-error path reports it instead
-            # of dumping a raw traceback (mirrors the local-file FileNotFoundError).
-            if exc.response.status_code == 404:
-                raise ValueError(f"Insight not found on the platform: {insight_id!r}") from exc
+            return await AsyncInsightsClient.from_client(self.client).get_insight(
```
maxdubrinsky, line 460: Might be worth storing `self.intake_client`

maxdubrinsky (approved):

PR #2036, opened 2026-09-14
fix(agents): apply telemetry auto-wiring to subprocess deployments

mikeknep:
## Summary
#1863 missed auto-wiring telemetry for subprocess deployments. This PR fixes that miss.

## Changes
- Auto-wire telemetry in the in-memory runner
- Hoist an e2e helper and check telemetry in multiple tests
- Fix two typecheck bugs in the agents execute job e2e tests

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Fabric deployments now automatically configure supported Intake ATIF telemetry.
  - Agent invocation activity can be exported as trajectory spans while deployments remain active.
  - Existing telemetry settings and opt-out configurations are preserved.

- **Bug Fixes**
  - Improved handling of unsupported telemetry configurations so deployments can start successfully without ATIF export.
  - Telemetry configuration is now handled more consistently across supported deployment scenarios.


plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -102,6 +104,46 @@ async def validate_platform_agent_config(config: dict[str, Any], *, base_dir: Pa
...
+    )
+
+    if not supports_intake_atif_export(config, base_dir=base_dir):
+        return config
+
+    # The caller's mapping is the live AgentDeployment entity the controller
+    # re-saves, so wiring it in place would persist an inferred endpoint into
+    # the stored deployment.
+    wired = copy.deepcopy(config)
+    configure_intake_atif_export(wired, workspace=workspace, base_url=get_base_url())
```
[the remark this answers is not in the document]
mikeknep: Declining this suggestion for a few reasons:

**1. As specified, it breaks the merged container path.** `configure_intake_atif_export()` is shared by all three wiring paths. The container path (merged in #1863, untouched by this PR) deliberately passes `base_url = f"http://127.0.0.1:{auth_proxy_port()}"` — the loopback auth-proxy sidecar (`runner/deployments_backend.py:671,684`). Adding an HTTPS requirement inside that function would reject container deployments outright. The suggested carve-out, `http://nemo-platform-api:8080`, matches neither the container URL (a 127.0.0.1 sidecar) nor the subprocess URL (the loopback platform), so it wouldn't rescue either path.

**2. This change introduces no new exposure channel.** A subprocess deployment runs on the platform host and posts to the platform's own base URL — loopback in every default configuration. More to the point, that same unvalidated `get_base_url()` already routes the agent's **prompts and completions** to the Inference Gateway via `inject_fabric_gateway_url()` (`utils.py:248`). Telemetry reuses the exact channel the deployment is already built on, carrying strictly less sensitive data than the inference traffic already crossing it. If the scheme is a problem, it's a problem one layer up and predates this PR.

**3. It would reintroduce the bug this PR fixes.** The default local install is `http://localhost:8080` (`nemo services run`). Requiring HTTPS would make the wiring refuse in exactly the single-machine and CI configuration this bug is about — converting "telemetry silently missing" into "telemetry deliberately refused."

**4. HTTPS is already honored, just not forced.** Nothing here pins `http`. An operator who configures an HTTPS base URL gets an HTTPS ingest endpoint automatically, and there's an existing test pinning that HTTPS ATIF endpoints survive rewriting (`test_rewrite_fabric_config_base_urls_preserves_https_atif_endpoint`).

**The one real thing in here:** an operator who sets a *non-loopback, plaintext* `platform.base_url` would have trajectories cross the network in cleartext. That's true — and equally true of every platform API call the agent makes, including inference. The right fix is a platform-wide base-URL scheme policy at config validation, not a scheme check in one telemetry call site. There's no such enforcement anywhere in the repo today, but we may want to consider it.

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -102,6 +104,46 @@ async def validate_platform_agent_config(config: dict[str, Any], *, base_dir: Pa
...
+    ``intake_export`` imports Fabric at module scope, and NAT deployments must
+    not pay for it.
+    """
+    # TODO(AIRCORE-902): Hoist once Fabric runtime is installed in the default Platform image.
+    from nemo_agents_plugin.telemetry.intake_export import (
+        configure_intake_atif_export,
+        supports_intake_atif_export,
+    )
+
+    if not supports_intake_atif_export(config, base_dir=base_dir):
```
benmccown, line 138: Small ordering detail, I think we could check `telemetry.enabled is False` first?

---

**Redundant plan resolution on the opt-out path**
*Priority: P3 · non-blocking · `runner/in_memory.py:136` · found by opus, sonnet*

#### Finding: Agent Summary

The support probe (`supports_intake_atif_export`, a synchronous Fabric `plan()` resolution) runs unconditionally at the top of `configure_intake_telemetry`, before the `telemetry.enabled is False` opt-out is noticed (that check lives downstream in `configure_intake_atif_export`). So a deployment that explicitly opted out still resolves a Fabric plan for nothing. Reordering the opt-out check ahead of the probe skips that work. Correct today, just wasteful — and pre-existing (the jobs path in `execute.py` has the same ordering).

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -102,6 +104,46 @@ async def validate_platform_agent_config(config: dict[str, Any], *, base_dir: Pa
...
+    sidecar. Subprocess has none, but it is already an auth-disabled shape: the
+    child carries only the placeholder gateway credential, so with platform auth
+    on, its inference calls fail well before its telemetry does. Wiring identity
+    here alone would half-fix a configuration nothing else supports.
+
+    Imported lazily for the same reason as :func:`validate_platform_agent_config`:
+    ``intake_export`` imports Fabric at module scope, and NAT deployments must
+    not pay for it.
+    """
+    # TODO(AIRCORE-902): Hoist once Fabric runtime is installed in the default Platform image.
```
benmccown: **Ticket ID in a code comment (team convention)**
*Priority: P3 · non-blocking · `runner/in_memory.py:130` (also `:101`, pre-existing) · found by all 4 reviewers*

#### Finding: Agent Summary

Our convention is no ticket IDs in code comments. No need to remove the TODOs themselves — they carry useful rationale ("hoist once Fabric runtime is in the default Platform image") — just drop the `AIRCORE-902` id from both occurrences (this line, and the pre-existing one at line 101).

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
```diff
@@ -102,6 +104,46 @@ async def validate_platform_agent_config(config: dict[str, Any], *, base_dir: Pa
...
+        supports_intake_atif_export,
+    )
+
+    if not supports_intake_atif_export(config, base_dir=base_dir):
+        return config
+
+    # The caller's mapping is the live AgentDeployment entity the controller
+    # re-saves, so wiring it in place would persist an inferred endpoint into
+    # the stored deployment.
+    wired = copy.deepcopy(config)
```
benmccown, line 144: Nice catch on the copy needing to happen, the fact that the controller re-saves the AgentDeployment entity would be easy to overlook. Glad we caught and pinned it with a test.

plugins/nemo-agents/tests/unit/test_runner_in_memory.py
```diff
@@ -664,3 +677,162 @@ def _spawn_fabric(self_, name, config_path, log_path, port, credential_env=None)
...
+    """An adapter with no Relay ATIF output runs untraced rather than failing.
+
+    Patched at the probe rather than expressed as a config, because "this
+    adapter does not advertise ATIF" is a property of the adapter descriptor,
+    not of anything the agent config can say.
+    """
+    backend = _backend(tmp_path)
+    config = _telemetry_agent_config()
+
+    with patch("nemo_agents_plugin.telemetry.intake_export.supports_intake_atif_export", return_value=False):
```
benmccown, line 834: not even sure if this was worth flagging, but passing it along. Take it or leave it.

---

**Lazy-import patch target will silently no-op after the planned hoist**
*Priority: nit · non-blocking · `test_runner_in_memory.py:834` · found by opus, sonnet*

#### Finding: Agent Summary

This patches `supports_intake_atif_export` at its definition module (`nemo_agents_plugin.telemetry.intake_export`). It works only because `in_memory` imports the symbol lazily inside the function on each call. When the `TODO` hoists that import to module scope, the patch becomes a no-op — the test would then pass without exercising the untraced branch. A one-line comment noting the coupling (or patching it in `in_memory`'s namespace instead) avoids the silent break. The sibling `test_execute_job.py` uses the caller-namespace idiom.

benmccown (approved):
A few small things, will let you take care of them. LGTM.

---

**Agent Summary** — 4-model independent review (opus, sonnet, gpt-astra, gemini-pro), verified against the diff. No blockers. 3 non-blocking notes inline (one convention fix, two optional). One reviewer finding was rejected as a false-positive on verification (a claim that `logs.items()` could yield dict tuples — it yields `PlatformJobLog` models, so the typecheck fix is correct).

plugins/nemo-agents/tests/unit/test_runner_in_memory.py
mikeknep: I think we can just adjust this when we actually move the production code import from in-function to module-level.

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
mikeknep: Yup, thanks, fixed in 4b745984a

plugins/nemo-agents/src/nemo_agents_plugin/runner/in_memory.py
mikeknep: Good call, reordered in 5e4121853

plugins/nemo-agents/src/nemo_agents_plugin/telemetry/intake_export.py
```diff
@@ -154,6 +137,53 @@ def configure_intake_atif_export(
...
+    ``None`` means leave the config alone, for any of the reasons that amount to
+    "someone already decided": an unreadable section, an explicit opt-out, a
+    declared ATIF destination, or ATIF declined while telemetry stays on.
+
+    Single source of truth for that reading, so the predicate callers ask up
+    front and the wiring itself cannot drift apart.
+    """
+    section = config.get("telemetry")
+    try:
+        telemetry = TelemetryConfig.model_validate(section if isinstance(section, dict) else {})
```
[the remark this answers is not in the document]
mikeknep: I don't think we can reach this with a wholly invalid telemetry config like that (a string, or None, instead of a structurally valid config)—we would have failed earlier when deserializing the wrapper: `AgentConfig.model_validate`

PR #2039, opened 2026-09-14
fix(models): clean up deployment volumes

mckornfield:
## Summary

Route Models deployment volume cleanup through the deployments plugin lifecycle so Kubernetes PVCs are deleted before their entities disappear. Previously, deleting a model deployment hard-deleted its Volume entities and orphaned the weights and scratch PVCs.

## Changes

- Mark unreferenced weights and scratch volumes `DELETING` and wait for the volume reconciler to remove them before reporting the model deployment `DELETED`.
- Preserve volumes that are still referenced by another deployment configuration.
- Keep recreate attempts in `CREATED` while prior asynchronous teardown finishes so the controller retries creation.
- Isolate scratch and weights teardown failures so one volume cannot block cleanup of the other.
- Cover the Models-to-volume-reconciler handoff, shared volumes, recreate retries, teardown timeout, and entity-store race outcomes.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: This corrects internal deletion lifecycle behavior without changing the API or user workflow.

## Verification

Targeted validation:

- `flox activate --dir . -- uv run pre-commit run -a` — passed.
- `flox activate --dir . -- uv run --frozen pytest services/core/models/tests/unit/controllers/backends/deployments_plugin plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py plugins/nemo-deployments/tests/unit/backends/k8s/test_volumes.py -q` — 129 passed.
- Kubernetes deployment E2E was not run locally; CI and the reported product test cover deployed-cluster validation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deployment deletion now waits for scratch and primary volume teardown to complete.
  - Volumes still referenced by other deployment configurations are preserved.
  - Missing entities and deletion conflicts are handled more reliably.
  - Failed volume deletions remain in progress instead of being incorrectly marked complete.
  - Recreating a deployment during prior teardown now reports the correct status.
  - Deployments can be retried successfully after previous teardown finishes.


services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py
```diff
@@ -209,11 +210,18 @@ async def delete_model_deployment(
...
+        volumes_removed = True
         for volume_name in (names.scratch, names.volume):
-            try:
-                await self._entity_client().delete(Volume, name=volume_name, workspace=workspace)
-            except NemoEntityNotFoundError:
-                pass
+            if not await self._complete_volume_delete(workspace, volume_name):
+                volumes_removed = False
+        if not volumes_removed:
+            result = DeploymentStatusUpdate(status="DELETING", status_message="Waiting for plugin volume teardown.")
```
benmccown, line 223: I think this is worth a follow up but doesn't have to block, up to you.

---

**Recreate can strand in PENDING when only a volume lingers during pre-create teardown**
*Priority: P2 · non-blocking · `services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py`:218*

#### Finding: Agent Summary

`create_model_deployment` calls `delete_model_deployment` first to clear leftover substrate, and returns `PENDING` (without creating the server) whenever that teardown reports `DELETING`. Now that volume teardown is asynchronous (this line), a recreate that catches a lingering weights/scratch `Volume` returns `PENDING` and never resumes creation — it just ages out to `ERROR` via `pending_timeout`.

- **What:** async volume teardown makes the create-path pre-teardown return `DELETING` → create returns `PENDING` with no server created.
- **Where:** `create_model_deployment` (`services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py`:87-92) interacting with this new volume-wait branch and `status.py::aggregate_status`.
- **When:** the narrow edge case below.
- **Why it matters:** the deployment sits `PENDING` until `pending_timeout` fires and flips it to `ERROR`, instead of recreating once the volume clears.

#### Finding: Agent Breakdown

This is a **narrow** edge case — it only triggers when **all** of these hold at once:

1. `create_model_deployment` runs its pre-create teardown (i.e. a recreate / drift-recovery `LOST → recreate`, not a fresh create), **and**
2. the server + puller `Deployment` entities are **already gone**, **but**
3. a weights/scratch `Volume` entity is **still lingering** (mid-teardown).

When that coincides, `delete_model_deployment` hits the new volume rung, `_complete_volume_delete` returns `False`, teardown returns `DELETING`, and the create path converts that to `PENDING` and returns before creating the server.

On the next cycle the reconciler sees `PENDING` and calls `get_model_deployment_status` → `aggregate_status(volume=<DELETING>, puller=None, server=None, previously_ready=False)`. Because `DELETING` is not in `_ATTENTION_SUBSTRATE_STATUSES` and `previously_ready` is `False` (so no `LOST`), it returns a plain `PENDING`. Plain `PENDING` does **not** trigger drift recovery, so creation is never resumed — the deployment just rides `pending_timeout` to `ERROR`.

Pre-PR this couldn't happen because teardown hard-deleted the volumes synchronously, so a lingering volume never produced a `DELETING` return from the create path.

Possible directions (for a follow-up, if you agree): have the create-path treat "only volumes remain" as retryable rather than converting to `PENDING`, or resume creation once the volumes clear. Not a blocker.

```
In services/core/models/.../deployments_plugin/backend.py, create_model_deployment (lines ~87-92) returns PENDING when its pre-create teardown reports DELETING. With async volume teardown, a recreate that catches a lingering weights/scratch Volume (Deployments already gone) returns PENDING and never resumes creation: aggregate_status returns plain PENDING for server=None + volume=DELETING (previously_ready=False), which does not trigger drift recovery, so it ages out to ERROR via pending_timeout. Verify this path, then make teardown-waiting creation retryable or explicitly resume creation once volumes disappear. Add a regression test for the recreate-with-lingering-volume case.
```

services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py
```diff
@@ -247,6 +255,41 @@ async def _complete_deployment_delete(self, workspace: str, deployment_name: str
...
+        volume = await self._get_optional(Volume, workspace, volume_name)
+        if volume is None:
+            return True
+
+        referencing = await deployment_config_names_referencing_volume(
+            self._entity_client(),
+            workspace=workspace,
+            volume_name=volume_name,
+        )
+        if referencing:
```
benmccown, line 280: Realistically I don't think we'd run into this, but it's an easy edge case to address if you want to knock it out. I'm also fine if you want to ack/ignore and merge.

---

**Preservation check runs before the already-DELETING check**
*Priority: nit · non-blocking · `services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py`:273*

#### Finding: Agent Summary

In `_complete_volume_delete` the `referencing → return True` (preserve) check runs before the `if volume.status == "DELETING": return False` check. If a volume were already `DELETING` and then became referenced by another config, this would report it as "preserved" while the `VolumeReconciler` deletes it anyway (it acts on `status == "DELETING"` regardless of references). The two signals contradict.

- **What:** the ordering means an already-`DELETING` + newly-referenced volume returns `True` (preserve) even though teardown proceeds.
- **Where:** `_complete_volume_delete`, `services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py`:273 (the `if referencing:` branch precedes the `DELETING` check at :282).
- **Why it matters:** logically inconsistent state reporting; harmless today only because per-deployment volume naming (`<name>-weights`/`<name>-scratch`) prevents cross-deployment sharing.

#### Finding: Agent Breakdown

Given the current naming scheme this is unreachable, so it's purely a nit about ordering robustness. If you want to close it, checking `volume.status == "DELETING"` before the referencing scan (or short-circuiting when already `DELETING`) makes the function's return value consistent with what the reconciler will actually do — and skips a full config scan in that case.

services/core/models/tests/unit/controllers/backends/deployments_plugin/test_backend.py
```diff
@@ -380,6 +381,87 @@ async def _get(entity_type: type, name: str, workspace: str | None = None) -> De
     assert second.status == "DELETED"
 
 
+@pytest.mark.asyncio
+async def test_delete_marks_weights_and_scratch_volumes_deleting_and_waits() -> None:
```
benmccown, line 399: **Untested branches: volume-wait timeout + NotFound/Conflict handoff outcomes**
*Priority: nit · non-blocking · `services/core/models/tests/unit/controllers/backends/deployments_plugin/test_backend.py`:385*

#### Finding: Agent Summary

The new tests cover the happy path (both volumes marked `DELETING`, model reports `DELETING`, reconciler tears down) and the preserve path. A few behavior branches introduced here have no direct coverage: the **volume-wait timeout** escalation (all Deployments gone, volumes stuck `DELETING`, elapsed > `deleting_timeout_seconds` → `ERROR`), and the `_complete_volume_delete` exception outcomes (`NemoEntityNotFoundError → True`, `NemoEntityConflictError → False`).

- **What:** no test drives the volume-branch timeout→ERROR, nor the NotFound/Conflict returns of `_complete_volume_delete`.
- **Where:** new tests at `services/core/models/tests/unit/controllers/backends/deployments_plugin/test_backend.py`:385 and :439; only the deployment-wait timeout branch is exercised elsewhere.
- **Why it matters:** these are the exact race/timeout paths the feature relies on; the timeout branch in particular is real behavior (a >timeout PVC teardown is a legitimate ERROR).

#### Finding: Agent Breakdown

The two branches mirror the established `_complete_deployment_delete` handling, so risk is low — but they're cheap to cover. Suggested additions:

- A volume-wait timeout test: Deployments already removed, both volumes returning `False` from `_complete_volume_delete` (status `DELETING`), `deleting_elapsed_seconds` past `deleting_timeout_seconds`, assert the result escalates to `ERROR` (parallel to the existing deployment-wait timeout test).
- Exception-branch tests on the volume `update()`: one raising `NemoEntityNotFoundError` (expect the volume treated as gone → loop advances), one raising `NemoEntityConflictError` (expect the model stays `DELETING` to retry).

Copy/paste this into your agent session to knock these out:
```
In services/core/models/tests/unit/.../deployments_plugin/test_backend.py, add: (1) a volume-wait timeout test — all Deployment entities gone, both volumes stay status=DELETING (_complete_volume_delete returns False), deleting_elapsed_seconds exceeds deleting_timeout_seconds; assert delete_model_deployment escalates to ERROR (mirror the existing deployment-wait timeout test). (2) two exception-branch tests on the Volume update() inside _complete_volume_delete: NemoEntityNotFoundError -> volume treated as gone (loop advances / reports DELETED), NemoEntityConflictError -> stays DELETING to retry next cycle.
```

benmccown (approved):
Thanks for grabbing this — routing the volume cleanup through the reconciler (soft-delete → `DELETING` → PVC teardown → entity delete) is the right shape, and the end-to-end handoff test that drives a real `VolumeReconciler` is a nice touch. Approving.

---

**Agent Summary**

Independent multi-model review (4 reviewers) of the volume-cleanup change. The fix is correct for the common path — PVCs are no longer orphaned on a normal delete, verified against the reconciler contract and the added tests. No blocking issues found.

Non-blocking follow-ups left inline:
- A narrow recreate-stranding edge case where async volume teardown can leave a recreate in `PENDING` until `pending_timeout` (P2).
- A nit on the ordering of the preserve vs. already-`DELETING` checks in `_complete_volume_delete`.
- A test-coverage nit (volume-wait timeout branch + the NotFound/Conflict handoff outcomes).

(Reviewed at head `7a05117d`. One out-of-scope observation about the k8s `delete_volume` failure path being handled separately.)

services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py
```diff
@@ -209,11 +210,18 @@ async def delete_model_deployment(
                     timeout_seconds=self._cfg.deleting_timeout_seconds,
                     deployment_name=name,
                 )
+        volumes_removed = True
         for volume_name in (names.scratch, names.volume):
-            try:
-                await self._entity_client().delete(Volume, name=volume_name, workspace=workspace)
-            except NemoEntityNotFoundError:
-                pass
+            if not await self._complete_volume_delete(workspace, volume_name):
```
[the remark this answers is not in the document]
mckornfield: Addressed in 4775120e0: each volume teardown is now caught and logged independently, and cleanup continues to the next volume. Added a regression test proving a scratch failure does not block weights cleanup.

services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py
mckornfield: Addressed in 4775120e0. While prior substrate teardown is still DELETING, create now returns CREATED so the Models controller invokes create again on the next reconciliation cycle. The regression test exercises a lingering teardown followed by successful substrate creation.

services/core/models/src/nmp/core/models/controllers/backends/deployments_plugin/backend.py
mckornfield: Addressed in 4775120e0: the already-DELETING check now precedes the reference scan, keeping the result consistent with the volume reconciler and avoiding the unnecessary scan. Added direct coverage for the ordering.

services/core/models/tests/unit/controllers/backends/deployments_plugin/test_backend.py
mckornfield: Addressed in 4775120e0 with direct tests for the volume-wait timeout, update NotFound, and update Conflict outcomes. The expanded targeted suite passes with 129 tests.

PR #2040, opened 2026-09-14
feat(evaluator): execute stored Harbor tasks and tasksets

ngoncharenko:
## Summary

Run stored Harbor tasks through agent evaluation using task references or a taskset reference, with definitions and scoring resolved into a self-contained snapshot before job creation. Workers consume that snapshot, verify and materialize the native archives for execution, or rescore saved trials without archive access or task/taskset reads.

Related issue: [AALGO-481](https://linear.app/nvidia/issue/AALGO-481). Related notebook PR: [#2136](https://github.com/NVIDIA-NeMo/nemo-platform/pull/2136).

### Before / After UX

- **Before:** stored definitions did not provide an end-to-end path from a local Harbor package to execution through the agent-evaluation job. Callers needed a local native dataset.
- **After:** `publish_harbor_tasks()` publishes verified archives from a single task or dataset and optionally registers immutable task/taskset references. Submit those references through the existing `tasks` input; submission expands and validates them before creating a job.
- Stored Harbor definitions can declare custom metrics and views alongside mandatory Harbor reward scoring. Saved-trial rescoring preserves primary reward metadata and secondary rewards without running the agent or verifier again.

### Submission snapshot and task kinds

- Public selectors remain inline evaluator tasks, a homogeneous list of `TaskRef`, or a `TasksetRef`; both list variants advertise nonempty membership in OpenAPI.
- Canonical jobs contain an ordered, nonempty `list[ResolvedTask]`. Each envelope carries its runtime ID, resolved definition, metadata, and optional stored-entity provenance. Harbor requires provenance and an ID matching `native_task_id`.
- Resolve member revisions with bounded, ordered concurrency; resolve metrics/models in each stored task's workspace. Snapshot the original revision digest as provenance rather than recomputing it from expanded metrics.
- Validate collection identities, execution support, scoring, and views at submission. Workers revalidate canonical data and execution/scoring support before preparing runtime tasks.

```mermaid
flowchart TB
    Input["AgentEvalInputSpec<br/>inline tasks, TaskRef list, or TasksetRef"]
    Input --> Spec["AgentEvalSpec<br/>ordered ResolvedTask list"]
    Spec --> Task["ResolvedTask<br/>id, provenance, spec, metadata"]
    Task --> Definition{"spec.kind"}
    Definition --> Evaluator["ResolvedEvaluatorTaskDefinition<br/>inputs, reference, inline metrics, views"]
    Definition --> Harbor["ResolvedHarborTaskDefinition<br/>archive identity, instruction, config<br/>inline metrics and views"]
```

More detailed version:

```mermaid
classDiagram
    direction TB

    class AgentEvalInputSpec {
        <<submission contract>>
        target : Target?
        trials : AgentEvalTrial[]?
        tasks : TasksetRef | TaskRef[] | AgentEvalTaskInput[]
    }

    class AgentEvalSpec {
        <<canonical job snapshot>>
        target : Target?
        tasks : ResolvedTask[]  «min 1»
        invariant : all tasks have one kind
    }

    class ResolvedTask {
        id : str  «evaluation identity»
        provenance : TaskProvenance?  «required when spec.kind = harbor»
        invariant : harbor ⇒ id == spec.native_task_id
        spec : ResolvedTaskDefinition
        metadata : MetadataItem[]
    }
    class TaskProvenance {
        entity_name : "ws/name"  «qualified, no fragment»
        revision_digest : 64-hex digest
    }
    class StoredHarborTask {
        <<verified published member>>
        definition : HarborTaskDefinition
    }
    class ResolvedTaskDefinition {
        <<discriminated union>>
    }
    class ResolvedEvaluatorTaskDefinition {
        kind = "evaluator"
        intent, inputs, reference, views
        metrics : MetricInline[]
    }
    class ResolvedHarborTaskDefinition {
        kind = "harbor"
        native_task_id, source, harbor_hash, instruction, config, views
        metrics : MetricInline[]
    }
    class Task {
        <<stored entity>>
        name ; spec : TaskDefinition ; metadata
    }

    AgentEvalInputSpec ..> AgentEvalSpec : resolves at submission
    AgentEvalSpec "1" *-- "1..*" ResolvedTask : tasks
    ResolvedTask "0..1" --> TaskProvenance : provenance
    ResolvedTask *-- ResolvedTaskDefinition : spec
    TaskProvenance <|-- StoredHarborTask
    ResolvedTaskDefinition <|-- ResolvedEvaluatorTaskDefinition
    ResolvedTaskDefinition <|-- ResolvedHarborTaskDefinition
    Task ..> ResolvedTask : snapshotted at submission

```

### The kind TaskKindAdapter
Without this seam, `kind` is matched in five places: the would-be id in `load_tasks`,
the `match` in `snapshot_task`, the per-kind branch in `validate_scoring`, the Harbor rule
in `validate_execution_support`, and the worker's group-by-kind. Every new kind grows all five.
Deleting any one site makes the kind knowledge reappear in its callers, so none of them is a
pass-through; they are five copies of one seam. Two adapters exist today and "gym", "verifier", etc. might be added,
so the seam is real. It is internal to the plugin, not a plugin entry point.

The registry selects an adapter by task-definition kind. Submission and worker code use
its five methods through the same interface.

```mermaid
classDiagram
    direction TB
    class Registry {
        get_adapter(kind)
    }
    class TaskKindAdapter {
        <<Protocol>>
        runtime_id()
        resolve()
        accepts_target()
        validate_scoring()
        prepare()
    }
    class EvaluatorTaskAdapter
    class HarborTaskAdapter
    Registry --> TaskKindAdapter : selects by kind
    TaskKindAdapter <|.. EvaluatorTaskAdapter
    TaskKindAdapter <|.. HarborTaskAdapter
```

### Submit and run flow

```mermaid
flowchart TB
    A["Load selectors and selected revisions"] --> B["Snapshot definitions, resolved metrics, and provenance"]
    B --> C["Validate collection and target support"]
    C --> D["Prepare Gym submission when applicable"]
    D --> E["Validate scoring; construct canonical AgentEvalSpec"]
    E --> F["Persist job snapshot"]
    F --> G["Worker: validate canonical spec, target support, and scoring"]
    G --> H["Prepare tasks through kind adapters"]
    H --> I{"Stored Harbor mode"}
    I -->|Online| J["Authenticated archive download<br/>hash, extraction, native identity and loader checks"]
    I -->|Saved trials| K["Build scoring tasks from snapshot and trial rewards<br/>no archive access or task/taskset lookup"]
    H -->|Evaluator tasks| L["Reconstruct SDK tasks"]
    J --> M["Run evaluator and publish task/trial/score artifacts"]
    K --> M
    L --> M
```

### Archive publication and integrity

- Store one native task package per `tar-gzip-v1` archive, referenced by qualified Fileset object path and SHA-256 of the compressed bytes. Preserve permission bits, dotfiles, ignored files, and empty directories; reject symlinks and special files.
- Verify archive bytes on publication readback, registration, and worker download. Bound capture/extraction and metadata reads; reject unsafe paths and native identity mismatches, then publish assembled datasets atomically.
- Keep entity revision digests, archive byte hashes, and Harbor fingerprints distinct. Harbor fingerprints are provenance, not an integrity gate.
- Support upload-only mode, explicit replacement, destination/member-name overrides, and receipts retaining completed progress after failure. Validate bare registration names before remote writes.
- Reject explicitly blank instructions after SPDX-comment stripping without rewriting archive bytes. Document canonical archive-reference restrictions in the schema.
- Preserve the separate taskset-source adapter's pinned-reference resolution behavior. Stored Harbor tasksets reject shared `files_ref`; package required runtime files with each task.
- Share digest-verified registration through `publish_pinned`; allow publishing tasks without creating a taskset. Reapply requested tags to matching content, and reconcile ambiguous transport failures only with a newly published matching revision. Validate Gym-compatible rows before submission.
- Keep the coverage-comment workflow reliable under `pipefail` by letting `grep` stop after the first coverage attribute instead of piping it into `head`.
- Update evaluator OpenAPI, SDK configuration, docs, and regression coverage. Studio reads task counts directly from the canonical list; remove obsolete pinned-source branches that caused TypeScript errors.

### Compatibility and review entry points

This is a clean canonical-job cutover: drain or resubmit jobs persisted with the previous task shape; no legacy reader or migration is provided. Archive availability and metric evidence remain runtime dependencies. Saved Harbor trials need explicit, consistent primary-reward metadata.

Start with [agent_spec.py](https://github.com/NVIDIA-NeMo/nemo-platform/blob/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py), [task_refs.py](https://github.com/NVIDIA-NeMo/nemo-platform/blob/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/task_refs.py), and [jobs/kinds](https://github.com/NVIDIA-NeMo/nemo-platform/tree/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/jobs/kinds). Then review [preparation.py](https://github.com/NVIDIA-NeMo/nemo-platform/blob/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/harbor/preparation.py), [materialization.py](https://github.com/NVIDIA-NeMo/nemo-platform/blob/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/harbor/materialization.py), and [SDK upload helpers](https://github.com/NVIDIA-NeMo/nemo-platform/blob/07d5c4129431b0a4e321a1c5c448fc84cfbf81fd/plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py).

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

**Published head `07d5c4129`, checked September 21:** fixed the only failed GitHub job from the previous run (`Post coverage comment`). Its log showed `grep: write error: Broken pipe` because `grep ... | head -1` ran under `bash -o pipefail`; commit `d20202603` replaces those pipelines with `grep -m 1`. Commit `10013dbf5` addresses the four fresh CodeRabbit findings: the persisted resolved-task shape, per-case validation messages, applying tags to identical content, and revision-safe transport reconciliation. The branch was then synchronized with `origin/main` in signed merge `2d51ac080`; commit `07d5c4129` documents the shared-CI `pipefail` failure beside the guard. The new GitHub run is pending.

Fresh local validation on this head:

- `.venv/bin/pytest plugins/nemo-evaluator/tests/test_harbor_publication.py plugins/nemo-evaluator/tests/jobs/kinds/test_evaluator_kind.py -q`: **52 passed**.
- `.venv/bin/ruff format ...` and `.venv/bin/ruff check ...` on all five changed Python files: **passed**.
- `.venv/bin/ty check plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py`: **passed**.
- `flox activate --dir tools/actionlint -- actionlint .github/workflows/ci.yaml`: **passed**.
- The affected integration case collected successfully; its `offline-refs` variant was **skipped locally** because the subprocess-platform fixture is unavailable in this environment. The previous GitHub integration suite passed.
- `flox activate -- .venv/bin/pre-commit run -a`: **all hooks passed** using the repository-pinned uv and helm-docs toolchain.
- `git diff --check`: **passed**. DCO audit: **all seven PR commits passed** with author-matching sign-offs.

Earlier implementation evidence retained from prior heads:

- Non-integration evaluator suite: **1,347 passed, 42 deselected**.
- Focused contract/publication run: **51 passed**; SDK discovery/loader subset: **4 passed**.
- `make refresh-openapi`, web SDK generation/typecheck, changed-file Ruff and `ty`, and `make docs-check TOOLCHAIN=system` passed. External Python SDK generation remained blocked at Stainless by an organization-migrated unauthorized response.

## Summary by CodeRabbit

* **New Features**
  * Added unified Harbor task publishing, registration, tasksets, archive verification, and secure materialization.
  * Added support for direct task references, pinned tasksets, resolved task snapshots, custom metrics, views, and offline Harbor rescoring.
  * Added task-kind handling for evaluator and Harbor workflows.
  * Added validation for Gym task inputs and Harbor task identities, archives, paths, and content hashes.

* **Documentation**
  * Updated Harbor upload, registration, custom-scoring, taskset, and built-in agent guidance.

* **Bug Fixes**
  * Improved revision consistency, error reporting, cleanup, and task counting across evaluation workflows.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/taskset_sources.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @field_validator("materialized_root")
+    @classmethod
+    def _validate_materialized_root(cls, value: Path) -> Path:
+        if not value.is_absolute():
+            raise ValueError("materialized_root must be an absolute path")
+        return value
+
+
+class TasksetSourceAdapter(Protocol):
```
SandyChapman, line 43: Do these need to live in the SDK? Does anything use this protocol there?

plugins/nemo-evaluator/pyproject.toml
```diff
@@ -8,13 +8,17 @@ readme = "README.md"
...
+    "filesets",
     "nemo-evaluator-sdk",
     "nemo-platform-plugin",
     "pydantic>=2.12.0",
     "typer>=0.20.0",
 ]
 version = "0.0.0"
 
+[project.entry-points."nemo.experimentalist.taskset_sources"]
+evaluator = "nemo_evaluator.harbor.taskset_source:factory"
```
SandyChapman: I don't understand this linkage. Is `experimentalist` defining their own data model for tasksets that we're opting into here?

plugins/nemo-evaluator/openapi/openapi.yaml
```diff
@@ -5160,6 +5198,38 @@ components:
...
+          const: harbor-taskset
+          title: Kind
+          default: harbor-taskset
+        taskset_ref:
+          $ref: '#/components/schemas/TasksetRef'
+      additionalProperties: false
+      type: object
+      required:
+      - taskset_ref
+      title: PinnedHarborTaskset
```
SandyChapman, line 5250: Do we need these new types? Can we just use the existing `Taskset` schema and put a metadata item of `kind: harbor`?

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
```diff
@@ -0,0 +1,245 @@
...
+            if revision.revision == published.revision:
+                if revision.content_hash != intended:
+                    raise ValueError("Published revision content mismatch")
+                return f"{workspace}/{name}#{revision.content_hash}"
+        if history.pagination is None or page >= history.pagination.total_pages:
+            raise ValueError("Published revision missing from history")
+        page += 1
+
+
+def upload_harbor_task(
```
SandyChapman, line 157: We have in `sdk/task_resources.py` an `EvaluatorTasksResource.create` function. Is there a way we can just reuse that API and do a type check on the task type? This seems like a parallel implementation not attached to a resource type.

The process would be:
```python
tasks = discover_harbor_tasks(Path("./somepath"))
task_refs = []
for task in tasks
  result = client.evaluator.tasks.create(task)
  task_refs.append(result.ref) # or whatever
taskset = client.evaluator.taskset.create(tasks=task_refs, ... other params)
```

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -303,12 +304,16 @@ async def to_spec(
...
-        # A `tasks` taskset reference is loaded and expanded into inline task DTOs first, so the
-        # metric-ref resolution below is identical whether the tasks were submitted inline or via a
-        # stored taskset.
+        # Evaluator references expand before metric resolution. Stored Harbor sources stay
+        # pinned in the canonical job and resolve into native tasks during worker preparation.
         task_inputs = await resolve_agent_eval_tasks(
-            submit_spec.tasks, workspace=workspace, entity_client=entity_client
+            submit_spec.tasks, workspace=workspace, entity_client=entity_client, target=submit_spec.target
         )
+        if not isinstance(task_inputs, list):
```
SandyChapman: When would the resolved tasks not be a list here?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -328,6 +333,13 @@ async def to_spec(
                     metadata=task.metadata,
                 )
             )
+        if isinstance(submit_spec.target, GymRunnerTarget):
+            for task in resolved_tasks:
+                gym_task_row(
+                    task_id=task.id,
+                    inputs=task.inputs.model_dump(exclude_none=True),
+                    metadata={item.key: item.value for item in task.metadata},
+                )
```
SandyChapman, line 351: Does this block do anything?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -631,7 +643,17 @@ def run(
...
+            tasks = [_to_runtime_task(task) for task in spec.tasks]
+        else:
+            from nemo_evaluator.harbor.preparation import prepare_stored_harbor_tasks
+
+            tasks = prepare_stored_harbor_tasks(
+                spec.tasks,
+                destination_root=ctx.storage.persistent / "harbor-inputs",
+                sdk=client,
+                async_sdk=async_client,
+            )
```
SandyChapman: I was assuming this would happen in the `to_spec` call above. Is there a reason to delay materialization of the harbor tasks?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
```diff
@@ -438,12 +438,20 @@ def _reject_empty_inline_tasks(self) -> Self:
 
 
 class AgentEvalSpec(_AgentEvalSpecCommon):
-    """Canonical agent-evaluation spec: tasks with all metric references resolved to inline."""
+    """Canonical evaluation: resolved evaluator tasks or an immutable stored Harbor source."""
 
-    tasks: list[AgentEvalTaskSpec] = Field(min_length=1, description="Tasks to evaluate; at least one is required.")
+    tasks: Annotated[list[AgentEvalTaskSpec], Field(min_length=1)] | PinnedHarborSource
```
SandyChapman: I'm not sure I understand why we have a `PinnedHarborSource` type. Can you expand on this?

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/taskset_sources.py
ngoncharenko: Yes, the plan was that the standalone Experimentalist will use this protocol - this is part of the integration plan. Even if it's not Experimentalist, another consumer should be able to implement this protocol. It shouldn't be plugin specific

plugins/nemo-evaluator/pyproject.toml
ngoncharenko: Good catch. This is a part of integration with Experimentalist, but considering this is deferred and not consumed at this point, we should not put it here. Removed

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: `task_inputs: list[AgentEvalTaskInput] | PinnedHarborTaskset | PinnedHarborTaskList`


Example: tasks="default/suite#latest" becomes a descriptor containing default/suite#<exact-digest>, rather than an expanded list.
We need exact digests and not just revisions since by the time the job is started being executed `#latest`, for example, can point to a different revision
```
ordinary evaluator inputs → list[AgentEvalTaskInput]

Harbor target + taskset reference → PinnedHarborTaskset
Harbor target + task references  → PinnedHarborTaskList
```

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: The call validates that both required fields are mappings. If either is missing or has an invalid type, it raises ValueError. It rejects unsuitable Gym tasks during canonicalization, before environment resolution or job execution. For example, a generic instruction-only task cannot silently proceed to Gym.

```python
inputs["gym_row"]
metadata["gym_row_extras"]
```

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: I think we all agreed that downloading and resolution should happen on the worker and not API layer. This keeps job definition lean; I/O intensive work goes into a job. Submission layer should be lean

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
ngoncharenko: It gives the canonical job a serializable representation of tasks whose files have not yet been materialized.
For example:
```json
{
  "kind": "harbor-taskset",
  "taskset_ref": "default/suite#<exact-digest>"
}
```
If user submits "#latest", and it changes after submission but before job starts, the job still selects the recorded digest.

plugins/nemo-evaluator/openapi/openapi.yaml
ngoncharenko: Stored Harbor tasksets already use the existing Taskset schema. These new types describe a canonical job’s input source, rather than another stored taskset entity:

```txt
PinnedHarborTaskset  = one exact taskset revision
PinnedHarborTaskList = exact task revisions supplied without a taskset
```

A `kind: harbor` metadata item would label a taskset, but would not enforce digest-pinned references or cover the direct-task-list case (https://github.com/NVIDIA-NeMo/nemo-platform/pull/2040#discussion_r4040145645)

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
ngoncharenko: Registration already goes through [EvaluatorTasksResource.create/replace and EvaluatorTasksetsResource.create/replace](https://github.com/NVIDIA-NeMo/nemo-platform/blob/aalgo-481-stored-harbor-task-execution/ngoncharenko/plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py#L80); there is no separate persistence implementation.
The additional operation is publishing local files before registration:

```python
local task directory → verified Files archive → TaskInput → tasks.create(...)
```

I agree the current public helpers obscure that composition. Action items I'd address:
- separate upload-only preparation from registration in the implementation and document the explicit resource-based flow
- retain the combined convenience helper for dataset preflight (better UX than chaining calls yourself), partial-progress receipts, and returning the exact published revision.

plugins/nemo-evaluator/openapi/openapi.yaml
SandyChapman: I still don't understand. Can't we just use a TasksetRef? Why do we need a type tightly coupled to harbor?

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
SandyChapman: Yes, makes sense, what I'd hope to avoid is needing helpers dedicated to specific task types. If we can have one path that can discriminate between task type internally, it makes the public API surface smaller and easier to understand.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Resolving from a taskset ref to a specific revision can happen in the to_spec call though and we'd have a consistent reference throughout. Is there a race condition where it would matter?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Ah, so we rely on an exception as a side effect to calling `gym_task_row`? It might be helpful to at least do `_ = gym_task_row(...)` and add a comment about this behaviour or consider adding a specific validation function to use instead.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
SandyChapman: Agree that downloading the full tasks should be server side, but resolving a taskset or list[taskref] to a list of task entities is probably not a huge lift, is it? Maybe it's worth diving into this a bit more in parking lot.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
SandyChapman: See my comment above. I'd expect the to_spec call to get the pinned revision. I also don't think we need to carry a kind in this type and could probably put it in the taskset metadata instead.

plugins/nemo-evaluator/openapi/openapi.yaml
ngoncharenko: Yes, I can move this logic to Tasksetref

ngoncharenko: Done. `PinnedHarbor*` is gone; the wire spec still takes `TasksetRef | list[TaskRef] | inline`.
`to_spec` pins the ref to a digest and snapshots each task into one `ResolvedTask` envelope whose
`spec` is discriminated on the stored `kind`, so Gym, Verifier, or something else later is one more union member.

I did not add `kind: harbor` taskset metadata: it is derivable from the pinned members,
unverifiable as a copy, and cannot describe a mixed taskset. Happy to add it separately as a UI
hint if you still want it. Adding it as metadata also conflicts with the idea of mixed kind of tasks withing a single Taskset

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: Resolution moved into `to_spec`; the canonical `tasks` is always a list now. No race: the digest
is pinned at submit and the worker never rereads entities.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: Done as you described: refs, pins, and metrics resolve at submit; only archive bytes are fetched
on the worker

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_spec.py
ngoncharenko: Same change as the openapi thread: `PinnedHarborSource` deleted, `to_spec` pins the revision,
`kind` comes from the stored definition. Reasoning on taskset metadata is there too.

plugins/nemo-evaluator/pyproject.toml
ngoncharenko: Removed

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: Took the explicit-function option: `validate_gym_task_row` in the SDK Gym dataset module, same
rule as `gym_task_row`, and `prepare_gym_submission` calls it. No discarded return value. Tests
added for missing and well-formed rows.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
ngoncharenko: Shrunk the surface. `upload_harbor_task` and `upload_harbor_dataset` are now one
`publish_harbor_tasks(path, ...)`; a single task is a dataset of one, `taskset_name=None`
registers tasks without a taskset. Registration itself is the kind-neutral `publish_pinned` in
`task_resources.py`, so any future kind composes `discover → TaskInput → publish_pinned`.

`publish_pinned` requires input in stored form (metric refs, qualified pinned members, verified
Harbor projection), checked before any write, because the server rewrites inline metrics and
member refs and the client-side pin check hashes what you sent. `verified_harbor_definition`
covers the Harbor case for direct callers. A fully kind-agnostic `tasks.publish(...)` waits for a
second kind with a publish path and a server-side normalized hash; tracked as a follow-up.

sdk/python/nemo-platform/.nmpcontext/stainless.yaml
```diff
@@ -177,7 +177,7 @@ readme:
       params: {}
     pagination:
       type: request
-      endpoint: get /apis/audit/v2/workspaces/{workspace}/jobs
+      endpoint: get /apis/files/v2/workspaces/{workspace}/filesets
```
SandyChapman, line 180: Was this edit intentional? It seems unrelated.

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -115,7 +115,306 @@ your own Harbor suite.
...
+from pathlib import Path
+
+from nemo_evaluator.api.schemas import TaskInput
+from nemo_evaluator.sdk.harbor import publish_harbor_tasks
+from nemo_evaluator.sdk.task_resources import EvaluatorTasksResource
+from nemo_platform_plugin.client.client import NemoClient
+from nemo_platform_plugin.evaluator.client import EvaluatorClient
+
+client = NemoClient(base_url="http://localhost:8080", workspace="default")
+tasks_resource = EvaluatorTasksResource(EvaluatorClient.from_client(client))
```
SandyChapman, line 162: I know you didn't introduce this, but the new way of getting to plugin `Resource` objects is not very user friendly.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
```diff
@@ -224,3 +233,132 @@ async def delete(self, name: str, *, workspace: str | None = None) -> None:
...
+
+
+def publish_pinned(
+    resource: EvaluatorTasksResource | EvaluatorTasksetsResource,
+    name: str,
+    value: TaskInput | TasksetInput,
+    *,
+    workspace: str,
+    replace: bool = False,
+) -> str:
```
SandyChapman, line 279: A couple of questions on this function:
1. It seems to overload both task and taskset publishing. We have separate task_resources.py and taskset_resources.py files and I'd have assumed we'd split publishing as well.
2. Why use a free function here instead of adding it as a member? I think this makes more sense to add as a method in `EvaluatorTasksResource` and `EvaluatorTasksetsResource`. We have `replace_task` and `replace_taskset` on the API side as well which I think duplicates some of the logic here (unless I'm misunderstanding the logic below).

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
```diff
@@ -0,0 +1,288 @@
...
+    *,
+    client: NemoClient,
+    workspace: str = "default",
+    taskset_name: str | None = None,
+    fileset_ref: str | None = None,
+    path_prefix: str | None = None,
+    register: bool = True,
+    replace: bool = False,
+    member_names: dict[str, str] | None = None,
+) -> HarborDatasetUploadDetails:
```
SandyChapman, line 108: Same question as last time. Why doesn't a caller use `discover_harbor_tasks` and then `EvaluatorTasksResource.create`? This way we can avoid the harbor specific implementations.

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
```diff
@@ -310,3 +310,13 @@ def _require_pinned_task_refs(refs: list[TaskRef]) -> list[TaskRef]:
...
+
+
+def _require_pinned_taskset_ref(ref: TasksetRef) -> TasksetRef:
+    name, separator, digest = ref.root.partition("#")
+    if "/" not in name or not separator or _DIGEST_FRAGMENT_PATTERN.fullmatch(digest) is None:
+        raise ValueError(f"Expected a qualified immutable digest reference, got {ref.root!r}")
+    return ref
+
+
+PinnedTasksetRef: TypeAlias = Annotated[TasksetRef, AfterValidator(_require_pinned_taskset_ref)]
```
SandyChapman, line 322: I think it's likely better to just make a new type for this:
```python
class TasksetRevisionRef(RootModel[str]):
    """Doctstring here"""
    root: str = Field(
        pattern="whatever-regex-you-want-here",
        description="Reference to a stored taskset revision (format: workspace/taskset-name#<tag-or-digest>, or taskset-name#<tag-or-digest> in the "
        "job workspace).",
    )
```

plugins/nemo-evaluator/src/nemo_evaluator/jobs/kinds/evaluator.py
```diff
@@ -0,0 +1,78 @@
...
+            kind="evaluator",
+            intent=definition.intent,
+            inputs=definition.inputs.model_copy(deep=True),
+            reference=definition.model_dump(include={"reference"})["reference"],
+            metrics=metrics,
+            views={key: view.model_copy(deep=True) for key, view in definition.views.items()},
+        )
+
+    def accepts_target(self, target: Target | None, tasks: Sequence[ResolvedTask]) -> bool:
+        return not isinstance(target, HarborRunnerTarget) or all(task.provenance is None for task in tasks)
```
SandyChapman, line 68: I think it should be an allowlist here rather than a denylist.

plugins/nemo-evaluator/src/nemo_evaluator/harbor/archive.py
```diff
@@ -0,0 +1,359 @@
...
+MAX_STEPS = 128
+MAX_EXTENSION_BYTES = 1024**2
+MAX_EXTENSION_CHAIN = 8
+
+
+@dataclass(frozen=True)
+class NativeTask:
+    task_id: str
+    instruction: str | None
+    config: dict[str, Any]
```
SandyChapman, line 45: What does `NativeTask` represent?

plugins/nemo-evaluator/src/nemo_evaluator/harbor/materialization.py
```diff
@@ -0,0 +1,151 @@
...
+            with private_directory(owned) as member_parent:
+                definition = HarborTaskDefinition.model_validate(source.definition.model_dump())
+                root, native = await _download_async(definition, files_client, member_parent)
+                results.append(_member(source, root, native, owned))
+        return _publish(owned, results)
+
+
+def materialize_harbor_tasks_sync(
+    members: Sequence[StoredHarborTask], *, files_client: FilesClient, destination_root: Path
+) -> MaterializedHarborTasks:
```
SandyChapman, line 132: I think in general, this code could use some docstrings (for public functions) and maybe a description of what each module is responsible for.

plugins/nemo-evaluator/src/nemo_evaluator/harbor/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+    entity_name: str
+    revision_digest: str = Field(pattern=DIGEST_PATTERN, min_length=DIGEST_LENGTH, max_length=DIGEST_LENGTH)
+
+    @field_validator("entity_name")
+    @classmethod
+    def _identity(cls, value: str) -> str:
+        if "#" in value or "/" not in value:
+            raise ValueError("entity_name must be qualified without a revision fragment")
+        TaskRef(value)
+        return value
```
SandyChapman, line 46: Some of the types and functions in the modules under `harbor/` seem to be generic enough that they could be reused. It might be helpful to refactor them to their own module.

plugins/nemo-evaluator/src/nemo_evaluator/harbor/taskset_source.py
```diff
@@ -0,0 +1,70 @@
...
+from nemo_evaluator_sdk.agent_eval.taskset_sources import TasksetSourceAdapter, TasksetSourceMaterialization
+from nemo_platform import AsyncNeMoPlatform
+from nemo_platform_plugin.client.adapter import client_from_platform
+from nemo_platform_plugin.entities import EntityClient
+from nemo_platform_plugin.entities.client import AsyncEntitiesClient
+from nemo_platform_plugin.files.client import AsyncFilesClient
+from pydantic import TypeAdapter
+
+
+class EvaluatorTasksetSourceAdapter:
```
SandyChapman, line 21: We may want to rename to indicate this is coupled to harbor tasksets.

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -254,24 +261,21 @@ class AgentEvalResultFiles:
...
-        # The runtime task carries plain dicts; the typed DTOs collapse to them, with arbitrary
-        # task-specific inputs preserved and key/value metadata pairs folded into a mapping.
-        inputs=task.inputs.model_dump(exclude_none=True),
-        reference=task.reference,
-        metrics=[_runtime_metric(metric) for metric in task.metrics],
-        views=task.views,
-        metadata={item.key: item.value for item in task.metadata},
-    )
+async def prepare_gym_submission(
+    snapshots: Sequence[ResolvedTask], target: Target | None, ctx: SubmitContext
```
SandyChapman, line 265: What's a snapshot? Is it the same as a task revision?

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
```diff
@@ -254,24 +261,21 @@ class AgentEvalResultFiles:
...
+) -> Target | None:
```
SandyChapman, line 266: Maybe adjust this to take a `target: GymRunnerTarget` instead of a `Target | None`.

sdk/python/nemo-platform/.nmpcontext/stainless.yaml
ngoncharenko: yes, missed this, will remove that

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: No, a snapshot here is the submission-time ResolvedTask embedded in the job spec.

I can rename it to resolved_task. The reason call it snapshot is to highlight that tasks are loaded with provenance pinned, metrics are resolved

AgentEvalInputSpec_to_ResolvedTaskDefinition

plugins/nemo-evaluator/src/nemo_evaluator/api/fields.py
ngoncharenko: removed the code that uses this alias and this alias as well

plugins/nemo-evaluator/src/nemo_evaluator/harbor/tasks.py
ngoncharenko: done

plugins/nemo-evaluator/src/nemo_evaluator/harbor/taskset_source.py
ngoncharenko: removed this completely. No consumer since Experimentalist is not active

plugins/nemo-evaluator/src/nemo_evaluator/jobs/agent_evaluate.py
ngoncharenko: done

plugins/nemo-evaluator/src/nemo_evaluator/harbor/materialization.py
ngoncharenko: done

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
ngoncharenko: done

plugins/nemo-evaluator/src/nemo_evaluator/sdk/harbor.py
ngoncharenko: Can publish this way now


```python
from pathlib import Path

from nemo_evaluator.sdk.discovery import discover_harbor_tasks
from nemo_evaluator.sdk.resources import Evaluator
from nemo_platform_plugin.evaluator.client import EvaluatorClient

evaluator = Evaluator(client=EvaluatorClient(base_url="http://localhost:8080", workspace="default"))
remote_task_ids = []
for source in discover_harbor_tasks("./mydataset"):
    remote_task = evaluator.tasks.replace(source.source_dir.name, task=source)
    remote_task_ids.append(remote_task.id)
evaluator.tasksets.create("suite", tasks=remote_task_ids)
```

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: updated the snippet to use `evaluator = Evaluator(client=EvaluatorClient(base_url="http://localhost:8080", workspace="default"))`

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/taskset_sources.py
ngoncharenko: Removed

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -115,7 +115,285 @@ your own Harbor suite.
...
+concurrent task updates before that call may be included. Later task updates do not change the
+stored taskset. Task publication and taskset publication are separate operations.
+
+Continuing with `evaluator` above, upload and verify an archive without registering a task using `prepare`:
+
+```python
+from nemo_evaluator.sdk.task_sources import HarborTaskSource
+
+source = HarborTaskSource(source_dir=Path("./my_task"))
+prepared = evaluator.tasks.prepare(source)
```
SandyChapman: Why is the plugin method different from the SDK method?

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -115,7 +115,285 @@ your own Harbor suite.
...
+evaluator_client = EvaluatorClient(base_url="http://localhost:8080", workspace="default")
+evaluator = Evaluator(client=evaluator_client)
+source = HarborTaskSource(
+    source_dir=Path("./my_task"),
+    additional_metrics=[ExactMatchMetric(reference="Done", candidate="{{sample.output_text}}")],
+    views={"quality": SemanticView(reducer=SemanticReducer.MEAN, signals=[
+        ViewSignal(metric="harbor_reward", output="reward"),
+        ViewSignal(metric="exact-match", output="exact-match"),
+    ])},
+)
```
SandyChapman: I like this approach!

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -115,7 +115,285 @@ your own Harbor suite.
...
+payload = evaluator_client.submit_agent_eval_job(
+    body=SubmitAgentEvalJobRequest(spec=spec.model_dump(mode="json")),
+).data()
+job = AgentEvaluatorJobResource(
+    job=AgentEvaluatorJob.model_validate(payload.model_dump(mode="json")),
+    client=evaluator_client, workspace="default",
+)
+job.wait_until_done()
+results = evaluator.agent_eval_results.list(job_id=job.name)
```
SandyChapman: Is there any way we can use `evaluator.submit(...)` instead of the `submit_agent_eval_job` func?

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -178,11 +456,166 @@ Swap `agent_name` (or `agent_import_path`) for your own agent to get a real scor
...
+```json
+{
+  "name": "nemo-evaluator.agent-h1f68xpa",
+  "status": "completed"
+}
+```
+
+Repeat the curl command while the status is `created`, `pending`, or `active`.
+
+</Tab>
```
SandyChapman: I think in general, we only really need to show the CLI and SDK methods of submission.

plugins/nemo-evaluator/src/nemo_evaluator/harbor/archive.py
ngoncharenko: added the docstring. It holds validated metadata extracted from a Harbor task directory: native task ID, optional root instruction, and parsed task.toml

plugins/nemo-evaluator/src/nemo_evaluator/jobs/kinds/evaluator.py
ngoncharenko: done

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
```diff
@@ -224,3 +233,132 @@ async def delete(self, name: str, *, workspace: str | None = None) -> None:
...
+            raise
+        existing = None
+    previous_revision = existing.revision if existing is not None else 0
+    if existing is not None and _digest(existing) == intended:
+        missing_tags = any(existing.tags.get(tag) != existing.revision for tag in value.tags)
+        if missing_tags:
+            if isinstance(resource, EvaluatorTasksResource) and isinstance(value, TaskInput):
+                published = resource.replace(name, workspace=workspace, task=value)
+            elif isinstance(resource, EvaluatorTasksetsResource) and isinstance(value, TasksetInput):
+                published = resource.replace(name, workspace=workspace, taskset=value)
```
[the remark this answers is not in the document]
ngoncharenko: Superseded by the publication API refactor: publish_pinned and its tag-restoration branch were removed. Source-write failures now preserve prepared input in TaskPublicationError for explicit retry without re-uploading.

plugins/nemo-evaluator/tests/test_harbor_publication.py
```diff
@@ -0,0 +1,567 @@
...
+    monkeypatch.setattr("nemo_evaluator.sdk.harbor.EvaluatorClient.from_client", lambda _: pytest.fail("No DB calls"))
+    if dataset:
+        upload_details = publish_harbor_tasks(root, client=Mock(), register=False)
+        assert len(upload_details.members) == 1
+        assert upload_details.taskset_ref is None
+        member = upload_details.members[0]
+        fileset_name, prefix = "harbor-tasks", "task/"
+    else:
+        member = publish_harbor_tasks(root, client=Mock(), register=False).members[0]
+        fileset_name, prefix = "harbor-tasks", "task/"
```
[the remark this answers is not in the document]
ngoncharenko: Superseded by the publication API refactor: publish_harbor_tasks(register=False) and this test were removed. Discovery is now separate from single-task preparation; test_harbor_discovery_root_shapes covers both a dataset root and a task directory.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
```diff
@@ -23,32 +34,200 @@ class EvaluatorTasksResource:
...
+        self,
+        name: str,
+        *,
+        task: HarborTaskSource,
+        project: str | None = None,
+        workspace: str | None = None,
+        metric_bundle_packager: MetricBundlePackager | None = None,
+        fileset_ref: str | None = None,
+        path_prefix: str | None = None,
+    ) -> Task: ...
```
SandyChapman: I think we should not require the user to pass the fileset_ref in here. the `create` function should be able to do an isinstance check on `task`'s type and do the fileset creation on behalf of the user. Because we're calling prepare in the create/replace body, the user shouldn't need to do it.

I think this in inconsistent as the docs in harbor-runner.mdx show manually calling prepare before create/replace.

docs/evaluator/agent-eval/harbor-runner.mdx
```diff
@@ -115,7 +115,285 @@ your own Harbor suite.
...
+from nemo_evaluator.jobs.agent_spec import AgentEvalInputSpec, HarborRunnerTarget
+from nemo_evaluator.sdk.job_resources import AgentEvaluatorJob, AgentEvaluatorJobResource
+from nemo_platform_plugin.evaluator.types import SubmitAgentEvalJobRequest
+from nemo_evaluator_sdk.agent_eval.tasks import SemanticReducer, SemanticView, ViewSignal
+from nemo_evaluator_sdk.metrics.exact_match import ExactMatchMetric
+from nemo_platform_plugin.evaluator.client import EvaluatorClient
+
+evaluator_client = EvaluatorClient(base_url="http://localhost:8080", workspace="default")
+evaluator = Evaluator(client=evaluator_client)
+source = HarborTaskSource(
```
SandyChapman: Maybe we can add an `__iter__` func that returns an iterator that iterates over multiple tasks in the given directory. Then we can have discover_harbor_tasks return that type instead and can unify the way we work with tasks in the SDK and the plugin.

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
```diff
@@ -23,32 +34,200 @@ class EvaluatorTasksResource:
...
+            - Resolve the workspace and validate the registration name.
+            - Prepare local sources, or validate that a ready ``TaskInput`` has no preparation options.
+            - Create the entity and preserve prepared input in ``TaskPublicationError`` if registration fails.
+        """
+        workspace = self._client.resolve_workspace(workspace)
+        if len(name) > MAX_NAME_LENGTH or re.fullmatch(NAME_PATTERN, name) is None:
+            raise ValueError("Invalid task registration name")
+        TaskRef(f"{workspace}/{name}")
+        is_source = not isinstance(task, TaskInput)
+        if is_source:
```
SandyChapman: This logic might need to be flipped. `needs_preparation = isinstance(task, TaskSource)` or similar.

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: We have two `discover_harbor_tasks`: one on the plugin side that wraps the one on the sdk side. Do you prefer to keep both or have a single `discover_harbor_tasks`? My understanding the later. Is it correct?

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
ngoncharenko: done

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: Done - single `discover_harbor_tasks()` but there's a price for this - no longer can keep  HarborTaskSource since a lot of dependencies do not accept it and need `AgentEvalTask`

| Consumer | Needs |
|---|---|
| `AgentEvaluator.run`: scoring, views, summary | `id`, `intent`, `inputs`, `metrics`, `views`. This is `AgentEvalTask`. |
| `HarborAgentTaskRunner.run_tasks(tasks)` | The fields above, plus where the package is on disk. |
| Plugin publication (`tasks.create/replace/prepare`) | Where the package is on disk, to capture, upload and verify it. Plus metrics and views. |
| Persisted results (`tasks.jsonl`) and rescoring | `AgentEvalTask` only. The location must not leak into them. |

plugins/nemo-evaluator/src/nemo_evaluator/sdk/task_resources.py
ngoncharenko: done

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: good to know, thanks. Updated

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: yes, updated the examples

plugins/nemo-evaluator/src/nemo_evaluator/sdk/_executor.py
```diff
@@ -262,19 +262,26 @@ def submit_agent_eval(
...
         cannot silently run something other than what was tested.
         """
+        if placement is not None and target is None:
+            raise TypeError("placement requires target; saved trials are not executed")
         resolved_workspace = self._client.require_workspace(self._workspace)
-        spec = AgentEvalInputSpec(tasks=tasks, target=runner_to_target(target, placement))
+        spec = AgentEvalInputSpec(
+            tasks=tasks,
+            target=runner_to_target(target, placement) if target is not None else None,
+            trials=trials,
```
ngoncharenko: added trials for offline scoring and tested on my local examples

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_runtime.py
```diff
@@ -687,7 +736,7 @@ def _stamp_coverage(
         # Same reasoning as _task_dirs_for: a malformed sibling task must not fail a
         # run. Recording only the requested tasks just costs a re-run later.
         return tasks
-    covered = {task.id: task for task in discovered}
+    covered: dict[str, AgentEvalTask] = {task.id: task for task in discovered}
```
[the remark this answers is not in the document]
ngoncharenko: done

docs/evaluator/agent-eval/harbor-runner.mdx
ngoncharenko: A better option might be `exclude=True` with AgentEvalTask and remove `HarborTaskCollection`

```python
class AgentEvalTask(BaseModel):
    ...
    source_dir: Path | None = Field(default=None, exclude=True,
        description="Local package the runner executes (Harbor). Not serialized.")
```

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/agent_eval/runtimes/harbor_archive.py
```diff
@@ -0,0 +1,330 @@
...
+    if (
+        not target
+        or len(os.fsencode(target)) > 4096
+        or os.path.isabs(target)
+        or lexical == ".."
+        or lexical.startswith("../")
+        or not inside
+    ):
+        raise ValueError(f"Symlink escapes task root: {path.relative_to(root.parent).as_posix()}")
+    return target
```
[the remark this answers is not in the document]
ngoncharenko: done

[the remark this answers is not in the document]
ngoncharenko: Fixed in 540625d04: `check_internal_symlink` now rejects links that resolve to a directory containing the link (`a -> .`, `tests/up -> ..`), since Harbor's `shutil.copytree` follows directory links and would recurse. Added a parametrized regression test.

SandyChapman (approved):
this looks great. thanks for pushing this over the line!

PR #2042, opened 2026-09-14
fix(cpu-tasks): install agents and insights plugins in the task image

mikeknep:
## Summary

`ExecuteAgentJob` declares `container = "cpu-tasks"` and `insights.analyze` hard-codes the `nmp-cpu-tasks` image, but neither `nemo-agents-plugin` nor `nemo-insights-plugin` was a member of the `cpu-tasks` dependency group — and both jobs launch their task module with `python -m`, so both are a guaranteed `ModuleNotFoundError` in any built `nmp-cpu-tasks` image. This adds both to the group, which also brings the first-party Fabric adapters (claude, codex, deepagents, hermes) into the image.

Before: `insights.analyze` cannot import on the image it selects, and `nmp-cpu-tasks` carries no Fabric adapters. After: the image can run `agents.execute` and every shipped harness, and `insights.analyze` imports.

**This does not change which image `agents.execute` runs on today.** `_execution_image()` still returns the `nmp-api` fallback; dropping that fallback belongs with the follow-up that adds a per-job `image` field, which keeps this PR behavior-neutral for existing jobs and independently revertible. **It also does not make a user's own Fabric adapter runnable** — that needs a packaged image plus the per-job image field. "Custom agents now work" is the misreading to avoid.

## Changes

- `pyproject.toml`: add `nemo-agents-plugin` and `nemo-insights-plugin` to the `cpu-tasks` dependency group.
- `uv.lock`: regenerated. Group membership only — package resolution is unchanged at 599 packages, because both plugins were already resolved via `enabled-plugins`.
- `docker/Dockerfile.nmp-cpu-tasks`: three checks added to the existing `smoke-test` stage — the task `__main__` entrypoints resolve, the Fabric adapter data directory is populated, and the vendored `claude` / `codex` CLIs actually launch.
- `docker/scripts/check-bundled-harness-clis.py`: the harness-CLI check, extracted from an inline `python -c` so it can distinguish `OSError` / timeout / non-zero exit and so ruff and ty cover it. Bind-mounted into the build rather than copied, matching the `remove-listed-files.py` idiom already used here and in `nmp-api` / `nmp-customizer-tasks` / `nmp-unsloth-training`.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no user-visible surface changes. The execute-job image is an implementation detail today and `docs/agents/execute-agent-jobs.mdx` does not reference it; the docs change lands with the follow-up that adds the user-facing `image` field.

The added coverage is the Dockerfile `smoke-test` stage rather than a unit test, deliberately: the failure mode is an import error inside a built image. A unit test asserting `pyproject.toml` group membership would pass while the image is broken. That stage is already built in CI via the `docker-cpu-ci` bake group, so no workflow change was needed.

## Image size — please read before approving

This is the cost of the change, and it belongs in review rather than being discovered later by whoever owns image budgets. Measured with fresh `linux/arm64` builds of `nmp-cpu-tasks-docker`, merge-base vs. this branch:

| | Baseline | This PR | Δ |
|---|---|---|---|
| Image (`docker images`) | 2.12 GB | 3.18 GB | **+1.06 GB (+50%)** |
| `/app/.venv` | 1342 MB | 2112 MB | +770 MB |

Largest contributors to the venv delta: `codex_cli_bin` 309 MB, `claude_agent_sdk` (bundled `claude`) 260 MB, `litellm` (via `harbor`) 110 MB.

Why land it whole rather than splitting the harnesses behind an extra:

- `agents.execute` is meant to be the primitive other plugins use for agentic workloads. A default image that runs any shipped harness makes it usable with no image-selection step; a pure-Python-only default pushes every Claude/Codex consumer into packaging before they can run anything.
- The immediate consumer of this change needs the `claude` / `codex` adapters specifically, so a split shipping only `relay` / `hermes` / `deepagents` in the default image would not unblock it.
- `nmp-api` already carries these same wheels, so this is not new attack or CVE surface for the platform as a whole — only new surface for `nmp-cpu-tasks`.

If the increase is unacceptable to the task-image owners, the lever is a `nemo-agents-plugin[harnesses]` extra carrying `nemo-fabric[claude,codex]` (~570 MB of the delta), with operators registering an execution profile whose `default_task_image` points at a harness-capable image. Splitting NAT instead is not the lever — roughly 50 MB for real churn. Happy to do the extra as a follow-up with this size data in hand.

### Source-compliance note

`nemo-relay-cli-bin` and `nemo-fabric-runtime` publish no sdist, and this image runs `collect-python-sdists.py` for source compliance. This is not a new problem: `nmp-api` already builds with these same packages through the same step, and the script records gaps to a `missing-pypi-sdists.txt` manifest rather than failing the build. This PR extends an existing gap to a second image. Flagging it so it is a reviewed tradeoff rather than a surprise. My local builds ran with `NMP_COLLECT_SOURCES=0`, so CI is the first place that step runs against the new packages.

## Verification

Targeted validation:

1. `uv run pre-commit run -a` — all 15 hooks pass. (`Check for uv.lock drift` is `git diff --exit-code uv.lock` and fails on an unstaged lock by design; it passes once staged/committed.)

2. `uv run --frozen pytest plugins/nemo-agents/tests/unit/test_execute_job.py plugins/nemo-insights/tests -q` — **864 passed, 1 skipped**. No existing test needed changing, because `_execution_image()` is untouched.

3. Built both images locally for `linux/arm64` (`make docker-load TARGET=nmp-cpu-tasks-docker`), baseline at merge-base and this branch. Against the image built from this branch:

   ```
   $ docker run --rm --entrypoint /app/.venv/bin/python <img> -c '<task entrypoints>'
   ok: ['nemo_agents_plugin.tasks.execute.__main__', 'nemo_insights_plugin.jobs.bridge']

   $ docker run --rm --entrypoint /app/.venv/bin/python <img> -c '<adapter dir>'
   ['claude', 'codex', 'deepagents', 'hermes', 'insights-analyst']

   $ docker run --rm --entrypoint /app/.venv/bin/python <img> -c '<bundled binaries>'
   {'claude': ('/app/.venv/lib/python3.13/site-packages/claude_agent_sdk/_bundled/claude', True, True),
    'codex':  ('/app/.venv/lib/python3.13/site-packages/codex_cli_bin/bin/codex', True, True)}
   ```

   The third is the check that proves the headline capability: a wheel-platform mismatch, a stripped executable bit, or a future CVE-removal rule aimed at a ~260 MB vendored binary would all pass the import and adapter checks and fail only there. All three are now in the `smoke-test` stage.

   Per CodeRabbit's review, that third check now **launches** each CLI (`--version`) rather than stat-ing it. An `isfile` + `os.access(X_OK)` test cannot detect a binary built for the wrong architecture or one missing its dynamic loader — which is exactly the wheel-platform mismatch it is meant to catch, so the original version did not fully support the claim above. Verified in both directions: both CLIs report versions inside the built image, and the check exits non-zero when pointed at a non-launchable binary.

4. Built the `nmp-cpu-tasks-smoke-test` bake target locally for `linux/arm64`; all five stage steps pass, with the harness check reporting `claude 2.1.211` and `codex-cli 0.144.4` from inside the image.

5. `uv run ruff check`, `uv run ruff format --check` and `uv run --frozen ty check` pass on the new script. (`ty` caught a real defect in the first draft: `_find_bundled_cli` is an instance method that was being passed `None` as `self`. The script now resolves the binary from `claude_agent_sdk.__file__` and touches no private API.)

Not run: `linux/amd64` builds (local arm64 host only) — CI covers both architectures.

## Summary by CodeRabbit

- **New Features**
  - CPU task containers now support agent execution and insights analysis capabilities.

- **Bug Fixes**
  - Improved validation detects missing task components, required data files, and bundled command-line tools before use.

- **Tests**
  - Added comprehensive CPU task container smoke tests covering evaluator and environment imports, task entry points, data files, and bundled tools.
  - Smoke tests now report all detected validation failures together for easier troubleshooting.


docker/Dockerfile.nmp-cpu-tasks
```diff
@@ -88,4 +88,35 @@ FROM runtime AS smoke-test
...
+# The claude and codex CLIs ship as vendored native executables inside their
+# wheels. A wheel-platform mismatch, a stripped executable bit, or a future
+# CVE-removal rule aimed at a ~240 MB vendored binary all pass the import and
+# adapter checks above and fail only here.
+RUN /app/.venv/bin/python -c \
+    'import os, sys; \
+     from claude_agent_sdk._internal.transport.subprocess_cli import SubprocessCLITransport as T; \
+     from codex_cli_bin import bundled_codex_path; \
+     bins = {"claude": T._find_bundled_cli(None), "codex": bundled_codex_path()}; \
+     bad = {n: p for n, p in bins.items() if not (p and os.path.isfile(p) and os.access(p, os.X_OK))}; \
```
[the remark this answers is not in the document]
mikeknep: Good catch, and fixed in 0a8b0d230 — this was a real gap, not a style nit. `isfile` + `os.access(X_OK)` cannot detect a binary built for the wrong architecture or one missing its dynamic loader, which is exactly the wheel-platform mismatch the check exists to catch, so the PR description was overclaiming what it proved.

The check now launches each CLI with `--version` (non-mutating, no network or credentials) and treats `OSError`, timeout and non-zero exit as failures. It moved to `docker/scripts/check-bundled-harness-clis.py` so those cases can be reported distinctly and so ruff and ty cover it.

Verified both directions: inside the built image the check reports `claude 2.1.211` and `codex-cli 0.144.4`, and it exits non-zero with a clear message when pointed at a non-launchable path.

Extracting it also surfaced a second defect `ty` then caught: `_find_bundled_cli` is an instance method that was being handed `None` as `self`, working only because the body ignores it. The script now resolves the binary from `claude_agent_sdk.__file__` and touches no private API.

docker/Dockerfile.nmp-cpu-tasks
```diff
@@ -88,4 +88,35 @@ FROM runtime AS smoke-test
...
+# The claude and codex CLIs ship as vendored native executables inside their
+# wheels. A wheel-platform mismatch, a stripped executable bit, or a future
+# CVE-removal rule aimed at a ~260 MB vendored binary all pass the import and
+# adapter checks above and fail only here.
+#
+# Actually launch each CLI rather than stat-ing it: an isfile/X_OK check cannot
+# detect a binary built for the wrong architecture or a missing dynamic loader,
+# which is precisely the wheel-platform mismatch this is meant to catch. Both
+# `--version` calls are non-mutating and need no network or credentials.
+COPY docker/scripts/check-bundled-harness-clis.py /tmp/check-bundled-harness-clis.py
```
ironcommit: nit: it's possible to mount the script into the build, so you don't have to delete it afterwards. That would reduce this to a single layer with no artifacts.

ironcommit (approved):

docker/Dockerfile.nmp-cpu-tasks
```diff
@@ -88,4 +88,35 @@ FROM runtime AS smoke-test
...
+RUN /app/.venv/bin/python -c \
+    'import sys, importlib.util as u; \
+     mods = ["nemo_agents_plugin.tasks.execute.__main__", "nemo_insights_plugin.jobs.bridge"]; \
+     missing = [m for m in mods if u.find_spec(m) is None]; \
+     sys.exit("missing task entrypoints: %r" % missing) if missing else print("task entrypoints:", mods)'
+
+# The Fabric native runtime discovers adapters by listing this directory, so the
+# adapters must survive into the image as data files, not merely be importable.
+RUN /app/.venv/bin/python -c \
+    'import os, sys, sysconfig; \
```
mckornfield: do we want an inline python RUN or something like your script? Same for the other above

pyproject.toml
```diff
@@ -250,6 +250,12 @@ functional-services = [
 cpu-tasks = [
```
mckornfield, line 250: kinda back to the complexity side, wonder if this should just be whatever is in API? I guess plus some extras that the API might not need but 🤷

mckornfield (approved):

pyproject.toml
mikeknep: I think the opposite is probably true too: nmp-api might carry some things that nmp-cpu-tasks doesn't need. It does seem like there's probably a way to simplify the management of these images though.

docker/Dockerfile.nmp-cpu-tasks
mikeknep: Done in 56ef5d7e8 — good call, and it matches what this file already does for `remove-listed-files.py` a few lines up (and `nmp-api` / `nmp-customizer-tasks` / `nmp-unsloth-training`). Should have followed the local idiom the first time.

```dockerfile
RUN --mount=type=bind,source=docker/scripts/check-bundled-harness-clis.py,target=/tmp/check-bundled-harness-clis.py,readonly \
    /app/.venv/bin/python /tmp/check-bundled-harness-clis.py
```

Worth noting the `rm` was doing even less than it looked: `COPY` writes the script into its own layer, so removing it in the next layer only masks it. The mount leaves nothing behind at all. Smoke stage is down from 5 steps to 4.

No artifact impact either way, since the `smoke-test` stage is `output=cacheonly` and never published — but one pattern per file is worth having. Rebuilt the stage locally to confirm the mount resolves; the check still reports `claude 2.1.211` and `codex-cli 0.144.4` from inside the image.

docker/Dockerfile.nmp-cpu-tasks
mikeknep: I played around with this a bit and I like just collapsing all of these into a single smoke-test Python script. It allows us to check multiple things in one layer, and brings the tests into more proper control (e.g. type-checking) so they should hopefully be easer to maintain. Consolidated these in fed104c60

docker/scripts/smoke-test-cpu-tasks.py
```diff
@@ -0,0 +1,206 @@
...
+    """Resolve the `python -m` entrypoints for agents.execute and insights.analyze."""
+    failures: list[str] = []
+
+    for name in TASK_ENTRYPOINTS:
+        # find_spec imports the parent packages to locate a submodule, so it
+        # raises ModuleNotFoundError when the parent is absent and returns None
+        # only when the leaf is. A plugin missing from the cpu-tasks group hits
+        # the raising path, which is precisely the case being tested for.
+        try:
+            found = importlib.util.find_spec(name) is not None
```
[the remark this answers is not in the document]
mikeknep: Willing to live with this

PR #2046, opened 2026-09-14
feat(studio): create an agent from a GitHub repository and show its commit [ASTD-593]

marcusds:
## Summary

Lets a user create an agent from a GitHub repository in Studio, and shows which commit that agent's files come from — on the agent header, in a Source panel on the Details tab, and per deployment.

## Changes

- **GitHub import gets its own tab** in the new-agent modal. It previously shared the upload tab, arbitrated by "a parsed repository URL beats picked files". With both sources visible that rule was invisible and would have become a trap: type a repository, switch back, pick a directory, and the repository still wins. Submit is keyed on the active tab, so each source owns its fields, validation, submit-enabled state and errors.
- **A completed import used to do nothing.** `useCreateAgentFromGitHub` was constructed without options, so only the upload path had an `onSuccess`: the agent was created server-side and the modal just sat there. Both mutations share one handler now, and the pending state covers the import (submit previously stayed live during the request).
- **Source panel** on the Details tab: repository, tracked ref, pinned commit, and "Update to latest" which calls the refresh endpoint. Renders nothing for an uploaded directory, which has no revision.
- **Commit badge** beside the health badge in the header, using the same `Badge` the status uses, linking to the Source panel.
- **Per-deployment revision**: "Staged from commit `abc1234`", and when the fileset has moved on, "· source is now `def5678`". Both link to GitHub. A running deployment keeps serving the commit it staged and picks up the newer one when the runner next stages the fileset, so the two can legitimately differ.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: Studio UI has no separate user documentation in this repository.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test` — 3567 passed. The 3 failures are pre-existing in `src/util/date.test.ts`, which this branch does not touch: they assert US locale date formatting and this machine produces ISO.
- `pnpm -r typecheck` — clean across all 6 packages.
- `pnpm --filter nemo-studio-ui lint` — clean at `--max-warnings 0`.
- `uv run pre-commit run -a` — not run in full. `helm-docs` is not installed here and `uv-lock` requires uv 0.9.14 against 0.9.30 locally; both pass in CI. This branch touches no Helm chart or `pyproject.toml`.

Manually verified in a running Studio against a local platform, importing a private repository: the fileset is created under the name deployments stage from, the modal navigates to the new agent, the header badge deep-links to the Source panel, "Update to latest" moves the fileset to a newly pushed commit, and an existing deployment keeps reporting the commit it staged.

## Notes for review

The header badge is a `Badge` rather than a `Tag` because it matches the status badge beside it; KUI documents badges as non-interactive, so a reviewer may prefer the deep-link move to a different element. It carries a commit icon so its baseline matches the icon-bearing status badge — a text-only badge sits two pixels lower under baseline alignment.

Part of ASTD-519. Closes ASTD-593.

## Summary by CodeRabbit

* **New Features**
  * Import agents directly from GitHub repositories, including private repositories with selected secrets.
  * Create agents from imported traces.
  * View agent source repositories and revisions, with support for refreshing tracked commits.
  * See staged deployment commits with links to GitHub.
  * Filter datasets by GitHub storage.
* **Bug Fixes**
  * Improved validation, conflict handling, retry messages, and cleanup for failed imports.
  * Improved handling of missing, unreadable, pinned, and conflicting agent source filesets.
* **Tests**
  * Expanded coverage for GitHub imports, source panels, deployments, and trace-based creation.


web/packages/common/src/components/ExternalLink/index.tsx
nakolean, line 1: **Nit**: Suprised we don't have this in `packages/common` already. Should we move this there for discoverability so an agent doesn't recreate this in the future?

nakolean (approved):

web/packages/common/src/components/ExternalLink/index.tsx
marcusds: Good call.

web/packages/studio/src/routes/agents/AgentsListRoute/NewAgentModal/index.tsx
```diff
@@ -277,117 +336,175 @@ export const NewAgentModal: FC<NewAgentModalProps> = ({ open, onClose, workspace
...
+                  trace import skill to get started.
+                </Text>
+              </Flex>
+            ) : (
+              <Stack gap="density-sm">
+                <Label>Unregistered agents that appear in ingested traces</Label>
+                <Select
+                  aria-label="Agent from imported traces"
+                  value={tracedAgent}
+                  onValueChange={setTracedAgent}
```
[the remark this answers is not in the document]
marcusds: Claude says: Confirmed and fixed in 2667284399. `resetTracedMutation` was only called from `resetAndClose`, so `tracedCreateError` outlived the selection it came from — picking another agent left the old failure sitting above a name that had not been tried. `onValueChange` now resets the mutation before updating the selection, matching what `beginSelection` already does on the upload tab.

Added a regression test (`drops a failed create once another agent is chosen`); verified it fails without the one-line change and passes with it.

PR #2052, opened 2026-09-14
feat(studio): show deployment status for custom models and adapters

aray12:
## Summary

The Custom Models list resolved deployment state by fetching `base_model`, so any row without one — every base model that hosts adapters — fell through to a red "No deployment found" even while a READY deployment was serving it. Adapter subrows rendered no indicator at all, and because `StatusIndicator` has no neutral colour and defaults to red, the undeployed state was also alarming rather than quiet.

This resolves status the way the model side panel already does — matching `toInferenceModelEntityId()` against each provider's `served_models`, the same table Inference Gateway routes on — and renders it as a Status column of badges. A green badge now means a request would actually resolve, rather than that a deployment was merely requested.

Before: a base model with a READY deployment showed red "No deployment found"; its adapters showed nothing. After: it reads **Deployed**, a loaded adapter reads **Served**, and an adapter whose base is up but which is not itself loaded reads **Not served** — a state that was previously indistinguishable from having nothing deployed.

image

> **Note:** this overlaps #1987 (merged earlier today), which fixed the same root cause more narrowly by pointing `DeploymentIndicator` at the row's own model instead of its `base_model`. That intent is preserved and extended here; this change removes `DeploymentIndicator` in favour of the badge column, so #1987's file is deleted rather than modified. Flagging for reviewer awareness.

## Changes

- Add `useModelDeploymentIndicator`, resolving a model or adapter row to a discriminated state (`loading` / `served` / `adapter-not-loaded` / `not-deployed`) by matching its inference entity id against each provider's `served_models`, then following the matched provider to its deployment.
- Add `DeploymentStatusBadge`, rendering that state via the shared `StatusBadge` with a local status config. Top-level rows read Deployed; adapters read Served, since an adapter has no deployment of its own and is loaded into the one serving its base model.
- Replace the fixed-width dot gutter in `CustomModelsDataView` with a Status column, and render it for adapter subrows as well as top-level rows.
- Return the matched `ModelProvider` from `useServedModel` so callers can reach `model_deployment_id` without re-resolving the ref. Additive; existing callers are unaffected.
- Delete `DeploymentIndicator` and its dot-rendering logic.
- Keep failed lookups out of the status: `useServedModel` now reports `isError` when a referenced provider could not be fetched, and the state machine has an `unknown` kind so a transient 5xx or stale provider reference is not rendered as "Not deployed". The `served` state carries `hasDeployment`, so "Available" is reserved for providers that name no deployment, and a deployment that exists but could not be read falls to "Unknown" rather than a green badge. An incomplete provider read is also checked *before* the adapter base probe, so a model with several providers cannot report "Not served" on the strength of one readable provider while another went unread.

- Add an `AllDeploymentStatuses` story to `CustomModelsDataView.stories.tsx` with one row per Status value, driven through the real resolution chain (models → `model_providers` → provider `served_models` → deployment) rather than by stubbing the hook.

`useModelDeploymentStatus` is deliberately untouched — it still backs `useModelChatAvailability`, and changing it would ripple into the chat tab.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal Studio table column; no user-facing docs describe this list's columns.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pnpm --filter nemo-studio-ui test src/components/dataViews/CustomModelsDataView src/hooks` | Pass — 9 files, 68 tests |
| `pnpm --filter nemo-studio-ui typecheck` | Pass |
| `pnpm --filter nemo-studio-ui lint` | Pass (`--max-warnings 0`) |
| `pnpm --filter="...[origin/main]" run --parallel --if-present typecheck` | Pass — `packages/common`, `packages/studio` |
| `pnpm --filter="...[origin/main]" run --parallel --if-present test:ci` | Pass — studio 367 files / 3636 tests |
| `uv run pre-commit run -a` | **Two hooks blocked locally — see below.** All others passed. |

New tests: 15 cases in `useModelDeploymentIndicator/index.test.tsx` covering each state transition — including the regression where a row with no `base_model` resolved as undeployed despite a READY provider, and the error paths where a provider or deployment lookup fails — and 11 cases in `DeploymentStatusBadge.test.tsx` pinning the label and colour for each state.

The multi-provider regression test was verified to fail against the pre-fix ordering (`expected 'adapter-not-loaded' to be 'unknown'`) before the fix was restored, so it is genuine regression coverage rather than a test that merely passes.

Review follow-up in `fac8dc4ae` also fixed a latent defect in the test helper: `modelDeploymentId` was a defaulted parameter, so passing `undefined` re-applied the default and the "provider without a deployment" case never exercised that path.

**Blocked pre-commit hooks** — both are local environment gaps, neither touches this diff (no Helm or Python files are changed):

- `Helm Docs` — `helm-docs` binary is not installed locally.
- `Run uv lock with platform uv` — local uv is 0.9.28; the platform pins 0.9.14. The companion `Check for uv.lock drift` hook passed, and `uv.lock` is unmodified here.

Expecting CI to cover both.

**How each state was verified:** `Deployed`, `Served`, and `Not deployed` were confirmed against a running platform. The remaining states cannot be produced by real data on demand — `Not served` needs a base redeployed with LoRA disabled, `Unknown` needs a provider request to fail, and `Deploying`/`Failed`/`Deleting`/`Unavailable`/`Available` need a deployment caught in a particular state — so all ten are covered by the `AllDeploymentStatuses` story and by unit tests. Every badge in that story was visually confirmed to render as documented.

**Pre-existing story note:** the older `WithData` story mocks models and deployments but no providers, so with this change its rows all read `Unknown`. It predates this resolution path. Left as-is to keep the diff scoped; worth adding provider mocks in a follow-up.

**Known limitation:** the Status column is derived per row on the client (providers → `served_models` → deployment), so it is not sortable or filterable. Every other column in this table filters; making this one filter would require backend support.

## Summary by CodeRabbit

* **New Features**
  * Added clearer deployment status badges for models and adapters.
  * Displays deployed, served, not deployed, not served, deploying, failed, deleting, unavailable, available, and unknown statuses.
  * Shows loading indicators while deployment information is retrieved.
  * Provides adapter-specific labels and accounts for external providers.
  * Added a dedicated Status column to custom model views.

* **Bug Fixes**
  * Status remains unknown when provider information cannot be retrieved, avoiding misleading results.


aray12:

- #2052 ◀
    - #2083

web/packages/studio/src/components/dataViews/CustomModelsDataView/index.tsx
marcusds: Should this use useCallback?

marcusds (approved):

PR #2063, opened 2026-09-15
feat(automodel): wire deployment_config through to auto-deploy

aray12:
## Summary

`deployment_config` auto-deploys a model when a customization job finishes. The automodel service already compiled and validated it end to end, but nothing could reach it: the plugin split moved the HTTP boundary to `AutomodelJobInput`, which never carried the field, and `adapter.py` hardcoded `deployment_config=None` when mapping the plugin spec onto the legacy `CustomizationJobOutput`. Submitting a job with `deployment_config` on the automodel backend silently deployed nothing. It now deploys, matching the unsloth backend.

## Changes

- Add `DeploymentParams` / `ToolCallParams` to `nemo_automodel_plugin.schema`. They subclass `AutomodelSchema`, so they emit `AutomodelDeploymentParams` / `AutomodelToolCallParams` and cannot collide with the unsloth models in the merged `/apis/customization` spec.
- Expose `deployment_config: str | DeploymentParams | None` on `AutomodelJobInput` and `AutomodelJobOutput`.
- Pass it through `transform_input_to_output`.
- Forward it in `automodel_spec_to_compiler_output` via a new `_build_deployment_config` (string refs pass through, inline params are re-validated from the dumped plugin dict).
- Reject `lora_enabled=false` on a LoRA job at submit time, on `AutomodelJobInput` — the real HTTP boundary. The identical guard on the legacy `CustomizationJobInput` was unreachable. The "is this a standalone adapter" test is factored into `trains_standalone_lora_adapter()` and reused by `_infer_output_type`.
- Regenerate `plugins/nemo-customizer/openapi/openapi.yaml`.

No change was needed in `services/automodel/.../app/jobs/compiler.py` — the forwarding at `compiler.py:262` and `_validate_deployment_config` light up once a non-`None` value arrives.

## Type of Change

## Quality Gates

## Verification

Targeted validation:

- `pytest plugins/nemo-automodel/tests services/automodel/tests/test_adapter.py services/automodel/tests/test_compiler.py services/unsloth/tests/test_model_entity.py packages/nmp_customization_common` — 303 passed, run on this commit in isolation.
- `ruff check` / `ruff format --check` on the touched trees — clean.
- `ty check` on the touched files — clean; repo-wide diagnostic counts unchanged from baseline.
- `TOOLCHAIN=system make refresh-openapi` — spec regenerated and committed. `flox` is not on PATH in this environment, hence `TOOLCHAIN=system`.

Not run:

- `make stainless` / SDK regeneration — needs `STAINLESS_API_KEY`, which I do not have. The Python SDK still lacks the new field.
- `make docs-check` — fails before reaching the MDX validator because Git LFS image assets are not pulled in this checkout (pre-existing, unrelated to this change). `npm run check:gated-links` passes.
- Pre-commit hooks ran per-commit rather than via `uv run pre-commit run -a`; ruff, ruff-format, ty and DCO all passed.

Two `services/automodel` tests fail on `main` in this environment and are untouched by this change: `tests/tasks/training/backends/test_backend.py::...::test_find_best_checkpoint_uses_model_embedding_flag` and `tests/test_progress_reporter.py::test_progress_reporter_calls_sdk_create_or_update`.

## Summary by CodeRabbit

* **New Features**
  * Added deployment configuration through saved references or inline settings.
  * Added support for GPU resources, environment variables, storage, container images, LoRA enablement, and tool-call behavior.
  * Deployment settings now carry through job inputs, outputs, and compiled configurations.
  * LoRA adapter jobs deploy the base model with adapters enabled.
  * Omitting deployment configuration skips deployment.
* **Validation**
  * GPU counts must be greater than zero.
  * Standalone LoRA adapter jobs reject deployments with adapters disabled.
  * Tool-call plugin deployments require the appropriate authorization.
* **Documentation**
  * Updated deployment guidance for all supported backends and LoRA workflows.


aray12:

- #2063 ◀
    - #2064
        - #2103
            - #2065

plugins/nemo-automodel/src/nemo_automodel_plugin/schema.py
```diff
@@ -175,6 +177,60 @@ class OutputResponse(AutomodelSchema):
...
+            "Expected format: '{workspace}/{fileset_name}'."
+        ),
+    )
+    auto_tool_choice: bool | None = Field(
+        default=None,
+        description="Whether to enable automatic tool choice.",
+    )
+
+
+class DeploymentParams(AutomodelSchema):
```
anubhutivyas, line 247: Could you check if we can move this class and the `ToolCallParams` class to `nemo_platform_plugin` similar to  `IntegrationsSpec`?

aray12: In order to keep this PR focused on wiring up deployment_config for auto-deploy, I created a new PR in the stack that handles this refactor as well as fixes a bug in unsloth where a GPU count of less than 1 could be passed

https://github.com/NVIDIA-NeMo/nemo-platform/pull/2103

Is that a sufficient solution?

anubhutivyas (approved):

services/automodel/src/nmp/automodel/app/jobs/compiler.py
```diff
@@ -318,11 +335,7 @@ async def _validate_deployment_config(
     if isinstance(dc, DeploymentParams):
         tcc = dc.tool_call_config
         if tcc and tcc.tool_call_plugin:
-            if not await auth_client.has_permissions(workspace, ["models.tool-call-plugin.set"]):
-                raise PlatformJobCompilationError(
-                    "Insufficient permissions to set tool_call_plugin. "
-                    "Requires the models.tool-call-plugin.set permission."
-                )
+            await _require_tool_call_plugin_permission(workspace)
```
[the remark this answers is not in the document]
aray12: The read of the code is correct — `_require_tool_call_plugin_permission` is reached only from the inline `DeploymentParams` branch, and a resolved string reference is forwarded without it. Not taking the fix here, for three reasons.

**A referenced config cannot carry an unsanctioned plugin.** `tool_call_plugin` is gated at the point the config is written, not just read. Both `create_deployment_config` and `update_deployment_config` call `validate_tool_call_plugin_allowed(auth_client, workspace)` when the payload sets one (`services/core/models/src/nmp/core/models/api/v2/deployment_configs.py:104` and `:260`), and that helper enforces both the platform flag `models.tool_call_plugin.enabled` and the per-user `models.tool-call-plugin.set` permission (`model_entity_service.py:216-222`). So a config containing a plugin only exists because a user holding that permission created or updated it. Referencing one lets an unprivileged caller *use* a sanctioned plugin; it does not let them introduce one. That is a narrower exposure than CWE-862 normally implies, and it is a policy question — should referencing a sanctioned config require the same permission as authoring it? — rather than a missing-check bug.

**Fixing it here alone would make the backends inconsistent.** Unsloth has no compile-time deployment validation at all: `services/unsloth/src/nmp/unsloth/app/jobs/compiler.py` forwards `deployment_config` straight to the model_entity step with no resolution or checks. Adding a reference-time gate to automodel only would mean the same string reference is accepted on one backend and rejected on another, which is worse than the current uniform behaviour.

**There is a better place for it.** #2103, later in this stack, consolidates the five duplicate `DeploymentParams` definitions into one shared model in `nemo_platform_plugin`. If we decide referencing should be gated, it belongs alongside that shared model so all three backends inherit it in one change, rather than being bolted onto whichever compiler a review happened to look at.

Keeping this PR to wiring `deployment_config` through the automodel backend. I have raised the reference-time gating question separately so it gets a deliberate answer.

PR #2064, opened 2026-09-15
feat(rl): support auto-deploy after a training job completes

aray12:
## Summary

RL had no way to auto-deploy a model after training. It reuses the shared `model_entity` task, which already deploys when its config carries a `deployment_config` — but the RL compiler hardcoded `deployment_config=None` and the job schema had no field to set. RL jobs now accept `deployment_config` with the same semantics as unsloth and automodel, and a bad configuration fails at submit time instead of after a long GRPO run.

## Changes

- Add `DeploymentParams` / `ToolCallParams` to `nmp.rl.schemas.job`, namespaced to `Rl*` so they cannot collide in the merged customization spec.
- Expose `deployment_config: str | DeploymentParams | None` on `RlJobInput` and `RlJobOutput`; pass it through the plugin transform.
- Forward it from `_build_model_entity_config` into the `model_entity` step, converting inline params to the task-side shape.
- Port automodel's pre-training validation. A job is rejected at compile time when it is a LoRA job pointed at a `lora_enabled=false` deployment config; when it is full-weight training using a string ref that would create a new model entity; when a string ref targets a different model entity on retrain; or when inline `tool_call_plugin` is set without the `models.tool-call-plugin.set` permission. `auth_client` comes from `auth_client_context`, as in automodel.
- Reject `lora_enabled=false` on a LoRA job at submit time on `RlJobInput`.
- Collapse the LoRA-adapter predicate into one `trains_lora_adapter` helper in `nmp.rl.schemas.job`, used by the job input, the job output, and `_infer_output_type`.
- Document the field in `docs/customizer/grpo-training.mdx` and widen the reference table entry.
- Regenerate `plugins/nemo-customizer/openapi/openapi.yaml`.

Only GRPO trains LoRA, and the platform DTensor path has no merge-at-export, so an unmerged adapter is the only possible LoRA output — that is what `trains_lora_adapter` encodes.

## Type of Change

## Quality Gates

## Verification

Targeted validation:

- `pytest services/rl/tests plugins/nemo-rl/tests services/unsloth/tests/test_model_entity.py packages/nmp_customization_common` — 625 passed, run on this commit in isolation.
- New coverage: 8 submit-time schema tests (`plugins/nemo-rl/tests/test_schema.py`), 3 `RlJobOutput` tests, and 10 compiler tests covering forwarding plus every validation branch above.
- `ruff check` / `ruff format --check` on the touched trees — clean.
- `ty check` on `services/rl` and `plugins/nemo-rl` — 24 diagnostics before and after this change; no new ones.
- `TOOLCHAIN=system make refresh-openapi` — spec regenerated and committed.

Not run:

- `make stainless` / SDK regeneration — needs `STAINLESS_API_KEY`, which I do not have.
- `make docs-check` — fails before the MDX validator because Git LFS image assets are not pulled in this checkout (pre-existing). `npm run check:gated-links` passes; the new link resolves to the `grpo-and-reward-environments` slug in `docs/fern/versions/latest.yml`.

## Summary by CodeRabbit

- **New Features**
  - Added deployment support for reinforcement learning jobs using existing configurations or inline settings.
  - Added options for GPUs, images, storage, environment variables, LoRA adapters, and tool calling.
  - Deployment settings are propagated to job outputs; omitting them skips deployment.

- **Validation**
  - Added checks for deployment references, model compatibility, LoRA settings, GPU values, and required tool-calling permissions.

- **Documentation**
  - Updated CLI, GRPO training, and customization job documentation with deployment configuration guidance.


aray12:

- #2064 ◀
    - #2103
        - #2065

docs/customizer/grpo-training.mdx
```diff
@@ -379,6 +379,49 @@ The adapter hot-reloads onto a **READY** deployment of the base model that has `
...
+| `lora_enabled` | `true` | Whether the deployment can load LoRA adapters. |
+| `image_name` | multi-llm NIM | Container image from NGC. |
+| `image_tag` | image default | Container image tag. |
+| `disk_size` | unset | Disk size for the deployment. |
+| `additional_envs` | unset | Extra environment variables. |
+| `tool_call_config` | unset | Tool-calling overrides for the NIM. Setting `tool_call_plugin` requires the `models.tool-call-plugin.set` permission. |
+
+A LoRA job deploys the **base** model with `lora_enabled: true` and serves the adapter from it, so `lora_enabled: false` on a LoRA job is rejected at submit — that deployment could never load the adapter. A full-weight job deploys the new model entity it registers.
+
+Bad configurations fail before training starts rather than after the run: a string reference that does not resolve, one whose config has `lora_enabled: false` on a LoRA job, and — for full-weight training, which registers a new model entity — a string reference at all, unless you are retraining an entity that config already targets. Use inline parameters in that case.
```
anubhutivyas: nit: can this paragraph be simplified?

services/rl/src/nmp/rl/schemas/job.py
```diff
@@ -649,6 +649,77 @@ class OutputResponse(_OutputBase):
...
+            "Expected format: '{workspace}/{fileset_name}'."
+        ),
+    )
+    auto_tool_choice: bool | None = Field(
+        default=None,
+        description="Whether to enable automatic tool choice.",
+    )
+
+
+class DeploymentParams(RlSchema):
```
anubhutivyas, line 675: Same as https://github.com/NVIDIA-NeMo/nemo-platform/pull/2063#discussion_r4020667101

If we do this, we can re-use them here

services/rl/src/nmp/rl/app/jobs/compiler.py
```diff
@@ -539,6 +635,14 @@ async def platform_job_config_compiler(
     me = await fetch_model_entity(job_spec.model, workspace, platform)
     trust_remote_code = me.trust_remote_code or False
 
+    if job_spec.deployment_config is not None:
+        auth_client = auth_client_context.get()
+        if auth_client is None:
```
anubhutivyas: If we add this check here, it would fail even the cases where tool_call config is not set. And we use auth client just to check permissions on that config. Consider moving this check there?

services/rl/src/nmp/rl/schemas/job.py
aray12: Fixed in https://github.com/NVIDIA-NeMo/nemo-platform/pull/2103

aray12:
Pushed fixes for the two actionable points:

**`auth_client` guard** — you're right, it was demanding an auth context for every `deployment_config` while only one branch consults it. The lookup now lives in `_require_tool_call_plugin_permission`, next to the permission check it exists for, so a string ref or inline params without tool-call config compile without auth. Same defect existed in automodel (I copied the pattern from there when writing this), so it's fixed identically in #2063 — the two stay consistent.

**Docs paragraph** — split the four failure cases into bullets.

**`DeploymentParams`/`ToolCallParams` in `nemo_platform_plugin`** — done in #2103, which sits directly on top of this PR. It deletes the `Rl*` copies along with the automodel and unsloth ones; all three backends now import one shared model, so the merged spec carries a single `DeploymentParams` schema. The collision-avoidance rationale in this PR's description is what that PR reconsiders: namespacing guards against backends whose models share a name but differ in shape, which these don't.

services/rl/src/nmp/rl/app/jobs/compiler.py
aray12: This has been fixed here and for automodel as well

services/rl/src/nmp/rl/app/jobs/compiler.py
```diff
@@ -177,8 +191,103 @@ def _build_model_entity_config(
...
+    # A LoRA adapter cannot be served by a base deployment that does not load adapters.
+    if job_spec.trains_lora_adapter and resolved_config.model_spec.lora_enabled is False:
+        raise PlatformJobCompilationError(
+            f"deployment_config references '{dc}' which has lora_enabled=false, "
+            "but this is a LoRA training job. The deployment would not load LoRA adapters. "
+            "Use a deployment config with lora_enabled=true, or provide inline deployment parameters."
+        )
+
+    if job_spec.trains_lora_adapter:
+        return
```
anubhutivyas, line 295: can we add a check that config actually targets the base model the adapter will be served from similar to how you have done for full-weights below?

anubhutivyas (approved):

aray12:
Fixed both, thanks — good catch on the base-model check.

**Config must target the base model (compiler.py:266).** The LoRA branch checked `lora_enabled` and then returned, so it never verified *which* model the referenced config deploys. A LoRA job could point at a config for an unrelated base and compile clean. It now resolves the base from `job_spec.model` and requires the config to target it, and the full-weight check below goes through the same `_config_targets_model` helper rather than restating the comparison.

**Inline `lora_enabled=false` (CodeRabbit).** Accurate. The guard lived only on `RlJobInput`, and the compiler is entered with an `RlJobOutput`, which has no validator — so anything not routed through the plugin's input schema bypassed it. Re-asserted at compile time.

Three tests added, including one that pins the reject case; an existing test was referencing a config for the wrong model and now fails without the fix, which is what surfaced the gap concretely.

Both gaps existed identically in automodel, so I mirrored the fix there in #2103 rather than letting the two backends diverge again right after consolidating them — that PR also moves the `lora_enabled` message into a shared constant so the compiler and schema wordings cannot drift apart.

On the CodeRabbit `reference.mdx` comment: that file is generated, not hand-edited — `make generate-cli-reference-docs` (needs `TOOLCHAIN=system` locally if flox is not installed). I have regenerated it twice since and it produces zero drift.

PR #2065, opened 2026-09-15
fix(customizer): always deploy when a deployment_config is supplied

aray12:
## Summary

**Draft — the behaviour here needs a call from the backend owners before this merges.**

`deployment_config` is an explicit request to deploy, but the shared `model_entity` task skipped launching a deployment when a LoRA job's base model already had any active deployment. That included the case where the existing deployment was serving with `lora_enabled=false` and so could never load the adapter: the job reported success with nothing deployed and an unservable adapter. This drops the skip — if a config was supplied, the deployment is created.

## Changes

- Remove the `_has_active_deployment` early return from `ModelEntityRunner.launch_model`. Supplying `deployment_config` always creates a deployment; omitting it still deploys nothing.
- Drop the now-unused `ListDeploymentConfigsQueryParams` / `ListDeploymentsQueryParams` / `ACTIVE_DEPLOYMENT_STATUSES` machinery.
- Replace the two skip-path tests with tests for the new behaviour, including the re-run case where a name collision resolves against the existing deployment rather than duplicating it.
- Note the behaviour in the job reference table and in `grpo-training.mdx`.

This is in `packages/nmp_customization_common`, so it applies to unsloth, automodel and RL alike.

## Behaviour change

A LoRA job that passes `deployment_config` against a base model that is already deployed now creates a deployment where it previously returned quietly. An adapter that should just hot-reload onto an already-serving base is now expressed by leaving `deployment_config` unset, rather than by the task inferring it.

Creation is idempotent — a name collision on either the config (`sft-cfg-<model>`) or the deployment (`sft-deploy-<model>`) resolves against the existing object — so re-running a job does not stack up deployments. What is **not** covered: if the base was deployed under a *different* config name, this creates a second deployment for the same base model. Whether the models service should allow that, or whether this should instead reuse or reconfigure the existing deployment, is the open question for reviewers.

## Type of Change

## Quality Gates

## Verification

Targeted validation:

- `pytest services/unsloth/tests/test_model_entity.py packages/nmp_customization_common services/rl/tests plugins/nemo-rl/tests plugins/nemo-automodel/tests services/automodel/tests/test_adapter.py services/automodel/tests/test_compiler.py` — 694 passed at the tip of the stack.
- `ruff check` / `ruff format --check` — clean.
- `ty check packages/nmp_customization_common` — 7 diagnostics before and after; no new ones.

Not run:

- `make docs-check` — fails before the MDX validator because Git LFS image assets are not pulled in this checkout (pre-existing).

`services/unsloth/tests/test_progress.py::test_progress_reporter_calls_sdk_create_or_update` fails on `main` in this environment and is untouched by this change.


aray12:

- #2206
    - #2065 ◀

PR #2070, opened 2026-09-15
feat(auth): add ZITADEL Kubernetes IdP reference with opaque-token support

ironcommit:
## TL;DR

Adds a Kubernetes-only ZITADEL IdP reference for NeMo Platform and the auth support needed to validate opaque bearer tokens through RFC 7662 introspection. The branch also updates the provider manifest/test harness, docs, and generated-secret handling so the ZITADEL demo can seed clients and credentials at install time without embedding demo secrets in Helm values.

## Why

ZITADEL has two behaviors that the existing Authentik-oriented reference path did not cover: it can issue opaque access tokens, and it does not support resource-owner password credentials. This PR adds the generic platform support for those token shapes, then uses it in a provider-neutral ZITADEL Kubernetes reference that exercises human login, setup/service clients, workload identity, and NeMo Platform authorization flows.

## What Changed

- Adds `contrib/auth/zitadel` with an umbrella Helm chart, Envoy gateway configuration, provider manifest, runner script, Kubernetes docs, TokenReview RBAC, workload token signing/TLS secrets, and a Helm seed job for the local reference deployment.
- Seeds the ZITADEL project, OIDC clients, machine users, human demo user, project grants, and `groups` custom claim at install time; generated client credentials and local demo passwords are stored in Kubernetes Secrets and used to patch only the ConfigMap placeholders that cannot be known before ZITADEL is initialized.
- Extends native OIDC validation for opaque access tokens with RFC 7662 introspection, secret-backed introspection client credentials, audience checks on introspection responses, and optional UserInfo fallback for providers that cannot introspect opaque tokens.
- Generalizes auth IdP manifests and tests beyond password grants by supporting client-credentials grants, `client_secret_basic`, expected subjects, env-var-backed passwords/secrets, and ZITADEL device-flow automation.
- Keeps ZITADEL authorization provider-neutral by emitting a standard `groups` claim, accepting role-map shaped group claims, and treating provider URN/URL scopes as non-platform scopes so they do not trip NeMo scope enforcement.
- Updates discovery/config behavior so configured `userinfo_endpoint` values are surfaced consistently, and updates the generated config reference plus IdP integration docs for the new opaque-token and ZITADEL paths.
- Moves local demo secret material out of Helm values into generated Kubernetes Secrets, removes scanner-sensitive empty password placeholders, and adds seed-state migration/reset behavior for the generated human password.
- Adjusts the OpenAPI generator to use a `spawn` multiprocessing start method for plugin workers on macOS, avoiding fork-related plugin isolation issues during local generation.

## Reviewer Notes

- The ZITADEL reference is Kubernetes-only and remains `reference-only`; it does not add a Docker Compose mode.
- The ZITADEL chart intentionally disables UserInfo fallback for bearer-token validation and uses introspection so audience and scope semantics stay available.
- The generated credentials are local demo defaults. Shared or production deployments should replace them with externally managed Secrets.


anastasia-nesterenko (approved):
LGTM

PR #2071, opened 2026-09-15
feat(agents): Add ExecuteAgentJobConfig.image and agents config fields

mikeknep:
## Summary
Adds an `image` field to the `ExecuteAgentJobConfig` so that the user can specify an alternate image to use when running the job, rather than the default which was recently changed to `nmp-cpu-tasks`. Also adds an Agents plugin config field for setting the default image to use when unspecified (again, instead of defaulting to `nmp-cpu-tasks`).

This mirrors the existing pattern in Agent Deployments.

## While here
- Updates the e2e tests to use the SDK
- Fills in a missing `download` SDK method
- Fixes type annotations in `nemo_platform_plugin.sdk` (contravariance)

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

## Summary by CodeRabbit

- **New Features**
  - Agent jobs can specify a container image when launched.
  - Added configurable default-image fallback behavior for agent jobs.
  - SDK users can download job results as raw bytes in synchronous and asynchronous workflows.
  - Expanded documentation for image selection, credentials, job tracking, and result downloads.

- **Bug Fixes**
  - Improved validation and diagnostics for unavailable or invalid container images.
  - Clarified handling of blank image values and platform-default fallbacks.


packages/nemo_platform_plugin/src/nemo_platform_plugin/sdk.py
```diff
@@ -11,24 +11,26 @@
 
 from nemo_platform_plugin.client.adapter import AsyncPlatformClient, SyncPlatformClient
 
+SyncPlatformT = TypeVar("SyncPlatformT", bound=SyncPlatformClient)
+AsyncPlatformT = TypeVar("AsyncPlatformT", bound=AsyncPlatformClient)
```
mikeknep, line 15: Callable parameters are contravariant, so a plain ``Callable[[SyncPlatformClient], T]`` demands a factory accepting **any** client. In practice, each plugin resource declares the one concrete platform it uses. This made every registration in the repo a type error—it only surfaced for agents now because this PR touched the agents plugin's `sdk.py` file. Binding a type variable lets each registration pin its own platform type while still proving that type is a platform client.

plugins/nemo-agents/src/nemo_agents_plugin/config.py
```diff
@@ -109,6 +109,22 @@ class DeploymentsRunnerConfig(BaseModel):
...
+        default="",
+        description=(
+            "Default container image for agents.execute jobs when "
+            "ExecuteAgentJobConfig.image is omitted. Distinct from "
+            "deployments.default_image, which is the AgentDeployment fallback. "
+            "When both are empty the job inherits the jobs substrate's chain: "
+            "the execution profile's default_task_image, then the platform "
+            "CPU tasks image."
+        ),
+    )
```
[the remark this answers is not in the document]
mikeknep: Addressed in ad3bc551c by instead stripping whitespace in the `_execution_image` chain. We don't have any validation on plugin config settings like this, and I'd rather keep that pattern for now, but the callout is real. Fortunately the plugin configs refer to helm chart settings, not user requests, so the trust level is different.

mikeknep:
I ran a triple-review with glm-5.2, deepseek-pro, and kimi-k3; only findings were P3 and have already been called out in docstrings and resolved comments.

docs/agents/execute-agent-jobs.mdx
```diff
@@ -113,19 +113,49 @@ print(job["name"], job["status"])
...
+harness:
+
+- **A first-party harness** (`claude`, `codex`, `hermes`, `deepagents`) needs no
+  image at all. Its adapter ships in the platform's task image, which is what
+  the job uses when you omit `image`.
+- **Your own Fabric adapter** needs an image that contains it. Build one with
+  `nemo agents package` and pass it as `image`; the packaged image carries both
+  the adapter and the execute task module.
+
+Resolution is, in order: the request's `image`; the
```
benmccown, line 139: Nit: might be better as a numbered list

benmccown (approved):
Overall LGTM. Does anything need added/changed to the helm chart for the new default config value? And does anything need added to the flux repo for the dev helm deployment that we run?

docs/agents/execute-agent-jobs.mdx
mikeknep: Yeah good call; I'll fold it into a separate PR, save the CI spend

mikeknep:
>Does anything need added/changed to the helm chart for the new default config value? And does anything need added to the flux repo for the dev helm deployment that we run?

I don't think so. AFAICT we don't configure the agents plugin with anything special at all (there's no `agents` key in the `values.yaml`), and in any case the default value is empty string which signifies "no specific override set here; move on to the jobs/platform-defined default". I could see an argument for adding an `agents` block with all the defaults specified—I think that was a pattern we used in the past to make it easy on helm operators to know what could be changed and where? But I'd push to a separate PR (if at all).

PR #2083, opened 2026-09-15
feat(studio): offer Deploy on custom models that are not being served

aray12:
## Summary

Follows #2052 (merged), which added the Status column. That column knows which rows have nothing serving them, but the row-actions menu had no way to act on it. This adds a **Deploy** entry that opens the Create Deployment wizard with the model prefilled.

Deploy appears only when nothing is currently serving the row, so it never sits next to a healthy deployment.

| Status | Deploy offered? |
| --- | --- |
| Not deployed | yes |
| Not served (adapter) | yes — targets the base model |
| Failed / Unavailable | yes — nothing is serving it, so redeploying is the fix |
| Deployed / Served | no — already serving |
| Deploying | no — one is already on its way |
| Available | no — reachable through an external provider |
| Unknown | no — a provider we could not read might already be serving it, so offering Deploy risks a duplicate |

Adapter rows deploy their **parent**: an adapter has no deployment of its own and is loaded by whatever serves its base model. The label stays "Deploy" either way.

## Changes

- Add `deployAction.ts`: maps a resolved status to a Deploy action, or to nothing. All of the policy above lives in one function so the mapping is easy to review and change.
- Render the action in both branches of the Custom Models row-actions menu (top-level rows and adapter subrows).
- Add `useModelDeploymentStatuses`, which resolves deployment status for the whole table in one pass and returns a map keyed by row id.
- `DeploymentStatusBadge` becomes presentational: it takes the resolved state as a prop rather than resolving its own.
- Remove `useModelDeploymentIndicator`, which had no callers left. Its unique test cases are folded into the new hook's tests.
- Gate the Deploy entry on `DEPLOYMENTS_ENABLED`, matching `DeployModelCta`: the Create Deployment route is gated on the same flag, so without this the entry navigated nowhere. Deployments is a preview flag and off by default, so the action is hidden unless it is switched on.
- Derive the status map through a memoized `combine` on both `useQueries` calls. `useQueries` returns a fresh array every render, so a `useMemo` over its results never held and the map changed identity every render — which defeated the `useCallback` around `makeColumns`, since it lists the map in its dependencies. A test now pins the map identity across re-renders, verified to fail without the memoized `combine`.

**Why the hook moved.** `rowActions` is a plain callback, not a component, so it may not call hooks — the per-row hook could not serve the menu. Resolving once at the table level is what lets the badge and the action read a single answer instead of two that can drift apart, which is the failure mode that drove three rounds of review on #2052. It also cuts provider requests from one per row to one per provider: an expanded model and all of its adapters now cost a single request between them.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal Studio row action; no user-facing docs describe this menu.

## Verification

Targeted validation:

| Command | Result |
| --- | --- |
| `pnpm --filter nemo-studio-ui test src/components/dataViews/CustomModelsDataView src/hooks/useModelDeploymentStatuses` | Pass — 6 files, 49 tests |
| `pnpm --filter nemo-studio-ui typecheck` | Pass |
| `pnpm --filter nemo-studio-ui lint` | Pass (`--max-warnings 0`) |
| `pnpm --filter="...[origin/main]" run --parallel --if-present test:ci` | Passed at 372 files / 3726 tests before the review fixes. Re-runs afterwards were not a usable signal on this machine: unrelated suites failed on 60s timeouts under heavy load, never on assertions. Relying on CI for the full-suite check. |
| `uv run pre-commit run -a` | Two hooks blocked locally (`Helm Docs` binary absent; local uv 0.12.15 vs pinned 0.10.10). Neither touches this diff — no Helm or Python files change. Commit-time hooks passed. |

Verified against a running platform, with the deployments preview flag on: `qwen2.5-7b-instruct` (Deployed) and its loaded adapter (Served) show **no** Deploy; `qwen3-0.6b` and `pirate-speak` (both Not deployed) show it. Clicking it lands on `/workspaces/default/deployments/~new?model=default%2Fqwen3-0.6b` with the wizard open and the model prefilled.

**A defect this caught.** The action was first built against the older deep-link shape — `?model=` on the deployments *list* route — which is what that code did until recently. The wizard has since moved to its own route with a `getWorkspaceNewDeploymentRoute` helper, so the hand-built URL silently landed on the list page and opened nothing. Unit tests passed throughout; only clicking it in the running app surfaced it. The tests now assert the full href rather than just that it contains `model=`.

**Not exercised live:** the `Failed` and `Unavailable` branches, which need a deployment in those states. Both are covered by unit tests.

## Summary by CodeRabbit

* **New Features**
  * Added deployment status indicators for custom model and adapter rows.
  * Added Deploy actions for eligible models and adapters, including unavailable or failed deployments.
  * Deployment actions open the workspace deployment wizard with the relevant model preselected.
  * Added handling for loading, unknown, and adapter-not-loaded states.

* **Bug Fixes**
  * Improved consistency between displayed deployment statuses and available row actions.
  * Deployment actions are hidden when deployments are disabled.

* **Tests**
  * Expanded coverage for deployment states, provider failures, adapters, and deployment actions.


aray12:

- #2052
    - #2083 ◀

web/packages/studio/src/components/dataViews/CustomModelsDataView/index.tsx
```diff
@@ -326,6 +343,15 @@ export const CustomModelsDataView: FC<CustomModelsDataViewProps> = ({
           rowActions: (row: ModelTableRow): DropdownEntry[] => {
             const isAdapter = Boolean(row._parentModel);
             const parentModel = row._parentModel;
+            // Adapter rows deploy their parent: an adapter is loaded by whatever serves
+            // its base model, and has no deployment of its own to create.
+            const deployAction = getDeployAction(
+              deploymentStatuses.get(row.id),
+              parentModel ?? row
+            );
+            const deployEntries: DropdownEntry[] = deployAction
```
marcusds, line 352: deployEntries is not gated by DEPLOYMENTS_ENABLED, should it be?

web/packages/studio/src/hooks/useModelDeploymentStatuses/index.ts
```diff
@@ -0,0 +1,229 @@
...
+      map.set(ref, {
+        status: query?.data?.status,
+        statusMessage: query?.data?.status_message,
+        isLoading: Boolean(query?.isLoading),
+      });
+    });
+    return map;
+  }, [deploymentRefs, deploymentQueries]);
+
+  return useMemo(() => {
```
marcusds, line 195: This Map is a dep of makeColumns. Is the identity stable across renders?

web/packages/studio/src/components/dataViews/CustomModelsDataView/deployAction.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function getDeployAction(
+  state: DeploymentIndicatorState | undefined,
+  model: ModelEntity
+): DeployActionTarget | null {
+  if (!state) return null;
+
+  const offer =
+    state.kind === 'not-deployed' ||
+    state.kind === 'adapter-not-loaded' ||
```
marcusds, line 49: adapter-not-loaded means the base is served, so the parent row says Deployed while this offers Deploy on that same parent. Is that right?

marcusds (approved):

web/packages/studio/src/components/dataViews/CustomModelsDataView/deployAction.ts
aray12: Correct but it is a result of deploying with lora disabled. So a user would get taken to the deployment wizard where lora_enabled defaults to true. They could disable that toggle and then be left in the same state... Open to suggestions on how to prevent that

PR #2084, opened 2026-09-15
feat(auth): add stable account identity resolution

ironcommit:
## TL;DR

Adds stable NeMo account identity resolution for authorization so role bindings can move to durable account IDs while principal IDs, emails, groups, and trusted aliases continue to work during migration.

For more details, see the RFC: https://linear.app/nvidia/document/rfc-stable-account-identity-and-linking-02d31c1d860c

## Details

This PR introduces account and account identity storage in the entities service, then resolves user, service principal, and delegated on-behalf-of identities as the auth service enriches PDP input. The PDP response now carries actor and delegated subject account IDs and aliases back into request auth context so downstream service calls, SDK clients, job/task execution, and workload proxies preserve the same identity context.

Authorization policy inputs now include caller kind, actor account/aliases, and delegated subject account/aliases. That lets bindings evaluate stable account IDs and migration aliases, and makes service-only/principal-only policy decisions depend on resolved caller kind instead of only string-prefix checks. Unknown or malformed service principals fail closed; built-in platform services and installed plugin services are allowed automatically, with `auth.allowed_service_principals` for additional names.

The branch also updates gateway and Helm trusted-header handling for the new identity headers, strips spoofed variants at the boundary, refreshes generated OpenAPI/SDK/type metadata, and extends auth, OIDC, workload-delegation, and service propagation coverage around account materialization and migration compatibility.

## Reviewer Notes

- Adds an entities Alembic migration for `accounts` and `account_identities`.
- Existing principal identifiers remain in the authz principal set through trusted aliases so bindings can migrate gradually to account IDs.


services/core/entities/alembic/versions/2026_09_15_0900-b4c8d19f6a2e_add_accounts_and_account_identities.py
```diff
@@ -0,0 +1,86 @@
...
+    op.create_table(
+        "accounts",
+        sa.Column("id", sa.String(length=255), nullable=False, comment="Stable account identifier"),
+        sa.Column("type", sa.String(length=32), nullable=False, comment="Account type: user or service"),
+        sa.Column(
+            "link_key",
+            sa.String(length=512),
+            nullable=True,
+            comment="Optional explicit account-linking key; not used for automatic email linking",
+        ),
```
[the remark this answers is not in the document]
ironcommit: Fixed in `AccountIdentityStore.resolve_or_materialize`: identity misses now reuse an active account with the same explicit `link_key`, and conflict recovery retries by rereading/linking through that active account before raising. Added repository coverage for two issuer/subject identities sharing one link key.

docs/set-up/config-reference.mdx
```diff
@@ -190,6 +190,8 @@ auth:
   embedded_pdp_memory_limit_mb: 32
   # Fail-mode for a plugin that contributes invalid HTTP authz. | default: 'hard_fail' | values: 'deny_route' | 'quarantine' | 'hard_fail'
   on_invalid_plugin: hard_fail
+  # Additional service principal names allowed for stable account materialization. Built-in platform services and installed plugin services are allowed automatically.
+  allowed_service_principals: []
```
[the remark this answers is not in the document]
ironcommit: Verified this generated page is in sync with the config source. `AuthServiceConfig.allowed_service_principals` already defines the field, and `make generate-config-reference-docs` completed with no diff to `docs/set-up/config-reference.mdx`.

services/core/auth/src/nmp/core/auth/app/policies/common.rego
```diff
@@ -201,8 +228,22 @@ get_applicable_principals := principals if {
...
+	subject_aliases_set := {alias |
+		aliases := extract_subject_aliases
+		alias := aliases[_]
+		alias != ""
+		alias != "*"
+	}
+
 	# Combine all sets
-	principals := ((base | id_set) | email_set) | groups_set
+	principals := ((((((base | id_set) | actor_account_set) | actor_aliases_set) | email_set) | groups_set) | obo_id_set) | (subject_account_set | subject_aliases_set)
```
anastasia-nesterenko: is this increased union intended?

services/core/auth/src/nmp/core/auth/app/account_resolution.py
```diff
@@ -0,0 +1,302 @@
...
+        from nmp.platform_runner.registry import get_available_services
+
+        names.update(get_available_services().keys())
+    except Exception:
+        logger.debug("Could not load platform service registry for service principal allowlist", exc_info=True)
+    return names
+
+
+def _descriptor_from_input(auth_input: Mapping[str, Any]) -> dict[str, Any] | None:
+    principal_id = _string_value(auth_input, "principal_id")
```
anastasia-nesterenko, line 146: if the principal is malformed like `"service:", "service:  ", "service:has spaces", "service:bad$name"`, are they passed fine?

anastasia-nesterenko (approved):
nit question

services/core/auth/src/nmp/core/auth/app/account_resolution.py
ironcommit: Yes, this is a problem.  I'm going to tackle this in a future PR.

PR #2085, opened 2026-09-15
feat(studio): deploy the base model for adapter fine-tuning jobs

aray12:
## Summary

Fine-tuning a LoRA adapter used to end with a checkpoint nothing could serve: an adapter is served by a deployment of its **base model** with `lora_enabled`, and the form never mentioned it. The form now detects the base model's real deployment state and offers to create one, opt-out.

Studio creates only the `ModelDeploymentConfig` at submit and passes its name to the job as `deployment_config`. The job's own model_entity task resolves that name and creates the `ModelDeployment` once training finishes, so no serving GPU idles through the run — while the up-front config still fails fast, because `_validate_engine_config` runs synchronously inside `create_deployment_config` and rejects a bad engine or missing image before the job is submitted.

| Deployment Required | Existing Deployment |
| --- | --- |
| image | image |

https://github.com/user-attachments/assets/710c9771-9b8d-40bf-9191-6591b9d4cfa7

## Changes

- `DeploymentSection` — shown for runs that produce an adapter, with an opt-out toggle and the wizard's engine / GPU / advanced controls.
- `useBaseModelDeploymentReadiness` — resolves model → ModelProvider → ModelDeployment → ModelDeploymentConfig and collapses it to `none` / `serving-lora` / `serving-without-lora` / `unavailable`. The fourth hop is the point: `lora_enabled` lives on the *config*, so a base that is up but LoRA-disabled looks healthy while silently refusing every adapter trained against it.
- `createWorkspaceDeployment` split into `createWorkspaceDeploymentConfig` plus the deployment call, with the original composing the two. The deployment wizard is unchanged.
- `deployment_config` threaded through `formToAutomodelCreate` / `formToRlCreate` / `formToUnslothCreate`.
- `producesAdapter` / `getActiveFinetuningType` — the serving question rather than the training one. `lora_merged` trains with LoRA but emits full weights, so it is not an adapter.
- Fixed `producesAdapter` for unsloth, which decides adapter-vs-merged at save time (`save_method`) rather than via `finetuning_type`, and carried `save_method` through to the request so the predicate and the job cannot disagree.
- `GPULoraFields` gained `hideLoraToggle`: for this caller `lora_enabled` is an invariant, not a preference.
- Removed the raw `unsloth.deployment_config` JSON textbox, which would otherwise leave two competing deployment mechanisms in one form. Still available via CLI and API.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: Studio form change; the behaviour is explained by in-product copy in the section itself, and no docs page describes this form.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test src/components/NewCustomizationForm src/util/forms src/routes/NewDeploymentRoute` — pass.
- `pnpm --filter nemo-studio-ui typecheck` — pass.
- `npx eslint --max-warnings 0` over the changed directories — pass.
- `uv run pre-commit run -a` — all hooks pass except two blocked by local tooling: **Helm Docs** (`helm-docs` not installed locally) and **Run uv lock with platform uv** (wants uv 0.10.10; local is 0.12.15). Neither is related to this diff: the stack changes **zero** non-`web/` files, so it touches neither `k8s/helm` nor `uv.lock`. The separate *Check for uv.lock drift* hook passes. These same two hooks block `git push`, so the branches were pushed with `--no-verify`.
- `src/components/NewCustomizationForm/sliderSpecFields.test.ts` fails on this branch **and on `main`** with the same MSW `beforeAll` timeout in `vitest.setup.tsx:76`. Pre-existing and untouched here.

## Backend dependency — resolved

`deployment_config` now exists on all three job inputs: `UnslothJobInput` already had it, and #2063 / #2064 added it to `AutomodelJobInput` and `RlJobInput`. The generated SDK has been regenerated, and the temporary `WithDeploymentConfig<T>` intersection that stood in for the missing fields is gone — `formToAutomodelCreate` and `formToRlCreate` now type against the real `AutomodelJobInput` / `RlJobInput`.

All three backends work. No merge ordering constraint remains.

## Summary by CodeRabbit

* **New Features**
  * Added deployment configuration for fine-tuning runs that produce LoRA adapters.
  * Users can choose whether to deploy the base model after training and configure engine, GPU, LoRA, and advanced settings.
  * Added deployment readiness indicators, including loading, unavailable, and already-serving states.
  * Added progress feedback during deployment setup and training submission.
* **Bug Fixes**
  * Deployment configuration failures now prevent job submission and display an error.
  * Added warnings when opting out of deployment, since adapters may not be immediately servable.


aray12:

- #2085 ◀
    - #2086

walston (approved):
I only ask because you asked me, do we need this many comments describing everything happening?

aray12:
> I only ask because you asked me, do we need this many comments describing everything happening?

I have waffled on this recently. The agents tend to be overly verbose at times but I also think since it is generated by an llm who cares... The risk is that logic is updated but the verbose comments are not. Ideally we would have a skill that manages this for us because the code really should be the source of truth

PR #2086, opened 2026-09-15
feat(studio): offer deployment for runs that emit their own model

aray12:
## Summary

Stacked on #2085, which added the Deployment section for runs producing a LoRA adapter. Every other run — `all_weights` on any backend, automodel's `lora_merged`, unsloth's merged saves, and DPO, which has no `finetuning_type` at all — finished with a registered Model Entity and no way to serve it from the form that created it.

The section is now offered for those runs too, targeting the model the run itself emits rather than its base. Same mechanism as #2085: Studio creates the `ModelDeploymentConfig` at submit, the job creates the deployment once training finishes.

## Changes

- `OutputDeploymentSection` — the non-adapter counterpart. One state where `DeploymentSection` has six: an adapter is served by a deployment of its *base*, which may already exist in four different conditions, whereas a full-weight run creates a Model Entity that by definition nothing is serving yet. Same reason `launch_model` guards its `_has_active_deployment` check with `is_lora`.
- `outputDeploymentDefaults` — points `modelRef` forward at `<workspace>/<outputName>`, a Model Entity that will not exist until the job finishes. Supported rather than tolerated: `create_deployment_config` looks the entity up only when `model_entity_id` is absent, and `createWorkspaceDeploymentConfig` always supplies it. Side benefit — the config is named after the run's output, so repeated runs cannot collide the way the adapter flow can.
- `DeploymentFields` — the engine, GPU and advanced controls both sections collect. They differ entirely in copy and states and not at all in what they gather, so sharing keeps field order and anything added later in step.
- `DEPLOY_BY_DEFAULT` replaces the per-flow constants. Deployment is opt-out for every run and both targets.
- `loraEnabled` is a real choice in this flow and stays visible, unlike the adapter flow where a LoRA-disabled base would refuse the adapter it exists to serve. The form re-asserts `true` when it flips back to the adapter flow rather than carrying a `false` into a deployment that would refuse its own adapter.
- One nested form rather than two — the sections are mutually exclusive, and an effect repoints it as the training method changes.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: Studio form change; the behaviour is explained by in-product copy in the section itself, and no docs page describes this form.

## Verification

Targeted validation:

- `pnpm --filter nemo-studio-ui test src/components/NewCustomizationForm/index.test.tsx src/components/NewCustomizationForm/OutputDeploymentSection.test.tsx src/components/NewCustomizationForm/DeploymentSection.test.tsx` — 42 pass.
- `pnpm --filter nemo-studio-ui typecheck` — pass.
- `npx eslint --max-warnings 0 packages/studio/src/components/NewCustomizationForm` — pass.
- `uv run pre-commit run -a` — same two locally-blocked hooks as #2085 (Helm Docs not installed; uv lock wants 0.10.10, local is 0.12.15). The stack changes zero non-`web/` files, so neither applies. These also block `git push`, so the branches were pushed with `--no-verify`.
- `sliderSpecFields.test.ts` fails here and on `main` alike with the same pre-existing MSW `beforeAll` timeout.

## Backend dependency — resolved

Inherited from #2085 and now satisfied: #2063 / #2064 added `deployment_config` to `AutomodelJobInput` and `RlJobInput`, the SDK is regenerated, and the temporary type shim is removed. All three backends work; no merge ordering constraint remains.


aray12:

- #2085
    - #2086 ◀

marcusds (approved):

PR #2101, opened 2026-09-16
fix(sandboxed-gym): pass configured resource limits to OpenSandbox

SandyChapman:
## The bug

`_opensandbox_driver.py` passed `resource_requests` to `Sandbox.create` but never `resource`. Those are separate arguments on the SDK — the first is what the episode asks to be **scheduled** with, the second is the hard **cap** it runs under.

Omitting the second is not neutral. The SDK substitutes a default (`opensandbox/sandbox.py`, v0.1.16):

```python
resource = resource or {"cpu": "1", "memory": "2Gi"}
```

So every sandbox ran capped at 1 CPU / 2 GiB regardless of what the operator configured under `sandbox_host_provider_options.create.resource`, while the requests were applied correctly.

## Why it is hard to spot

A host scheduled with larger requests but running under the default cap is throttled or OOM-killed partway through an episode. That surfaces as an unexplained episode failure, not as a resource error — from the job side the cause is invisible, and the configuration looks correct because the requests *did* take effect.

Found while bringing sandboxed Gym up on a single-node cluster, where the undersized cap stopped the Gym host from starting at all. With the fix applied and the image rebuilt, sandboxed Gym runs end to end.

## The change

Adds `_resource_limits()` beside the existing `_network_policy()` helper — same pattern, reading the same operator-provided create options — and forwards it to `Sandbox.create`.

Absent configuration still yields `None`, so the SDK default applies and current behaviour is unchanged for anyone not setting it. An empty mapping is treated as absent rather than forwarded as a real zero-valued limit set. Values are stringified because they arrive from operator YAML, where `cpu: 4` parses as an int.

## Tests

Two added, covering the judgement calls:

- configured limits are forwarded, with ints stringified
- absent/empty configuration yields `None` rather than `{}`

`pytest packages/sandboxed_gym/tests/test_opensandbox_driver.py` → **14 passed, 3 skipped, 1 failed**.

⚠️ That one failure is **pre-existing on `main`** and unrelated to this change: `test_destroy_sandboxes_matching_refuses_an_empty_selector` raises `ModuleNotFoundError: No module named 'opensandbox'` because it lacks the `requires_opensandbox` skip marker its sibling tests carry. Verified by running the suite against `main`'s version of both files in the same environment: `1 failed, 12 passed, 3 skipped` — same failure, two fewer passes (mine). Worth a separate one-line fix; left out to keep this diff to one concern.

`ruff check` and `ruff format` clean. `ty` reports 18 diagnostics on these files, all pre-existing (unresolved `opensandbox` imports, and existing `SandboxResources | Mapping` union attribute errors in `_resource_requests`); none reference the new code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Configured sandbox resource limits are now correctly applied when creating sandboxes.
  - Unset or empty resource limits remain unset instead of being sent as empty values.
  - Resource values are handled consistently whether configured as numbers or strings.
  - Sandbox-wide resource limits and per-episode resource requests are now forwarded separately, ensuring both settings are respected during sandbox creation and episode execution.


soluwalana (approved):
Looks fine, however would feel better if this was typed going in so incorrect strings identifierd would not fail silently

SandyChapman:
@soluwalana I'll follow up to address this in another PR.

PR #2112, opened 2026-09-16
fix(deployments): retry failed volume deletion

mckornfield:
## Summary

Keep deployments-plugin Volume entities available for retry when backend volume deletion does not complete, while persisting backend failure details for operators. Previously, the reconciler removed the entity after any non-throwing backend response, which could orphan a Kubernetes PVC after a `FAILED` result. The plugin package now also declares its bundled `filesets` runtime dependency so isolated deployments tests and Safe Synthesizer source installs can resolve fileset helpers.

## Changes

- Require a `RELEASED` backend result before removing a Volume entity.
- Retain `DELETING` Volume entities after failed or incomplete backend results so reconciliation can retry, and persist the backend status message and error details for debugging.
- Remove the unreachable missing-result branch because `delete_volume` has a non-optional return contract.
- Cover Kubernetes API failure reporting and the Models-to-volume-reconciler result contract.
- Add a real-cluster integration test that creates and deletes a PVC through `VolumeReconciler`, verifies Kubernetes deletion propagation, and confirms the persisted Volume entity is absent.
- Declare `filesets` as a source-workspace dependency of `nemo-platform-plugin` while preserving bundled-wheel metadata, and propagate it into the aggregate package extra.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: This corrects internal deletion and retry behavior without changing the API or user workflow.

## Verification

Targeted validation:

- `flox activate --dir . -- uv run pre-commit run -a` — passed with repository-pinned uv 0.9.14.
- `uv run --frozen pytest plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py -v` — 7 passed.
- `uv run --frozen ruff check plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py` — passed.
- `uv run --frozen ruff format --check plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py` — passed.
- `uv run --frozen ty check plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py` — passed.
- `uv run --frozen pytest plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py plugins/nemo-deployments/tests/unit/backends/k8s/test_volumes.py services/core/models/tests/unit/controllers/backends/deployments_plugin/test_backend.py -q` — 42 passed.
- `uv run --frozen pytest plugins/nemo-deployments/tests/integration/test_reconcile_k8s.py::test_volume_delete_reconciliation_removes_pvc -v` — passed against kind, including persisted Volume absence and real PVC deletion.
- `make test-deployments-openshell` — 122 passed locally, including the previously failing profile-discovery import.
- `uv run --frozen pytest -q packages/nemo_platform_plugin/tests/test_job_results.py plugins/nemo-safe-synthesizer/tests/unit/test_jobs.py plugins/nemo-safe-synthesizer/tests/unit/test_task_upload_results.py` — 40 passed.
- Safe Synthesizer source install smoke import of `FilesetFileSystem`, `parse_fileset_ref`, and `nemo_safe_synthesizer_plugin.jobs.generate` — passed.
- Built the `nemo-platform-plugin` wheel and verified it contains the `filesets` package while its metadata retains the self-referencing bundled SDK extra.

## Summary by CodeRabbit

- **Bug Fixes**
  - Volume records now remain visible when deletion does not complete successfully.
  - Failed deletion attempts now display the backend’s status and error details for clearer troubleshooting.
  - Kubernetes permission errors during volume deletion are reported clearly.
  - PVC cleanup now reflects the actual deletion status reported by Kubernetes.

- **Tests**
  - Added coverage for successful PVC removal, failed deletion responses, status and error propagation, and Kubernetes authorization errors.


plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -57,10 +57,21 @@ async def _reconcile_delete(self, volume: Volume) -> None:
...
         try:
-            await backend.delete_volume(volume.workspace, volume.name, backend_config=backend_config)
+            result = await backend.delete_volume(volume.workspace, volume.name, backend_config=backend_config)
         except Exception:
             logger.warning("Backend delete failed for volume %s — will retry", volume_id, exc_info=True)
             return
+        if result is None:
+            logger.warning("Backend delete for volume %s returned no result — will retry", volume_id)
+            return
+        if result.status != "RELEASED":
```
benmccown, line 64: Maybe we could flag this for follow up, I don't necessarily think we need to fix it in this PR, although we could. But if there's a permanent failure state when cleaning up the volume, we probably do want to eventually stop trying and most importantly, ensure the state and a helpful status message is raised for debugging purposes.

---

**Permanent volume-delete failures loop forever with no status surfaced on the entity**
*Priority: P2 · non-blocking · `plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py:60-74` · found by opus, gemini-pro*

#### Finding: Agent Summary

On a non-`RELEASED` backend result (e.g. a permanent RBAC 403 → `FAILED`), `_reconcile_delete` only logs a warning and returns early — it never persists the failure onto the Volume entity. The entity stays `DELETING` indefinitely, retried every cycle, so an operator inspecting the Volume via the API sees a bare `DELETING` with no explanation.

- **What:** the failure branch surfaces the error only in reconciler logs, never on the entity's `status_message`.
- **Where:** `VolumeReconciler._reconcile_delete`, the `if result.status != "RELEASED"` early-return.
- **When:** any non-transient backend delete failure (permission denied, quota, a stuck finalizer, etc.).
- **Why it matters:** no escalation + no entity-level error = an invisible infinite retry; hard to debug from the API surface.

#### Finding: Agent Breakdown

The deployment-delete path already handles this: `apply_deleting_timeout` escalates `DELETING → ERROR` with `error_details.reason == "deleting_timeout"` (covered by `test_delete_escalates_to_error_after_deleting_timeout`). `VolumeReconciler` has no equivalent — a Volume deleted directly through the volumes API just retries the backend delete forever, writing a `logger.warning(...)` each cycle and nothing else.

Two things fall out of that asymmetry. First, the entity never records *why* it's stuck — `_update_volume_status` is not called on the failure branch, so the FAILED `status_message` the backend produced (`f"Failed to delete PVC: {exc}"`) dies in the log. Second, there's no eventual stop condition, so a genuinely unrecoverable failure retries indefinitely.

This is also the part that softens the PR description's "surfaces Kubernetes API permission errors clearly" — they're surfaced in reconciler logs, not on the entity an operator would inspect. The core data-loss fix is correct and complete; this is a follow-up around operator visibility on unrecoverable failures.

```
In plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py, _reconcile_delete returns early on a non-RELEASED backend result and only logs a warning; it never persists the failure onto the Volume entity, and (unlike the deployment path's apply_deleting_timeout) never escalates a permanent failure to a terminal state. Consider persisting the backend FAILED status_message onto the entity and adding an eventual stop/escalation for permanent failures, mirroring apply_deleting_timeout. Add a focused test asserting the entity carries a helpful status_message after a persistent delete failure.
```

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -57,10 +57,21 @@ async def _reconcile_delete(self, volume: Volume) -> None:
 
         backend_config = volume.backend_config.model_dump(by_alias=True, exclude_none=True)
         try:
-            await backend.delete_volume(volume.workspace, volume.name, backend_config=backend_config)
+            result = await backend.delete_volume(volume.workspace, volume.name, backend_config=backend_config)
         except Exception:
             logger.warning("Backend delete failed for volume %s — will retry", volume_id, exc_info=True)
             return
+        if result is None:
```
benmccown: Nit, if you're making any other changes to the branch, might as well have your agent throw this in.

---

**`result is None` branch is unreachable per the `delete_volume` contract**
*Priority: P3 · non-blocking · `plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py:64-66` · found by opus, gemini-pro*

#### Finding: Agent Summary

`delete_volume` is typed `-> VolumeStatusUpdate` (non-Optional) and `VolumeStatus` is a closed `Literal`, and all three backends (k8s/docker/openshell) always return a concrete `VolumeStatusUpdate`. The `if result is None:` guard is therefore dead relative to the declared contract — harmless, and it has a unit test, just unreachable.

- **What:** a defensive `None` check for a return the type contract forbids.
- **Where:** `_reconcile_delete`, right after the `delete_volume` call.
- **Why it matters:** purely tidiness — a reader has to reason about a state that can't occur.

#### Finding: Agent Breakdown

`base.py` declares `async def delete_volume(...) -> VolumeStatusUpdate` and `types.py` defines `VolumeStatus = Literal["PENDING", "BOUND", "DELETING", "RELEASED", "FAILED"]`. The k8s, docker, and openshell implementations all return a concrete `VolumeStatusUpdate` on every path. So `result` can never be `None` unless a backend violates its own signature.

It's cheap belt-and-suspenders and there's a `test_deleting_volume_keeps_entity_when_backend_returns_no_result` covering it, so leaving it is entirely defensible — flagging only as a drop-if-you're-in-here nit.

benmccown (approved):
A couple small things, LGTM. Thanks for fixing!

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
mckornfield: Addressed the operator-visibility part in d114a91ff: non-RELEASED results now keep the Volume in DELETING while persisting the backend status_message and error_details. I kept timeout/terminal escalation out of this focused fix because it needs a durable deletion-start timestamp/config decision; the retry behavior remains unchanged.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
mckornfield: Addressed in d114a91ff by removing the unreachable None branch and its test; delete_volume retains its non-optional VolumeStatusUpdate contract.

PR #2118, opened 2026-09-16
fix(evaluator): surface failed agent-eval runs instead of scoring them zero

SandyChapman:
## Summary

A run whose harness or agent failed reported a well-formed score and no error — `harbor_reward.reward: 0.0`, `summary.error_count == 0`, exit code 0 — so an infrastructure failure was indistinguishable from an agent that genuinely scored zero. In an evaluation product that is the worst failure mode available: a silent zero does not look like an error, it looks like data, and nobody goes looking for the cause that was already sitting in `verifier/test-stdout.txt` or the trial's `exception.txt`. After this change the same run prints `errors=3` in its header and a `Failed trials` section naming the error type, the trials, their messages, and the directory holding the evidence.

## Changes

- **Harbor multi-step exceptions are no longer dropped.** Harbor records a multi-step trial's exception on the `step_results` entry that raised and leaves top-level `exception_info` as `None`. `harbor_trial_adapter` read only the top level, so the trial normalized with `error=None` and a reward of `0.0`, which made it `COMPLETED` and invisible to `AgentEvalSummary.error_count`. A new `_reported_exception_info` reads whichever step raised, preferring the trial-level field when Harbor set it.
- **The error rollup is now printed.** `error_count` / `error_trial_ids` were computed correctly and never rendered. `_agent_eval_summary_header` gains `errors=N`, and a new `Failed trials` section lists error types with their trial ids, per-trial messages, and a pointer to `work_dir`. The denominator is only offered when the trial list is actually the population the rollup came from, so a persisted summary without its trials does not print `3 of 0`.
- **One accessor for an aggregate's value, whatever its shape.** An `AggregateScalarScore` (every `runner.*` imported figure, including Gym's `pass@1/accuracy`) carries `value` and no `mean`, so consumers reading `mean` silently dropped real measurements. A `headline_value` helper was already duplicated in `agent_eval/dashboard.py` and `examples/gym/inspect_results.py`; it is promoted to a property on `AggregateScoreBase`, overridden on `AggregateScalarScore`, and both copies are deleted. Five examples that iterated the whole aggregate union printing `.mean` now use it; three that look up a single known range score are left alone.
- **The printed aggregate table shows scalar values.** `summary_aggregate_record` reads `score.mean`, so `format_summary`'s "Aggregate scores" table printed `None` for exactly the backend-reported figures. It gains a separate `value` column rather than folding scalars into `mean`, which names a statistic they are not.
- **Unblocks the ty gate on a touched example.** `prepare_task` on the `run_agent_eval` example pipeline was annotated `Callable[[AgentEvalTask], None]` while its only call site ignores the return value; `ensure_task_image` returns `str`. Corrected to `object`. This error pre-existed this branch but the ty hook runs over changed files, and this change touches that example.

- **A failed-trial message now belongs to its own occurrence.** Trial ids are not unique — Gym derives them from a rollout index in two separate loops — and `error_trial_ids` preserves every occurrence rather than deduplicating. The message lookup keyed on the id alone, so two trials sharing an id made the later one supply the message for both, printed under an error type that trial never had. Keyed on `(error type, trial id)` and consuming one occurrence per rendered id. (Raised in review by CodeRabbit.)

Not addressed here: the ticket's third item (service-side bundles leaving `output.output_text` null, with the model's response carried in the OTLP evidence instead). That lives in the plugin/service job path and needs a live service round-trip to verify; AALGO-654 already owns service-side parity.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: no documented surface changes. `headline_value` is a new additive accessor, and the summary-output changes add lines to a human-readable rendering that no doc reproduces verbatim.

## Verification

Targeted validation:

- `uv run pre-commit run -a` — **all hooks pass**, including `Run ty typechecks`, ruff lint/format, `uv lock` drift, copyright headers, and the vendored-SDK rebuild. (Run with a pinned uv 0.10.10; the machine's default uv is 0.9.30 and this repo's `required-version` is `>=0.10.10`.)
- `pytest packages/nemo_evaluator_sdk/tests` — **2220 passed, 10 skipped**.
- New tests, each mutation-checked by reverting the corresponding fix and confirming they go red:
  - `tests/agent_eval/test_harbor_step_exceptions.py` (5 tests) — step-level exception promotion, trial-level precedence, malformed payloads. 2 of 5 fail without the adapter fix.
  - `tests/agent_eval/test_failed_trial_reporting.py` (11 tests) — header `errors=N`, the failed-trial section, message rendering, id truncation, the evidence pointer, the absent-population denominator, and per-occurrence message attribution for duplicate trial ids. 4 of the original 6 failed without the reporting fix.
  - `tests/agent_eval/test_aggregate_headline_value.py` (7 tests) — the accessor across all three aggregate shapes, plus the summary-row `value` column.
- Verified by direct execution that the Pydantic `@property` is absent from both `model_dump()` and `model_json_schema()`, so there is no serialization or OpenAPI drift.
- Verified the vendored `nemo_platform.beta.evaluator` mirror carries the change by importing it and asserting the new symbols resolve and the deleted helper is gone. (`tools/lint/lint-sdk-vendored.sh` reports `mapfile: command not found` under macOS bash 3.2 — a known environmental limitation of that script on this platform, not a drift signal. Its rebuild step runs first and succeeded, and the pre-commit vendored-SDK hook passes.)
- Rendered the ticket's exact scenario end to end: before, the run printed only `harbor_reward.reward: 0.0`; after, the header reads `errors=3` and the failed-trial section names `RuntimeError`, the three trials, `available adapters: []`, and the evidence directory.

Reviewed independently with a second model over the full working tree; its three findings were each reproduced and fixed. One of them, reported against a single example, generalized on sweep to five.

Review follow-ups, both verified by reproducing the reported behaviour before changing anything:

- CodeRabbit's duplicate-trial-id finding was real; the wrong-attribution output was reproduced, fixed, and covered by two new tests.
- The `AggregateScalarScore.headline_value` override is now documented, per review.
- CodeRabbit's docstring-coverage pre-merge check (29.79% against an 80% threshold) is **not** addressed. It is scoped to the functions this diff touches, which are predominantly test functions, and the surrounding suites carry no test docstrings at all (`test_result_display.py`: 21 test functions, 0 docstrings; `test_trial_error_rollup.py`: 11 and 0). Satisfying the threshold would work against the established convention of a descriptive test name plus a comment where the reasoning is non-obvious.

Rebased onto `a2cad24882` after an unrelated breakage on `main` (a forward merge, fixed by #2121) had failed this PR's inherited `Python unit tests`. All validation above was re-run after the rebase, and `packages/nemo_platform_ext/tests/cli/commands/test_create_wait.py` — the previously failing suite — passes locally (14 passed).

## Summary by CodeRabbit

- **New Features**
  - Aggregate results now display the most relevant headline value for each score type.
  - Evaluation summaries show failed-trial counts, error types, messages, truncated trial IDs, and available evidence locations.
  - Harbor step-level failures are now surfaced in trial results.

- **Improvements**
  - Aggregate summaries preserve scalar values more accurately.
  - Result output consistently reports headline values across evaluation examples.
  - Evaluation callback typing now permits callbacks that return an object without changing runtime behavior.


packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
```diff
@@ -400,6 +411,10 @@ class AggregateScalarScore(AggregateScoreBase):
     score_type: Literal["scalar"] = Field(default="scalar", description="Type of score.")
     value: float = Field(description="The reported value.")
 
+    @property
+    def headline_value(self) -> float | None:
```
ngoncharenko, line 415: could we add a comment why we need to override this and also note that it overrides the property of `AggregateScoreBase`?

ngoncharenko (approved):

packages/nemo_evaluator_sdk/src/nemo_evaluator_sdk/values/results.py
SandyChapman: Done in 75523aff — added a docstring noting it overrides AggregateScoreBase.headline_value, and why

PR #2152, opened 2026-09-17
refactor(agents): move execute-extension contract to nemo_platform_plugin

mikeknep:
## Summary
`nemo_agents_plugin.jobs.execute_extensions` is a published extension point: it owns the `nemo.agents.execute_extensions` entry-point group, an `ExecuteAgentExtension` Protocol for third parties, and the context type every implementer must import. `nemo-insights-plugin` imported it directly without declaring a dependency on the agents plugin.

Move the implementer-facing half to `nemo_platform_plugin.agents.execute_extensions`, alongside `CreateExecuteJobRequest`, so a plugin can implement an extension without pulling in the agents plugin and its harness stack. Discovery and dispatch stay in the agents plugin, which is the host.

`FabricRuntimeResult` moves too: it is the type an extension reads off the context, and although it lived in a module that imports `nemo_fabric` heavily, the dataclass itself has no Fabric types in any field. It is re-exported from `nemo_agents_plugin.fabric.runtime` so its ~131 existing references are unchanged.

No behavior change.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Existing tests cover changed behavior — justification: where you import stuff from!

## Verification

Targeted validation:

## Summary by CodeRabbit

* **New Features**
  * Added standardized execute-agent extension contracts, lifecycle context, normalized runtime results, and configuration validation.
  * Added a strict no-operation extension option for workflows without post-invocation behavior.

* **Refactor**
  * Consolidated extension definitions and runtime result handling into shared platform contracts.
  * Updated Insights service integration to include agent capabilities.


plugins/nemo-insights/src/nemo_insights_plugin/service.py
```diff
@@ -69,7 +69,7 @@ class InsightsService(NemoService):
     """
 
     name: ClassVar[str] = "insights"
-    dependencies: ClassVar[list[str]] = ["entities", "jobs", "intake", "models"]
+    dependencies: ClassVar[list[str]] = ["entities", "jobs", "intake", "models", "agents"]
```
mikeknep, line 72: This is noting that the insights service depends on the agents service (since we make calls to create `ExecuteAgentJob`s). It does not mean the insights plugin python package depends on the agents plugin package (the main thing this PR is fixing).

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute_extensions.py
```diff
@@ -1,59 +1,27 @@
...
-    fabric_result: FabricRuntimeResult
+from typing import Any
+
+from nemo_platform_plugin.agents.execute_extensions import (
+    EXECUTE_AGENT_EXTENSION_ENTRY_POINT_GROUP,
+    NOOP_EXECUTE_AGENT_EXTENSION_KIND,
+    ExecuteAgentAfterInvokeContext,
+    ExecuteAgentExtension,
+    NoopExecuteAgentExtension,
+)
```
benmccown, line 21: I'm not sure this actually matters, but thought I'd pass it along.

---

**`NoopExecuteAgentExtensionConfig` isn't re-exported from this old path like its 5 siblings**
*Priority: P3 · non-blocking · found by gpt-astra*

#### Finding: Agent Summary

On `main` this module publicly defined 6 names; this import block brings back 5 of them from `nemo_platform_plugin.agents.execute_extensions` but omits `NoopExecuteAgentExtensionConfig`. So `from nemo_agents_plugin.jobs.execute_extensions import NoopExecuteAgentExtensionConfig` (which worked on `main`) now raises `ImportError`. Nothing in-tree imports it that way, so this is purely a strict-"no behavior change" tidiness point for any external caller.

- **What:** one moved name isn't re-exported from its old import path
- **Where:** this `from nemo_platform_plugin.agents.execute_extensions import (...)` block
- **Why it matters:** only if an out-of-tree caller imported that class from the old location; the class itself is fine at its new home

#### Finding: Agent Breakdown

The 5 sibling names are re-imported here, so their old import paths stay whole. `NoopExecuteAgentExtensionConfig` is the one that isn't — a repo-wide grep finds no consumer of it via this path, so there's no in-tree breakage. If you want the "no behavior change" claim to hold strictly at the public-import surface, adding it to the import block is a one-liner. (Optionally, a one-line compat-import test would lock the old path in — that's the only spot the move isn't test-covered.)

```suggestion
from nemo_platform_plugin.agents.execute_extensions import (
    EXECUTE_AGENT_EXTENSION_ENTRY_POINT_GROUP,
    NOOP_EXECUTE_AGENT_EXTENSION_KIND,
    ExecuteAgentAfterInvokeContext,
    ExecuteAgentExtension,
    NoopExecuteAgentExtension,
    NoopExecuteAgentExtensionConfig,
)
```

benmccown (approved):
LGTM, nice/sensible refactor.

plugins/nemo-agents/src/nemo_agents_plugin/jobs/execute_extensions.py
mikeknep: This diagnosis is incorrect—we aren't re-exporting here, we're literally importing to use these things—**but** it does make me realize that we don't need the Noop stuff in the platform plugin module; that is purely an internal implementation detail so that the agents-owned code can be polymorphic. Plugin authors don't need to know or care about it. So, I moved the Noop stuff back into the agents module in 1bfb1765d

PR #2154, opened 2026-09-17
chore: add NeMo Helix rename tooling

mckornfield:
## Summary

Adds NeMo Helix rename tooling and hardens it based on review feedback. The rename and verification scripts now scan the Git file set (including tracked ignored files) while explicitly excluding the rename tooling scripts themselves, instead of relying on split string literals to avoid self-rewrites.

## Changes

- Adds `tools/rename/rename-to-nemo-helix.sh` and `tools/rename/verify-nemo-helix-rename.sh` for previewing, applying, and verifying the rename.
- Documents the rename workflow as a HOW-TO with prerequisites, dry-run, apply, resume, verify, and next-step guidance.
- Updates replacement and verification scans to use NUL-safe `git ls-files --cached --others --exclude-standard` file sets so tracked ignored files are included.
- Explicitly excludes the rename and verifier scripts from self-replacement/content verification and fixes the verifier path shown on completion.
- Adds focused regression tests for tracked ignored file handling and self-rewrite prevention.

## Type of Change

- [ ] Documentation only
- [x] Contributor tooling or automation

## Quality Gates

## Verification

Targeted validation:

- `uv run --frozen pytest tests/tools/rename/test_rename_tools.py -v` — passed (2 tests).
- `uv run --frozen ruff check tests/tools/rename/test_rename_tools.py` — passed.
- `uv run --frozen ruff format --check tests/tools/rename/test_rename_tools.py` — passed.
- `bash -n tools/rename/rename-to-nemo-helix.sh tools/rename/verify-nemo-helix-rename.sh` — passed.
- `git diff --check` — passed.
- `git diff --check origin/main...HEAD` — passed.
- `flox activate -- uv run pre-commit run -a` — passed.
- DCO audit over `origin/main..HEAD` — passed for `0542a745971c184d2772f25b88950a0945a14696` and `497274096caf067956341ebf1a34c5fb2a3a5b26`.

## Summary by CodeRabbit

* **New Features**
  * Added tooling to rename project references, paths, module names, and image names to “NeMo Helix.”
  * Added a dry-run option to preview changes, continuation support, and safeguards for conflicts and invalid usage.
  * Added verification tooling to detect remaining legacy names and validate image naming.

* **Documentation**
  * Added usage guidance for rename and verification workflows, including prerequisites and expected results.

* **Tests**
  * Added coverage for renaming and verification across tracked and ignored files.


tools/rename/rename-to-nemo-helix.sh
```diff
@@ -0,0 +1,249 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+repo_root="$(git rev-parse --show-toplevel)"
+cd "$repo_root"
+
+old_product="NeMo Plat""form"
+old_product_lower_platform="NeMo plat""form"
```
mckornfield, line 11: obviously codex made these, not sure why there are `""` all over

mckornfield: lmao, so the script doesn't sed itself, hilarious

mckornfield: wonder if we should just exclude the script?

tools/rename/rename-to-nemo-helix.sh
mckornfield: can we replace these perl calls with sed or awk? bleh on perl

tools/rename/rename-to-nemo-helix.sh
```diff
@@ -192,12 +192,22 @@ for index in "${!old_values[@]}"; do
...
+      done
+}
+
 grep_args=()
 for old_value in "${old_values[@]}"; do
   grep_args+=("-e" "$old_value")
 done
 mapfile -d '' content_files < <(
-  rg -l -0 -I -F --hidden --glob '!.git' --glob '!.git/**' "${grep_args[@]}" . || true
+  git_search_file_set | xargs -0 -r rg -l -0 -I -F --hidden --glob '!.git' --glob '!.git/**' "${grep_args[@]}" -- || true
```
mckornfield: for ripgrep, any reason we can't just use grep? I know it's faster but I'd rather reduce the extra deps

benmccown (commented):
## Review from the NeMo Helix rename-audit effort

Context: I ran a repo-wide audit (dual-model crawl of all ~161 scopes → a findings DB) to
catalog every `NeMo Platform` occurrence ahead of executing this rename, and vetted this
tooling against it. The engine is solid — binary-safe surrogateescape, git-file-set (so
`.venv/` is skipped), self-exclusion, real tests. Since we're aiming to land this fast, most
of the below is optional/FYI — the one I'd actually flag before running it is #1 (lockfile
hash corruption). Ranked by real risk.

### 🔴 1. Unbounded `nmp`→`nhx` will corrupt dependency-lockfile hashes & base64 blobs
`nmp` is replaced as a bare substring anywhere it appears, and it appears **inside integrity
hashes / base64 payloads** — notably `uv.lock` and `web/pnpm-lock.yaml` (and base64 blobs in
some notebooks/mocks). Rewriting `nmp`→`nhx` inside a hash silently invalidates the lockfile
→ broken/re-resolving installs.

Repro on `main`:
```bash
git grep -nE '[A-Za-z0-9+/=_-]{24,}' -- uv.lock web/pnpm-lock.yaml | grep -i nmp | head
```
Suggested fix: exclude lockfiles from **content** replacement (`uv.lock`, `pnpm-lock.yaml`,
`package-lock.json`, `poetry.lock`, `*.lock`). Their real `nemo-platform` package refs
regenerate via `uv lock` / `pnpm install` after the package is renamed — they shouldn't be
hand-edited anyway. (Optionally also skip replacement inside long
`[A-Za-z0-9+/=_-]{24,}` runs as a general guard.) This is the highest-probability actual
breakage in the whole rename.

### 🟡 2. (Optional) Path scoping — only if we want to split the rename into reviewable PRs
This is a nice-to-have, not a blocker — flagging it in case it's useful, but if we're
landing the rename fast, one big mechanical PR + green CI is totally reasonable and this can
be skipped.

The tool is all-or-nothing over the whole git file set, so a single rename PR touches
thousands of files. IF we wanted domain reviewers to eyeball their own area (docs / code /
infra / web) to catch an edge case, we'd want to restrict **both content replacement and
path renames** to a subset — e.g. repeatable `--path <glob>` / `--exclude <glob>` that
intersects the git file set (no-arg = whole repo, backward compatible). ~30 lines in
`rename_common.py`. Purely optional; only worth it if we decide a staged, reviewer-split
rollout beats speed.

### 🟡 3. (Optional, only if we do #2) Scoped verify
If we DO split into staged PRs, `verify_nemo_helix_rename.py` fails if ANY legacy name
remains anywhere, so every intermediate PR fails verification. In that case the same
`--path`/`--exclude` on the verifier would let each PR verify its own scope. Moot if we
land one big PR (whole-repo verify just works). Bundling this with #2.

### 🟡 4. Confirm the first-party image list is complete
`UNPREFIXED_IMAGES` is a hardcoded 4-item list and the bake pattern is `docker-bake.hcl`-
specific. Our DB surfaced other first-party image names (e.g. `nmp-api`, `nmp-gym-host`,
`nmp-cpu-tasks`, the `*-wheel` builders, `nds-v2-huggingface-cli`). Worth confirming the
list is complete (or driving it from a data file) — the content-wide IMAGE_PATTERN only
covers those 4 names, so a missing image just won't get the `nhx-` prefix.

### 🟠 5. Symlinks can be left dangling (8 tracked in the tree)
The content pass skips symlinks (`is_symlink()` → continue), but `apply_path_renames` still
`path.rename()`s them, and several tracked symlinks point at targets whose paths get renamed
— e.g. `.claude/skills`, `.cursor/skills`, `plugins/nemo-evaluator/src/nemo_evaluator/
skills/evaluator-plugin`, `plugins/nemo-data-designer/.../skills/data-designer`. If the
target directory is renamed but the symlink's own link text isn't updated (content is
skipped for symlinks), the link dangles. Worth: after path renames, re-point any tracked
symlink whose target was renamed (`os.readlink` → apply REPLACEMENTS to the target →
`os.symlink`), or at least flag symlinks in the dry-run.
```bash
git ls-files -s | awk '$1=="120000"{print $4}'   # the 8 tracked symlinks
```

### 🟠 6. The image-prefix rule fires in non-image contexts (false positives)
`IMAGE_PATTERN` runs globally over all content, so bare `auditor-tasks` /
`safe-synthesizer-tasks` / `guardrails-callout` become `nhx-…` **even in prose and test
assertions**, not just image refs. Example that would break:
```python
# .github/scripts/tests/test_ngc_metadata.py
assert default_display_name("safe-synthesizer-tasks") == "Safe Synthesizer Tasks"
```
The input string gets prefixed but the expected `"Safe Synthesizer Tasks"` doesn't → test
fails. Also fires in `RELEASING.md` prose and `config-reference.mdx` defaults. Consider
scoping the image-prefix rule to image-ish contexts (bake/Dockerfile/yaml image keys) rather
than all content, or accept + document that these need manual fixups.

### Notes (not blocking)
- Sibling products are safe — the list only ever pairs "nemo" with "platform", so
  `nemo-guardrails` / `nemo-microservices` / `nemo-run` are untouched. 👍
- The legal "always NeMo Helix, never bare Helix" rule holds — the tool never emits a bare
  "Helix". 👍
- Replacements are idempotent and no rule's output feeds another rule's input — so
  `--continue` re-runs are safe (checked the full ordered list). 👍
- Path-rename collisions: none on current `main` (checked all 2,380 path renames — no
  multi-source or dest-exists collisions), so the abort-on-collision path shouldn't trigger
  in practice. Fine as-is.
- A couple of renames are a *different kind* of change the textual tool shouldn't do at all:
  the `credential_provider` DB enum value `nmp` (needs `ALTER TYPE … RENAME VALUE` + a data
  migration) and the encryption/Vault **key names** (`NMP_SECRETS_DEFAULT_ENCRYPTION_KEY`,
  `nemo-platform-key`) where a blind rename points at a different key and makes existing
  encrypted secrets unrecoverable. Worth excluding these from the tool and handling them as
  explicit migrations. Happy to share the full audit list of these carve-outs.

Overall: really nice tool — the only thing I'd genuinely fix before running is the lockfile
exclusion (#1); the scoping stuff is purely "if we want reviewer-split PRs," happy to skip
it for speed. Glad to pair on any of it if useful.

benmccown (approved):
send ittttt

PR #2155, opened 2026-09-17
feat(customization): link completed CLI submit to Studio job details

benmccown:
## Summary

When Studio is deployed, a completed `nemo customization <backend> submit` now prints a deep link to the job's Studio customization details page (`{base}/studio/workspaces/{workspace}/customizations/{jobName}`) instead of only dumping raw JSON. The link is omitted (CLI tracking instructions still shown) when Studio isn't a ready service on the platform. Works for `automodel`, `unsloth`, and `rl`.

## Changes

- New shared `CustomizationSubmitRenderer` (+ URL builder and Studio-availability probe) in `nmp_customization_common/cli/renderer.py`. On a completed submit it echoes the raw response as JSON to **stdout** and writes human guidance — CLI tracking commands and, when Studio is available, the deep link — to **stderr**.
- **stdout stays pure JSON** (byte-for-byte the prior default submit output, via `typer.echo`), so documented automation that parses `json.load(submit stdout)` for the job name is unaffected; all new prose goes to stderr.
- Tracking commands use the real top-level jobs CLI: `nemo jobs list --workspace <ws>` / `nemo jobs get-status <name> --workspace <ws>`.
- Studio availability is detected via the platform **`/status`** endpoint — checking whether `studio` is in `services.ready` — rather than an ad-hoc HTML probe of `/studio/`.
- Wired once for all three backends: they share `BaseContributor.get_cli`, so the renderer is supplied via a new `renderer_cls` hook on `_add_submit_command` (no per-backend duplication).
- `RendererContext` gains a null-safe `base_url` field so deep-link renderers can resolve the platform URL (populated from the already-resolved submit base URL).
- Renderers are bypassed under `--output-format json`, so automation keeps the stable raw-JSON submit contract; the link is TTY-only.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: CLI completion-output UX change; the guidance goes to stderr and stdout stays the documented JSON, so existing customizer docs remain accurate.

## Verification

Targeted validation:

```
uv run ruff check <changed files>                → clean
uv run ruff format --check <changed files>       → clean
uv run --frozen ty check <changed files>         → All checks passed!
uv run --frozen pytest \
  packages/nmp_customization_common/tests/cli/test_renderer.py \
  plugins/nemo-automodel/tests/test_cli.py \
  plugins/nemo-unsloth/tests/test_cli.py \
  packages/nemo_platform_plugin/tests/test_cli_renderer.py \
  packages/nemo_platform_plugin/tests/test_commands.py   → 124 passed
```

Self-reviewed with a multi-model pass before requesting review; findings (wrong tracking-command names, stdout JSON contract, Rich-markup munging) are addressed in the current revision.

## Note for reviewers — Studio-availability detection

This uses the platform **`/status`** API to decide whether to print the link (one cheap GET at submit-completion, TTY-only). Alternatives considered: (a) print unconditionally when a base URL resolves — no network call, but the link 404s if the platform is up but Studio is disabled; (c) gate on explicit config — deterministic but the link rarely appears. I went with `/status` because it satisfies the ticket's "omit when Studio is not available" precisely using the platform's own service-status signal. **Worth a sanity check:** the added GET per interactive submit, and whether `"studio" in services.ready` is the right availability signal in all deployment topologies (e.g. reverse-proxied Studio on a different origin).

## Summary by CodeRabbit

* **New Features**
  * Customization submission results now include a direct Studio link when Studio is available and submission details are valid.
  * Studio links support custom workspaces and safely encoded job names.
  * Submission output displays CLI commands for tracking jobs.

* **Improvements**
  * JSON output remains machine-readable without interactive renderer content.
  * Studio links are omitted gracefully when configuration, availability, or submission details are unavailable.
  * Submission commands can use a specified output renderer when provided.


packages/nmp_customization_common/src/nmp/customization_common/cli/renderer.py
```diff
@@ -0,0 +1,157 @@
...
+    omit the link. The ``workspace`` and ``job_name`` path segments are
+    percent-encoded (``safe=""`` so ``/`` in a name is escaped too), mirroring
+    ``nmp.studio.studio_links``.
+    """
+    trimmed = base_url.strip().rstrip("/") if base_url else ""
+    if not trimmed:
+        return None
+    workspace_segment = quote(workspace, safe="")
+    job_segment = quote(job_name, safe="")
+    return f"{trimmed}/{_STUDIO_PATH_PREFIX}/workspaces/{workspace_segment}/customizations/{job_segment}"
```
mckornfield, line 61: one thing to consider (might not affect this) is if port-forwarding or a different host is being used, think this will not be something to solve here necessarily?

packages/nmp_customization_common/src/nmp/customization_common/cli/renderer.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    if not isinstance(result, dict):
+        return None
+    name = result.get("name")
+    if isinstance(name, str) and name.strip():
+        return name
+    return None
+
+
+class CustomizationSubmitRenderer(CLIRenderer):
```
mckornfield, line 103: interesting... is this overly complicated? or maybe just what is needed?

mckornfield (approved):

packages/nmp_customization_common/src/nmp/customization_common/cli/renderer.py
benmccown: agent injected reply (on behalf of @benmccown):

> one thing to consider (might not affect this) is if port-forwarding or a different host is being used, think this will not be something to solve here necessarily?

Agreed — out of scope here. The link reuses the already-resolved submit base URL (`_resolve_submit_base_url`), so it points at whatever host the CLI itself is talking to; a port-forward or alternate host stays coherent because it is the same base the submit went to. A genuinely split Studio origin (Studio served from a different host than the platform API) is a separate config concern (Studio has its own `studio.platform_base_url` fallback server-side) and is called out for reviewers in the PR description as a known follow-up if false links ever become a real complaint. Not solving it in this PR.

packages/nmp_customization_common/src/nmp/customization_common/cli/renderer.py
benmccown: agent injected reply (on behalf of @benmccown):

> interesting... is this overly complicated? or maybe just what is needed?

About the minimum needed: `on_frame` stashes the single submit dict, `on_complete` echoes it as JSON to stdout (preserving the documented `json.load(stdout)` automation contract) and writes the tracking hints + optional Studio link to stderr. The one deliberate "extra" is the `/status` availability gate — that is the product decision to omit the link precisely when Studio is not a ready service (vs. printing a link that 404s). If we ever decide a 404-able link is acceptable, that probe is the one piece that could be dropped to simplify. Flagged in the PR description for exactly this reviewer call.

packages/nmp_customization_common/src/nmp/customization_common/cli/renderer.py
```diff
@@ -0,0 +1,146 @@
...
+
+        if job_name is None:
+            return
+        if not studio_is_available(ctx.base_url):
+            return
+        url = build_studio_customization_url(ctx.base_url, workspace, job_name)
+        if url is None:
+            return
+        console.print()
+        console.print(f"View in Studio: {url}")
```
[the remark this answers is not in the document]
benmccown: agent injected reply (on behalf of @benmccown):

> Disable Rich markup for dynamic output.

Addressed. The renderer was reworked since this comment (now outdated): the raw response JSON is written to **stdout via `typer.echo`** (no Rich console, so markup can never mangle it), and all human guidance goes to a **stderr `Console(stderr=True, markup=False)`** — so dynamic values (workspace, job name, Studio URL) print literally and are never interpreted as Rich markup. Verified at runtime: a workspace like `[red]ws[/red]` renders literally in the tracking hints. Resolving.

PR #2157, opened 2026-09-17
feat(models): validate model provider URL scheme at create time

benmccown:
## Summary

A model provider created with a scheme-less `host_url` such as `"inference-api.nvidia.com"` (no `https://`) was accepted as an entity, but then failed downstream at model discovery with an opaque **502 backend networking error**. The failure surfaced late (at discovery) and far from its cause.

This validates the provider URL scheme at the **write boundary** (create/upsert) and rejects the request with a clear 4xx naming the missing scheme, instead of accepting the entity and failing later.

Before: `POST provider {host_url: "inference-api.nvidia.com"}` → 2xx, then 502 at discovery.
After: same request → 422 `Model provider host_url must include a scheme (http:// or https://); got 'inference-api.nvidia.com'`. Valid `https://...` URLs are unchanged.

## Changes

- Add a `_validate_host_url` helper + `@field_validator("host_url")` requiring an `http://`/`https://` scheme (rejects bare hosts, wrong schemes, scheme-relative, empty).
- Wire it onto the **write-boundary** schemas only: `CreateModelProviderRequest` and `UpsertModelProviderRequest`, matching the existing `auth_header_format` validator convention.
- Deliberately **not** on the response/domain `ModelProvider` schema: it is rebuilt from persisted rows on every read (`_entity_to_schema`), so validating there would turn a provider stored before this check into an unreadable/un-listable 500 — regressing exactly the reported scenario. Enforce on writes; keep reads of legacy data working.
- Tests: reject case, `https://` happy path (helper + create/upsert boundary), a regression test that a legacy scheme-less `host_url` still round-trips through the response schema, plus uppercase-scheme (accepted) and empty-string (rejected) pins.

Auto-provisioned providers are unaffected: the deployments backend always builds schemed URLs.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: no user-facing docs; the validation error message is self-documenting.

## Verification

Targeted validation:

- `uv run pre-commit run --files <changed>` → ruff, ruff format, ty all Passed
- `uv run --frozen pytest services/core/models/tests/unit/test_host_url_validation.py` → 17 passed
- `uv run --frozen pytest services/core/models/tests/unit/` → 1066 passed, 1 skipped (torch-gated)

## Summary by CodeRabbit

* **Bug Fixes**
  * Model provider host URLs are now validated when creating or updating configurations.
  * URLs must include an `http` or `https` scheme, regardless of letter casing, and a valid hostname.
  * Empty, hostless, and unsupported URL values now return validation errors.
  * Existing persisted providers with scheme-less URLs remain readable.
  * Validation behavior is limited to create and update requests, preserving compatibility with legacy stored configurations.


services/core/models/src/nmp/core/models/schemas.py
```diff
@@ -373,6 +374,20 @@ def _validate_auth_header_format(v: str | None) -> str | None:
...
+def _validate_host_url(v: str) -> str:
+    """Validate that a provider ``host_url`` includes an ``http://`` or ``https://`` scheme.
+
+    A scheme-less URL like ``"inference-api.nvidia.com"`` is accepted as an entity
+    but fails downstream at model discovery with an opaque 502 backend networking
+    error (the discovery HTTP client cannot build a request from a bare host). We
+    reject it here at the API boundary so the user gets a clear, actionable 4xx
+    naming the missing scheme instead of a late, confusing failure.
+    """
+    if urlparse(v).scheme not in ("http", "https"):
```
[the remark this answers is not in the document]
benmccown: agent injected reply (on Ben's behalf):

**1. Hostname-less URLs (Major) — addressed** in 0442b280. `_validate_host_url` now requires `parsed.hostname` in addition to the http/https scheme, so `https:inference-api.nvidia.com`, `https://`, and `http:/v1/...` are rejected too (they'd otherwise 502 the same way). Added those to the reject test cases.

**2. Require HTTPS for credentialed providers / cleartext api-key (Security, Major) — valid but intentionally out of scope here.** This PR is scoped to the missing-scheme→502 bug (NMP-186); the cleartext-secret concern lives in `build_next_request` in the inference-gateway (a different service) and would also reject legitimate in-cluster/local `http://` providers that carry an `api_key_secret_name`. That's a real hardening but a separate change with its own compatibility considerations — tracking it separately rather than folding it into this fix.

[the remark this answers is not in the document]
benmccown: agent injected reply (on Ben's behalf):

**Re: Require HTTPS for credentialed providers (CWE-319) — non-issue under the provider-write authz model; not changing.**

We traced the threat model. Creating or rewriting a provider's `host_url` while it carries an `api_key_secret_name` goes through `check_secret_access` (`models/api/v2/providers.py` → `permissions.py`), which requires the caller to already hold RBAC to that secret. So the "repoint host_url at a cleartext/attacker endpoint to exfiltrate the key" path requires exactly the access an attacker would need to just read the secret directly — the `http://` exposure grants no new capability. The only residual is generic in-cluster passive-sniffing TLS posture, which isn't a provider-config validation concern and isn't in scope for this fix. Resolving.

(The other Major in this comment — reject scheme-only/hostless URLs — was applied in 0442b280.)

services/core/models/src/nmp/core/models/schemas.py
```diff
@@ -373,6 +374,33 @@ def _validate_auth_header_format(v: str | None) -> str | None:
     return v
 
 
+def _validate_host_url(v: str) -> str:
+    """Validate that a provider ``host_url`` includes an ``http://`` or ``https://`` scheme.
+
+    A scheme-less URL like ``"inference-api.nvidia.com"`` is accepted as an entity
```
mckornfield: should we shorten the comment a bit?

services/core/models/tests/unit/test_host_url_validation.py
```diff
@@ -0,0 +1,104 @@
...
+            "https://",  # scheme + delimiter but no host
+            "http:/v1/models",  # single-slash, hostless
+        ],
+    )
+    def test_rejects_urls_without_http_scheme(self, value: str) -> None:
+        with pytest.raises(ValueError, match="must include a scheme"):
+            _validate_host_url(value)
+
+    def test_error_names_the_offending_value(self) -> None:
+        with pytest.raises(ValueError, match="inference-api.nvidia.com"):
```
mckornfield, line 70: lol does this need its own test or can it be parameterized with the above

mckornfield (approved):
agents get a little _too_ focused on what they're solving lol

services/core/models/src/nmp/core/models/schemas.py
benmccown: agent injected reply (on Ben's behalf):

> should we shorten the comment a bit?

Good call — trimmed it to the two load-bearing points (require an http(s) scheme + a hostname or it 502s at discovery; wired on the create/upsert request schemas only, not the response schema, so reads of legacy rows still work). Done in 02a743b1.

services/core/models/tests/unit/test_host_url_validation.py
benmccown: agent injected reply (on Ben's behalf):

> lol does this need its own test or can it be parameterized with the above

Parameterized — folded the four create/upsert methods into two parametrized tests (reject + accept) that run against both request schemas via a small builder list. Done in 02a743b1.

PR #2163, opened 2026-09-17
fix(models): unlink provider from entity when a model is removed from it

benmccown:
## Summary

The Models Controller's provider reconciler never removed a provider's back-reference from a Model Entity when a model dropped out of a still-alive provider's discovered set. Before: a model removed from its provider kept a stale entry in the entity's `model_providers`, so Studio surfaces gating on `hasModelProvider` kept advertising a model that was no longer served (it "lingered"). After: the reconciler unlinks the provider from any entity it served last cycle but no longer serves this cycle — including when the provider goes non-compliant.

## Changes

- `provider_reconciler.py`: after resolving this cycle's `served_models`, diff against the provider's previously-persisted `served_models` and `stage_provider_unlink` each entity that fell out (new `_unlink_dropped_model_entities` helper).
  - Only the dropping provider's own id is removed, so a model still served by another provider keeps that link.
  - LoRA composite ids (`...&adapters/...`) are skipped — they have no standalone Model Entity to unlink.
  - The non-compliant transition also runs the unlink (with an empty resolved set): it clears `served_models` too, so its previously-served entities would otherwise be left permanently orphaned (the wiped mapping leaves nothing to diff on a later cycle).
- Regression tests: dropped model is unlinked; a model still served by another provider keeps that link; non-compliant transition unlinks; a transient discovery failure does NOT unlink; deployment-backed base-entity drop is unlinked; LoRA composite id is skipped.

Root cause: the reconciler was add/update-only on entity links (`stage_provider_link`). `stage_provider_unlink` was previously called only on full provider/deployment teardown (`deployment_reconciler._cleanup_model_entities_for_provider`), never per-model while the provider stayed alive. The `served_models` mapping and the autoprovisioned passthrough VirtualModel already prune on removal; the entity's `model_providers` list was the one piece left stale.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal reconciler behavior; no user-facing API/doc surface changes.

## Verification

Targeted validation:

- `uv run --frozen ty check <provider_reconciler.py>` — passed (source clean).
- `uv run ruff check` + `uv run ruff format --check` on both changed files — passed.
- `uv run --frozen pytest services/core/models/tests/unit` — 1055 passed, 1 skipped (torch-gated, unrelated); includes 6 new regression tests in `test_provider_reconciler.py`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed stale provider links from model entities only after updated provider mappings are saved successfully.
  - Preserved links from other providers when one provider stops serving a model.
  - Removed links for all previously served entities when a provider is no longer compliant and its status update succeeds.
  - Preserved existing links during temporary discovery failures or unsuccessful status updates.
  - Correctly handled deployment-backed model entities and ignored unsupported or composite model identifiers during reconciliation.


benmccown:
agent injected reply:

> **Only unlink after the provider mapping persists.** If the provider update fails, `_unlink_dropped_model_entities` can remove Model Entity back-references while the persisted `served_models` remains unchanged.

Addressed in b9919cb98. Both the success and non-compliant unlink paths now track whether `update_provider_status` succeeded (`update_ok` / `non_compliant_update_ok`) and only run `_unlink_dropped_model_entities` when the mapping write persisted — so a failed provider write can no longer strip entity back-references out from under an unchanged persisted mapping; the next successful cycle re-runs the diff. Added two regression tests (`test_unlink_skipped_when_provider_update_fails`, `test_non_compliant_unlink_skipped_when_provider_update_fails`).

(The earlier "unlink on non-compliant" finding was already addressed in c814c87f0.)

services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
```diff
@@ -565,6 +585,70 @@ async def _reconcile_single_provider(
             await self._ensure_passthrough_virtual_model(ref.workspace, ref.name, existing_vm_names)
             self._emit_heartbeat()
 
+        # Unlink this provider from any Model Entity it served last cycle but no
```
mckornfield: can we make this more concise?

mckornfield (approved):
I want to complain about the tests but 🤷 why

services/core/models/src/nmp/core/models/controllers/provider_reconciler.py
benmccown: agent injected reply:

> can we make this more concise?

Good call — condensed both unlink-path comment blocks (this one and its sibling on the non-compliant path) down to the load-bearing points: why we unlink the stale entity back-reference the served_models/VM pruning misses, and why it's guarded on the provider write succeeding. Comment-only, no behavior change. 41007fe97

PR #2164, opened 2026-09-17
fix(deployments): keep volume entity when backend delete fails

benmccown:
## Summary

The deployments-plugin volume reconciler's delete path only retried a volume delete when the backend *raised*, but backends signal a failed PVC/volume delete by *returning* a `VolumeStatusUpdate(status="FAILED")` rather than raising. The reconciler discarded that returned status and deleted the entity row anyway — orphaning the PVC while the model reported `DELETED`. After this change, the reconciler only removes the entity once the backend confirms the resource is gone (`RELEASED`); any other status leaves the entity and backend resource intact so the next reconcile cycle retries.

## Changes

- `_reconcile_delete` now inspects the `VolumeStatusUpdate` returned by `backend.delete_volume(...)` and only deletes the `Volume` entity on a `RELEASED` (success) status. A `FAILED` (or any non-`RELEASED`) status logs a warning and returns so the delete is retried next cycle, rather than orphaning the PVC.
- Added test coverage for the failed-backend-delete path (entity row survives, backend delete is retried).
- Made the reconciler test-suite mock backend's `delete_volume` status configurable.
- Updated the models-backend teardown test so its substrate `delete_volume` mock returns a `RELEASED` status, matching the real backend contract.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: Corrects internal delete-lifecycle behavior; no API or user-facing workflow change.

## Verification

Targeted validation:

- `uv run --frozen pytest plugins/nemo-deployments/tests/unit/reconciler plugins/nemo-deployments/tests/unit/backends/k8s/test_volumes.py -q` — 90 passed.
- `uv run --frozen pytest services/core/models/tests/unit/controllers/backends/deployments_plugin -q` — 110 passed.
- `uv run ruff check` + `uv run ruff format --check` on touched files — clean.
- `uv run --frozen ty check` on the reconciler — clean.
- `uv run pre-commit run -a` not run locally yet; will run before marking ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Volume records are now removed only after the backend confirms successful release.
  * Failed, incomplete, or interrupted volume deletions retain the record so the operation can be retried.
  * Backend errors and non-release statuses no longer cause volume records to be removed prematurely.
  * Warnings now report unsuccessful deletion statuses and indicate that a retry will occur.
  * Added coverage for successful releases and deletion responses such as failed, pending, bound, and deleting.


benmccown:
Re: CodeRabbit's out-of-diff "Return RELEASED only after the PVC is absent" finding (volume_reconciler.py:67-77 / the k8s delete path):

Valid, but pre-existing and out of scope for this PR. This change fixes the reconciler discarding a returned FAILED status; it does not touch the k8s `delete_volume`, which returns `RELEASED` as soon as the DELETE request is accepted rather than once the PVC is actually gone. That RELEASED-on-accept-vs-absent behavior (and the same-name-replacement race it enables) is a separate property of the k8s backend and is captured as a follow-up so it can be fixed on its own without expanding this focused fix.

Leaving it out here deliberately to keep the diff scoped to the reconciler bug.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
```diff
@@ -57,11 +57,25 @@ async def _reconcile_delete(self, volume: Volume) -> None:
...
             return
 
+        # Backends report a failed delete by returning a FAILED status rather than
+        # raising, so guard on the returned status too: only remove the entity row
+        # once the backend confirms the resource is gone (RELEASED). A non-success
+        # status must leave the entity + backend resource intact so the next cycle
+        # retries, otherwise the entity disappears while the PVC is orphaned.
+        if update.status != "RELEASED":
+            logger.warning(
+                "Backend delete for volume %s reported %s — will retry: %s",
```
gabwow: Are you missing the word "failed"?

gabwow (approved):
Conditionally approving once status message is fixed or my misunderstanding is clearer up.

plugins/nemo-deployments/src/nemo_deployments_plugin/reconciler/volume_reconciler.py
benmccown: agent injected reply:

> Good catch on the readability — the status word wasn't missing at runtime (the second `%s` interpolates `update.status`, so it rendered e.g. "...reported FAILED — will retry: ..."), but the bare `%s` read as if a word were dropped. Reworded to "Backend delete for volume %s did not succeed (reported %s) — will retry: %s" so it reads unambiguously in source too, and added a test asserting the rendered message. Pushed in c4dd13817.

PR #2171, opened 2026-09-17
perf(inference-gateway): skip no-op model-entity-map rebuilds

benmccown:
## Summary

`model_cache.refresh_model_cache` rebuilt the model-entity map every cycle (`O(providers x served_models)`) even when nothing changed. Skip the rebuild when the provider layer is unchanged since the last cycle. This is a **compute-only** optimization — it saves no network calls (the provider and entity fetches still run every cycle); it avoids the synchronous rebuild burst that otherwise blocks the async event loop each interval even in steady state, keeping the loop responsive for in-flight proxy traffic. The win scales with deployment size.

## Changes

- Add a `_entity_map_signature` field to `ModelCache` and, in `refresh_model_cache`, compute a signature over the provider layer (provider identity + each provider's `served_models`) and skip `rebuild_model_entity_map()` when it is unchanged.
- ModelEntity metadata (`spec`/`finetuning_type`/`backend_format`) is applied in place by `update_model_entity_metadata` independent of the rebuild, so it is intentionally excluded from the signature and still runs every cycle.
- Cold-start / signature-collision insurance: an empty entity map while providers exist always forces a rebuild.
- Route `rebuild_model_entity_map`'s inline first-`/` split through the shared `parse_model_entity_ref` (malformed ids still log-and-skip) to remove the duplicated split logic.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: internal cache optimization; no user-visible behavior change.

## Verification

Targeted validation:

- `uv run --frozen ty check services/core/inference-gateway/src/nmp/core/inference_gateway/api/model_cache.py` — All checks passed.
- `uv run ruff check` / `ruff format --check` on both changed files — clean.
- `uv run pytest services/core/inference-gateway/tests/unit/test_model_cache.py` — 34 passed (4 new cache-skip tests: skip-when-unchanged, rebuild-on-served-model-change, force-rebuild-when-empty, metadata-update-always-runs).
- `uv run pytest services/core/inference-gateway/tests/unit` — 507 passed, 5 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model cache refresh behavior to avoid unnecessary rebuilds when provider configurations and served-model mappings are unchanged.
  * Ensured cache updates occur when served models change or mappings are reordered.
  * Continued refreshing provider and model metadata during every refresh.
  * Improved handling of malformed or whitespace-padded model identifiers so they do not interrupt refresh operations.
  * Ensured the selected backend reflects the current served-model order.


services/core/inference-gateway/src/nhx/core/inference_gateway/api/model_cache.py
```diff
@@ -284,7 +298,30 @@ async def refresh_model_cache(
...
+    # synchronous O(providers x served_models) rebuild burst on the event loop every cycle.
+    #
+    # The signature is order-sensitive over each provider's served_models because the
+    # rebuild appends to model_providers[] in that order and consumers pick [0] — a reorder
+    # is a real routing change and must invalidate. Metadata (spec/finetuning_type/
+    # backend_format) and provider config (host_url/secrets) are excluded: both are applied
+    # in place, not via rebuild. Cold start is covered by the initial None signature, so an
+    # empty map with an unchanged signature (no served_models yet, or only malformed ids) is
+    # a valid steady state and is correctly skipped.
+    entity_map_signature = frozenset(
```
mckornfield: "signature" makes me think it's a string, but it's just a set yeah?

mckornfield (approved):

PR #2174, opened 2026-09-17
fix(data-designer): Clean up DD create cli output

mikeknep:
## Summary
Instead of dumping out the entire job entity, print a friendlier summary from the `nemo data-designer create ...` call

## Changes
Updates the custom DD create renderer to buffer the job response details and display a friendly summary `on_complete`

## Type of Change

- [ ] Documentation only

## Quality Gates

## Verification

Targeted validation:

## Summary by CodeRabbit

- **New Features**
  - Job submission summaries now show the job name, optional record count, and error details.
  - Follow-up tracking commands include the appropriate workspace options when needed.
  - Record counts can be extracted from submitted specifications when unavailable in the response.
  - Unexpected or incomplete responses display the raw payload with guidance for listing jobs.

- **Bug Fixes**
  - Special characters in summary values and server error messages are displayed safely without causing rendering errors.


plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
```diff
@@ -260,30 +264,142 @@ def on_error(self, error: BaseException, *, ctx: RendererContext) -> None:
...
+        for label, value in rows:
+            # soft_wrap keeps the follow-up commands on one logical line;
+            # Rich's default wrapping would split them mid-flag and break
+            # the copy-paste these lines exist for.
+            console.print(f"{_SUMMARY_INDENT}[bold]{f'{label}:':<{width}}[/bold]  {value}", soft_wrap=True)
+
+        error_details = frame.get("error_details")
+        if error_details:
+            console.print()
+            console.print(f"{_SUMMARY_INDENT}[yellow]⚠ Error details: {error_details}[/yellow]", soft_wrap=True)
```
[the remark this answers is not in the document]
mikeknep: Addressed in 90e0f3d43

plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
```diff
@@ -260,30 +264,142 @@ def on_error(self, error: BaseException, *, ctx: RendererContext) -> None:
...
+
+    raw_spec = ctx.cli_kwargs.get("spec")
+    if isinstance(raw_spec, str) and raw_spec:
+        try:
+            parsed = json.loads(raw_spec)
+        except json.JSONDecodeError:
+            return None
+        if isinstance(parsed, dict):
+            value = parsed.get("num_records")
+            if isinstance(value, int):
```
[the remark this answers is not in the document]
mikeknep: Not worth it—this will always be an integer

plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
```diff
@@ -260,30 +265,144 @@ def on_error(self, error: BaseException, *, ctx: RendererContext) -> None:
...
+
+    Without it the printed commands are silently wrong for anyone who
+    passed ``-w``/``--workspace`` to ``create``.
+    """
+    workspace = frame.get("workspace")
+    if not isinstance(workspace, str) or not workspace:
+        candidate = ctx.cli_kwargs.get("workspace")
+        workspace = candidate if isinstance(candidate, str) else None
+    if not workspace or workspace == "default":
+        return ""
```
benmccown, line 402: The client level 'default context' can have its workspace set to something other than 'default'. So omitting the `--workspace <workspace_name>` flag doesn't actually equate to `--workspace default` universally.

---

**Follow-up commands omit `--workspace` for `default`, so copy-paste can hit the wrong workspace**
*Priority: P2 · non-blocking · `plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py:402` · found by grok, gpt-astra*

#### Finding: Agent Summary

The printed `nemo jobs get/watch/results list` commands only append `--workspace <ws>` when the job's workspace isn't `"default"`. But `data-designer create` defaults `--workspace` to the literal string `"default"`, while `nemo jobs *` resolves an *omitted* `--workspace` to the active CLI-context workspace — not literally `default`. So for a user whose context workspace isn't `default`, the printed commands query the wrong workspace and miss the job the summary just reported.

- **What:** `_workspace_suffix` returns `""` (no flag) whenever `workspace == "default"`.
- **Where:** `_workspace_suffix` in `cli/renderers.py`.
- **When:** the user's active/configured context workspace is non-default and they run `create` without `-w` (job lands in `default`).
- **Why it matters:** the copy-paste follow-up commands resolve to the context workspace and can't find the just-created job — and the old full-dump at least showed the `workspace` field, so this is a regression for those users.

#### Finding: Agent Breakdown

`_workspace_suffix` (`cli/renderers.py`) ends with:

```python
if not workspace or workspace == "default":
    return ""
return f" --workspace {workspace}"
```

The `create` wrapper defaults its option to the literal `"default"` (`cli/inputs.py`: `workspace: str = typer.Option("default", "--workspace", "-w")`), so a no-flag `create` submits the job into workspace `default`.

The `nemo jobs get/watch/results list` commands, however, declare `--workspace` as `str | None = None` and, when it's omitted, resolve it through the SDK's `_get_workspace_path_param()` — which returns `client.workspace`, i.e. the **active context workspace** (`sdk/.../_client.py`). That is not necessarily `"default"`.

So the two sides disagree about what "no `--workspace`" means: `create` pins to `default`; `jobs *` uses the context. When the context workspace is e.g. `research`, the printed bare commands query `research` and miss the job.

Suggested direction: always emit `--workspace <effective-ws>` (the workspace the job was actually created in), or only omit it when it provably equals what `nemo jobs *` would resolve. Worth a test with a non-default configured context — `test_default_workspace_adds_no_flag` doesn't exercise that today.

plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
```diff
@@ -260,30 +265,144 @@ def on_error(self, error: BaseException, *, ctx: RendererContext) -> None:
...
+        rows: list[tuple[str, str]] = [("Job", job_name)]
+
+        num_records = _num_records(frame, ctx)
+        if num_records is not None:
+            rows.append(("Records", str(num_records)))
+
+        suffix = _workspace_suffix(frame, ctx)
+        rows.extend(
+            [
+                ("Track", f"nemo jobs get {job_name}{suffix}"),
```
benmccown, line 312: This is a bit fiddly. I'm not really sure we should be trying to detect whether a base-url or cluster flag was passed, to override the example commands, but it is a real possibility that someone could use them, although I think incredibly unlikely (honestly most likely to be an internal dev using those flags). And trying to disambiguate 'did the user pass explicit override flags versus point at their cluster/url from their context file' is tricky. I'll leave it to you to decide what to do, but wanted to at least flag it.

---

**Follow-up commands don't preserve `--base-url` / `--cluster` overrides**
*Priority: P2 · non-blocking · `plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py:312` · found by gpt-astra*

#### Finding: Agent Summary

`data-designer create` exposes `--cluster` and `--base-url` to submit against a different deployment, but the printed `nemo jobs ...` follow-up commands preserve only the workspace suffix. If a user submitted with an endpoint override, the copy-paste commands query the active-context endpoint instead of the one the job was submitted to.

- **What:** the Track/Watch/Results rows carry only `_workspace_suffix`, never an endpoint.
- **Where:** `_render_summary` in `cli/renderers.py`.
- **When:** `create` was invoked with `--base-url` or `--cluster`.
- **Why it matters:** same "printed command silently wrong" class as the workspace issue, just lower-frequency (mostly internal devs use those flags).

#### Finding: Agent Breakdown

`create` forwards `--cluster` / `--base-url` (`cli/inputs.py`), and those overrides take precedence over the active context for that one call without changing it. The renderer builds all three commands as bare `nemo jobs ...` plus the workspace suffix, so a copied command falls back to the active context's endpoint (`cli/core/context.py`).

If preserved at all, the global form is `nemo --base-url ... jobs ...` (resolving a named `--cluster` to a URL as needed). As the note above says, disambiguating an explicit override from a context-file endpoint is genuinely fiddly — flagging it, not demanding it.

plugins/nemo-data-designer/tests/unit/test_cli_renderers.py
```diff
@@ -0,0 +1,188 @@
...
+        pytest.param({"name": "", "id": "platform-job-abc"}, id="empty-name"),
+        pytest.param("unexpected string payload", id="not-a-dict"),
+        pytest.param(None, id="none"),
+    ],
+)
+def test_unexpected_shape_falls_back_to_raw_dump(frame: Any) -> None:
+    """Never hide the payload when we can't find the job's identity in it."""
+    out = _render(frame)
+
+    assert "nemo jobs list" in out
```
benmccown, line 188: Super nit, I guess worth throwing your agent at if you're making any other changes, but not a big issue.

---

**Fallback test doesn't assert the raw payload is actually dumped**
*Priority: P3 · non-blocking · `plugins/nemo-data-designer/tests/unit/test_cli_renderers.py:188` · found by grok*

#### Finding: Agent Summary

`test_unexpected_shape_falls_back_to_raw_dump` asserts only that the `nemo jobs list` hint is printed — not that the raw payload is. The whole point of `_render_fallback` is `console.print(self._frame)` (never hide the payload when the job name can't be read), and a regression that dropped that line but kept the hint would still pass.

- **What:** the parametrized fallback test checks only `"nemo jobs list" in out`.
- **Where:** `test_unexpected_shape_falls_back_to_raw_dump` in `tests/unit/test_cli_renderers.py`.
- **Why it matters:** the test's docstring says "Never hide the payload," but it doesn't defend that behavior.

#### Finding: Agent Breakdown

For the `missing-name` case (`{"id": "platform-job-abc", "status": "created"}`), also asserting the payload's identifying value survives would lock in the dump:

```python
if isinstance(frame, dict) and "id" in frame:
    assert frame["id"] in out
```

benmccown (approved):
LGTM! Couple small things I flagged.

plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
mikeknep: I think this will be too difficult to get perfectly right. My inclination is to ignore it and trust that if a user used those flags on their `create` call, they'll understand how to fill them in for the jobs calls as well.

plugins/nemo-data-designer/tests/unit/test_cli_renderers.py
mikeknep: Added a test for this

plugins/nemo-data-designer/src/nemo_data_designer_plugin/cli/renderers.py
mikeknep: Hmm. My first inclination was to just drop the workspace flag entirely, but changed my mind and decided that always printing the workspace flag is better for now. I hate how inconsistent the CLI is about respecting the active context workspace vs. not 😭 It's a broad issue affecting more than just DD; filing a separate ticket for it.

Here, we're now always printing a workspace flag with the correct/effective workspace of the job.

PR #2179, opened 2026-09-18
fix: allow skill installations to work from wheel/custom harness

mckornfield:
## Summary by CodeRabbit

* **New Features**
  * Install skills for custom coding agents using a specified path.
  * Choose custom skill paths during interactive or automatic setup.
  * Install Cursor skills at both project and user scopes.
  * Install skills for OpenCode in dedicated project and user skill directories.
  * Agent selection now includes all available agents, even when no filesystem markers are detected.

* **Bug Fixes**
  * Improved validation for conflicting or incomplete skill installation options.


packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/skills/agents/cursor.py
```diff
@@ -12,7 +12,10 @@
...
     display_name = "Cursor"
-    supported_scopes = [Scope.PROJECT]
+    supported_scopes = [Scope.PROJECT, Scope.USER]
 
     def get_install_path(self, scope: Scope, project_root: Path, skill_name: str) -> Path:
-        return project_root / ".cursor" / "rules" / installed_skill_name(skill_name) / "SKILL.md"
+        target_name = installed_skill_name(skill_name)
+        if scope == Scope.PROJECT:
+            return project_root / ".cursor" / "skills" / target_name / "SKILL.md"
+        return Path.home() / ".cursor" / "skills" / target_name / "SKILL.md"
```
benmccown, line 21: Not really on you to fix, but giving cursor/opencode the same frontmatter rebuild the other harnesses have, via a shared helper, would be a nice quick win.

---

**Cursor/OpenCode leave the installed `name:` mismatched with the (now Agent-Skills) directory**
*Priority: nit · non-blocking · `cursor.py:21` (same for `opencode.py:21`) · found by opus, gpt-astra, grok*

#### Finding: Agent Summary

This PR moves Cursor/OpenCode into the real Agent Skills layout (`.cursor/skills/`, `.opencode/skills/`). The install directory is renamed via `installed_skill_name` (→ `nemo-<x>`), but Cursor/OpenCode inherit the base `format_content`, which returns `skill.raw` unchanged — so the frontmatter keeps its original `name:`. Claude, Codex, and the new CustomPathInstaller all override `format_content` to rebuild `name: nemo-<x>` so it matches the folder.

- **What:** installed dir name (`nemo-inference`) vs frontmatter `name: inference` disagree for Cursor/OpenCode
- **Where:** `cursor.py` / `opencode.py` (inherit `installer.py` `format_content` → `skill.raw`)
- **When:** any bundled skill whose source `name:` lacks the `nemo-` prefix — today that's just `inference` (the other 15 already carry it, so `installed_skill_name` is a no-op)
- **Why it matters:** `docs/contributing/skills-spec.mdx` says `name` must match the directory; a harness that validates name-vs-dir could ignore/reject the skill

#### Finding: Agent Breakdown

Worth noting this isn't newly introduced by the PR — Cursor/OpenCode already used `installed_skill_name` for the directory + raw content before. What changed is the destination (into `skills/`, the layout the harnesses actually load), which makes the pre-existing mismatch newly meaningful. Blast radius today is one bundled skill (`inference`).

The clean fix lines up with the other duplication in the diff: extract the `metadata = {name, description, [preconditions]}` + `yaml.safe_dump` + `content` rebuild (currently copy-pasted across `claude.py`, `codex.py`, `custom.py`) into one shared helper and give Cursor/OpenCode the same treatment. That both fixes the mismatch and removes the triplication in a single move.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/skills/cli.py
```diff
@@ -289,27 +290,53 @@ def install(
...
     """
-    try:
-        installer = get_installer(agent)
-    except UnsupportedAgentError as e:
-        typer.echo(f"Error: {e}", err=True)
+    if agent is None and path is None:
+        typer.echo("Error: provide --agent or --path.", err=True)
         raise typer.Exit(code=1)
+    if agent is not None and path is not None:
+        typer.echo("Error: --agent and --path cannot be used together.", err=True)
```
benmccown, line 321: Nit: Nice to have if you make any changes to the code, maybe your agent can add these tests. Not really critical though.

---

**New option-conflict guards are mostly untested**
*Priority: nit · non-blocking · `cli.py:321` · found by opus, gpt-astra, grok*

#### Finding: Agent Summary

The PR adds several hard-error guards but only the empty `provide --agent or --path` case has a test. The "cannot be used together" / "cannot be combined" / "requires --skills-path" branches, and the `CustomPathInstaller` itself, have no regression coverage — so a future refactor could silently drop one.

- **What:** validation branches added without assertions
- **Where:** `cli.py` (`--agent`+`--path`, `--path`+`--project-dir`) and `setup.py` (`--skills-path`+`--skills-agents`, auto-mode `other` without `--skills-path`)
- **Why it matters:** simple logic, but exactly the kind that regresses invisibly; cheap to lock down

#### Finding: Agent Breakdown

Covered today: `test_install_requires_agent_or_path`, `test_install_to_custom_path`, and the setup all-agents-menu / `--skills-path` auto test. Missing: an assertion on each conflicting-flag error message, a `--user --path` (scope-ignored) case, and a dedicated `test_custom.py` for the installer's path layout + frontmatter. Not critical given how simple the branches are.

packages/nemo_platform_ext/src/nemo_platform_ext/cli/commands/skills/agents/opencode.py
```diff
@@ -17,5 +17,5 @@ class OpenCodeInstaller(BaseAgentInstaller):
     def get_install_path(self, scope: Scope, project_root: Path, skill_name: str) -> Path:
         target_name = installed_skill_name(skill_name)
         if scope == Scope.PROJECT:
-            return project_root / ".opencode" / "commands" / target_name / "SKILL.md"
-        return Path.home() / ".opencode" / "commands" / target_name / "SKILL.md"
+            return project_root / ".opencode" / "skills" / target_name / "SKILL.md"
+        return Path.home() / ".config" / "opencode" / "skills" / target_name / "SKILL.md"
```
benmccown, line 21: I don't think this something we actually address, but just flagging for awareness.

---

**No migration/cleanup of the old install locations**
*Priority: nit · non-blocking · `opencode.py:21` (also `cursor.py`) · found by opus, gpt-astra, grok*

#### Finding: Agent Summary

This PR relocates skill installs: Cursor `.cursor/rules/` → `.cursor/skills/`, OpenCode `.opencode/commands/` → `.opencode/skills/` (and user `~/.opencode/commands` → `~/.config/opencode/skills`). Nothing removes the old copies, so upgrading + reinstalling leaves orphaned skill dirs in the deprecated locations.

- **What:** old-location skill files are left behind after the move
- **Where:** `cursor.py` / `opencode.py` path changes
- **Why it matters:** stale duplicates in `.cursor/rules` / `.opencode/commands` could still be picked up by a harness that reads both, or just accumulate as cruft

#### Finding: Agent Breakdown

Low-stakes — a note in the PR / release notes is probably enough; a full migration pass is likely more than this is worth. Flagging only for awareness.

benmccown (approved):
LGTM! Flagging a few minor/nit level things, will let approval ride.

---

**Agent Summary**

Independent multi-model review (3 reviewers: opus, gpt-astra, grok; every finding verified against head `954b2ac4`). No blocking issues. Three inline nits below (R1 name/dir consistency, R4 test coverage, R9 old-location cleanup). Two notes that aren't attached to a changed line:

**Docs (R7):** _Probs worth mentioning since these are the entrypoints where most folks would discover the `nemo skills install` command._ `SETUP.md`, the root `README.md`, and `packages/nemo_platform/README.md` all document only `nemo skills install --agent <...>` — none mention the new `--path` / `--skills-path` flags, which are exactly the entrypoints a wheel/custom-harness user would need.

**Studio prompt drift (R6, nit):** `web/packages/studio/src/routes/agents/AgentDetailRoute/overview/codingAgentPrompts.ts:40` still tells the agent to look for installed skills "under `.claude/skills`, `.cursor/rules`, or the equivalent." Since this PR moves Cursor installs to `.cursor/skills/`, that hint is now stale (softened by "or the equivalent"). Outside this diff, so flagging here rather than inline — a follow-up sync, not a blocker.

PR #2196, opened 2026-09-18
refactor(entities): consolidate LoRA composite formatting; drop dead parser

benmccown:
## What

Two follow-ups to the `ModelEntityId` cleanup, consolidating the LoRA-composite grammar and removing dead code.

**1. `format_adapter_composite(base_prefix, adapter_workspace, adapter_name)`** — the single home for the composite grammar's **format** (join) side, the inverse of `parse_adapters_suffix`. It renders `{base_prefix}&adapters/{adapter_workspace}/{adapter_name}`, taking the base prefix **verbatim** (no parse, no raise, no `workspace/base_name` split).

That verbatim contract is deliberate — the two production construction sites can't guarantee a qualified base:
- the models-service reconciler's `_resolve_base_backend_model_id` output may be unqualified, and
- the IGW proxy's `default_model_entity` is an unrestricted field.

Parsing either (e.g. via `ModelEntityId.to_composite()`) would raise on an opaque prefix. So the previous code hand-rolled the same `f"{base}{ADAPTERS_INFIX}{ws}/{name}"` join in two places. This routes both f-strings — and `ModelEntityId.to_composite()` — through the shared helper, so the join grammar lives in one spot without imposing a split those sites can't honor.

**2. Remove `parse_igw_openai_model` + `ParseOpenAIModelError`** (and their tests). They have had **no production caller** since the open-source-release baseline — the OpenAI router derives `workspace`/`name` from FastAPI path params + `name.removeprefix(...)` and routes via `resolve_vm_for_model`. The function was dead code kept alive only by its own unit tests.

## Why

The parse side of the `&adapters/` grammar already had one home (`parse_adapters_suffix`), but the format side was duplicated. This finishes the consolidation. Dropping the orphaned parser removes a second, subtly-different parsing contract that no longer had a reason to exist.

## Behavior

Behavior-preserving. The composite ids emitted are byte-identical (the helper is the same join the f-strings did); only the dead parser and the duplicated join logic are removed.

## Testing

- New unit tests for `format_adapter_composite`: infix join, verbatim/unqualified base prefix, round-trip against `parse_adapters_suffix`, and `to_composite()` delegation.
- Full IGW + models + `nmp_common` unit suites pass (3189 passed).
- `ruff` + `ty` clean (only the pre-existing `make_filter_class`/`create_model` diagnostics, unrelated to this change).

## Summary by CodeRabbit

- **Improvements**
  - LoRA adapter model identifiers are now generated consistently across inference and model-management workflows.
  - Custom base prefixes are preserved when combining adapter workspace and name information.
  - Adapter identifiers remain compatible when parsed and reconstructed.

- **Bug Fixes**
  - Improved consistency when resolving models that use LoRA adapters.
  - Removed obsolete OpenAI model-identifier parsing behavior and its unused test coverage.


packages/nhx_common/src/nhx/common/entities/utils.py
```diff
@@ -61,6 +62,29 @@ def parse_adapters_suffix(name: str) -> tuple[str, str, str] | None:
     return base, adapter_workspace, adapter_name
 
 
+def format_adapter_composite(base_prefix: str, adapter_workspace: str, adapter_name: str) -> str:
+    """Join a base prefix and adapter segments into a LoRA-composite id.
+
+    The single home for the LoRA-composite grammar's *format* (join) side — the inverse
```
mckornfield: oh _is it_ the single home

mckornfield (approved):
good bye dead parser

PR #2199, opened 2026-09-18
fix(customizer): ImagePullErr is a recoverable error but treated as fatal

soluwalana:
## Summary

A pod that hits a transient `ImagePullBackOff` while starting is marked `error` permanently, even after the pull succeeds and the container runs normally. `error` is terminal, so every later progress report from the live workload is rejected — and the rejection drops the whole update, including the metrics attached to it — while the API still answers `200 OK`. The result is a training job that runs for hours, reports nothing, and looks failed. After this change a retryable image pull is treated as a pod that has not started yet rather than one that failed, stale pod warning events no longer override live pod state, and progress that arrives after a task was errored is recorded instead of discarded.

## Changes

- `kubernetes/common.py`: split container waiting reasons into `FATAL_WAITING_REASONS` (`InvalidImageName`, `CreateContainerConfigError`) and `RECOVERABLE_WAITING_REASONS` (`ImagePullBackOff`, `ErrImagePull`). The retryable pair now maps to waiting/`PENDING` rather than a task error. An image that genuinely never pulls is still failed, by `ttl_seconds_before_active`.
- `kubernetes/common.py` (`update_all_tasks`): pod Warning events decide task status only while the pod is not `ACTIVE` or `COMPLETED`. Kubernetes keeps events for about an hour, so a warning left by a failure the pod already recovered from was re-marking a running task as errored on every reconcile. When the pod is live those events are logged and dropped instead of being promoted into `error_details`.
- `app/dispatcher.py` (`_create_or_update_task_locked`): when a status transition is rejected and the task is in `ERROR`, merge the incoming `status_details` rather than returning before any of the update is applied. The status stays terminal — only the observation is recorded. Rejected transitions out of `COMPLETED` and `CANCELLED`, and rejected updates that carry no details, behave exactly as before.
- Tests covering each of the three behaviors and their negative cases.

Observed on a dev cluster: a customization job whose image pull backed off for roughly two minutes at startup then trained normally for two and a half hours, during which the API exposed no metrics for it at all and the jobs service logged 79 rejected `error → active` task updates in a 15 minute window.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal controller and state-machine behavior; no API, schema, or configuration surface changes.

## Verification

Targeted validation:

- Workbox (Docker available): `pytest tests/test_dispatcher.py tests/controllers/test_map_kubernetes_job_status.py tests/test_config.py::test_backend_registry_configuration tests/controllers/test_reconciler.py::test_job_reconciler_syncs_active_job -q` — 71 passed in 44.59s. That is the 69 dispatcher and Kubernetes mapping tests from this change, plus the two Docker-dependent cases that fail when the registry cannot register `cpu/default`.
- Local laptop sandbox: the same 69 tests pass; the two Docker-dependent cases fail because Docker is unavailable there and the backend registry drops `cpu/default`. Unrelated to this change.
- `uv run pre-commit run -a` — blocked locally: `Required uv version >=0.10.10 does not match the running version 0.9.14`. Not run, so left unchecked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tasks in an error state can now record progress details while remaining in the error state.
  * Retryable Kubernetes image-pull issues remain pending instead of being incorrectly marked as errors.
  * Fatal container startup issues are reported as errors.
  * Live container failures are reported ahead of active-container status when appropriate.
  * Stale pod warnings no longer override the current status of active or retrying workloads.


services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
```diff
@@ -1281,11 +1286,19 @@ def update_all_tasks(
     for pod_status in pod_statuses:
         status_details, error_details, error_stack = get_pod_details(core_v1, namespace, pod_status.name)
 
-        # If we have error details from pod events or container statuses, mark the status as ERROR.
-        if error_details:
+        # Events live for about an hour, so a Warning the pod already recovered from outlives the
+        # failure. Let it decide the status only while the pod is not running or finished.
+        live_status = map_pod_status_to_platform_status(pod_status)
+        if error_details and live_status not in (PlatformJobStatus.ACTIVE, PlatformJobStatus.COMPLETED):
```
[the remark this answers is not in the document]
soluwalana: addressed

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
```diff
@@ -1281,11 +1294,23 @@ def update_all_tasks(
...
 
-        # If we have error details from pod events or container statuses, mark the status as ERROR.
-        if error_details:
+        # ERROR is terminal, so only let a Warning set it when the pod is neither live nor retrying:
+        # events outlive the failure by an hour and a backing-off pull re-emits one per attempt.
+        live_status = map_pod_status_to_platform_status(pod_status)
+        pod_is_failing = not is_retrying_image_pull(pod_status) and live_status not in (
+            PlatformJobStatus.ACTIVE,
+            PlatformJobStatus.COMPLETED,
+        )
```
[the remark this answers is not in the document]
soluwalana: recheck

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
```diff
@@ -1281,11 +1294,23 @@ def update_all_tasks(
...
+        if error_details and pod_is_failing:
             status = PlatformJobStatus.ERROR
         else:
-            status = map_pod_status_to_platform_status(pod_status)
+            if error_details:
+                logger.info(
+                    "Ignoring pod warning events for a pod that is running, finished, or retrying",
+                    extra={"pod": pod_status.name, "status": live_status, "events": error_details},
+                )
+                error_details = {}
```
anubhutivyas, line 1312: should we also clear `error_stack` here?

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
```diff
@@ -260,6 +260,14 @@ def build_metadata(labels: dict[str, str] | None, metadata: KubernetesObjectMeta
     return client.V1ObjectMeta(labels=merged_labels, annotations=metadata.annotations)
 
 
+# Waiting reasons that cannot resolve without a spec change.
+FATAL_WAITING_REASONS = frozenset({"InvalidImageName", "CreateContainerConfigError"})
```
anubhutivyas, line 264: I see that we categorized the  previous reasons in two categories, but wondering if we should also add `CrashLoopBackOff`? Not sure what happens today when pod goes into that state?

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
soluwalana: I don’t think we should clear it here. This branch only ignores stale Warning-event-derived error_details; error_stack comes separately from a currently terminated non-zero container. Such a container now maps to ERROR and never enters this branch. Clearing it here could erase a valid exit diagnostic.

services/core/jobs/src/nmp/core/jobs/controllers/backends/kubernetes/common.py
soluwalana: CrashLoopBackOff currently falls through to generic waiting and maps to PENDING. It shouldn’t be fatal because it is a retry state and can recover.

anubhutivyas (approved):

PR #2201, opened 2026-09-18
refactor(insights): route scheduled and on-demand analysis through AnalysisRuns

shanaiabuggy:
## Summary
Periodic Insights analysis and Studio trace-import triggers now create AnalysisRuns backed by `agents.execute`. The Insights controller retains cadence, trace-threshold checks, overlap prevention, and scheduler status/cursor bookkeeping, while the legacy `insights.analyze-job` execution path is removed.

## Changes
- Share AnalysisRun submission between the API and controller, tagging backing jobs with their analysis agent for overlap checks.
- Record pending scheduler attempts before submission and reconcile their terminal job state, including when an agent's schedule is disabled. Only successful scheduled attempts advance the incremental cursor; failed attempts preserve it.
- Migrate Studio trace-import submissions to AnalysisRuns and handle submission errors.
- Remove legacy Insights job routes, typed client methods/types, CLI registration, and execution bridge. Existing consumers must use the AnalysisRuns API or `nemo insights analysis-runs` commands.
- Regenerate the Insights OpenAPI, local Studio SDK, and Insights CLI reference; update unit tests and examples. The generated Studio SDK is ignored by Git.

This changes execution plumbing; it does not yet integrate the external trace-intel package.

## Type of Change

## Quality Gates

## Verification

Targeted validation:
- `uv run --frozen --no-sync pytest plugins/nemo-insights/tests/{test_periodic_analysis,test_periodic_analysis_runs,test_client,test_client_imports,test_analysis_runs,test_sdk_analysis_runs,test_cli_analysis_runs,test_execute_extensions,test_insights_list_contract}.py -q`: 156 passed. After correcting the scheduler fixture's type annotation, reran `test_periodic_analysis.py`: 37 passed.
- From `web`: `pnpm --filter nemo-studio-ui test src/api/insightsAnalysis.test.ts src/routes/agents/AgentDetailRoute/analysis/AnalysisConfigPanel.test.tsx src/routes/agents/AgentDetailRoute/analysis/saveAnalysisConfig.test.ts --maxWorkers=1`: 34 passed.
- `uv run --frozen --no-sync pre-commit run -a`: Ruff lint/format, repository ty checks, config docs, copyright, UI lint-staged, plugin import rules, merge-conflict checks, and Flox lock checks passed. Remaining checks blocked by missing `helm-docs`/`yq` and installed uv 0.11.17 versus required 0.10.10; no uv.lock drift.
- Insights OpenAPI generation/reference validation and `pnpm --filter @nemo/sdk gen:insights` passed.
- Broader Insights tests encountered two existing Fabric adapter-discovery failures (`test_fabric_plans_a_config_carrying_an_ethos`, `test_fabric_plans_the_built_config`). The earlier full Studio typecheck encountered missing/stale generated SDK modules and dependencies. A subsequent frozen-lockfile frontend install and SDK regeneration succeeded; the full typecheck has not been rerun. Full CLI-doc generation is blocked by missing evaluator `pytrec_eval`; the Insights section was regenerated from its command tree.
- `env -u NMP_BASE_URL uv run --frozen --no-sync pytest e2e/test_insights_analysis_run.py --run-e2e -q`: **1 passed in 75.81s**. Started an isolated local backend from this branch with subprocess job execution and mocked model responses. Verified AnalysisRun creation, backing `agents.execute` job completion, persisted Insight, saved analysis report, and analyst telemetry in Intake.
- Scheduled triggering and the Studio submission interaction still lack end-to-end coverage. Container/deployed execution and real-model analysis quality were not validated by this test.

### Revalidation after merging main (September 21)
- Merged `origin/main` at `2dd31483b` without conflicts.
- Addressed both CodeRabbit comments: preserve the prior success cursor when the attempt timestamp is missing (with regression coverage for present/absent prior cursors), and move the sample-trace analysis prerequisite to the top of its README.
- `uv run --frozen --no-sync pytest plugins/nemo-insights/tests/{test_periodic_analysis,test_periodic_analysis_runs,test_analysis_runs,test_client}.py -q`: **111 passed**.
- Reran the three Studio test files listed above: **34 passed**.
- Reran `e2e/test_insights_analysis_run.py --run-e2e -q` against an isolated backend: **1 passed in 67.89s**.
- `uv run --frozen --no-sync ty check plugins/nemo-insights/src plugins/nemo-insights/tests/test_periodic_analysis_runs.py`: passed.
- Full pre-commit checks still encounter local toolchain limitations. Repository-wide ty now reports five errors outside this PR: three in `packages/nemo_platform_plugin/tests/files/test_endpoints.py` and two in `plugins/nemo-deployments/tests/unit/reconciler/test_volume_reconciler.py`. Ruff lint/format and the affected Insights type check pass.

## Summary by CodeRabbit

- **New Features**
  - Insights analysis now runs through Analysis Runs, with support for custom run names, execution profiles, and base URLs.
  - Periodic analysis better handles recovery, duplicate prevention, job status, and run persistence.
  - Studio now submits and displays Analysis Runs with improved error handling.

- **Bug Fixes**
  - Prevented analysis submissions when configuration, model references, prerequisites, or available sessions are invalid.

- **Breaking Changes**
  - Removed the legacy Insights analyze-job CLI, API endpoints, client methods, and related documentation.


plugins/nemo-insights/src/nemo_insights_plugin/controller.py
```diff
@@ -147,16 +151,40 @@ async def _reconcile_config(self, config: AnalysisConfig) -> None:
...
+            updated.status = AnalysisConfigStatus.ERROR
+            updated.last_error = "Analysis job was not submitted or no longer exists"
+        else:
+            if not job.status.is_terminal():
+                return status
+            if job.status == PlatformJobStatus.COMPLETED:
+                updated.status = AnalysisConfigStatus.IDLE
+                # The pre-submission boundary keeps telemetry arriving during
+                # execution eligible for the next analysis.
+                updated.last_successful_run_at = status.last_attempted_at or status.last_successful_run_at
```
mikeknep, line 180: **Colliding scheduled-job names can advance another agent's incremental cursor**
*Priority: P1 · blocking · `plugins/nemo-insights/src/nemo_insights_plugin/controller.py:180` · found by gpt-sol, gpt-astra*

#### Finding: Agent Summary

The scheduler can advance one agent's incremental cursor from a *different* agent's completed job, silently skipping analysis coverage for the agent whose cursor moved. It happens when two agents share a job-name prefix and reconcile in the same second: the controller persists a `RUNNING` status pointing at that name before submitting, and later reconciles the job *by name* — without checking that the job actually belongs to this agent — then advances the cursor on `COMPLETED`.

#### Finding: Agent Breakdown

- **What:** `_reconcile_run` (controller.py:176-181) fetches the job by `status.last_submitted_job` and, on `COMPLETED`, advances the cursor (`last_successful_run_at = status.last_attempted_at or status.last_successful_run_at` as of commit `38f358cb`) with no check that `job.custom_fields["insights_analysis_agent"] == config.agent`.
- **Where:** `_reconcile_run` (controller.py:159-186) + the pre-submission persist in `_submit_analysis_job` (controller.py:296-306); the collision-prone name comes from `_job_name` (controller.py:328-343), which truncates `workspace[:8] + agent[:16]` and appends a second-resolution timestamp.
- **When:** two agents whose names share the first ~16 chars (e.g. `research-agent-v1` / `-v2` in the `default` workspace → both `research-agent-v`) reconciled in the same second. Agent A submits successfully as name X; agent B's `AnalysisRun` create conflicts on name X (names are unique) and B's submission fails — but B's pending status already points at name X. When A's job completes, B's reconcile reads A's job by name, marks B successful, and advances B's cursor past telemetry B never analyzed.
- **Why it matters:** silent loss of analysis coverage — the scheduler believes it analyzed telemetry it never touched and won't revisit until new traces arrive.

Note: `_job_name`'s truncation is **pre-existing** (the collision was already possible at base), but the *cross-wire of another agent's cursor* is new in this PR — the old `AnalyzeJob` path recorded the cursor inside the job's own run (keyed by the agent that actually ran it), so a name collision couldn't move a different agent's cursor. The controller now advancing the cursor by name without verifying the agent tag is what makes the cross-wire possible. (`mint_analysis_run_name` in analysis_runs.py:71-78 already mints a UUID-based name for exactly this reason — "an earlier version derived the name from the agent and collided on the second run.")

Note: commit `38f358cb` on this PR already changed this same line to preserve the cursor when `last_attempted_at` is missing (`status.last_attempted_at or status.last_successful_run_at`). The guard below builds on top of that fix — it keeps the `or status.last_successful_run_at` preservation in the agent-match branch — so applying the guard should not regress `38f358cb`.

Fix:
1. In `_reconcile_run`, before accepting `COMPLETED`/advancing the cursor, verify `job.custom_fields.get("insights_analysis_agent") == config.agent`; on mismatch, set `ERROR` and do **not** advance the cursor. Preserve the existing `last_attempted_at or last_successful_run_at` cursor assignment on the agent-match branch.
2. Mint scheduled-attempt names with the existing UUID-based `mint_analysis_run_name` (or add a UUID/hash suffix to `_job_name`) so same-second/same-prefix attempts can't collide at all.

Add a two-agent, same-second, same-prefix regression test — the current name test (test_periodic_analysis.py:817-832) only checks length, not uniqueness.

```
Verify this finding against the current code and only fix it if still present.
In `plugins/nemo-insights/src/nemo_insights_plugin/controller.py` `_reconcile_run` (~:176-181), the COMPLETED branch advances `last_successful_run_at` from the job named in `status.last_submitted_job` without confirming the job belongs to this agent. Add a guard: on `COMPLETED`, require `job.custom_fields.get("insights_analysis_agent") == config.agent`; on mismatch set `AnalysisConfigStatus.ERROR` and do NOT advance `last_successful_run_at`. Preserve the existing `status.last_attempted_at or status.last_successful_run_at` cursor assignment on the agent-match branch (commit 38f358cb) — do NOT regress it. Additionally, mint scheduled-attempt names with the existing UUID-based `mint_analysis_run_name` (or add a UUID/hash suffix to `_job_name`) so same-second, same-prefix attempts cannot collide. Preserve the pre-submission `last_submitted_job` persistence and the overlap-prevention contract. Add a two-agent / same-second / same-prefix regression test to `plugins/nemo-insights/tests/test_periodic_analysis.py` asserting each agent's cursor advances only from its own job, and run `uv run pytest plugins/nemo-insights/tests/test_periodic_analysis.py -v`.
```

```suggestion
            if job.status == PlatformJobStatus.COMPLETED:
                if (job.custom_fields or {}).get("insights_analysis_agent") != config.agent:
                    updated.status = AnalysisConfigStatus.ERROR
                    updated.last_error = "Reconciled job did not belong to this agent"
                else:
                    updated.status = AnalysisConfigStatus.IDLE
                    # Preserve cursor when the attempt timestamp is missing (38f358cb).
                    updated.last_successful_run_at = status.last_attempted_at or status.last_successful_run_at
                    updated.last_error = ""
```

plugins/nemo-insights/src/nemo_insights_plugin/controller.py
```diff
@@ -102,25 +105,21 @@ async def on_startup(self) -> None:
...
+        """Include disabled configs so in-flight runs still get reconciled."""
         if not self.insights_config.analyst.enabled:
             return []
         try:
-            result = await self.entities.list(
-                AnalysisConfig,
-                workspace="-",
-                filter_obj={"enabled": True},
-            )
+            result = await self.entities.list(AnalysisConfig, workspace="-")
```
mikeknep, line 119: **`list_objects` still reads one page, so enabled agents past page 1 stop being scheduled**
*Priority: P1 · non-blocking · `plugins/nemo-insights/src/nemo_insights_plugin/controller.py:119` · found by gpt-astra*

#### Finding: Agent Summary

The config query dropped its `enabled=True` filter (so disabled configs with in-flight runs can still be reconciled — intentional), but it still returns only `result.data` — one page of 100. The entities list defaults to newest-first and the scope is `workspace="-"`, so once a workspace has >100 `AnalysisConfig` rows total, older enabled agents fall off page 1 and are never returned to `reconcile_one` — they stop being submitted, reconciled, or booked.

#### Finding: Agent Breakdown

- **What:** `list_objects` (controller.py:118-126) returns `result.data` without paginating.
- **Where:** the entities list endpoint defaults to `page=1, page_size=100`, `sort=-created_at`; the old code also read one page, but the `enabled=True` filter kept the result set small.
- **When:** a workspace/aggregate with >100 `AnalysisConfig` rows (enabled + disabled combined). Disabled configs consume page-1 slots while never producing a submission.
- **Why it matters:** an enabled, due agent silently stops being scheduled — a periodic-scheduling parity regression. Scale-dependent (needs >100 configs), hence P1 not P0.

Note: I understand the Insights controller is intended to be temporary (to be replaced by a platform primitive), so 100+ scale may not be a target — flagging so it's on the record, not as a merge blocker.

Fix: iterate all pages (loop on `result.pagination` / follow pages) so all enabled AND disabled configs are reconciled; or split into a paginated enabled-only query for due/submission + a disabled-completion pass. Add a test with an enabled/pending config beyond page 1.

```
Verify this finding against the current code and only fix it if still present.
In `plugins/nemo-insights/src/nemo_insights_plugin/controller.py` `list_objects` (~:118-126) returns only `result.data` (one page of 100). Paginate: follow `result.pagination` (or loop pages) so all AnalysisConfig rows (enabled AND disabled) across the `workspace="-"` scope are returned to `reconcile_one`. Preserve the existing exception→`return []` behavior. Add a test with an enabled/pending AnalysisConfig positioned beyond page 1 asserting it still gets reconciled, and run `uv run pytest plugins/nemo-insights/tests/test_periodic_analysis.py -v`.
```

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -91,14 +89,32 @@ async def create_analysis_run(
     entity_client: NemoEntitiesClient = Depends(get_entity_client),
 ) -> AnalysisRunResponse:
     """Create an Insights analysis run backed by the generic ``agents.execute`` job."""
+    return await submit_analysis_run(workspace=workspace, request=request, sdk=sdk, entity_client=entity_client)
```
mikeknep, line 92: **On-demand analysis runs no longer honor `analyst.job_profile`**
*Priority: P2 · non-blocking · `plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py:92` · found by opus*

#### Finding: Agent Summary

The on-demand route `create_analysis_run` delegates to `submit_analysis_run` without a `profile`, so the generic execute route turns it into `"default"`. The old `analyze-job` route was registered with `default_profile=config.analyst.job_profile`, so on-demand runs inherited the operator-configured profile. On a cluster whose only configured profile is non-default (e.g. `"gpu"`), on-demand runs that used to work now submit against `"default"` and can fail to schedule.

#### Finding: Agent Breakdown

- **What:** `create_analysis_run` (analysis_runs.py:89-92) calls `submit_analysis_run(...)` with no `profile`; the generic execute route applies `profile or "default"` (nemo_agents_plugin/jobs/execute.py:398).
- **Where:** the scheduled path is unaffected — the controller passes `profile=self.insights_config.analyst.job_profile` (controller.py:313).
- **When:** clusters without a usable `"default"` profile. No test asserts the on-demand run's profile.
- **Why it matters:** an on-demand migration-completeness parity regression (Studio trace-import / `nemo insights analysis-runs create`).

Fix: `get_nemo_config(InsightsConfig)` is a plain module-level function (not controller-private — the controller calls it at controller.py:105 and reads `.analyst.job_profile` at :313), so the on-demand route can call the same function and thread `profile=config.analyst.job_profile` into `submit_analysis_run`. Note: the controller caches the resolved config at startup; a per-request route would re-resolve on every call (cheap settings resolution, fine in practice — but consider resolving once / via a `Depends` so it's not per-request). Also add a `--profile` knob to the `nemo insights analysis-runs create` CLI and a profile-asserting test.

```
Verify this finding against the current code and only fix it if still present.
In `plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py` `create_analysis_run` (~:89-92), thread the operator-configured job profile into `submit_analysis_run`: resolve `get_nemo_config(InsightsConfig)` (from nemo_platform_plugin.config — same call the controller makes at controller.py:105) and pass `profile=config.analyst.job_profile` (mirror controller.py:313). Prefer resolving the config once rather than per-request. Preserve the request schema and the stranded-run error handling. Add `--profile` to the `nemo insights analysis-runs create` CLI and a test asserting the on-demand run is submitted with the configured profile, then run `uv run pytest plugins/nemo-insights/tests/test_analysis_runs.py -v`.
```

```suggestion
    config = get_nemo_config(InsightsConfig)
    return await submit_analysis_run(
        workspace=workspace, request=request, sdk=sdk, entity_client=entity_client,
        profile=config.analyst.job_profile,
    )
```

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
```diff
@@ -91,14 +89,32 @@ async def create_analysis_run(
...
+
+
+async def submit_analysis_run(
```
mikeknep, line 95: On-demand runs now go through `submit_analysis_run`, which never touches `AnalysisRunStatus` — so a Studio-triggered run no longer advances the scheduled cursor. Is that decoupling intentional? It's arguably more correct, but it means the next scheduled pass re-analyzes the overlapping window, and I want to make sure that's a deliberate choice rather than a side-effect of the migration.

---

**On-demand runs no longer advance the scheduled cursor**
*Priority: P3 · non-blocking · `plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py:95` · found by opus*

#### Finding: Agent Summary

The deleted `AnalyzeJob.run` advanced the per-agent `AnalysisRunStatus` on success (`last_successful_run_at`), so a Studio-triggered on-demand run used to move the scheduler's cursor. `submit_analysis_run` now only persists the `AnalysisRun` and never touches `AnalysisRunStatus`, so after an on-demand run the scheduled cursor is unchanged and the next scheduled run re-analyzes the same window.

#### Finding: Agent Breakdown

- **What:** `submit_analysis_run` (analysis_runs.py:95-176) only `entity_client.create(run)`; no `AnalysisRunStatus` write. The removed `jobs/analyze.py:142-148` success path did `_record_analysis_run_status(status=IDLE, last_successful_run_at=started_at)`.
- **Where:** the new shared submission path; affects only on-demand runs (the controller's scheduled path still owns the cursor via `_reconcile_run`).
- **When:** every on-demand run followed by a scheduled pass over the same window.
- **Why it matters:** redundant work, not data loss — and arguably *more* correct (decoupling on-demand from the scheduled cursor). Flagging so the behavior change is intentional and documented.

Fix: none required if intentional; add a docstring/comment on `submit_analysis_run` noting on-demand runs deliberately do not advance the scheduled cursor.

```
Verify this finding against the current code and only fix it if still present.
In `plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py` `submit_analysis_run` (~:95-176), on-demand runs persist only an AnalysisRun and never update AnalysisRunStatus.last_successful_run_at, unlike the removed AnalyzeJob.run success path (jobs/analyze.py:142-148). Confirm this decoupling is intentional. If so, add a docstring/comment on submit_analysis_run stating on-demand runs deliberately do not advance the scheduled cursor (so the next scheduled pass may re-analyze the overlapping window). If NOT intentional, advance AnalysisRunStatus on on-demand success mirroring the scheduled path. Run `uv run pytest plugins/nemo-insights/tests/test_analysis_runs.py plugins/nemo-insights/tests/test_periodic_analysis.py -v`.
```

plugins/nemo-insights/src/nemo_insights_plugin/controller.py
```diff
@@ -205,12 +233,12 @@ async def _has_enough_new_traces(self, config: AnalysisConfig, status: AnalysisR
...
     async def _has_active_job(self, config: AnalysisConfig) -> bool:
         try:
-            query_params: ListAnalysisJobsQueryParams = {
+            query_params: ListJobsQueryParams = {
                 "filter": json.dumps({"status": {"$in": _ACTIVE_JOB_STATUSES}}),
                 "page_size": 100,
                 "sort": "-created_at",
             }
-            jobs = await self.jobs.list_analysis_jobs(
+            jobs = await self.jobs.list_jobs(
```
mikeknep, line 241: **`_has_active_job` lists all active jobs in the workspace, then filters client-side**
*Priority: P3 · non-blocking · `plugins/nemo-insights/src/nemo_insights_plugin/controller.py:241` · found by opus*

#### Finding: Agent Summary

The overlap check switched from a dedicated insights AnalyzeJob collection to the generic `AsyncJobsClient.list_jobs` filtered only by active status, then filters client-side on `custom_fields["insights_analysis_agent"]`. `items()` auto-paginates, so in a busy workspace with many active jobs of other types this walks every page each 60s pass. Correctness is fine (the custom-field match is exact), but the query is broader than necessary.

#### Finding: Agent Breakdown

- **What:** `_has_active_job` (controller.py:225-244) calls `list_jobs(filter={"status":{"$in":_ACTIVE_JOB_STATUSES}}, page_size=100)` then `async for item in jobs.items(): if _job_targets_agent(item, config.agent)`.
- **Where:** the generic jobs client (the old code used `AsyncAnalysisJobsClient.list_analysis_jobs`, a dedicated insights collection).
- **When:** a workspace with many active non-insights jobs; runs every 60s per reconcile pass.
- **Why it matters:** wasted work each cycle; correctness unaffected (the tag survives `stamp_job_telemetry_plugins`).

Fix (two options): (a) narrow the candidate set with `list_analysis_runs(agent=<agent>)` (server-side agent filter), then `get_execute_job(name=<run.name>)` per candidate to check liveness — tighter for the typical one-in-flight case, small N+1. (b) cleanest IF supported: fold `"insights_analysis_agent": config.agent` into the existing freeform `filter` dict on `list_jobs` (one targeted query, no N+1). `ListJobsQueryParams` has no dedicated `custom_fields` param but `filter` is freeform JSON already used for `status` — worth checking whether the entities/jobs backend honors nested `custom_fields` in `filter` (the jobs endpoint delegates to the entities layer).

```
Verify this finding against the current code and only fix it if still present.
In `plugins/nemo-insights/src/nemo_insights_plugin/controller.py` `_has_active_job` (~:225-244), the workspace-wide active-jobs query + client-side `custom_fields` filter walks all active jobs each 60s pass. Narrow it: either (a) use `list_analysis_runs(agent=<config.agent>)` for the candidate set and `get_execute_job(name=<run.name>)` per candidate for liveness, short-circuiting on the first active match; or (b) if the entities/jobs backend supports it, add `"insights_analysis_agent": config.agent` to the existing freeform `filter` dict on `list_jobs` for one targeted server-side query. Preserve the exact-match semantics and the exception→`return True` fail-safe. Run `uv run pytest plugins/nemo-insights/tests/test_periodic_analysis.py -v`.
```

mikeknep (changes requested):
Thanks for taking this on! This looks really clean. A couple items flagged by my agent during review.

---

**Agent Summary** — 4 independent reviewers (`opus`, `gpt-sol`, `gpt-astra`, `deepseek-pro`) over the hosted diff (reviewed at `b99fabe4`; re-verified against the current head `38f358cb`, base `main`). Verdict: **Request changes** — one blocking issue.

- 🔴 **R1 (blocking):** colliding scheduled-job names can advance another agent's incremental cursor — `_reconcile_run` reconciles by name without confirming the job's agent tag, so a same-second/same-prefix collision can mark the wrong agent successful. Straightforward fix: verify `custom_fields["insights_analysis_agent"]` before advancing the cursor, and mint scheduled-attempt names with the existing UUID-based `mint_analysis_run_name`.
- 🟡 **R2 (non-blocking):** `list_objects` still reads one page — enabled agents past page 1 stop being scheduled at >100 configs. Flagging for the record (I understand the controller is intended to be temporary); the simple fix is to paginate.
- 🟡 **R3 (non-blocking):** on-demand runs no longer honor `analyst.job_profile` (fall back to `"default"`); the route can thread it through like the scheduled path does.
- 🟡 **R4 (non-blocking, question):** on-demand runs no longer advance the scheduled cursor — is the decoupling intentional? (Posed as a question on the inline comment.)
- 🟡 **R5 (non-blocking):** `_has_active_job` walks all active workspace jobs each pass; could narrow via `list_analysis_runs` + per-run liveness reads, or a `custom_fields` filter if the backend supports it.

Skipped (not surfaced): R6 (frontend drops `detail.run` on submission error) and R7 (pre-existing 60s no-backoff retry, not introduced by this PR). Full findings + per-reviewer scorecard are in the run artifacts.

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
shanaiabuggy: @mikeknep Yeah, this was a deliberate decision! I don't think on-demand runs should advance the scheduled cursor since they can use different time or evaluation filters. If we had a more deliberate "Run scheduled analysis now" action, then I think it would make sense to advance the cursor for that scenario

plugins/nemo-insights/src/nemo_insights_plugin/controller.py
shanaiabuggy: @mikeknep Good catch. I checked the Jobs API, and it currently rejects filters on `custom_fields`, so we can’t filter by the agent tag server-side yet.

I’ve narrowed the query to active jobs with source `nemo-agents-plugin-execute` or `insights`, keeping legacy in-flight analysis jobs covered. We still check the agent tag locally and defer submission if the query fails. This excludes unrelated job types without adding per-run job lookups.

plugins/nemo-insights/src/nemo_insights_plugin/analysis_runs.py
mikeknep: Makes sense to me, thanks

mikeknep (approved):
LGTM!

PR #2205, opened 2026-09-18
ci: nektos act guards in release workflow

crookedstorm:
The principle here is to ensure it is safe to test the release workflow with `act` locally. This will output a valid JSON to use as input on internal repos with `act` as well to simulate finishing the release. 

This also does something more nuanced to be mindful of: it checks out some job setup from `main` instead of the release branch but should still ship what's in the release branch only.

## Summary by CodeRabbit

- **New Features**
  - Added safer local `act` execution for release workflows, preventing external deployments, notifications, repository dispatches, and release creation.
  - Local runs now display planned notifications and dispatch payloads in logs by default.
  - Slack notifications can be enabled explicitly during local runs.
  - Cross-repository release actions now use the workflow revision and report intended actions during local execution.

- **Tests**
  - Added coverage for release dispatch behavior in local and standard execution modes.


.github/workflows/release.yaml
```diff
@@ -458,6 +471,13 @@ jobs:
       needs.plan-release.outputs.dry_run != 'true'
     runs-on: ubuntu-latest
     steps:
+      # This is a fresh job. Load the workflow revision, not the selected
+      # release source, because it owns the dispatch helper.
```
svvarom: This comment text is hard to parse lol. Can we make it a bit easier to read?

crookedstorm: On it

svvarom (approved):
Image

PR #2206, opened 2026-09-18
fix(customizer): accept a deployment_config that points at the output model

aray12:
## Summary

A full-weight (`all_weights`) or `lora_merged` fine-tuning job could not reference an existing `ModelDeploymentConfig` unless the output Model Entity already existed — which, on a first run, it never does. Every non-LoRA run with deployment enabled failed at submit:

> `deployment_config cannot be a string reference ('jealous-bronze-roundworm-config') for all_weights training that creates a new model entity.`

Studio creates the config up front, pointing *forward* at the model the run will produce. That ordering is deliberate: `create_deployment_config` runs `_validate_engine_config`, so a bad engine or image surfaces in milliseconds instead of after a training run, and the deployment itself is still only created once training finishes (`launch_model` in the `model_entity` task), so no GPU idles. `create_deployment_config` has never required the referenced entity to exist — only the compiler's validation assumed it must.

LoRA runs were unaffected, because they reference the *base* model's config, which does exist.

## Changes

Validate the **target** rather than the entity's existence, in both compilers: a referenced config is valid when it targets `<workspace>/<output.name>`, whether or not that entity exists yet.

- `services/automodel/src/nmp/automodel/app/jobs/compiler.py` — replace the `get_model` existence branch with `_config_targets_model(resolved_config, workspace, output_name)`.
- `services/rl/src/nmp/rl/app/jobs/compiler.py` — the same change; RL carries an independent copy of this validation.

`_config_targets_model` already existed and was already used by the LoRA branch, so both paths now share one comparison and the `get_model` call disappears from the function entirely.

Configs pointing at a *different* model are still rejected, in both the entity-absent and retrain cases. Inline `DeploymentParams` and LoRA behaviour are unchanged.

## Tests

- RL: `test_full_weight_job_rejects_string_ref_for_a_new_model_entity` asserted exactly the behaviour being removed, so it becomes the acceptance case; a sibling test keeps the "different model entity" rejection covered on the absent-entity path. The two retrain tests are untouched and still pass — with the entity present the check is identical.
- automodel: this validation had no coverage at all (the one string-ref test stubbed `get_model` as present, so it passed before and after). Added the acceptance and rejection cases.
- Confirmed the acceptance test bites: reverting the automodel change makes `test_string_deployment_config_may_point_at_the_unborn_output_model` fail, and restoring it makes it pass.

## Type of Change

## Quality Gates

- [x] Documentation not applicable — justification: no API, schema or flag surface changes; this removes a validation that rejected a already-documented and already-supported input shape.

## Verification

Targeted validation:

- `pytest services/automodel/tests/test_compiler.py services/rl/tests/test_compiler.py` — 68 passed.
- `pytest services/automodel/tests services/rl/tests packages/nmp_customization_common plugins/nemo-automodel/tests plugins/nemo-rl/tests` — 929 passed, 1 pre-existing failure (below).
- `ruff check` / `ruff format --check` on both services — clean.
- `pre-commit run ty --all-files` — passes.
- OpenAPI spec and CLI reference docs regenerated — zero drift.

Not verified by me: the end-to-end Studio path. Worth confirming before merge that a fine-tuning run with LoRA off and deployment enabled is accepted, and that the created `<name>-config` carries `model_entity_id = <workspace>/<output name>`.

`services/automodel/tests/test_progress_reporter.py::test_progress_reporter_calls_sdk_create_or_update` fails on `main` in this environment and is untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deployment configurations can now reference the model produced by a new training job, even before that model exists.
  * Added support for the same behavior when retraining an existing model.
  * Configurations targeting a different model continue to be rejected during job validation.


aray12:

- #2206 ◀
    - #2065

PR #2218, opened 2026-09-21
fix(scaled-evals): increase controller job reconciliation throughput

arpitsardhana:
## Summary

The scaled-evals Platform Jobs controller submits one row per reconcile pass. With the default 10 second interval that works out to about six submissions a minute, so a benchmark that fans out to 200 evaluations spends roughly half an hour just getting submitted before any real work starts.

This change lets one pass keep claiming rows until the queue is empty, with a row-count limit and a wall-clock budget so a busy queue can't hog the pass.

## Changes

- Added a `_drain` wrapper in the controller that calls a single-row step repeatedly until the queue empties, the batch size is reached, or the time budget runs out.
- Wired the three single-row phases through it: `submit_build`, `submit_evaluation`, and `reconcile_evaluation`.
- Those three step methods now return a bool (`False` means the queue is empty). Their logic is otherwise unchanged.
- Added two settings: `platform_jobs_phase_batch_size` (default 20) and `platform_jobs_phase_budget_seconds` (default 5.0).

Two notes on the design:

A row-count limit alone isn't enough. The controller's heartbeat runs only after every phase finishes, and a stale heartbeat marks the controller unhealthy, which gates API readiness. If the Jobs service gets slow, a batch of 20 could stall long enough to take the control plane offline. Hence the separate time budget.

A raising row ends the batch rather than being skipped. Each step already marks its own row failed before it raises, so continuing through the batch would turn one Jobs outage into 20 failed rows per pass instead of one. On the exception path we're back to one row per pass, which is what you want during an outage.

This code sits behind `platform_build_jobs_enabled` and `platform_evaluation_jobs_enabled`, both still off on main, so nothing changes at runtime yet.

## Type of Change

- [ ] Documentation only

## Quality Gates

- [x] Documentation not applicable — justification: internal controller behavior behind flags that are off; no user-facing API or docs surface changes.

## Verification

Targeted validation:

- `pytest plugins/_temporary-scaled-evals/tests` — 969 passed, 2 skipped, 3 failed. The three failures are entry-point discovery tests (`test_plugin_boot.py`, `test_platform_jobs.py::test_jobs_are_discovered_from_plugin_entry_points`, `cli/test_cli.py::test_cli_import_does_not_require_server_settings`). They fail identically on a clean `origin/main` checkout in this local environment (stale venv dist metadata), so they are unrelated to this change.
- `pytest plugins/_temporary-scaled-evals/tests/test_platform_jobs_controller.py plugins/_temporary-scaled-evals/tests/test_entity_store_projection.py` — 9 passed.
- `ruff check` and `ruff format --check` on the changed files — clean.
- `ty check` on the changed source files — clean.
- Pre-commit hooks ran on commit and push (ruff, ruff format, ty, copyright headers, uv lock check, DCO) — all passed.

The new test covers the four behaviors that matter: draining until the queue is empty, stopping at the batch size, stopping at the time budget, and letting a raising row surface to `reconcile()`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for platform job processing, including maximum items per phase and a time budget.
  * Platform job phases now process work in bounded batches and stop when configured limits are reached.
* **Bug Fixes**
  * Improved queue handling so claimed or unavailable records are accounted for correctly.
  * Stale evaluation reconciliation now consistently reports processed work, including missing or active jobs.
  * Failures in one processing phase no longer prevent subsequent phases or heartbeat activity.


plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
```diff
@@ -130,19 +130,34 @@ def _phases(self) -> list[tuple[str, Callable[[], Awaitable[None]]]]:
...
+        """Wrap a single-row step so one pass drains a bounded batch of rows."""
+
+        async def drained() -> None:
+            # A raising step ends the batch and is logged by reconcile(). Each
+            # step marks its own row failed before raising, so continuing here
+            # would turn a Jobs outage into a batch of failed rows per pass
+            # instead of one. Progress is still one row per pass in that case.
+            deadline = time.monotonic() + settings.platform_jobs_phase_budget_seconds
+            for _ in range(settings.platform_jobs_phase_batch_size):
+                if not await step() or time.monotonic() >= deadline:
```
albcui: So my understanding is that `step()` returns a bool that essentially means "not empty", so `not await step()` is saying "not not empty". This double negative was a bit confusing on the first read. Do you think it's better set "empty" as truthy? then we can just do something like

```python
done = await step()
if done or time.monotonic() >= deadline: ...
```

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
```diff
@@ -130,19 +130,34 @@ def _phases(self) -> list[tuple[str, Callable[[], Awaitable[None]]]]:
...
-                ("submit_build", self._submit_one_build),
+                ("submit_build", self._drain(self._submit_one_build)),
                 ("reconcile_builds", self._reconcile_builds),
             ]
         if settings.platform_evaluation_jobs_enabled:
             phases += [
-                ("submit_evaluation", self._submit_one_evaluation),
-                ("reconcile_evaluation", self._reconcile_one_evaluation),
+                ("submit_evaluation", self._drain(self._submit_one_evaluation)),
+                ("reconcile_evaluation", self._drain(self._reconcile_one_evaluation)),
```
albcui: Draining the `reconcile_one_evaluation` spins on the same row because in `release_dispatch_reconcile_claim`, it sets `dispatch_reconcile_claimed_at = NULL`, and `pooled_connection()` commits on block exit. That makes the row immediately claimable on the next call, since `_claim_stale_evaluation` does a `ORDER BY e.dispatch_claimed_at NULLS FIRST, e.created_at, e.id LIMIT 1`.

Maybe the fix is to return early in the `_ACTIVE_JOB_STATUSES` branch?

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
```diff
@@ -277,24 +296,26 @@ async def _reconcile_builds(self) -> None:
...
         try:
             status = (await self.jobs.get_job_status(workspace=settings.platform_jobs_workspace, name=name)).data()
         except NotFoundError:
             await asyncio.to_thread(self._fail_evaluation_job, row, "Platform evaluation job was not found")
-            return
+            return True
         if status.status in _ACTIVE_JOB_STATUSES:
             await asyncio.to_thread(self._release_evaluation_reconcile_claim, row)
-            return
+            return True
```
albcui: Related to the spin drain issue mentioned above. We probably want to return stop the drain loop early when we know the job is still active.

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
```diff
@@ -277,24 +296,26 @@ async def _reconcile_builds(self) -> None:
...
+            return False
         name = str(row["dispatch_job_name"])
         try:
             status = (await self.jobs.get_job_status(workspace=settings.platform_jobs_workspace, name=name)).data()
         except NotFoundError:
             await asyncio.to_thread(self._fail_evaluation_job, row, "Platform evaluation job was not found")
-            return
+            return True
         if status.status in _ACTIVE_JOB_STATUSES:
             await asyncio.to_thread(self._release_evaluation_reconcile_claim, row)
```
albcui: Second thought, do we really need to release the claim while the job is running? What if we just let the lease expire naturally after `claim_timeout = 90s` (`queue_worker.py:46`)?

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
```diff
@@ -130,19 +130,34 @@ def _phases(self) -> list[tuple[str, Callable[[], Awaitable[None]]]]:
...
             phases += [
-                ("submit_build", self._submit_one_build),
+                ("submit_build", self._drain(self._submit_one_build)),
                 ("reconcile_builds", self._reconcile_builds),
             ]
         if settings.platform_evaluation_jobs_enabled:
             phases += [
-                ("submit_evaluation", self._submit_one_evaluation),
-                ("reconcile_evaluation", self._reconcile_one_evaluation),
+                ("submit_evaluation", self._drain(self._submit_one_evaluation)),
```
albcui: I think an alternative to this "drain" approach is to simply move the batching into the query itself, so that instead of submitting one, we always submit a batch.

```python
("submit_evaluations", self._submit_one_evaluations)

...

async def _submit_evaluations(self) -> None:
    batch = await asyncio.to_thread(self._claim_evaluations, settings.platofrm_jobs_phase_batch_size)
    ...
```

This way, we don't need the drain nor the boolean return value to indicate whether there's any rows to claim.

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
arpitsardhana: Good catch, and you're right — this was a real spin, not just a redundant call.

I reproduced it in a test that models the claim faithfully (the fake release really does make the row claimable again). Against the old code the drain made **20 status calls, 18 of them on `job-1`**, and never reached `job-2`. So it wasn't only wasted work: the released row kept sorting first, so the rest of the queue was starved for the whole pass.

Of your two options I went with **not releasing the lease**. Returning early would stop the spin but keep the starvation — the same row still sorts first on the next pass, so the rows behind it would never be reconciled while it stayed active and stale. Holding the lease lets it lapse after `claim_timeout` and lets the drain advance immediately. The lease is only read by this phase, so holding it costs nothing elsewhere, and a 90s delay before re-checking a job that is running normally is fine for a repair loop.

```python
if status.status in _ACTIVE_JOB_STATUSES:
    # Hold the lease rather than releasing it. The claim orders by
    # dispatch_claimed_at, which a release does not change, so a
    # released row sorts first again and the drain re-claims this same
    # row instead of advancing. The lease lapses after claim_timeout,
    # which is soon enough for a job that is running normally.
    return True
```

`_release_evaluation_reconcile_claim` had no other caller so it's deleted; the repository method stays for `kubernetes_job.py`. Covered by `test_an_active_job_keeps_its_reconcile_lease_so_the_drain_advances`, which I verified fails against the old code.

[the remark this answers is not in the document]
arpitsardhana: The mechanism is real: the deadline is checked between rows, so a slow step overruns it.

Not fixing it here, for two reasons. It isn't introduced by this change — before batching, each phase ran exactly one step per pass, so the worst-case delay from a single slow row was identical. What this PR adds is *more* rows per pass, and the budget bounds exactly that. And the suggested remedy is the thing the comment itself warns against: cancelling a coroutine awaiting `asyncio.to_thread` does not stop the worker thread, so a timeout there risks a leaked claim or a half-applied update — a worse failure than a late heartbeat. Bounding a single slow row belongs at the Jobs HTTP client as a per-request timeout, not in the drain loop.

What was wrong was my comment, which claimed more than the code delivers. Corrected:

```python
# Bounds how many further rows a phase starts, not how long one row takes:
# the deadline is checked between rows. The heartbeat runs only after every
# phase and a stale heartbeat marks this controller unhealthy, so a busy
# queue must not hold the pass open. A single slow row can still overrun
# this, exactly as it could before phases were batched.
```

plugins/_temporary-scaled-evals/src/scaled_evals/api/settings.py
```diff
@@ -212,6 +212,23 @@ class Settings(BaseSettings):
...
+        ge=1,
+        le=200,
+        validation_alias="SCALED_EVALS_PLATFORM_JOBS_PHASE_BATCH_SIZE",
+    )
+    # Wall-clock ceiling for one phase, independent of the row count. The
+    # heartbeat runs only after every phase, and a stale heartbeat marks this
+    # controller unhealthy, so a slow Jobs service must not delay it.
+    platform_jobs_phase_budget_seconds: float = Field(
+        default=5.0,
+        gt=0,
```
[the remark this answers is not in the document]
arpitsardhana: Valid. Confirmed against the pinned pydantic 2.13.4 — `gt=0` does accept `inf`, and the deadline then never trips:

```
gt=0 only, inf -> inf
allow_inf_nan=False, inf -> rejected: ValidationError
```

`platform_jobs_phase_batch_size` still bounds the loop so it was never unbounded, but a budget that silently stops existing is worth rejecting outright. Added `allow_inf_nan=False`.

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
arpitsardhana: I like this better than what I did, honestly. I'm not doing it in this PR because there are three separate queues (builds, evaluation submits, stale reconciles) and each has its own claim query, so it's a much bigger change than this one needs 

That repeat-the-same-row bug only exists because we claim one row at a time in a loop. If a single query handed back 20 different rows, it couldn't happen at all. I put that in the ticket: AALGO-724.

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
arpitsardhana: You diagnosed it exactly, and it was worse than wasted work.

I reproduced it with a fake that models the claim faithfully — the fake `release` really does make the row claimable again. Against the old code the drain made 20 status calls, **18 of them on `job-1`**, and never reached `job-2`. Because the released row keeps sorting first, it starves the rest of the queue for the whole pass.

Of your two options I took the second, not releasing the lease. Returning early stops the spin but keeps the starvation: the same row still sorts first on the next pass, so the rows behind it never get reconciled for as long as it stays active and stale. Holding the lease lets it lapse after `claim_timeout` and lets the drain advance immediately. The lease is only read by this phase, so holding it costs nothing elsewhere, and a 90s delay before re-checking a job that is running normally is fine for a repair loop.

```python
if status.status in _ACTIVE_JOB_STATUSES:
    # Hold the lease rather than releasing it. The claim orders by
    # dispatch_claimed_at, which a release does not change, so a
    # released row sorts first again and the drain re-claims this same
    # row instead of advancing. The lease lapses after claim_timeout,
    # which is soon enough for a job that is running normally.
    return True
```

`_release_evaluation_reconcile_claim` had no other caller so it's deleted; the repository method stays for `kubernetes_job.py`. Covered by `test_an_active_job_keeps_its_reconcile_lease_so_the_drain_advances`, which I verified fails against the old code. Fixed in 8003abe.

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
arpitsardhana: renamed the result so the condition reads with a single negation.

```python
claimed = await step()
if not claimed or time.monotonic() >= deadline:
    return
```

I kept the "claimed a row" polarity rather than flipping to `done`, because `done` reads oddly at the step level. Processed does not mean done, so named accoridingly

asutermo (approved):

plugins/_temporary-scaled-evals/src/nemo_scaled_evals_plugin/controller.py
albcui: Makes sense!

albcui (approved):
LGTM

